PR #60, opened 2015-03-13
typo fixes

jdv:


moritz:
Thanks @jdv, I've invited you to the perl6 org, so that you can push directly in future.

PR #61, opened 2015-03-20
Prettier usage output for p6doc and p6doc-index.

wolverian:
Before:

~~~
$ p6doc 
What documentation do you want to read?
Examples: /Users/iv/src/rakudo-star-2015.02/install/bin/p6doc Type::Str
          /Users/iv/src/rakudo-star-2015.02/install/bin/p6doc Type::Str.split
~~~

After:

~~~
$ ~/src/doc/bin/p6doc
What documentation do you want to read?
Examples: p6doc Type::Str
          p6doc Type::Str.split
~~~

I only tested this on OS X.

I'm not sure we actually want this change, but I opened the PR for discussion.

### DIFF
--- a/bin/p6doc
+++ b/bin/p6doc
@@ -65,9 +65,11 @@ sub show-docs(Str $path, :$section) {
 }
 
 multi sub MAIN() {
+    my $me = $*PROGRAM_NAME.IO.basename;
+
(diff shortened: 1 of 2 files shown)


moritz:
I want it :-)

wolverian:
Thanks!

PR #63, opened 2015-03-25
Emphasize unorderedness of "Sets, Bags, and Mixes".

edwinst:
Emphasize the unorderedness of these containers. Avoid calling them "lists".
Add some words about immutable/mutable variants.


paultcochrane:
Thanks!

You have a commit bit now, so you can freely update the docs pages (or any other repo under the perl6 group).

PR #82, opened 2015-05-15
fix anchor link

rjbs:
The old link was to `#Object Construction` but the live page has an underscore in place of space.


moritz:
Since the heading doesn't use an underscore either, IMHO it would be a good idea to fix the tooling instead of generate the correct URL from the link without the underscore

zoffixznet:
I've came across this PR many times, but just now noticed that we can't fix the tooling for this case for these reasons:
1) The hash of the URL comes from an `id` attribute, which is [forbidden to have a space character in it](http://www.w3.org/TR/html-markup/global-attributes.html#common.attrs.id)
2) Even if those were allowed, they are not allowed to be non-encoded in URLs, which would make writing the URLs harder (` ` -> `%20`), they'd also look uglier.

PR #89, opened 2015-06-06
Avoid blowing up if a directory in @*INC doesn't exist

ab5tract:
Not sure how the nonexistent directories ended up in @*INC anyway, but this seems like something that should not result in a blowup.

Arguably it would be useful to add a warning output when an @*INC directory does not actually exist.


ab5tract:
This was resolved through a separate commit.

PR #90, opened 2015-06-07
Whatever

ab5tract:
I wasn't doing proper branching and so the @*INC 'fix' is mixed in with some changes that I hope will make Whatever a bit more grokkable to newcomers.


jonathanstowe:
Hi, The first one needed a manual merge so I pushed them separately the first one as .575666c
and the others as the came in.  Thanks.

PR #103, opened 2015-07-21
Use $*TMPDIR to create temporary files

hoelzro:
`$*TMPDIR` will follow the operating system's best practices for
temporary file management


perlpilot:
Ah, so it does.

PR #104, opened 2015-07-22
Increased scope of modules.pod by adding "Creating and Using" section

LLFourn:
This is simply as I understood it from experimenting and reading s10 and s11. Not sure if some of it wasn't rakudo specific.

### DIFF
--- a/lib/Language/modules.pod
+++ b/lib/Language/modules.pod
@@ -2,9 +2,148 @@
 
 =TITLE Modules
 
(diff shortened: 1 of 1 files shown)


moritz:
Thanks @LLFourn!

If you want to improve a bit on it, please use the active voice more (and less passive voice).

LLFourn:
Thanks @mortiz. I will take another look at it soon.

PR #108, opened 2015-07-27
Attempted to address passive vs active voice and added examples for need,use,import and require

LLFourn:
@moritz is this better? I think my intuition to use passive voice stems from my school days. It's pretty much what we were taught unfortunately :(.


paultcochrane:
Thanks!

PR #128, opened 2015-09-16
Added docs for with/without

dha:
May want to change it to a =head2 heading, but for now have put it as a
=head3 under the =head2 for C<if>

### DIFF
--- a/lib/Language/control.pod
+++ b/lib/Language/control.pod
@@ -238,7 +238,7 @@ two differences C<unless> works the same as L<if>:
 Implementation note: Currently, Rakudo will say "1 Nil 3 0" for the last
 example because it is not caught up to this part of the design yet.
 
-=head3 with and without
+=head3 with, orwith, without
 
 The C<with> statement is like C<if> but tests for definedness rather than
 truth.  In addition, it topicalizes on the condition, much like C<given>:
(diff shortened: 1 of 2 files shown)


lib/Language/control.pod
```diff
@@ -238,6 +238,32 @@ two differences C<unless> works the same as L<if>:
 Implementation note: Currently, Rakudo will say "1 Nil 3 0" for the last
 example because it is not caught up to this part of the design yet.
 
+=head3 with, orwith, without
+
+The C<with> statement is like C<if> but tests for definedness rather than
```
vendethiel: should `defined` be a link here?

lib/Language/variables.pod
```diff
@@ -513,24 +513,25 @@ Therefore, the subroutine
 
 sub a {
     state @x;
-    @x.push(++$)
+    state $l = 'A'
+    @x.push($++)
```
vendethiel: shouldnt that be `$l`?

dha: Urk. Yeah, it should. I must have had a bad copy paste. Will fix. Thanks!

lib/Language/control.pod
dha: I don't think so, unless we have something to link to that's a description of definedness. I don't think it should be a link to C<defined> if that's what you're thinking.

PR #132, opened 2015-09-19
Added docs for C<while> and C<until> in control.pod

dha:


dha:
Added example for Int.chr

jonathanstowe:
:) That works

Issue #143, opened 2015-10-07
Doc from cosine linking to cosine wikipedia page but displayed as sine


tirkarthi:
As I can see from the reddit comments (https://www.reddit.com/r/programming/comments/3nr0xo/larry_wall_unveils_perl_600/cvqo4k1) the doc for routine cos has the text `Coerces the invocant (or in sub form, the argument) to Numeric, interprets it as radians, returns its sine.` The last word sine links to the wikipedia page on cosine and the text could be changed too.

PR #144, opened 2015-10-07
Update cosine wikipedia link with correct text

tirkarthi:
Fixes https://github.com/perl6/doc/issues/143 . Change sine to cosine.


labster:
thanks :D

PR #164, opened 2015-10-15
s/EnumMap/Map/

LLFourn:
PR incase there is a reason no one has done this yet :)


LLFourn:
err I didn't get them all. need case sensitive search.

PR #170, opened 2015-10-19
Remove blob literal section.

bnoordhuis:
Blob literals are described in S02 but the section in syntax.pod uses
the wrong syntax and blob literals don't appear to be implemented in
Rakudo or tested for in Roast.  Avoid confusion, remove the section.


LLFourn:
Thanks!

PR #172, opened 2015-10-21
Remove vim settings bits from a number of documents.

djgoku:
I found the issue by seeing `*.pod` files being broken in github. Remove all instances I could find.

### DIFF
--- a/bin/p6doc
+++ b/bin/p6doc
@@ -190,5 +190,3 @@ sub prompt-with-options( %options, %found ) {
     }
     return $final-docee;
 }
-
-# vim: expandtab shiftwidth=4 ft=perl6
--- a/bin/p6doc-index
+++ b/bin/p6doc-index
@@ -83,5 +83,3 @@ multi sub MAIN('lookup', $key) {
         exit;
     }
 }
-
-# vim: expandtab shiftwidth=4 ft=perl6
--- a/doc/Language/5to6.pod
+++ b/doc/Language/5to6.pod
@@ -1409,5 +1409,3 @@ add it to the document, even if we don't have a good answer yet. That will
 be better than losing the information about a real need.
 
 =end comments
-
-# vim: expandtab shiftwidth=4 ft=perl6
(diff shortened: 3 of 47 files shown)


moritz:
The modelines are there for a reason; rejecting.

(Also pod being broken on github is a symptom of github not rendering Perl 6 pod, not of the modeliness)

djgoku:
Can you point me to a document the reasoning/explanation?

If they are supposed to be there, is there a reason it isn't present on every file?

Jonathan Otsuka

moritz:
It's not documented. It's simply handy that every contributor (at least those using vim) uses the same editor settings.

And it's not in every file because we haven't bothered to automate it.

paultcochrane:
Would it be generally worthwhile to automate it?  If so, I could add a check for the appropriate modelines and add them where necessary.

zoffixznet:
Broken `.pod`s are unrelated to vim files (See #167)

djgoku:
Could we not add a recommended vimrc file with all the settings and point people to it?

FROGGS:
How do you do that? And how nice will that play with ppl jumping between projects?

For example we added the vimlines to the core setting files to not let hard tabs sneak into our codebase. Even when I'm not using vim, I understand that we want to keep them.

djgoku:
I don't use vim so I can't relate. I use the defaults for vi, sublime text and atom.

moritz:
What's the big issue here? Why can't we just keep the status quo? What's the problem?

jonathanstowe:
There isn't one.  In any other text editor the modeline is just a comment, if something is attempting to parse the file and getting confused by a comment then that thing is broken not the document.

zoffixznet:
Reading the discussion, I think this was left un answered:

> moritz: And it's not in every file because we haven't bothered to automate it.

> paultcochrane: Would it be generally worthwhile to automate it? If so, I could add a check for the appropriate modelines and add them where necessary.

jonathanstowe:
So I'd say :+1: to @paultcochrane's suggestion :)

PR #179, opened 2015-10-28
Comment about new method to not confuse new users

teodozjan:
Due to first example contains custom `new` it may give impression that writing `new` is something that programmer want to do. Maybe it would be good to mark it as non standard behavior and say that probably want to write build method?

Or maybe better to close this pull request and rewrite this tutorial in way that will not contain neither BUILD nor new at start. Perl has many ways to initialize attributes without them and should be shown first

Example question http://www.nntp.perl.org/group/perl.perl6.users/2015/10/msg2309.html

### DIFF
--- a/doc/Language/classtut.pod
+++ b/doc/Language/classtut.pod
@@ -17,6 +17,8 @@ is interesting and, at times, useful.
         has Task @!dependencies;
         has Bool $.done;
 
+        # Normally don't need to written
+        # BUILD is equivalent of constructor in other languages
         method new(&callback, *@dependencies) {
             return self.bless(:&callback, :@dependencies);
         }


jonathanstowe:
I agree that the language and examples in the classtut need to be changed to make it clearer that for nearly all cases creating a custom new is un-necessary, and in even in cases where it might be (such as adding a constructor that takes positional arguments,) it is probably not necessary to create a completely new constructor rather than delegating to the existing multi.

Probably a "Constructors in the raw" section toward the end (I would hesitate to use "Advanced Constructors" as there will always be people that think "Advanced? that's what all experts uses right, so I should too ....")

LLFourn:
Thanks for bringing this up @teodozjan.

I agree starting the tutorial with **BUILD** and custom new doesn't really work. It's the cpan doc "synopsis" style (trying to show all the features in one block of code at the top) but in this case it's caused confusion.

re your commit, actually custom **new** is preferable to **BUILD**  in most situations IMO. As @jonathanstowe said custom **new** is fine idea if you are just creating a shorthand for the original default constructor with **multi** and then redispatching. **BUILD** is usually unnecessary but there are situations where you have attributes in a role that really should be private and need to be set with more complex logic at construction time.

awwaiid:
mmm.... in the meantime can we either merge or close this?

zoffixznet:
@awwaiid would you mind fixing the typos; I'm away from anything that's got git on it:
`# Normally don't need to written:` -> Normally **doesn't** need to **be** written.
`# BUILD is equivalent of constructor in other languages` -> of **a** constructor

or something like that :) :smile_cat:

awwaiid:
@zoffixznet fixed

PR #180, opened 2015-10-30
Make clear that is export should appear after function args list

ungrim97:
A simple clarification that is export appear after a subs args


jonathanstowe:
Looks good. Thanks!

Issue #101, opened 2015-07-06
C<given> not searchable.


dha:
Unlike while, given is actually documented in doc. However, entering it in the search field does not give any results. Not sure if this is an issue with given per se, or a deeper problem with the search feature overall.

smls:
Things only show up in the search if
* there's either a full page dedicated to it, or
* it's a routine defined by a section with a structured name like `=head2 method foo` which .htmlify can detect
* It's marked as an indexable item via the X<> markup code

The third option is appropriate for language features like `given` which are documented as free-form sections inside `/lib/Language` pages. We already do it for things like regex syntax; for example in `/language/regex.pod` you'll find headers like:

    =head2 X<One or more: +|regex,+>

`One or more: +` Is the header title that is actually rendered there; `+` is the index item that is being added; and `regex` is the index category to add it to.

Doing the same for the syntax elements defined in `/lib/Language/control.pod` would be trivial: The question is just, what the "index category" should be. Remember that they'll show up in search results as `Category: thing`:

![screenshot of search result](http://i.imgur.com/vrhv9Li.png)

From what I can tell from a simple ack, we already use these index categories:

category         | used for
-----------------|---------------------------------------------
syntax           | things like `.WHAT`, `.HOW`, `.VAR`
infix            | special infixes like `X`
declarator       | things like `sub`, `class`, `role`
quote            | quoting constructs like `qw`, `< >`
regex            | regex syntax elements like `*`, `\s`
regex adverb     | modifiers like `:i`, `:ignorecase` passed to regexes
subscript adverb | adverbs like `:exists`, :delete passed to subscripts

So, what category name should be use for things like `if`, `while`, `given`?

I guess `block statement` would fit, but maybe there's a shorter/better alternative? I'd like to avoid using the generic `syntax` category unless really necessary.

LLFourn:
There is also `use`, `import`,`need` and `require` that are in want of a category while we are on the subject. `compunit statement`?

LLFourn:
Hey I'm just revisiting this and I'm thinking "control" (something that controls whether a block is or isn't executed and how many times and with what arguments each time) is is a fine name for if,given,while etc?

vendethiel:
Yeah, control or control flow seems fine

LLFourn:
PR: https://github.com/perl6/doc/pull/181

moritz:
Now shows up as "Control Flow: given" in the search, thanks to  @LLFourn.

PR #181, opened 2015-10-30
Made headings in control.pod searchable

LLFourn:
```
1. Added X<> blocks around everything control.pod (separating via ' '
instead of '/'. '/' conflicts with url path atm)
2. Made X<if|control flow> work as shorthand for X<if|control flow,if> in htmlify.p6
3. Got rid of extra `while` heading, it was already documented in `while/until`
```

Making PR just incase 2. is not desirable. I X<>'d everything except for LABELs as "control flow".

### DIFF
--- a/doc/Language/control.pod
+++ b/doc/Language/control.pod
@@ -4,7 +4,7 @@
 
 =SUBTITLE Statements used to control the flow of execution
 
-=head2 blocks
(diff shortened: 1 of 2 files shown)


LLFourn:
close when merge :) https://github.com/perl6/doc/issues/101

PR #187, opened 2015-11-09
Fix/add a few links

awwaiid:
I'm a little skeptical of the /type/Blah full path links, but that is what is being used all over.


zoffixznet:
Thanks!

> I'm a little skeptical of the /type/Blah full path links

As I recall, currently, if you don't use full path links, the thing that makes individual bit pages out of large documents fails to make proper links (the link would point to that individual page instead of the full doc the link is originally from)

PR #189, opened 2015-11-09
Clarify Hash literal syntax. Expand on other things.

awwaiid:


doc/Language/syntax.pod
```diff
@@ -395,6 +406,23 @@ You can also assign subroutines to variables.
...
+Subroutines can be declared with multiple signatures.
+
+    multi sub foo() { say "Hello!" }
+    multi sub foo($name) { say "Hello $name!" }
+
+Inside of a class, you can also declare multi-dispatch methods.
+
+    multi method greet { }
+    multi method greet(Str $name) { }
+
```
LLFourn: somehow link to: http://docs.perl6.org/language/functions#Multi-dispatch "for more info" or you could just L<> the heading.

I plan to document `proto` there soon too.

LLFourn:
Thanks again!

doc/Language/syntax.pod
awwaiid: I suppose I should learn what `proto` is :)

LLFourn: Hmm I just realised that `L<>` in headings looks like a bad idea. As you can see it's in the `Int literals` heading, which seems to goof the table of contents link.

Probably best to purge them from headings for now.

PR #196, opened 2015-11-12
Replace lolcat by neutral title

nxadm:
While I love memes as the next guy, I don't think they have a place in the standard documentation, specially now Christmas is arriving.


zoffixznet:
:+1: I too found it a bit out of place when I was reading that doc, because the surrounding documentation is *not* written in this loose style :smile_cat:

PR #199, opened 2015-11-16
Document recent changes to allow non-inclusive ranges as quantifiers

ungrim97:
Recent changes allowed 1^..^10 non inclusive range syntax as general quantifiers in regexes. These were documented.


zoffixznet:
Thanks. Would you fix a tpo tho: `carat` -> `caret` (Unless "carat" is a Perl 6 term I'm not familiar with)

ungrim97:
@zoffixznet Fixed that typo

zoffixznet:
Thanks a lot!

PR #210, opened 2015-11-23
Changed "item" to "scalar" in section on scalar sigils

dha:
Not sure how "item" got in here, as it's not used anywhere else in this context. So, reverting to the more traditional "scalar".


zoffixznet:
Thanks a lot!

PR #212, opened 2015-11-24
add Mu.put and IO::put, add some printable newlines

gfldex:


doc/Type/IO.pod
```diff
@@ -30,6 +30,12 @@ displays
 To print text implicitly including the trailing newline character, use
 L<say|/type/IO#sub_say>.
 
+=head2 sub put
+
+Print the given text on C<$*OUT> (standard output) with appended C<$*OUT.nl-out>. The default for the latter is the platform dependent newline sequence. 
+
+    print 'Merry 1.0!'     # Merry 1.0!␤
```
zoffixznet: That should read `put` ... doesn't it?

PR #213, opened 2015-11-24
better examples for IO, little CSS change that may have fallout. Beware the dragons!

gfldex:


zoffixznet:
Thanks a lot!

There might be some issues with that CSS on mobiles, but we'll cross that bridge when we get there.

Personally, I don't think there's a need to `put` or print all 101 elements in those examples :) A handful would be just as illustrative.

zoffixznet:
Ah, I see the point of them now when looking at say, but those could easily be made something like (0 1 2 [...and so on...] 99 100 101) and )0 1 2 [...and so on...] 99...)

PR #214, opened 2015-11-25
clarify IO::Handle.getc

gfldex:


moritz:
It might be helpful to explain when Nil is returned (is it at EOF?)

gfldex:
S32 thinks: Returns a Str type object if no more lines to be read. (And yes, it says 'lines' not chars.)
rakudo returns Nil if nqp::getcfh returns an empty str.
roast doesn't test .getc beyond EOF.
That's the reason why I don't explain where Nil. I simply don't know. But if you follow the spec you might end with wrong code. So stating what  rakudo does is the best option for now.

zoffixznet:
I'm merging this as is. If anyone can describe the cases for when `Nil` is returned, please make a commit, submit a PR, or just comment here and I'll make a commit.

@gfldex, thanks for the PR! :+1:

Issue #178, opened 2015-10-27
Operator ::( ) not officially documented


ronaldxs:
The ::( ) operator seems to be used in a couple of modules by now including LWP::Simple to test for SSL, lately URI and maybe Panda.  I haven't found it documented anywhere except for a very brief mention in S11.

awwaiid:
::( ) isn't quite an operator -- rather the :: is tapping into the package name system, and the ( ) is doing a lookup. I started some docs in #224 which will hopefully clarify.

awwaiid:
#224 has been merged, which at least starts to document this.

PR #224, opened 2015-12-05
Add package (namespace) documentation, taken from S02

awwaiid:
Looking at #178, I realized we have basically nothing about package namespaces spelled out. I found some material in S02 and started to chop it down -- this is my first pass. We can merge it as "better than nothing" if we like. I'll continue refining in any case.

Mostly I'd like feedback about WHERE are some good places to put this info. Language/modules was the first place I hit upon, but I also considered Language/packages as possibly it's own document.


LLFourn:
mmm yeah I suggest "packages and namespaces" as a new .pod file. Writing modules is not the same as understanding packages and namespaces in p6 so I don't think it belongs in modules.pod.

:+1:  for new file

LLFourn:
`import` might be a nice thing to explain on this new page. Just had someone in irc looking for import and not finding it in the docs.

awwaiid:
@LLFourn if this is tollerable I'd like to merge it where it is and then incrementally improve it

jonathanstowe:
Yeah' I'll go with merging it now.  If people can see it they'll be more inclined to suggest improvements.

PR #230, opened 2015-12-08
Add use-ok

nxadm:
Test.pm's use-ok is not documented. Added.

If no reaction, I will merge this small PR.


zoffixznet:
Thanks a lot.

PR #233, opened 2015-12-09
add trap of adverb precedence

gfldex:


vendethiel:
Maybe this could include !exists for completeness?

PR #238, opened 2015-12-12
Clarify that CStructs are exchanged by reference

sylvarant:
This commit clarifies that the CStructs of NativeCall are always pass by reference. 
This wasn't immediately clear to me when I worked on Compress::Brotli, leading to some odd behaviour.


moritz:
Thanks @sylvarant. This here sounds like it needs a small grammar tweak: "must also by return CStruct objects by reference as well"

sylvarant:
Thanks should be resolved

moritz:
@sylvarant Thanks! Also you can already directly push to this repo without opening a pull request if you want.

Issue #227, opened 2015-12-07
lib/Perl6/Documentable/Registry.pm fails at runtime


gfldex:
on This is rakudo version 2015.11-402-g67202d6 built on MoarVM version 2015.11-34-gc3eea17 implementing Perl v6.b.

Processing Language Pod files ...
   1/32: doc/Language/5to6-nutshell.pod           => language/5to6-nutshell
   2/32: doc/Language/5to6-perlfunc.pod           => language/5to6-perlfunc
   3/32: doc/Language/5to6-perlop.pod             => language/5to6-perlop
   4/32: doc/Language/5to6-perlsyn.pod            => language/5to6-perlsyn
   5/32: doc/Language/5to6-perlvar.pod            => language/5to6-perlvar
   6/32: doc/Language/about.pod                   => language/about
   7/32: doc/Language/classtut.pod                => language/classtut
   8/32: doc/Language/concurrency.pod             => language/concurrency
   9/32: doc/Language/containers.pod              => language/containers
  10/32: doc/Language/control.pod                 => language/control
  11/32: doc/Language/exceptions.pod              => language/exceptions
  12/32: doc/Language/faq.pod                     => language/faq
  13/32: doc/Language/functions.pod               => language/functions
  14/32: doc/Language/glossary.pod                => language/glossary
  15/32: doc/Language/grammars.pod                => language/grammars
Type check failed in assignment to @!categories; expected Str but got Array
  in method add-new at /home/dex/projekte/perl6/doc/lib/Perl6/Documentable/Registry.pm:12
  in block  at htmlify.p6:451
  in sub find-definitions at htmlify.p6:358
  in sub process-pod-source at htmlify.p6:220
  in sub process-pod-dir at htmlify.p6:176
  in sub MAIN at htmlify.p6:123
  in block <unit> at htmlify.p6:808

sylvarant:
I have the exact same issue for:
```
$ perl6 --version
> This is rakudo version 2015.11-477-gde52f97 built on MoarVM version 2015.11-34-gc3eea17 implementing Perl v6.b.
```

smls:
See also http://irclog.perlgeek.de/perl6/2015-12-17#i_11735487

sylvarant:
I can confirm that 
```
make html
```
now works on my machine.
Btw, shouldn't travis run make html, to keep track of it working ?

jonathanstowe:
:+1:  to testing this on travis

awwaiid:
Seems ok now, closing

PR #240, opened 2015-12-13
Fix type check failing (gh #227)

pdl:
This fixes an issue described in #227, however `make html` still doesn't complete due to another issue which I've so far not been able to fix.


zoffixznet:
Thanks a lot.

PR #245, opened 2015-12-15
fix tail docs

abraxxa:
clarify that tail always returns a Seq as its signature already correctly shows


zoffixznet:
Thanks a lot!

Issue #241, opened 2015-12-13
make html fails: Cannot unbox a type object


pdl:
After #240, another issue has cropped up later in the build process

```
Composing doc registry ...
Cannot unbox a type object
  in method lookup at /home/d/f/p6-doc/lib/Perl6/Documentable/Registry.pm:26
  in sub MAIN at htmlify.p6:133
  in block <unit> at htmlify.p6:820
```

```
$ perl6 --version
This is rakudo version 2015.11-513-g5308479 built on MoarVM version 2015.11-44-ge7edb06 implementing Perl v6.b.
```

sylvarant:
I think 6eeef84 and bd1480c have fixed this issue?
I no longer encounter this problem

jonathanstowe:
It is definitely generating the docs okay now.

moritz:
For the record, you can check http://doc.perl6.org/build-log/?C=M;O=D to see the logs of the builds that are used to display https://doc.perl6.org/

PR #250, opened 2015-12-19
Travis runs make html

sylvarant:
This PR extends the travis configuration to also run make html.
I didn't commit this directly because it extends the travis build time from 11 min to **40 min**.

The Pro's are that issues such as  #227 and #240 and #241 will be immediately noticed when they arise.
The Con's are that this duplicates some info captured in the build logs http://doc.perl6.org/build-log/?C=M;O=D  and slows down the CI system substantially

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,6 @@ perl6:
 install:
   - rakudobrew build-panda
   - panda installdeps .
+  - panda install Pod::To::HTML
+
+script: make test && make html
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+addons:
+  apt:
+    packages:
+      - graphviz
 branches:
   except:
     - gh-pages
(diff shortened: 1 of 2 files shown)


zoffixznet:
:+1: on merging. It twice happened that build was broken for several days before anyone noticed.

moritz:
Thanks @sylvarant!

moritz:
Can travis capture build artifacts and make them available through an API? If yes, we could use that for the official website builds :-)

sylvarant:
Travis can upload artifacts to  github (https://docs.travis-ci.com/user/deployment/releases/) 
Have not yet tried it

jonathanstowe:
I guess you could potentially have it upload the output anywhere, though obviously some mechamisms may be more or less secure than others :-\  But potentially one could make it upload to a github gh-pages site and have the existing web site redirect or proxy to that :)

PR #251, opened 2015-12-19
Remove some unneeded unspace

awwaiid:
(I hate unspace)

### DIFF
--- a/htmlify.p6
+++ b/htmlify.p6
@@ -157,11 +157,11 @@ sub MAIN(
 sub process-pod-dir($dir, :&sorted-by = &[cmp], :$sparse) {
     say "Reading doc/$dir ...";
     my @pod-sources =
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks a lot!

awwaiid:
mmm... it may not have been a good idea to merge this so quickly :) . I'm checking that I didn't break anything now...

zoffixznet:
Sorry, I jumped the gun :)

It does seem [this line](https://github.com/perl6/doc/pull/251/files#diff-16f2390767c1219dc1e184afbec86235R587) now [breaks the build](https://travis-ci.org/perl6/doc#L1607). I think it's the same issue as when a hyper op is used with a method chain on multiple line you get errors too...

zoffixznet:
Hm... I reverted that part, but it's still failing. no idea why

PR #255, opened 2015-12-21
Clear as mud - be explicit about the definition of 'module'.

ChoHag:


jonathanstowe:
Looks good to me, I'll merge now and if anyone feels strongly enough they can fix later :)

Issue #47, opened 2015-02-06
TODO: Explanation of compilation phases and phasers


moritz:
A question on #perl6 was roughly: from a `CHECK` phaser inside a class, can I use the class?

To answer this (and other) questions, the various compilation phases, phasers, and type composition time should be explained.

awwaiid:
Closing, since we have at least the basic doc now. Further issues can be on more specific deficiencies.

PR #260, opened 2015-12-24
Phasers

awwaiid:
Mostly adaptation from S04 or so, addresses #47. Maybe could use some further refinement but might be better than nothing as-is.

### DIFF
--- a/doc/Language/phasers.pod
+++ b/doc/Language/phasers.pod
@@ -69,6 +69,16 @@ example, but run the statements as a whole at the indicated time:
 (Note, however, that the value of a variable calculated at compile time may
 not persist under run-time cloning of any surrounding closure.)
 
+When multiple phasers are scheduled to run at the same moment, the general
+tiebreaking principle is that initializing phasers execute in order
+declared, while finalizing phasers execute in the opposite order, because
+setup and teardown usually want to happen in the opposite order from each
+other.  When phasers are in different modules, the C<INIT> and C<END>
+phasers are treated as if declared at C<use> time in the using module.  (It
+is erroneous to depend on this order if the module is used more than once,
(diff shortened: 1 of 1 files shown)


doc/Language/phasers.pod
```diff
@@ -0,0 +1,357 @@
...
+Here is a summary:
+
+      BEGIN {...} #  * at compile time, ASAP, only ever runs once
+      CHECK {...} #  * at compile time, ALAP, only ever runs once
+       LINK {...} #  * at link time, ALAP, only ever runs once
+       INIT {...} #  * at run time, ASAP, only ever runs once
+        END {...} #  at run time, ALAP, only ever runs once
+
+      ENTER {...} #  * at every block entry time, repeats on loop blocks.
+      LEAVE {...} #  at every block exit time (even stack unwinds from exceptions)
```
vendethiel: Stack unwindings?

awwaiid: Later it goes into some info on this -- I think it means that if something way outside catches an exception, AFTER the exception is done it will run all of the intermediate LEAVE phasers. If, however, you cancel the exception (I don't even know how to do that), then the original execution point isn't completed, so the stacks are not popped and execution continues without calling LEAVE until later.

vendethiel: No, I just meant that "unwinds" should probably be "unwindings". I'll leave it to someone who does speak english to confirm :)

jonathanstowe:
I'll merge and reword unwinding later.  In a bar in a shopping centre at the moment B-)

jonathanstowe:
Ha ha, pull requests that have a failing check can't be merged from a phone. Nice touch github :-D

PR #262, opened 2015-12-24
doc Assignment Operators

gfldex:


awwaiid:
Are there any operators that can't be combined with assignment?

gfldex:
> Are there any operators that can't be combined with assignment?

S03: The limitation is actually based on whether the left side can 
function both as an rvalue and an lvalue by the usual correspondence.

That being said, the actual operator is created on the fly in the 
background (on first spotting by Rakudo). If it can't create one, it will 
complain.

PR #278, opened 2015-12-28
Some more documentation about put

mortie:
timotimo in IRC wanted some more text about put in the FAQ. Thoughts?


zoffixznet:
Thanks a lot

PR #290, opened 2015-12-30
Also allow p6doc <filename>

awwaiid:
I wanted to preview some files, and this helped. I can update the usage if it is liked.


jonathanstowe:
Looks like a plan :)

PR #301, opened 2015-12-30
Add -n flag for no pager

MadcapJake:
Adds `-n` flag which pipes the output into `cat` instead of a pager.


MadcapJake:
Can this be merged?

An alternative would be to move the pager to a flag and make the default print to standard out.

sylvarant:
Thanks, I don't know why this wasnt merged sooner

PR #311, opened 2016-01-03
Allow -n flag on method/routine docs

MadcapJake:
Previously would just revert to `more`/`less` when faced with a method/routine,
now it properly passes the flag along to `MAIN` or `show-docs`.

Also wondering if `type` would do the same thing in Windows.  Anyone able to confirm? I could add another ternary like L57 i.e.,
```perl6
($*DISTRO.name eq 'mswin32' ?? 'type' !! 'cat')
```

### DIFF
--- a/bin/p6doc
+++ b/bin/p6doc
@@ -69,6 +69,7 @@ sub show-docs(Str $path, :$section, :$no-pager) {
         shell "$*EXECUTABLE-NAME -I$i --doc=SectionFilter $path $fmt-string | $pager";
     }
     else {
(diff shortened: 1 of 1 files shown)


jonathanstowe:
Looks good.

zoffixznet:
> Also wondering if type would do the same thing in Windows. Anyone able to confirm

Confirming `type` seems to be the equivalent of `cat` on Windows 7.

jnthn:
Also, $*DISTRO.is-win is the more idiomatic way to do the OS check than the string comparison.

PR #312, opened 2016-01-03
Update Windows support in show-docs sub

MadcapJake:
Made original ternary more idiomatic as per @jnthn's comment in
PR #311.  Added another ternary to support no-pager mode in Windows.

Please confirm that `type` works as `cat` in this situation under Windows.

I did some research last night and another alternative might be `findstr "^"` but that seems a bit hackish, would prefer to use `type` if possible.

### DIFF
--- a/bin/p6doc
+++ b/bin/p6doc
@@ -53,25 +53,26 @@ sub locate-module(Str $modulename) {
 }
 
 sub show-docs(Str $path, :$section, :$no-pager) {
-    my $pager = do if $no-pager.defined { 'cat' } else {
-      %*ENV<PAGER> // ($*DISTRO.name eq 'mswin32' ?? 'more' !! 'less');
-    }
+    my $pager;
+    $pager = %*ENV<PAGER> // ($*DISTRO.is-win ?? 'more' !! 'less') unless $no-pager;
     my $which-fmt = qq:x{which fmt}.chomp;
     my $fmt-string = $which-fmt.IO.e ?? "| $which-fmt -w 80 " !! "";
     if not open($path).lines.grep( /^'=' | '#|' | '#='/ )  {
         say "No Pod found in $path";
         return;
     }
+    my $doc-command-str = "$*EXECUTABLE-NAME";
     if $section.defined {
         %*ENV<PERL6_POD_HEADING> = $section;
         my $i = findbin() ~ '../lib';
-        say "launching '$*EXECUTABLE-NAME -I$i --doc=SectionFilter $path $fmt-string | $pager'" if DEBUG;
-        shell "$*EXECUTABLE-NAME -I$i --doc=SectionFilter $path $fmt-string | $pager";
+        $doc-command-str ~= " -I$i --doc=SectionFilter"
+    } else {
+        $doc-command-str ~= " --doc"
     }
-    else {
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Let's try this again.

Why not use ``run`` and pipe everything together in Perl 6? This will get you straight to a working solution without the need to find workarounds for different systems.

See this (and discussions before and after it): http://irclog.perlgeek.de/perl6/2016-01-03#i_11821299

There is one thing though. **ugexe** reported that ``:in()`` does not work on his setup. Obviously, it should work and you should rely on it (e.g. there are tests for that in roast). But perhaps there is some bug that should be fixed, that's ok, I don't think that it should stop you from using Perl6-ish way of piping stuff together.

MadcapJake:
Can you explain how using `run` allows us to avoid finding an alternative for `cat`?  The issue that this PR is addressing is that `cat` is not available in cmd.exe.  Currently there is no issue with how the eventual shell command is assembled on lines 71 and 75.

Not that I don't agree that this should be done in a more Perl6-ish way, but rather that perhaps that's for another PR refactoring the function.

AlexDaniel:
You are right, good point.

.oO(*well, at least we now have “piping stuff in shell is better!!!” thing figured out…*)

So, now on topic. It seems like you are trying to use ``cat`` as a no-op. In that case just use nothing.

Something like this:
```perl6
my $pager;
$pager = %*ENV<PAGER> // ($*DISTRO.is-win ?? 'more' !! 'less') unless $no-pager;

… … …

my $shell_str = “$*EXECUTABLE-NAME --doc $path $fmt-string”;
$shell_str ~= “| $pager” if $pager;
say “launching “$shell_str”” if DEBUG;
shell $shell_str;
```
Surely adapt that to your liking, but IMO code that attempts to find a no-op for specific platform is not the best solution.

There's also a chance that the code is relying on some kind of a side effect that is caused by ``cat``. If that's the case (although unlikely), it should be explained in a comment.

AlexDaniel:
Since you've already got a grip on it, here's another thing to consider (again off-topic). Perhaps it will be easy to you to address that in the future. Here:
```perl6
%*ENV<PAGER> // ($*DISTRO.is-win ?? 'more' !! 'less');
```
What completely escapes me is a blind idea that windows has ``more`` and everything else has ``less``. Well, perhaps that is true most of the time, but why not check it first? E.g. if ``less`` exists then let's use it. If not, let's check ``more`` and use it, etc.

In fact, if I read that correctly, that's already what happens with ``$which-fmt`` thing (although the way it does it is just beyond me as well, ``$which-fmt.IO.e`` is checking whether the file exists, but if it doesn't then the string is going to be empty. So we are checking if file ``''`` (empty string) exists, what an interesting way to do that!).

MadcapJake:
@AlexDaniel Refactored based on your advice, let me know what you think.  I left your last comment's advice for another PR.

AlexDaniel:
:+1: :+1: 

By the way, one way to move towards ``run`` is to make an array of args first. That is, right now everything is adding to ``$doc-command-str`` string, there would not be a big difference if it was appending to ``@doc-command-args`` array. When using shell it would be just ``join ‘ ’, @doc-command-args`` and when using ``run`` it will be ``run(|@doc-command-args)`` plus a line or two to handle the pipe. Maybe this idea could ease up the refactoring. But sure, that's for another PR.

PR #313, opened 2016-01-03
clarifying push usage

daleevans:
added examples of the other two forms of calling, and documented some non-obvious behaviour of what perl5 programmers might expect from fat commas


moritz:
Explanation added in b8749945d69c0ea26928b3908d3facd6bfe4e8b6

PR #321, opened 2016-01-06
Add another way of entering unicode characters in Emacs

CIAvash:
I also changed a few words in Emacs section. Hope that's OK.


moritz:
Thanks @CIAvash!

LLFourn:
Thanks. That is handy.

PR #324, opened 2016-01-07
doc Hash subscript adverbs

gfldex:


uvtc:
I think you're missing the adverbs in the first two code examples.

Also, the message here says it was merged, but I don't see your changes at <https://doc.perl6.org/type/Hash>. Is it supposed to be automatically updated there?

jonathanstowe:
It gets updated on a cron job, I'm not certain of the frequency

PR #334, opened 2016-01-09
Precompile .pod files

LLFourn:
They get put in a directory named `precompiled`. They are only
re-precompiled if changes are made (like .precomp). This reduces memory
usage (works around RT #127020), and speeds it up on second run (despite
slowing it on the first).

run|time|peak mem
-----|-----|--------------
htmlify before |    594s | 1.1GB
after, first run |  634s | 770MB
after, second run | 584s | 730MB

Making a PR so at least one person agrees with me that this is a good idea.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ html/language/
 html/syntax/
 html/images/type-graph*
 html/js/search.js
+precompiled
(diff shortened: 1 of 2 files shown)


azawawi:
Thanks. On my machine I got the following promising results:

Run|Real Time
-----|-----
1 |10m35.300s
2 |7m48.964s
3 |7m45.885s

azawawi:
Thanks :+1:

PR #346, opened 2016-01-13
EXPORT can take arguments |  	fix link

gfldex:


LLFourn:
hey @gfldex, I don't know if you noticed but the example before the one you added showed how EXPORT can take arguments. With both it's kinda repetitive IMO.

gfldex:
Well spotted, thanks.

PR #347, opened 2016-01-14
merge redundant examples and add type capture

gfldex:


ab5tract:
I find the new example hard to follow. Could the use case perhaps be elaborated further?

PR #356, opened 2016-01-16
Cross meta operator

dbrunton:
Some of the meta operator documentation is repeated in the infix operator sections. This makes the cross section match the other meta operators, though.


sylvarant:
Thanks!

PR #363, opened 2016-01-23
Add documentation for s/// operator

drforr:
The =head and =item markers may need some TLC.

### DIFF
--- a/doc/Language/regexes.pod
+++ b/doc/Language/regexes.pod
@@ -652,7 +652,7 @@ all named captures:
 There is a more convenient way to get named captures which is discussed in
 the section on Subrules.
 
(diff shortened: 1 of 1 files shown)


AlexDaniel:
> Note that neither the colon C<:> nor balancing delimiters such as C<{}> or C<()> can be delimiters.

Hm, well. What about this:
```perl6
my $x = ‘foo’;
$x ~~ s{oo} = ‘öö’;
say $x
```
It's a little bit different thing but I think that it should be mentioned (instead of saying that you can't use ``{ }``.

drforr:
I wasn't aware that form existed until I read the synopsis, I'll make a PR for that later.

PR #368, opened 2016-01-25
To, too.

dbrunton:


jonathanstowe:
Something's up with travis right now

zoffixznet:
> Something's up with travis right now

It's been happening since some time yesterday. Happening to Pod::To::HTML too: https://travis-ci.org/perl6/Pod-To-HTML/builds/104747418

I think we need to add `~/.rakudobrew/moar-nom/install/share/perl6/site/bin` to `PATH`, but I'm unsure as to why, all of a sudden

PR #370, opened 2016-01-27
remove 'autofocus' from search input field

fireartist:
The search field autofocus provides a very bad UX on mobile (I tested android/firefox) - causing the virtual keyboard to open on every new page load.

It also breaks the desktop/firefox feature "Search for text when I start typing", whereby typing no longer searches within the page by default.


moritz:
Originally I was in favor of autofocos, but I guess it simply didn't work out.

PR #371, opened 2016-01-27
minor typo edits

molecules:


AlexDaniel:
Hm, another option would be to do this:
```perl6
perl6 -e 'my $a = 2; say join ‘,’, ($a, ++$a)'
```
Perhaps it's worth to demonstrate another benefit of using smartquotes :)

PR #373, opened 2016-01-29
Index operators

pdl:
Mostly, this just adds `X<>` tags, but sometimes adds a name for the operator if it's missing.


awwaiid:
Thanks!

LLFourn:
I think we want to do this by binding to headings rather than bits of text in the body. Each of these add search results but when you click them you don't go anywhere useful. Probably requires some code in `htmlify.p6`.

PR #374, opened 2016-01-30
wont -> won't

dbrunton:
So it fails for the right reason.


LLFourn:
but won't it fail at compile time? Anyway declaring `won't-work` will fail at compile time anyway.

If I understand this correctly they should both be `wont-work`

LLFourn:
I fixed it #375

LLFourn:
I fixed my "fix" https://github.com/perl6/doc/commit/67a4e1235f84408db71aa7e5fca9b2b7ef4efd97

*curls up in a ball*

PR #377, opened 2016-02-03
Provide a simpler example for reversed operators

pdl:
Reversed operators are taught before operator reduction, so give an example that does not depend on operator reduction


zoffixznet:
Thanks a  lot!

PR #383, opened 2016-02-05
Fix Build.pm for proper Windows support

salortiz:
In windows File::Find::&find returns files with backslash (\\) separator, so $dest.split fails to find '/doc' in them.
See:
```
> say $dest, '=>', $dest.split("$workdir/doc")[1];
C:\rakudo\tests\doc\doc\Type\X\Syntax\Signature\InvocantMarker.pod=>Nil
> say $dest, '=>', $dest.split("$workdir\\doc")[1];
C:\rakudo\tests\doc\doc\Type\X\Syntax\Signature\InvocantMarker.pod=>\Type\X\Syntax\Signature\InvocantMarker.pod
```

### DIFF
--- a/Build.pm
+++ b/Build.pm
@@ -12,14 +12,15 @@ class Build is Panda::Builder {
     method build($workdir) {
 
         my $dest-pref = Panda::Installer.new.default-prefix() ~ '/../doc';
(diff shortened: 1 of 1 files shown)


sylvarant:
Thanks :+1:

PR #386, opened 2016-02-08
Fix example grammar to allow leading spaces so that it parses the exa…

tenatus:
…mple text


sylvarant:
Seems good to me :+1:

PR #387, opened 2016-02-09
Enhance the 5-to-6 documents.

shlomif:
Add some missing features and idioms.

### DIFF
--- a/doc/Language/5to6-nutshell.pod
+++ b/doc/Language/5to6-nutshell.pod
@@ -1112,16 +1112,33 @@ e.g.:
 Whereas in Perl 5 you would do:
 
     my $arg = 'Hello';
-    my $captured = `echo "$arg"`;
-    my $captured = qx(echo "$arg");
+    my $captured = `echo \Q$arg\E`;
+    my $captured = qx(echo \Q$arg\E);
 
-In Perl 6, you can do:
+Or using String::ShellQuote (because C<\Q…\E> is not completely right):
+
+    my $arg = shell_quote 'Hello';
+    my $captured = `echo $arg`;
+    my $captured = qx(echo $arg);
+
+In Perl 6, you will probably want to run commands without using the shell:
 
     my $arg = 'Hello';
-    my $captured = qq:x(echo "{$arg}");
+    my $captured = run('echo', $arg, :out).out.slurp-rest;
+    my $captured = run(«echo "$arg"», :out).out.slurp-rest;
 
-Beware of interpolating strings with special shell characters! (Which is a
-problem in Perl 5 as well).
+You can also use the shell if you really want to:
+
+    my $arg = 'Hello';
+    my $captured = shell("echo $arg", :out).out.slurp-rest;
+    my $captured = qqx{echo $arg};
+
+But beware that in this case there is B<no protection at all>! C<run> does
+not use the shell, so there is no need to escape the arguments (arguments
+are passed directly). If you are using C<shell> or C<qqx>, then everything
+ends up being one long string which is then passed to the shell. Unless you
+validate your arguments very carefully, there is a high chance to introduce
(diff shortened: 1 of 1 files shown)


AlexDaniel:
```perl6
my $captured = qq:x(echo "{$arg}");
```
This example does not work with ``()``. Moreover, can we get rid of that kind of insanity?

OK you're slamming a variable into the shell but you say ``Beware of …``, fine, but what's the point of ``"{ … }"``?

shlomif:
@AlexDaniel : hi there! Thanks for your comments. I see that it doesn't work with “(...)” and I can revise it . Which insanity do you wish to get rid of? The “{ ... }” are used to interpolate a variable - they can be omitted in this case.

AlexDaniel:
My suggestion (just a draft, feel free to improve/rephrase/reformat):
```
=head2 Trapping the standard output of executables.

Whereas in Perl 5 you would do:

    my $arg = 'Hello';
    my $captured = `echo \Q$arg\E`;
    my $captured = qx(echo \Q$arg\E);

Or using String::ShellQuote (because ``\Q…\E`` is not completely right):

    my $arg = shell_quote 'Hello';
    my $captured = `echo $arg`;
    my $captured = qx(echo $arg);

In Perl 6, you will probably want to run commands without using the shell:

    my $arg = 'Hello';
    my $captured = run('echo', $arg, :out).out.slurp-rest;
    my $captured = run(«echo "$arg"», :out).out.slurp-rest;

You can also use shell if you really want to:

    my $arg = 'Hello';
    my $captured = shell("echo $arg", :out).out.slurp-rest;
    my $captured = qqx{echo $arg};

But beware that in this case there is B<no protection at all>! C<run> does
not use the shell, so there is no need to escape the arguments (arguments
are passed directly). If you are using C<shell> or C<qqx>, then everything
ends up being one long string which is then passed to the shell. Unless you
validate your arguments very carefully, there is a high chance to introduce
shell injection vulnerability with such code.
```

AlexDaniel:
> The “{ ... }” are used to interpolate a variable - they can be omitted in this case.

Yup. But why? This example is 100% equivalent to the one you have provided:
```perl6
my $captured = qq:x{echo $arg};
```
Except that it does not introduce a false sense of security (fancy quoting does not help to escape variables). It is also almost identical to what *some people do* in perl5.

shlomif:
Thanks for your comment - I replaced the old text with the new text which I revised a little.

AlexDaniel:
Nice! Here are some other things to think about (these don't necessarily have to be done in this PR (or at all), it's just some food for thought):
* ``qx`` exists in Perl 6, but it is not identical to ``qx`` in Perl 5. Maybe the difference should be mentioned?
* ``qq:x`` could be written as ``qqx``. Not a big deal, but perhaps people switching from Perl 5 would like to read about that? Perhaps a link to quoting adverbs would be useful? Or maybe we should just hide it from them for now by using ``qqx``?

AlexDaniel:
Also, if you are willing to improve other docs associated with ``:x``, ``run`` and ``shell`` perhaps you might want to look at http://doc.perl6.org/language/quoting#Shell_quoting%3A_qx. There are various things that could be improved (e.g. it talks about an environment variable being available inside ``qx``. Well, sure enough it would be available, that's how you would expect shell to behave. But to me it  *sounds* like ``qx`` is doing the interpolation of env variables, which is not the case. Perhaps a little bit of rewording will help). Personally I'm interested in seeing the security aspect to be improved, but there other are things that could be done better too.

AlexDaniel:
I've built the docs and it looks OK. I think that this pull request is good to go, so I'll merge it. Other possible improvements can be done in other PRs.

Thanks for your work!

shlomif:
@AlexDaniel : Thanks for merging, and you're welcome.

PR #389, opened 2016-02-10
Bool() doesn't return a Match object.

dbrunton:
Bool() returns a Bool.


zoffixznet:
Thanks a lot.

dbrunton:
Why thank *you* :+1:

PR #390, opened 2016-02-10
link to Block

dbrunton:
Was displaying the link instead of linking the display.


zoffixznet:
Thanks a lot!

@moritz maybe it would be possible for you to give @dbrunton access to the repos so they could submit commits directly? I've seen a couple of decent-quality PRs fly in.

PR #404, opened 2016-02-20
Real: document how to convert string to number

rudis:


jonathanstowe:
Possibly, but this does need to be documented, so I'll merge this now and it can be finessed later.

PR #405, opened 2016-03-03
Fixed definition for ~

jjatria:
The sample definition for "~" was the same as that for "-"


zoffixznet:
Thanks a lot!

jjatria:
Thank _you_!

PR #415, opened 2016-03-06
Fix markup in nutshell headers

jjatria:
These headers had incorrect markup and were not displaying as such. Now they are fixed, but the is the content up to date, or should it be marked as TODO?


sylvarant:
Looks good to me :+1:

PR #421, opened 2016-03-13
Add documentation for pipeline operators

awwaiid:
I left out ==>> and <<== since they are not implemented in Rakudo.

### DIFF
--- a/doc/Language/operators.pod
+++ b/doc/Language/operators.pod
@@ -1893,6 +1893,56 @@ Same as C<infix //>, except with looser precedence.
 Returns the first defined argument, or else the last argument.
 Short-circuits.
 
+=head1 Sequencer Precedence
+
+=head2 infix C«==>»
+
+This X<pipeline> operator takes the result from the left and passes it as the
+last parameter to the right. This allows for a more natural left-to-right data
+flow similar to method chaining, but can be used with functions that are not
+methods of the data in the pipeline.
+
+    # Traditional structure, read bottom-to-top
+    my @result =
+        sort               # (4) Sort, result is <Earth People>
+        grep { /<[PE]>/ }, # (3) Look for P or E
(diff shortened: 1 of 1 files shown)


CIAvash:
Aren't they called `feed`operators?
http://design.perl6.org/S03.html#Feed_operators
http://design.perl6.org/S06.html#Feed_operators

AlexDaniel:
Same question :)

awwaiid:
Doh. I'll fix :)

MadcapJake:
The commas in the `<==` example I'm guessing is a bug right?

Also, perhaps instead of "left" and "right" you should say «the directly previous routine's -- or "leftmost" routine's -- last parameter» and «the directly next routine's -- or "rightmost" routine's -- last parameter» as I find the current verbiage slightly confusing. 

After some playing with the REPL, the `<==` example doesn't actually work and according to IRC, both of these are really YMMV and perhaps even NYI.

Maybe there should be a warning (though really these should be under the experimental pragma).

MadcapJake:
@masak was able to help me discover that the `<==` requires that you use parentheses:

```
my @result = sort() <== grep(/P|E/) <== map({ .tc }) <== qw{people of earth};
```

And this solves the syntax error however it gives me a new error:
```
rakudo-moar 42fb81: OUTPUT«===SORRY!=== Error while compiling /tmp/D7eBMaEp7Y␤Sorry, do not know how to handle this case of a feed operator yet.␤at /tmp/D7eBMaEp7Y:1␤------> <== map({ .tc }) <== qw{people of earth}⏏;␤»
```

MadcapJake:
Working on a PR that should address my concerns.

awwaiid:
Bleh, sorry. I had it working fine but then copy/pasted examples as I
updated the step-by-step comments :(

PR #424, opened 2016-03-14
Add, fix and reword feed operator docs

MadcapJake:
Update some verbiage, fix a few spelling errors, fix the examples, and
add some additional examples.


MadcapJake:
Alright I think this is good to go.

PR #438, opened 2016-04-03
change example lib name to agree with the explanatory text

tbrowder:


zoffixznet:
I don't believe this is correct. The "lib" prefix gets added automatically.

MadcapJake:
On linux, it justs looks in `LD_LIBRARY_PATH` for whatever you put in that string—with the added `so` extension.

zoffixznet:
Camelia runs on SUSE and I also tested on Bodhi Linux (fork of Ubuntu). The 'lib' prefix gets added automatically in both cases. What distro are you running that doesn't get the `lib` added?

```
<ZoffixWin> m: use NativeCall; sub foo is native('something') {}; foo
<camelia> rakudo-moar 868d8b: OUTPUT«Cannot locate native library 'libsomething.so': libsomething.so: cannot open shared object file: No such file or directory␤  in method setup at /home/camelia/rakudo-m-inst-1/share/perl6/sources/24DD121B5B4774C04A7084827BFAD92199756E03 (NativeCall) line 2…»

zoffix@zoffix-VirtualBox:~$ perl6 -e 'use NativeCall; sub foo is native("something") {}; foo'
Cannot locate native library 'libsomething.so': libsomething.so: cannot open shared object file: No such file or directory
  in method setup at /home/zoffix/.rakudobrew/moar-nom/install/share/perl6/sources/24DD121B5B4774C04A7084827BFAD92199756E03 (NativeCall) line 279
  in method CALL-ME at /home/zoffix/.rakudobrew/moar-nom/install/share/perl6/sources/24DD121B5B4774C04A7084827BFAD92199756E03 (NativeCall) line 290
  in block <unit> at -e line 1

zoffix@zoffix-VirtualBox:~$ 
```

MadcapJake:
My memory fails me, you're right @zoffixznet.

zoffixznet:
Alright. I reverted this commit and also added clarification that `lib` gets added automatically.

PR #441, opened 2016-04-04
Added X.pod to Types for p6doc to more accurately reflect what the X

dha:
package does (or, more to the point, doesn't do)

Further note: As X turns up in the list resulting from 'p6doc -l' it should probably reflect the fact that X isn't actually useful by itself, rather than just that it's a builtin type, as currently occurs.

### DIFF
--- a/bin/p6doc
+++ b/bin/p6doc
@@ -109,6 +109,7 @@ multi sub MAIN(Bool :$l!) {
     my @modules;
     for @paths -> $dir {
         for dir($dir).sort -> $file {
+            next if $file.d;
             @modules.push: $file.basename.Str.subst('.pod','');
         }
     }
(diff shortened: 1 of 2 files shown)


dha:
I note that @stmuk has suggested that making 'p6doc -l' not list directories may be a better solution for this problem. So maybe that should be investigated before merging this?

dha:
As noted, having p6doc -l not list directories seems the better solution for making X needlessly turn up in the listing. This also removes the duplicate listing of Backtrace. Therefore, I've deleted X.pod and added a directory test to p6doc.

salortiz:
:+1:

stmuk:
This should be fixed in f80ef79c4b41833373cc4620cee1b0636c4df7c4

Thanks for the error report dha++

PR #448, opened 2016-04-11
Adjust doc for ord to indicate it returns *first* code point.

dha:


zoffixznet:
I see a bunch of commits/changes, but when I view "Files changed" I only see the last commit that added `first` to the `ord` method. Does this PR include all the changes you meant to include?

dha:
On Mon, Apr 11, 2016 at 03:19:24PM -0700, Zoffix Znet wrote:
> I see a bunch of commits/changes, but when I view "Files changed" I only see the last commit that added `first` to the `ord` method. Does this PR include all the changes you meant to include?

I used the "create pull request" button on github.com, and apparently it
included all the changes that I'd made recently, instead of just the one
I was actually trying to make a pull request for. Maybe I don't fully
understand how that button works. :-/

FWIW, I think there was discussion of this already on IRC.

dha

-- 
David H. Adler - <dha@pobox.com> - http://www.panix.com/~dha/
If you think something is impossible, don't disturb the person who is
doing it!	- Amar Bose

PR #452, opened 2016-04-17
add example of a user-entered tolerance; emphasize Test is a core module

tbrowder:
Add a note about Test being a core module.  Also add an example of a user-supplied tolerance for numerical testing.

I find these additions helpful. It isn't clear in current docs that Test is a core module, and thus a casual user has to search a while for its source.  A section on core modules would be helpful, perhaps referenced on the page where the external modules are listed: <https://modules.perl6.org/>.

Another place to reference them would be to add another category, "Core Modules," on this page: <https://doc.perl6.org/>

### DIFF
--- a/doc/Language/testing.pod
+++ b/doc/Language/testing.pod
@@ -26,7 +26,7 @@ project's base directory.
 A typical test file looks something like this:
 
     use v6;
-    use Test; # a core module
+    use Test;      # a Standard module included with Rakudo
     use lib 'lib';
 
     plan $num-tests;
@@ -145,7 +145,7 @@ The use can also input a desired tolerance, e.g., testing $some-number:
 
     my $spec-num = 1.2345;
     my $spec-tol = 0.1;
-    is-approx $some-number, $spec-num, $tol, "special tolerance check";
+    is-approx $some-number, $spec-num, $spec-tol, "special tolerance check";
  
 =head2 By structural comparison
 
--- a/doc/Language/testing.pod
+++ b/doc/Language/testing.pod
@@ -141,11 +141,15 @@ C<$description> of the test.
     my $eulers-constant-approx = 2.71828;
     is-approx $eulers-constant-approx, e, "approximate Euler's constant";
 
-The use can also input a desired tolerance, e.g., testing $some-number:
+=item X<is-approx($value, $expected, $tol, $description?)|is-approx>
 
-    my $spec-num = 1.2345;
-    my $spec-tol = 0.1;
-    is-approx $some-number, $spec-num, $spec-tol, "special tolerance check";
+Marks a test as passed if the C<$value> and C<$expected> numerical values
+are within C<$tol> of each other.  The function accepts an optional
+C<$description> of the test.  For example, test for an approximation of pi:
+
+    my $pi-approx = 3.1416;
+    my $pi-tol = 0.0001;
+    is-approx $pi-approx, pi, $pi-tol, "approximate ";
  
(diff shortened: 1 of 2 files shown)


pmichaud:
Can we please not call modules like Test.pm "core modules"?   The term "CORE" already has a very specific meaning in Perl 6 (and Test.pm is not a part of CORE).

S24 refers to Test.pm as being a "standard module" -- that phrasing would sit much better with me.

Pm

tbrowder:
On Sunday, April 17, 2016, Patrick R. Michaud <notifications@github.com>
wrote:

> Can we please not call modules like Test.pm "core modules"? The term
> "CORE" already has a very specific meaning in Perl 6 (and Test.pm is not a
> part of CORE).
>
> S24 refers to Test.pm as being a "standard module" -- that phrasing would
> sit much better with me.
>

Sure, that's fine with me. I was just using Perl 5 lingo I guess. The point
is that the casual user will not easily find the docs for the module.

-Tom

>

tbrowder:
On Sunday, April 17, 2016, Tom Browder <tom.browder@gmail.com> wrote:

>
>
> On Sunday, April 17, 2016, Patrick R. Michaud <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
>> Can we please not call modules like Test.pm "core modules"? The term
>> "CORE" already has a very sp
>>
>

> S24 refers to Test.pm as being a "standard module" -- that phrasing would
>> sit much better with me.
>>
>
> Sure, that's fine with me. I was just using Perl 5 lingo I guess. The
> point is that the casual user will not easily find the docs for the module.
>

I plan to prepare a pull request for the website soon that I think will
help the situation.

-Tom

>

stmuk:
Maybe we need a glossary of Perl 5 terms to Perl 6 terms?

doc/Language/testing.pod
```diff
@@ -141,6 +141,12 @@ C<$description> of the test.
     my $eulers-constant-approx = 2.71828;
     is-approx $eulers-constant-approx, e, "approximate Euler's constant";
 
+The use can also input a desired tolerance, e.g., testing $some-number:
+
+    my $spec-num = 1.2345;
+    my $spec-tol = 0.1;
+    is-approx $some-number, $spec-num, $tol, "special tolerance check";
```
MadcapJake: `Variable '$tol' is not declared. Did you mean '$spec-tol'?`

tbrowder: On Sun, Apr 17, 2016 at 1:18 PM, Jake Russo <notifications@github.com> wrote:
>
> In doc/Language/testing.pod:
...
> > +    my $spec-tol = 0.1;
> > +    is-approx $some-number, $spec-num, $tol, "special tolerance check";
>
> Variable '$tol' is not declared. Did you mean '$spec-tol'?


Yes!  Good catch--sorry.

-Tom

tbrowder:
I just updated the commits to reflect comments in this conversation.

tbrowder:
I could merge this request, but am reluctant to do so without a more experienced dev's approval.

zoffixznet:
@tbrowder It's not at all clear how to use the tolerance argument. What's the meaning of value `0.1`? I'd expand on that.

Also, there's a typo: `The use can also` should probably be **user**.

As for merging... unless it's a huge change that modifies a lot of content, just merge or commit directly to master. If there are any issues, they can be fixed in another commit :smile_cat:

PR #463, opened 2016-04-25
Document "Hash vs. Block" Gotcha

DabeDotCom:
I stubbed my toe porting some Perl 5 code, and thought others might benefit from having it documented:

    my %fruits = apple => 'McIntosh', pear => 'Bartlett';
    given (%fruits) {
      my $fails = { red => $_<apple>, green => $_<pear> };
      say $fails.WHAT;          # Block
      say $fails<red>;          # ERROR: Type Block does not support associative indexing
    }


zoffixznet:
:-1: on string interpolation suggestion. Just use `%( )` instead of curlies:

```perl6
    my $yummy = %( red => $_<apple>, green => $_<pear> );
```

ab5tract:
In general the examples do not feel very idiomatic to me, but it is a bit of a corner case to begin with so I guess awkward syntax is the name of the game.

I agree with @zoffixznet that the interpolation example is less important than the `%( )` constructor. Personally I would rather not see that kind of non-idiomatic example in `doc`.

ab5tract:
Sorry about that, fat-fingered the 'close and comment' button!

I wanted to also point out that this is also, at the root, caused by being too rushed to bother with naming the topic variable. I think that would be better to just point out that you can get around the entire issue with 'given %fruits -> %f { ... }'.

But all that said, I want to thank you for documenting this corner case! Other than what I've noted, I think this a great commit 👍

DabeDotCom:
Thanks for the feedback!

As Just Another Perl 5 --> Perl 6 switcher, there's obviously a lot to "unlearn".  Coming to grips with the various nuances of the language can be challenging (but that *is*, after all, the purpose of `5to6-nutshell` in the first place, right? «grin»)

Until one groks the various constructors and type conversion/coercion operators, it can be tricky to know when to use `{ }` and when to use `%( )` — especially when the former **DIDN'T** Do What I Meant, even though the docs made it sound like it should have.  So I was trying to stay as close to the original construct as possible, just to ease the transition...

Anyway, I'm definitely okay with providing better *right* examples; I'm still learning the new idioms, myself, hehe!  :-D

PS: FWIW, the original code I was porting was actually in a `map `block — kind of like below — but I thought `given` made for a tighter (albeit contrived) example...

    # Perl 5
    use ojo;
    my @people = map {
        my $query = {
            firstName => $_->{'firstName'},
            lastName => $_->{'lastName'},
        };
    
        p( 'https://example.com/api/personLocator', json => $query ).json;
    } @{ g( 'https://example.com/api/rosterList' ).json };


Does the following strike a better balance between informative and idiomatic?

    B<NOTE:> If one or more values reference the topic variable, C<$_>, the
    right-hand side of the assignment will be interpreted as a L<Block|/type/Block.html>,
    not a Hash:
    
        my @people = [
            { id => "1A", firstName => "Andy", lastName => "Adams" },
            { id => "2B", firstName => "Beth", lastName => "Burke" },
            # ...
        ];
    
        sub lookup (Hash $h) { #`(Do something...) $h }
    
        my @names = map {
            my $query = { name => "$_<firstName> $_<lastName>" };
            say $query.WHAT;   # Block
            say $query<name>;  # ERROR: Type Block does not support associative indexing
    
            lookup($query);    # Type check failed in binding $h; expected Hash but got Block
        }, @people;
    
    Instead, you should either assign to a hash directly:
    
            my %query = name => "$_<firstName> $_<lastName>";  # No brackets
    
    or use the C<%( )> hash constructor:
    
            my $query = %( name => "$_<firstName> $_<lastName>" );
            say $query.WHAT;   # Hash
            say $query<name>;  # Andy Adams, Beth Burke, etc.
    
    and/or avoid the issue altogether by explicitly naming the topic variable:
    
        my @names = map -> $person {
            my $query = { name => "$person<firstName> $person<lastName>" };
            say $query.WHAT;    # Hash
            say $query<name>;   # Andy Adams, Beth Burke, etc.
    
           lookup($query);
        }, @people;
    
    Even in the final case, however, specifying the C<%()> hash constructor
    is still the most idiomatic and unambiguous solution:
    
        my @names = @people.map: -> $person {
            lookup( %( name => "$person<firstName> $person<lastName>" ) );
        };

zoffixznet:
> it can be tricky to know when to use 
I've been burnt by the `$_` inside curlies for Hash problem too, so it's definitely should be documented :smile_cat: 

> Does the following strike a better balance
:+1: looks good to me.

ab5tract:
This is some great `5to6-nutshell` writing, right there! I like the change to stringify-ing names in the example content too.

ab5tract:
> when to use { } and when to use %( ) — especially when the former DIDN'T Do What I Meant, 

This is a good point. Do you remember specific examples? This kind of thing should probably be elaborated when discussing the differences in Context between P5 and P6.

DabeDotCom:
> This is some great 5to6-nutshell writing, right there!

Aww, shucks...  Flattery will get you _**every**where!_ (But thanks!)  :-D

And you know what they say: _"The Penalty for Good Work is **MORE** Work!"_ «grin»

PR #468, opened 2016-04-27
define the code point properties in the description

tbrowder:


AlexDaniel:
>  C<Zl> ((Separator, line)

Two ``((`` here ↑.

Also, it seems like **Lm** is not mentioned.

Example:
```perl
my $ʹ = 42; say $ʹ    # prints 42
say ‘ʹ’.uniprop;      # prints Lm
say ‘ʹ’.uniname;      # prints MODIFIER LETTER PRIME
```

tbrowder:
Alex, all your comments have been addressed--thanks!

AlexDaniel:
>  Any codepoint with the C<Lo> (Letter, other) property can be used as an alphabetic where case distinctions do not matter.

↑ Does anybody know when case matters? Or what kind of “case distinctions” is it talking about?
Should we group **Lm** with **Lu** and **Ll** or with **Lo**?

> (These aren't decimal digit characters, so can't be combined.)

↑ I think that it does not make it clear enough that unicode digit characters can be combined. Perhaps it could be rephrased a bit?

Or maybe we can throw in some examples? Like:
```perl6
say ١٩   # prints 19
```

tbrowder:
Okay with merging what we have at the moment?  I can try to add something for your consideration on the next pull request.

AlexDaniel:
Of course!

PR #470, opened 2016-04-27
add some examples; format some text for ease of maintenance

tbrowder:
I think we should add a note about best practice:  for portable Perl 6 code, use of non-ASCII characters for operators and mathematical operations is not recommended.

### DIFF
--- a/doc/Language/unicode_texas.pod
+++ b/doc/Language/unicode_texas.pod
@@ -4,24 +4,33 @@
 
 =SUBTITLE Unicode symbols and their Texas (ASCII) counterparts
 
-The following unicode symbols can be used in Perl 6 without needing to load
-any additional modules.  Please note that the Since column applies to the
-version of Perl the symbol was known to be available.  It is only mentioned
-if it is different from C<6.c>.
+The following unicode symbols can be used in Perl 6 without needing to
+load any additional modules.  Please note that the Since column
+applies to the version of Perl the symbol was known to be available.
+It is only mentioned if it is different from C<6.c>.
 
 =head1 Alphabetic Characters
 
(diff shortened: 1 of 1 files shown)


jnthn:
Portability in what sense?

zoffixznet:
>  for portable Perl 6 code, use of non-ASCII characters for operators and mathematical operations is not recommended.

:-1: from me on that. I feel currently Perl 6 is the Unicode leader, really. Telling people to abstain from Unicode ops in the official docs is kinda backwards. Also, I'm unsure what you mean by portable Perl 6. The way you worded it, it implies if one uses unicode ops, their code won't run on some OS. Is that actually the case?

I think the only issue is font support, which I think is pretty self-evident, and will only improve in as time goes on.

jonathanstowe:
Actually no, there are many reasons why it might be better not to use unicode operators, font support is only one, others may include not being able to actually type them (or no consistent way to type them,) limited support for multi-byte characters in some terminal software and so on and so forth. However I don't think that we should discourage their use, rather encourage people to think before using them.

tbrowder:
On Wednesday, April 27, 2016, Jonathan Stowe <notifications@github.com>
wrote:

> Actually no, there are many reasons why it might be better not to use
> unicode operators, font support is only one, others may include not being
> able to actually type them (or no consistent way to type them,) limited
> support for multi-byte characters in some terminal software and so on and
> so forth. However I don't think that we should discourage their use, rather
> encourage people to think before using them.
>
Maybe I'll propose some words to that effect.  I know the last few days
I've had a time finding a reliable editor (and browser!) that will enable
easy Unicode entry as well as correct rendering.  The only one that I could
reliably enter 0x1d452 with (and see it afterward) was Emacs.

BTW, can anyone explain the extra white space associated with 0x1d452 in
some renderings?
>
>

PR #474, opened 2016-04-29
Add missing matching "/" at the end of regex

vadz:
Trivial change to avoid unterminated regex in an example.


zoffixznet:
Thanks a lot!

vadz:
Thanks to you for writing blog post which brought me to this page in the first place!

PR #489, opened 2016-05-03
Mostly whitespace/syntax cleanup

MasterDuke17:
One or two actual changes to code, but no operational changes.

### DIFF
--- a/bin/p6doc
+++ b/bin/p6doc
@@ -150,7 +150,7 @@ sub disambiguate-f-search($docee, %data) {
     }
 
     my $final-docee;
-    my $total-found = %found.values.map( +* ).sum;
+    my $total-found = %found.values.map( *.elems ).sum;
     if ! $total-found {
         say "No documentation found for a routine named '$docee'";
         exit 2;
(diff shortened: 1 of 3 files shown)


AlexDaniel:
It is great. But this:
```diff
-my $total-found = [+] %found.values.map( *.elems );
+my $total-found = %found.values.map( +* ).sum;
```
is more like code obfuscation. ``.sum`` is fine, but ``+*`` instead of ``*.elems`` is probably not a good idea. I think that it is worth changing it back to ``*.elems``.

MasterDuke17:
Yeah. I think I was trying to make it consistent with the rest of the code, but it looks like a roughly even split between `+<array_goes_here>` and `<array_goes_here>.elems`. Feel free to make that change, unless you'd rather I do it?

zoffixznet:
Rather you do it :) The documentation may be perused by novices so it's best to avoid clever code, even if it is more concise.

MasterDuke17:
That was ugly, but things seem ok now.

jonathanstowe:
Thanks!

PR #506, opened 2016-05-08
Adding docs for async LAST/QUIT phasers

zostay:
And also the same basic functionality in .tap.


jonathanstowe:
Works for me :) Cheers!

PR #509, opened 2016-05-09
Added samewith

molecules:
Based this on Brad Gilbert's example here: http://stackoverflow.com/a/37064186/215487. I recommend further elaboration by someone who knows more about it than I do.


jonathanstowe:
Cheers!

PR #513, opened 2016-05-13
@opts compilation error fix

Coleoid:
Lifted @opts to an outer scope so it could be seen in another spot it was used.


jonathanstowe:
It passes :)

PR #519, opened 2016-05-17
correct typos and grammar

tbrowder:


doc/Language/functions.pod
```diff
@@ -724,9 +724,9 @@ currently (2015.02) only implements them for subroutine parameters.
 
 =head1 C<sub MAIN>
 
-The sub with the special name C<MAIN> is executed after all relevant phaser
-and it's signature is the means by which command line arguments can be parsed.
-Multi methods are supported and a usage method is automatically generated and
+The sub with the special name C<MAIN> is executed after all relevant phases
```
zoffixznet: FWIW, the `phaser` is the correct word (just has to be pluralized to `phasers`). It's talking about phasers like BEGIN, INT, etc

tbrowder: On Tuesday, May 17, 2016, Zoffix Znet <notifications@github.com> wrote:

> In doc/Language/functions.pod
> <https://github.com/perl6/doc/pull/519#discussion_r63584053>:
>
> > @@ -724,9 +724,9 @@ currently (2015.02) only implements them for subroutine parameters.
> >
> >  =head1 C<sub MAIN>
> >
> > -The sub with the special name C<MAIN> is executed after all relevant phaser
> > -and it's signature is the means by which command line arguments can be parsed.
> > -Multi methods are supported and a usage method is automatically generated and
> > +The sub with the special name C<MAIN> is executed after all relevant phases
>
> FWIW, the phaser is the correct word (just has to be pluralized to phasers).
> It's talking about phasers like BEGIN, INT, etc
>

Thanks, that's comng back to me now.  (But the "its" vs. "it's" is still
valid--I've lost control of my git clone of  doc!!)

-Tom
>
>

PR #523, opened 2016-05-19
fix typo

tbrowder:


AlexDaniel:
> after all relevant phaser

phaser**s**?
Perhaps you can add one more commit :)

tbrowder:
changed back to phasers...

PR #524, opened 2016-05-19
replace two non-ascii operators with Texas equivalents; correct one typo

tbrowder:


vendethiel:
Why ?

AlexDaniel:
I agree with @vendethiel. ``>>`` instead of ``»`` is probably OK (although I still can't see why change it), but ``(<=)`` instead of ``⊆`` is not an improvement at all.

coke:
We don't have a style guide for the docs for this item, I don't think, but my preference is to use the unicode variants when available.

jonathanstowe:
FWIW I can't actually read ≤ in the standard font in my browser.  Though it is typable.

tbrowder:
Until the whole world has a completely unicode-capable, no-brainer system, I think public code should use the Texas variants.

tbrowder:
BTW, I think the Texas variant of the "(sub|method)»" fragment in the code seems NOT to be either of:

  "(sub|method)>>" 

or

  "(sub|method) '>>'"

so what is the correct way to Texify that?

AlexDaniel:
``»`` is a commonly used quotation mark in many languages. It is also part of ISO/IEC 8859-1. Although I agree that avoiding some characters may be beneficial, I can't see any reason to believe that this particular character may cause any problems.

tbrowder:
I'm just suggesting to stick with ASCII if you want to maximize participation in a public project.  If we all used the same editors (and keyboards), then maybe go whole hog.  Otherwise, either specify acceptable editor and keyboard requirements, or relax to ASCII.  Unicode can quickly lead to a Tower of Babel situation: it's wonderful for documents made primarily for reading in one's own language (or for mathematical texts), but it's not so wonderful for source code.

coke:
I respectfully disagree with the "stick with ASCII' line; Perl 6 has a history of pushing the boundaries on unicode. I think our primary examples should be with the unicode.

I would be 100% in favor, however, of having texas-style variants a click away for our HTML doc readers. I'll open a ticket for that.

tbrowder:
Coke, I hate to beat a dead horse, but I'm talking about source code, not documentation.  Also, if we ARE going to use the unicode operators in the source code, then we should use them all the time, i.e., replace all Texas operators with unicode (as in the htmlify.p6 I tried to change).

AlexDaniel:
I'm all for changing all texas operators to proper unicode ones.

coke:
My argument is the same for Perl 6 code in the perl6 repositories.

coke:
Stance is a better word than argument, regards

tbrowder:
One more comment: It would be helpful to document somewhere a list of free fonts that cover ALL the Perl 6 unicode operators.  I, for one, am trying to update my system fonts to ensure I have at least one monospaced font that can do that, but I am still looking.

AlexDaniel:
Another possibility is to create a new font that contains only perl 6 ops.

@tbrowder Try unifont. It is ugly, but it should contain almost everything.

tbrowder:
I complained earlier about unicode handling in Padre but it looks as if it's a bad choice of font.  The only character in the unicode/Texas list that does not show correctly is U+1D452 (the mathematical italic e).  Can anyone name the font they use that displays that correctly?

AlexDaniel:
@tbrowder Sometimes it is hard to tell which font exactly is rendering some particular character, but I think that [freefont](https://www.gnu.org/software/freefont/) has it.

tbrowder:
Thanks, that font looks like it should work, but it doesn't show as available in Padre--may be my system fonts--I'll check...

PR #535, opened 2016-05-21
Fix printf arguments

titsuki:
I think that there is no reason to use the list for the arguments of the printf subroutine.
It will confuse those who read the following section.
https://doc.perl6.org/type/Signature#Slurpy_%28A.K.A._Variadic%29_Parameters
And they will think that "The printf subroutine in Perl 6 is different from C-style printf".


zoffixznet:
Thanks a lot!

titsuki:
It's my pleasure :)

PR #555, opened 2016-05-28
Fix the example of the open method

### LINKED ISSUE #554: Example of the open method seems wrong
See: http://doc.perl6.org/type/IO#sub_open

This example is not work.
```
my $fh = open("test", :w);
spurt $fh, "stuff\n";
spurt $fh, "more stuff\n";
$fh.seek(0, 0);   # return to the start of the file
$fh.get();        # Reading from filehandle failed: bad file descriptor
```

titsuki:
Fix #554


zoffixznet:
Thanks.

PR #565, opened 2016-06-08
Add documentation for the new IO.mode method

MasterDuke17:
This requires Rakudo PR #773, which implements RT #125489


labster:
Link: rakudo/rakudo#773

I'd add: "The Str part of the result is the octal representation of the file permission, like the form accepted by the chmod(1) utility."  Of course, Perl 6 chmod() doesn't accept strings like this gracefully -- making that work is left as a exercise to the reader :)

PR #567, opened 2016-06-08
The Wombles were a British children's program

jsoref:
Include a reference so that everyone else is not totally befuddled.


stmuk:
I do wonder whether this might be better in the design glossary

http://design.perl6.org/S99.html

(even although design. isn't maintained much!)

labster:
I don't think this improves the signal to noise ratio of the documentation.  The job here is to document the language, not to document popular culture.  I'd recommend against merging.

AlexDaniel:
You can always use your favorite search engine to find such information if you're interested.

I'd actually vote for changing the example. Example that needs more explanation is probably a bad example.

AlexDaniel:
OK, let's close this. However, I have created issue https://github.com/perl6/doc/issues/577 because there is a problem.

PR #568, opened 2016-06-08
add links to design docs

jsoref:
I don't have perl6 handy, so I have no idea if this builds.
I also am not remotely confident that the links are correct.

I do know that as a reader, I expected links for these items.

### DIFF
--- a/doc/Language/5to6-nutshell.pod
+++ b/doc/Language/5to6-nutshell.pod
@@ -504,7 +504,8 @@ See L<S32/Containers|https://design.perl6.org/S32/Containers.html>
 
 =head1 Operators
 
-See S03-operators for full details on all operators.
+See L<S03/Operators|https://design.perl6.org/S03.html#Overview>
+for full details on all operators.
 
 Unchanged:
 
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Perhaps it's time to move ``design.perl6.org`` to ``speculations.perl6.org``. A big red warning is not enough.

First of all, we have a glossary in the docs: http://doc.perl6.org/language/glossary
**NYI** is not there, so please add it.

As for the other links, I think that they should be removed. Linking to the whole ``S32/Containers`` from a 5to6 page is just too much. ``S19#Removed_Syntactic_Features`` probably deserves a whole new page in the docs.

AlexDaniel:
As we now have “NYI” in the [glossary](http://doc.perl6.org/language/glossary#NYI), this pull request has to be reworked.

Please separates changes associated with links to the specs into something different, perhaps an issue? Maybe we should have a list of all references to the specs so that we can get rid of them (or leave them if there is a good enough justification for some particular ones).

AlexDaniel:
OK, I have added a proper link for NYI myself: https://github.com/perl6/doc/commit/5e8cf95c8809c7064c7c78e9c9035910a93e278a
And cherry-picked and squashed your commits: https://github.com/perl6/doc/commit/89d368d3fa3f4ddf5f4d441614e78b496dc1dd2e

I still think that we should get rid of references to speculations, but for now we can have proper links, indeed. This can be closed now.

Thank you for your work!

PR #573, opened 2016-06-09
htmlify: do not escape :: in site local links

jsoref:
Before this change,
http://doc.perl6.org/type/Pod::Item

```
Pod::Item inherits from class Pod::Block, which provides the following methods:
```
*Pod::Block* links to http://doc.perl6.org/type/Pod%3A%3ABlock

Ideally (not actually tested w/ a build, although *star-m:* approves of the general code) this should result in links to http://doc.perl6.org/type/Pod::Block

### DIFF
--- a/htmlify.p6
+++ b/htmlify.p6
@@ -250,6 +250,10 @@ sub process-pod-source(:$kind, :$pod, :$filename, :$pod-is-complete) {
 
 # XXX: Generalize
 multi write-type-source($doc) {
+    sub href_escape($ref) {
+        # only valid for things preceded by a protocol, slash, or hash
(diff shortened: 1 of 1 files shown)


moritz:
Everybody, please only merge this after verifying that it doesn't break any links.

zoffixznet:
Maybe @AlexDaniel can run their 404-finding program on the docs generated by the code in the fork.

AlexDaniel:
Sure, I'll try that today.

AlexDaniel:
OK finally got my hands on it. Well, I don't see any problems. In fact, looking at the code, I'm not sure how can it break anything.

Issue #484, opened 2016-05-01
｢.nl-in｣ and ｢.nl-out｣ show as routines, not methods


AlexDaniel:
These are mentioned in some places (some listed below), but they are not searchable.
- http://doc.perl6.org/type/IO::Handle#method_lines
- http://doc.perl6.org/type/IO::Handle#method_print
- http://doc.perl6.org/routine/put#role_IO
- http://doc.perl6.org/routine/say#role_IO

dogbert17:
This seems to be working, am I missing something?

AlexDaniel:
Yes! It guess @zoffixznet++ made it searchable at some point.

However, right now ``nl-in`` and ``nl-out`` are listed as a “Routine”, which is wrong.

zoffixznet:
> Yes! I guess @zoffixznet++ made it searchable at some point.

They were undocumented at the time and now they are.

> However, right now nl-in and nl-out are listed as a “Routine”, which is wrong.

Unsure why. They're both documented as methods: https://docs.perl6.org/routine/nl-out https://docs.perl6.org/routine/nl-in

Also, #1331 is not entirely unrelated here.

PR #575, opened 2016-06-09
Link fixes

jsoref:


doc/Language/objects.pod
```diff
@@ -65,7 +65,7 @@ return value of a method call. This is how read-writable attributes to
 objects are used:
 
 =for code :allow<L>
-$*IN.L<nl-in> = "\r\n";
+$*IN.C<nl-in> = "\r\n";
```
zoffixznet: Would the `:allow` line on line 67 need to be changed to `<C>` as well?

AlexDaniel: @zoffixznet Absolutely.

I'm not sure why it has to be changed though. ``nl-in`` is not documented yet (https://github.com/perl6/doc/issues/484), but it does not mean that we should not have a link like this.

AlexDaniel:
OK, so I cherry-picked everything besides the ``L<> → C<>`` change.

It also turned out that you had a commit that fixes wikipedia links, but I didn't notice it and did that myself… meh! We should merge such pull requests faster so that this kind of things do not happen.

PR #576, opened 2016-06-09
add 404 existence guards for make-run server

jsoref:


zoffixznet:
Does this PR bring any other benefits other than saving one redirect on 404s?

zoffixznet:
I [asked on IRC](http://irclog.perlgeek.de/perl6/2016-06-09#i_12637289), and it seems this app is not in use on the live site, but simply is a convenience feature for local use.

Since Mojolicious's static handler will already handle the 404's for non-existent files, there's no benefit for adding explicit check in this case.

I'll close this PR. If there are some benefits I missed, please reopen. Thanks.

jsoref:
I'm not sure how to reopen.

When using a link checker, a request for http://localhost:.../robots.txt gets sent to /robots.txt.html which gets sense to /robots.txt.html.html, etc. Thankfully my link checker gives up after 10 attempts, but it's pretty annoying.

coke:
If the error doesn't happen on the production site it's probably not worth fixing

jsoref:
I tried visiting http://localhost:3000/language/ and ended up at: http://localhost:3000/language/.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html.html

It's making trying to use the thing not worth my time. Hopefully you can see that I can make valuable contributions. But if the tools balk me, I'll leave.

zoffixznet:
Thanks for pointing out that issue. It's fixed in 9b72be06e6db0550bb7e7fe812d706c8d1d0459a. Try it out. If you're still experiencing problems, please open an Issue.

P.S.:

> I'm not sure how to reopen.

You just click this button:
png

jsoref:
image

PR #585, opened 2016-06-11
document the third is-approx signature

tbrowder:
try to improves description of all three versions

### DIFF
--- a/doc/Language/testing.pod
+++ b/doc/Language/testing.pod
@@ -133,24 +133,52 @@ of the test.
 
 =head2 By approximate numeric comparison
 
+Approximate numeric comparisons maye be tested by the C<is-approx>
+function which has three signatures. Note that in all three versions,
+the final C<$description> argument is optional (but recommended).
+
 =item X<is-approx($value, $expected, $description?)|is-approx>
 
-Marks a test as passed if the C<$value> and C<$expected> numerical values
-are within C<1e-5> of each other.  The function accepts an optional
-C<$description> of the test.
+The default version of the function calculates a tolerance and a
+comparison test in this way:
+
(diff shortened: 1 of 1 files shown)


jonathanstowe:
Cheers

zoffixznet:
@tbrowder, as I wrote on IRC, I really want to revert this, so we don't show extra functionality to the users until it's solidified and debugged (I'll submit a PR to Rakudo this weekend):

```
<Zoffix> tbrowder, would you think I'm a total dick were I to revert this? https://github.com/perl6/doc/pull/585  tbrowder, there are actually several bugs with that subroutine and it's unpollished. There were several attempts to fix it but they kinda drowned in decision paralysis. I finally wrote down a plan for what it's supposed to work like and Larry gave me a plausible excuse to at least submit a PR. With the documentation and tests lacking for the 
<Zoffix> "extra" functionality, it'd be easier to bikeshed what the sub should do, which is why I want to exclude it from the docs for now, lest someone actually uses it. Lastly, the rel/abs tol version in your docs is incorrect. Those are named args, not positionals.
````

tbrowder:
On Saturday, June 11, 2016, Zoffix Znet <notifications@github.com> wrote:

> @tbrowder <https://github.com/tbrowder>, as I wrote on IRC, I really want
> to revert this, so we don't show extra functionality to the users until
> it's solidified and debugged (I'll submit a PR to Rakudo this weekend):
>

Please revert it, Zoffix.  As I said later on irc, my interest in the funcs
comes from wanting to use them in a module I'm working on (Geo::Ellipsoid,
most wanted Perl 6).  And my own time line is not real fast...

Best,

-Tom

P.S. Good catch on errors.
>
>

PR #593, opened 2016-06-13
added blurb about python-dev in CONTRIBUTING

stevieb9:
CONTRIBUTING was missing the need to install Python dev components if they aren't already installed, before installing Inline::Python


zoffixznet:
Thanks!

PR #596, opened 2016-06-15
add NaN type

jsoref:
I could push this directly, but I'd like some eyes.

### DIFF
--- a/doc/Type/Complex.pod
+++ b/doc/Type/Complex.pod
@@ -86,7 +86,7 @@ Usage:
 
     COMPLEX.isNaN
 
-Returns true if the real or imaginary part is C<NaN> (not a number).
+Returns true if the real or imaginary part is L<C<NaN>> (not a number).
 
 =head2 method polar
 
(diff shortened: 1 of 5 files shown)


jonathanstowe:
Unless I'm missing something, NaN isn't actually an enum but a "special" value of Num:

```
[jonathan@coriolanus perl6]$ perl6 -e 'enum Foo <bar baz>; say baz.HOW'
Perl6::Metamodel::EnumHOW.new
[jonathan@coriolanus perl6]$ perl6 -e 'say NaN.HOW'
Perl6::Metamodel::ClassHOW.new
```

It's actually defined as constant of type Num with a value obtained by an NQP op.

jsoref:
@jonathanstowe : How do I write that using pod?

jonathanstowe:
Well, I think it probably wants to be a section in the Num page (along with Inf) as a special value of Num rather than as a separate type page.

The pattern for this would be something like that for Empty (which a special value of Slip) as in http://doc.perl6.org/type/Slip#Constants, this allows it to be searched and linked/.

but it's definitely a constant rather than an enum or a separate type.

gfldex:
NaN is not a type (you will confuse anyone who does type checks) and it is already doced and in the index at https://doc.perl6.org/type/Num#NaN .

PR #597, opened 2016-06-15
use pod link instead of url for 5to6-nutshell

jsoref:
I'm not sure this works. If someone can confirm it does, I'd appreciate it.


AlexDaniel:
5to6-nutshell**6**? Looks like a typo. ``/language/5to6-nutshell#Whitespace`` is supposed to work I think.

parabolize:
If you remove the trailing 6 typo AlexDaniel is talking about it will work:
`L<5to6-nutshell#Whitespace>`
You could also do this:
`See L<Whitespace|/language/5to6-nutshell#Whitespace> for details.`

PR #598, opened 2016-06-15
Corrections in Capture doc

parabolize:
A formatting correction and a code example correction.

I'm new to Perl 6, Pods and git so let me know if I could do any of these things better.


moritz:
Thanks @parabolize. I've invited you to the perl6 github organisation, so that you can push directly (or merge yourself) in the future, if you want.

Also while you are at Capture.pod, I've noticed a small grammar error: `Capture with one positional parts` should be "part". Maybe you want to fix that too.

PR #603, opened 2016-06-18
Perl6 docs

tbrowder:


zoffixznet:
Thanks a lot!

AlexDaniel:
It looks like there are a couple of formatting issues. [See yourself](http://doc.perl6.org/programs/00-running). But that's okay, these can be fixed as we go. Generally I really like the idea of having these pages.

tbrowder:
On Sat, Jun 18, 2016 at 12:49 PM, Zoffix Znet <notifications@github.com> wrote:
> Thanks a lot!

My pleasure--sorry it took so long.  As you can see, there are issues,
but I hope experts can cure them.

AlexDaniel:
Looking at it again, are we sure that we want a separate “Programs” tab? Can't we have these things as part of the “Language” page?

tbrowder:
On Sat, Jun 18, 2016 at 6:59 PM, Aleks-Daniel Jakimenko-Aleksejev
<notifications@github.com> wrote:
> Looking at it again, are we sure that we want a separate “Programs” tab?
> Can't we have these things as part of the “Language” page?

I think Zoffix agrees that the separate tab is friendlier to new
folks.  And executables don't really fit in Language.

zoffixznet:
Yeah, there's plenty of space for it.`Language` is overcrowded as it is. Maybe some things can be moved out of it and into `Programs`, like http://docs.perl6.org/language/modules-extra

PR #630, opened 2016-06-24
remove space between sub/method names and opening paren

tbrowder:


AlexDaniel:
For those wondering why: calling a function with a space before parens is problematic. If it is so, why not declare the function the same way you are going to call it? The hope here is that it will make it more straightforward for beginners.

The only weird thing with this commit is this text ``(Notice the space before the opening '(' which is normally not a good practice.)``, what space exactly is it talking about?

tbrowder:
AlexDaniel, I have removed that not since it is no longer valid, thanks for seeing that.

Issue #631, opened 2016-06-24
Duplicate id= in docs


jsoref:
I'm not sure if perl6/Pod-To-HTML#14 is really its fault or perl6/doc's fault (I think it's shared)

This code generates multiple `id=` for a given thing. `id=` needs to be unique within a document...
```perl6
multi sub node2html(Pod::Heading $node) {
    Debug { note colored("Heading node2html called for ", "bold") ~ $node.gist };
    my $lvl = min($node.level, 6); #= HTML only has 6 levels of numbered headings
    my %escaped = (
        id => escape_id(node2rawtext($node.contents)),
        html => node2inline($node.contents),
    );

    %escaped<uri> = uri_escape %escaped<id>;

    @indexes.push: Pair.new(key => $lvl, value => %escaped);

    return sprintf('<h%d id="%s">', $lvl, %escaped<id>)
                ~ qq[<a class="u" href="#___top" title="go to top of document">]
                    ~ %escaped<html>
                ~ qq[</a>]
            ~ qq[</h{$lvl}>\n];
}
```

https://doc.perl6.org/routine/new.html
```html
  <li class="indexItem indexItem1"><a href="#class_Thread">class Thread</a></li>
  <ol class="indexList indexList2">
    <li class="indexItem indexItem2"><a href="#method_new"><a href="/routine/new#class_Thread">method new</a></a></li>
  </ol>
  <li class="indexItem indexItem1"><a href="#class_Version">class Version</a></li>
  <ol class="indexList indexList2">
    <li class="indexItem indexItem2"><a href="#method_new"><a href="/routine/new#class_Version">method new</a></a></li>
  </ol>
</ol>
</nav>
<p>Documentation for routine new, assembled from the following types:</p>
<h1 id="class_Mu"><a class="u" href="#___top" title="go to top of document">class Mu</a></h1>
<p>From <a href="/type/Mu#new">Mu</a></p>
<h2 id="method_new"><a class="u" href="#___top" title="go to top of document"><a href="/routine/new#class_Mu">method new</a></a></h2>
<pre>multi method new(*%attrinit)</pre>
<p>Default method for constructing (create + initialize) new objects of a class. This method expects only named arguments which are then used to initialize attributes w
ith accessors of the same name.</p>
<p>Classes may provide their own <code>new</code> method to override this default.</p>
<p><code>new</code> triggers an object construction mechanism that calls submethods named <code>BUILD</code> in each class of an inheritance hierarchy, if they exist. S
[...]

Issue #561, opened 2016-05-31
Lots of broken links, invalid HTML and so on (checklink)


AlexDaniel:
I've used [checklink](http://search.cpan.org/dist/W3C-LinkChecker/bin/checklink.pod) utility on doc.perl6.org. **[Here is the output](https://gist.github.com/AlexDaniel/fc6cb8d6f66516c837268d43ecc47c33)**. I believe that all of these issues have to be fixed.

Some of the errors are repeated many times, that's why the output is so large.
## Current status

| Number of lines in the output | Date |
| --- | --- |
| 126591 | 2016-05-31 |
| 128555 | 2016-06-03 |
| 132005 | 2016-06-09 |
| 128771 | 2016-06-11 |
| 126707 | 2016-06-12 |
| 130491 | 2016-06-16 |
| 134073 | 2016-06-23 |
| 131489 | 2016-06-25 |
| 134542 | 2016-07-06 |
| 35561 | 2016-07-13 |
| 35005 | 2016-08-01 |
| 33926 | 2016-08-02 |
| 20540 | 2016-08-03 |
| 17432 | 2016-08-04 |
| 17002 | 2016-08-05 |
| 8253 | 2016-08-06 |
| 8615 | 2016-08-10 |
| 7374 | 2016-11-12 |
| 8157 | 2016-12-20 |
| 208521 | 2017-06-30 |
| 207664 | 2017-07-01 |
| 211899 | 2017-11-04 |
| 213547 | 2018-01-05 |
| 184904 | 2018-03-03 |
| 55393 | 2018-05-05 |
| 54232 | 2018-05-06 |
| 57626 | 2018-11-04 |
| 39726 | 2020-07-26 |

![checklink_graph](https://user-images.githubusercontent.com/5507503/88478516-a7a30200-cf51-11ea-8484-f4442860f356.png)

zoffixznet:
Are you sure you ran the program correctly? In the first item, `http://doc.perl6.org/type/X::TypeCheck::Splice`, the output says 

```
x:/js/main.js	
  Line: 703
  Code: 501 Protocol scheme 'x' is not supported
 To do: Could not check this link: method not implemented or scheme not
	supported.
```

Yet, no scheme `x:` is in use on that line: `<script type="text/javascript" src="/js/main.js"></script>`

I checked a couple of other errors, and they similarly are false positives.

zoffixznet:
[...]

PR #632, opened 2016-06-24
work around duplicate id= anchors

jsoref:
This should address  #631.

I'd rather someone else consider it (the output does change, it becomes wordier).

### DIFF
--- a/htmlify.p6
+++ b/htmlify.p6
@@ -552,7 +552,7 @@ sub find-definitions(:$pod, :$origin, :$min-level = -1, :$url) {
 
             my $new-head = Pod::Heading.new(
                 :level(@pod-section[$i].level),
-                :contents[pod-link "$subkinds $name",
+                :contents[pod-link "$origin.name() $subkinds $name",
                     $created.url ~ "#$origin.human-kind() $origin.name()".subst(:g, /\s+/, '_')
                 ]
             );


AlexDaniel:
It has been a while since you submitted this pull request. What do you think about it now?

AlexDaniel:
OK, so if one takes a look at https://github.com/perl6/doc/issues/561, it is easy to notice that duplicate anchors is the last big issue.

Therefore it is probably time to do something with this.

There are basically two options:

1. :+1: “Yes, let's fix the docs now!”
1. :-1: “No, we should fix it in ``Pod::To::HTML`` instead”… (or maybe change this pull request slightly?)

Below you can see before/after screenshots. It does become wordier, and it does introduce a bit of redundant information (the name of the class is displayed at least twice).

Also, “Any method” sounds really weird. I don't like it at all. Especially note these real cases:
* “Any method any”
* “Routine method multi”
* “Cool routine roots”
* “Any method end”

A bit funny sometimes, but we can probably live with that. Still, I think beginners will find it hard to read (or not? Am I the only one having troubles?), can we do something about it?

I am going to merge this if nobody dares to oppose, so **please speak up**.

By the way, it affects **all** pages. And that's for a good reason! For example, take a look at [IO::Handle#method_say](https://docs.perl6.org/type/IO$COLON$COLONHandle#method_say) – there are two methods “say” on that page, but the second link is broken because of duplicate ``id``s. So yeah, it makes sense.

----------

# Before
screenshot before

# After
screenshot after

coke:
Perhaps instead of

Any method eager

use

(Any) method eager

Altai-man:
The workaround is really small. When we'll fix our `Html::To::Pod`, it'll be super-easy to revert this back. So, generally, I'm +1 about this. But it is not a final solution.

We still have dead links at the `routine/` pages, it's my fault(or maybe not, since this part was completely broken, but just a little bit broken now because of me). I'll prepare a complete patch quickly(till today evening), so please, wait until it.

PR #633, opened 2016-06-24
add (passing) trailing whitespace test

coke:
branched from the tab-test branch. Adds a new test file to check for trailing whitespace.


AlexDaniel:
I think that we should give it a try. If nobody is going to say something against this then I'll work on merging this in the next few days.

coke:
All tests should pass now, also merged in all changes from master as of a minute ago.

jonathanstowe:
Is there any reason not to merge this?  It looks fine.

AlexDaniel:
@jonathanstowe no and yes. There are people who think that trailing whitespace is irrelevant, and therefore we should just ignore it. Such tests do exactly the opposite (it makes CI systems scream when the issue is not that big).

For that reason it was a good idea to give it a little bit of time, just to see if somebody has to say something about it. Since nobody did, you are right, there's no reason not to merge it.

``tws-test`` branch is now merged into master. This pull request was about merging it into ``tab-test``, but that is not very meaningful, so I am closing this.

PR #635, opened 2016-06-25
add section on func call problems

tbrowder:


doc/Language/traps.pod
```diff
@@ -248,6 +248,60 @@ The loose precedence of C<..> can lead to some errors.  It is usually best to pa
...
+  sub    foo(...) {} # where ... represent zero or more parameters
+  method foo(...) {} 
+
+but they may also may be declared with whitespace before the opening parenthesis:
+
+  sub    foo (...) {} # where ... represent zero or more parameters
+  method foo (...) {} 
+ 
+However, such whitespace can result in errors with certain argument
+lists when calling them using the function call form (see below).
```
moritz: I think this is factually wrong. In declarations, whitespace between the routine name and the signature is insignificant. If you disagree, please provide an example where it makes a difference.

moritz:
I'm sorry, but as is, I think this pull requests adds more confusion than it clears up. The rule is pretty simple: If you call a subroutine, the parenthesis must come directly after the subroutine name, without any whitespace. 

If you add a whitespace, the things inside the parenthesis become a list, and thus a single argument. Which isn't another rule, but merely normal semantics of `circumfix:<( )>`.

moritz:
I'm sorry, but as is, I think this pull request adds more confusion than it clears up.

The rule is pretty simple: If you call a function with parenthesis, no whitespace is allowed between the function name and the opening parenthesis.

If you do add witespace, then the parenthesis aren't part of the subroutine call, but a normal `circumfix:<( )>`, that is, constructing a list.

But when I read the new docs, this doesn't become clear, I just get a vague sense of "be careful with whitespace" without understanding why, or the mental model behind it.

doc/Language/traps.pod
tbrowder: Moritz, I agree, I didn't mean to say the declaration was wrong.  I confused a proposed "best practice" on the decl versus problems with the call.  Would a bit of re-do work, or should I just forget this.

tbrowder:
Working on another version.

Issue #545, opened 2016-05-24
Precomp not working for htmlify.p6 with doc/HomePage.pod


tbrowder:
I have been modifying doc/HomePage.pod but nothing in html generation from that file changed until I removed the precomp directory.  But changes in all other files seemed to result in changed html.

PR #636, opened 2016-06-25
Fix for issue https://github.com/perl6/doc/issues/545

dogbert17:
In order to speed up the doc generation the program precompiles pod files
into the directory precompiled. The code in the sub extract-pod loads the
precompiled files but if it detects that a pod file has been changed it
calls the method precompile in CompUnit::PrecompilationRepository. The
problem is that the precompile method checks if the file, that it is being
asked to recompile, exists in the repo, and if it does it won't recompile
it unless the named parameter :force is set to True. My suggested
change is that we set :force => True when calling the precompile method.

### DIFF
--- a/htmlify.p6
+++ b/htmlify.p6
@@ -186,7 +186,7 @@ sub extract-pod(IO() $file) {
 
     if not $handle {
         # precompile it
-        $precomp.precompile($file, $id);
+        $precomp.precompile($file, $id, force => True);
         $handle = $precomp.load($id)[0];
     }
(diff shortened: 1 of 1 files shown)


zoffixznet:
FWIW, a more canonical way to write that is `:force`, as in:

```perl6
    $precomp.precompile($file, $id, :force);
```

dogbert17:
Fixed :)

zoffixznet:
Thanks!

PR #637, opened 2016-06-25
add info on space probs with function call format (second attempt)

tbrowder:


doc/Language/traps.pod
```diff
@@ -248,6 +248,53 @@ The loose precedence of C<..> can lead to some errors.  It is usually best to pa
...
+the function call format.
+
+Note that there are clever ways to eliminate the error with the
+function call format and the space, but that is bordering on hackery
+and will not be mentioned here.  For more information, consult
+L<Functions|/language/functions#Functions>.
+
+Finally, note that, currently, when declaring the functions whitespace
+may be used between a function or method name and the parentheses
+surrounding the parameter list without problems (although that may
```
zoffixznet:
```
[19:03:44] <Zoffix> tbrowder, what are the actual chances of [space being problematic in the future] happening?
[19:03:49] <Zoffix> Seems this doesn't belong in the docs.
[19:06:16] <Zoffix> Such a change would literally break every piece of Perl 6 code I ever wrote. The benefits would have to be enormous for such a change to make it into the language.
```

AlexDaniel:
Yeah. I kinda liked everything besides this remark.

zoffixznet:
Thanks.

PR #649, opened 2016-06-28
Rename .pod files to .pod6

coke:
Rename all .pod to .pod6 to avoid issues in #167


AlexDaniel:
This was discussed long time ago and nobody opposed the change, so I think that we can merge it. If it causes any merge conflicts or whatever I'm ready to take care of it.

PR #655, opened 2016-07-01
IO::Path.mode doesn't belong in the timestamp section

MasterDuke17:


AlexDaniel:
grep -Ri 'File test operators' *
```
doc/Language/5to6-perlfunc.pod6:L<File Test operators|/type/IO::Path#File_Test_operators>
```

zoffixznet:
Thanks!

PR #661, opened 2016-07-03
Parallelize a bunch of htmlify.p6

MasterDuke17:
Convert the "Processing $dir Pod files ..." section to a bunch of start()s and await()s. Convert the "Writing [type graph images, specialized visualizations] to html/image" section to a bunch of Proc::Async.new.start()s, and awaits(). 

For me, this brings "time perl6 htmlify.p6 --no-highlight" from 6 min to 1.5 min. However, in my experience, the async/concurrency parts of Rakudo are definitely kind of weird crash prone, and this change does produce the occasional segfault.

### DIFF
--- a/htmlify.p6
+++ b/htmlify.p6
@@ -214,9 +214,20 @@ sub process-pod-dir($dir, :&sorted-by = &[cmp], :$sparse) {
     my $total = +@pod-sources;
     my $kind  = $dir.lc;
     for @pod-sources.kv -> $num, (:key($filename), :value($file)) {
-        printf "% 4d/%d: % -40s => %s\n", $num+1, $total, $file.path, "$kind/$filename";
-        my $pod = extract-pod($file.path);
-        process-pod-source :$kind, :$pod, :$filename, :pod-is-complete;
+        FIRST my @pod-files;
+
+        push @pod-files, start {
+            printf "% 4d/%d: % -40s => %s\n", $num+1, $total, $file.path, "$kind/$filename";
+            my $pod = extract-pod($file.path);
+            process-pod-source :$kind, :$pod, :$filename, :pod-is-complete;
+        }
+
+        if $num %% 10 {
+            await(@pod-files);
+            @pod-files = ();
+        }
+
+        LAST await(@pod-files);
     }
 }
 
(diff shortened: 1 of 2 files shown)


jonathanstowe:
Presumably the segfault on travis is due to resource constraints.

MasterDuke17:
I don't think it's just resource constraints. I get random errors on the occasional run on my machine even with low values for the parallelization. E.g., segfaults, "Unhandled exception: slurpy hash needs concrete key", "Use of uninitialized value $repo-id of type Any in string context", "Reading from filehandle failed: bad file descriptor", "Failed to open file /home/dan/Source/perl6/doc/html/routine/Metamodel::AttributeContainer.attributes.html: no such file or directory". Part of the reason I created this PR was to see if anybody could point out something I was doing wrong to hopefully fix the errors.

MasterDuke17:
As an FYI, I never get any of the errors while doing the image creation (using Proc::Async), it's all during the Pod processing (using start()).

AlexDaniel:
The idea of parallelizing htmlify is great, +1 for that.

Do I get it right that you just ``start`` as many things as possible and then just hope that it is not going to clog the CPU? If so, I don't think that it is optimal. In fact, as a user I probably want to specify myself whether I want it to be parallel or not, and if yes, how many threads it can use.

Instead, you should probably be using ``.race`` (which is almost completely broken, but I'm sure that you can find a way to make some use out of it).

As for the segfaults – yes, any kind of async code will crash rakudo if given enough time. But things will probably get fixed faster if we start using it, right? :)

AlexDaniel:
By the way, if you use .race with configurable params, then we could run some things (e.g. travis) sequentially, so that it always works.

MasterDuke17:
I just added an --parallel option to make the existing code configurable. I didn't start everything at once, I await()ed every hard-coded number of times, that's now specifiable on the command line. I don't remember why I didn't use .race (I just picked this back up after getting distracted for a couple weeks), I'll see if that makes sense.

MasterDuke17:
On my 8 core machine (4 physical + 4 hyperthreaded), --parallel=1 takes 7 real minutes and --parallel=10 takes 50 real seconds.

LLFourn:
I'm not sure if this is helpful but I tried to parallelize htmlify myself a
while ago but it didn't work at all because of:

https://rt.perl.org/Public/Bug/Display.html?id=127033 (dynamic variables
like $*DR disappear in start { })

But this doesn't seem to be the issue here (though I'm not sure why because
the bug still seems to exist).

Issue #686, opened 2016-07-10
ToC show/hide animation is gone, but the delay is still there


AlexDaniel:
[show]/[hide] buttons used to trigger a nice css animation, but it is no longer there.

However, the animation delay is still there. That is, you press [hide], then wait 1 second, and only then it actually hides.

Altai-man:
Does https://github.com/perl6/doc/pull/687 will fix the problem? If yes, merge it please.

AlexDaniel:
If I'm not mistaken, your pull request removes the animation completely. However, another fix would be to fix the animation.

Do we want the animation to be there or not?

gfldex:
CSS animations are fine in my eyes because the reader can make them go away with a userstyle. The less javascript we got the better. A botched setInterval can drain a battery rather quickly.

Altai-man:
Yes, you are right. I can revert js animation, firstly I thought title somewhat implies that delay should be removed.
If you want CSS animation as a substitution of js version, we can remove js version for now and create a ticket like "Implement css animation for ToC box". What do you think?
As for me, animation is unnecessary, but it doesn't bother me in any way if it's quick enough.

gfldex:
please remove the JS version

zoffixznet:
The show/hide feature was originally added as an easier way to get to content. Now that gfldex++ moved the ToC to the side, there's absolutely no reason for the existence of the show/hide feature and it should be yanked out entirely.

AlexDaniel:
> moved the ToC to the side, there's absolutely no reason for the existence of the show/hide feature and it should be yanked out entirely

And now I realized that it looks completely different on larger screens…

On a small screen the toc is still on the top, so [show]/[hide] feature still makes sense (or not?). Also, the links are blue, hmm…

zoffixznet:
> On a small screen the toc is still on the top, so [show]/[hide] feature still makes sense (or not?).

Oh, right, I forgot about smaller screens. I guess there is still use for it. The animation can likely be fixed by changing `ol` to `tbody` in here: https://github.com/perl6/doc/blob/master/html/js/main.js#L42 though I think there was a discussion above about removing the JS version.

Altai-man:
Now this situation looks more interesting.
So we have two choices:
1.Fix JS animation.
2.Remove JS animation and implement CSS animation(without state storing in the cookies) instead later.
I suppose, we can't throw away show/hide feature completely, because of small screens.
[...]

PR #687, opened 2016-07-10
Fix for https://github.com/perl6/doc/issues/686

Altai-man:
I can't test it on my local machine, but fix seems quite easy, so branch. It will be deleted after merging/unmerging.


Altai-man:
I close it because of incompleteness and somewhat wrong approach to the problem solution. I'll write CSS-based animation soon.

PR #725, opened 2016-07-20
Simple cache support

Altai-man:
This is a quite simple implementation of generated page caching.

It can generate cache, update cache for new and changed pod6-files. With --sparse=5 time difference between cached and non-cached run on my machine is about 8x, without is about 9-10x.

It brings new dependency - Digest::MurmurHash3, it's the only one quite consistent hash function I found in the ecosystem. If anyone knows better(and more light-weight) tool - it will be appreciated. But since it's a C bindings, it's quite fast and simple.

I prefer explicit typing style, but if some things can be more clear or concise, I'd like to know about them(I'm not an idiomatic Perl 6 pro).

I suppose, for a documentation main server we shouldn't use cache(it can be disabled using command-line argument), but for local make it seems good enough.

### DIFF
--- a/META.info
+++ b/META.info
@@ -2,7 +2,7 @@
     "name"        : "p6doc",
     "version"     : "*",
     "description" : "Perl 6 documentation (tools and docs)",
-    "depends"     : [ "URI", "File::Temp", "JSON::Fast" ],
+    "depends"     : [ "URI", "File::Temp", "JSON::Fast", "Digest::MurmurHash3" ],
     "build-depends" : [ "File::Find" ],
     "provides"    : {
         "Perl6::Type"                   : "lib/Perl6/Type.pm",
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: html html-nohighlight test help
 
 html:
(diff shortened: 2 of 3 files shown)


coke:
My preference would be to use make to handle building intermediate representations that we want to save, or barring that, fixing/expanding the caching that's already in htmlify that is trying to use the precomp features that are part of p6.

Altai-man:
I saw precomp-related code in htmlify, but I thought it has somewhat different purpose(since we re-generate all documentation files every time to matter how many times we build documentation).
If this piece of code just broken, we need to fix it and it will be more wise solution. Is current state of precomp feature good enough?

Altai-man:
This PR is already quite obsolete and nobody wants to merge/fix it(and we can't merge it anyway) including myself, so closed for now.

JJ:
Please @finanalyst @antoniogamiz check this implementation out. It uses a specific hashing algorithm to specify what needs to be cached. It was eventually deleted, but it could inspire us to do (or do not) create our own cache algorithms. Anyway, we need to acknowledge work that has been done before on this.

PR #731, opened 2016-07-21
syntax correction

molecules:
Correction replaced second smartmatch with double asterisk. Now the statement is true as advertised.


zoffixznet:
Thanks a lot.

PR #735, opened 2016-07-21
Glossary cleanup

Altai-man:
1.Entries that just point to wiki were removed(except MoarVM, since it's definitely Perl6-specific).
2.Entries which are already described were removed.

This is a part of https://github.com/perl6/doc/issues/728#issuecomment-234317210

### DIFF
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -1283,6 +1283,7 @@ compiled bytecode.
 Rakudo supports this only for modules so far.
 
 =head2 Importing specific functions from a module
+X<|import>
 
 In Perl 5 it is possible to selectively import functions from a given module
 like so:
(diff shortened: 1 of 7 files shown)


jonathanstowe:
Hi, would it be possible to split this into two PRs as the https://github.com/perl6/doc/pull/735/commits/7dbcc552be503a8ed0014c287efa3e1e18ad1873 is fairly uncontroversial, but the other one isn't quite so clear.

Of course it would just be possible to merge the one commit, but the discussion would be lost.

Altai-man:
Sure. I'll prepare them later, so this PR can be closed.

Thanks to @molecules for little fixing!

jonathanstowe:
The deletion of the items which are defined in external sources needs more discussion and should probably go in a new PR from a different branch.

The deleting of items which are defined in other pages in the documentation is the uncontroversial one and I'd say it can go straight in.

Altai-man:
Okay, I'll force push fixed PR here and then will create a new one.

PR #743, opened 2016-07-22
Second part of entries cleanup

Altai-man:
According to https://github.com/perl6/doc/issues/728 discussion, this is a patch that removes glossary items which just point to wikipedia pages.

**Should we remove them or not is still a debatable question, so don't merge this "Just because".**

Opinions are still needed.

### DIFF
--- a/doc/Language/glossary.pod6
+++ b/doc/Language/glossary.pod6
@@ -7,7 +7,7 @@
 =head1 Abstract Class
 X<|Abstract Class>
 
-The generic Computer Science term "abstract L<class|#Class>" defines the
+The generic Computer Science term "abstract class" defines the
 L<interface|#Interface> or L<#API> of a class.  In Perl 6, this is
 implemented using roles with L<stubbed|#Stub> methods.
 
@@ -134,13 +134,6 @@ L<callable block|#Callable>.
 The arity of a C<Callable> is one of the main selectors in
 L<multi-dispatch|#Multi-Dispatch>.
 
-=head1 AST
-X<|AST>
-
-Acronym for
-L<B<A>bstract B<S>yntax B<T>ree|https://en.wikipedia.org/wiki/Abstract_syntax_tree>.
-Used in many places, including actions, L<#PAST>, and L<#QAST>.
-
 =head1 Autothreading
 X<|Autothreading>
 
@@ -185,12 +178,6 @@ X<|bytecode>
 X<|Camelia>
 
 A butterfly image intended primarily to represent Perl 6, The Language.
-
-=head1 Class
-X<|Class>
-
-A basic software structure in L<#OOP>.
-See the L<Wikipedia entry|https://en.wikipedia.org/wiki/Class_%28computer_programming%29>.
 
 =head1 Colon Pair and Colon List
 X<|Color Pair> X<|Colon List>
@@ -423,7 +410,7 @@ No Such Thing
 X<|Opt>
 
 Short for "optimization", usually in either the context of L<spesh|#Spesh> or
-L<#JIT>.
+JIT.
 
 =head2 PB
 X<|PB>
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Well, I think that trying to create an alternative encyclopedia of commonly used IT terms just makes no sense, so I am all for it (as you might have guessed, given my #728 report).

However, could I ask you to work a bit more on this pull request?

* Given that you're removing ``Class``, and given that you touched the ``L<class|#Class>`` link, you should also change ``L<class|#Class>, L<module|#Module>, L<grammar|#Grammar>`` to point to appropriate places in the docs. For example, https://docs.perl6.org/language/objects#Classes, https://docs.perl6.org/language/modules and https://docs.perl6.org/language/grammars (or maybe there are better links, check it out yourself).
*  ``L<#Javascript>.`` assumes that we have a ``Javascript`` item (another term that we should not explain). We do not have a description for ``Javascript`` anyway.
* There is also a link to ``L<#VM>``, but we only have ``L<#Virtual Machine>``. Can we also get rid of “Virtual Machine” definition? For example, it claims that Javascript is a virtual machine… (what? Am I reading this correctly?)
* ``whitespace`` is listed but has no description. If one can write a description that will justify its existence – great! Until then, let's assume that one can find https://en.wikipedia.org/wiki/Whitespace_character. Actually, we can also make https://docs.perl6.org/language/5to6-nutshell.html#Whitespace searchable, as it gives a good idea of what whitespace is.
* What about ``Lexing``?

Also, have you tried grepping for these terms in the rest of the docs? We don't really want more broken links.

Altai-man:
1.Its already fixed with https://github.com/perl6/doc/commit/f96c6a1d3ebf6f9bfcdad9488b60a7e4ae1d24aa
2.Its already fixed with the same commit.
3.In this commit I changed it from `#VM` to `#Virtual machine`. We can remove it completely in this PR, I think.
4.I'll add commit that removes `whitespace` changes links to `5to6-nutshell` too, if you think it's appropriate.
5.Who knows.

Yes, I check it with
`grep -r "/language/glossary#" doc/*`
Current output is:
```
doc/Language/contributors.pod6:L<Pull Request|/language/glossary#Pull_Request> with the required change(s).
doc/Language/5to6-nutshell.pod6:C<no warnings> is currently L<NYI|/language/glossary#NYI>,
doc/Language/performance.pod6:Use the C<m: your code goes here> L<#perl6 channel evalbot|/language/glossary#camelia>
doc/Language/performance.pod6:Enter C<prof-m: your code goes here> in the L<#perl6 channel|/language/glossary#IRC>
doc/Language/performance.pod6:others using the L<#perl6|/language/glossary#IRC> evalbot
doc/Language/performance.pod6:L<camelia|/language/glossary#camelia>.
doc/Language/variables.pod6:Note: Slangs are L<NYI|/language/glossary#NYI> in Rakudo.
doc/Type/Str.pod6:attempt to construct the appropriate L<allomorph|/language/glossary#Allomorph>,
doc/Type/NumStr.pod6:The dual value types (often referred to as L<allomorphs|/language/glossary#Allomorph>)
doc/Type/IntStr.pod6:The dual value types (often referred to as L<allomorphs|/language/glossary#Allomorph>)
doc/Type/ComplexStr.pod6:The dual value types (often referred to as L<allomorphs|/language/glossary#Allomorph>)
doc/Type/RatStr.pod6:The dual value types (often referred to as L<allomorphs|/language/glossary#Allomorph>)
```
We have allomorph, NYI and IRC stuff described, so I'm assuming it's safe.

AlexDaniel:
Right. OK, good to go then.

jonathanstowe:
Any chance we could get that merge conflict fixed up?

jonathanstowe:
Looks good, I'd say all things being equal that we merge it and see where we stand after that.

AlexDaniel:
Speaking of something else, I think that it's time to merge this. :)

Nobody opposed really, so if there are some good reasons not to do that I guess that we will hear them after the merge.

PR #744, opened 2016-07-22
Bool.succ, .pred methods explained (please review)

molecules:
I first encountered the method "succ" today, and it was a method on Bool. At first I misunderstood it to mean "success", and wondered what type of failure or disaster "pred" stood for. Then I discovered that "succ" means "successor" and "pred" means "predecessor" and that they seem to totally make sense if we think of Bool in terms of it consisting of two possible enums.  So, that's what I tried to explain, but I would really like someone that feels more comfortable with this topic to please review my proposed changes (and please improve them if needed), before merging this pull request. (I could have committed directly, but I wanted someone to review this first.)

### DIFF
--- a/doc/Type/Bool.pod6
+++ b/doc/Type/Bool.pod6
@@ -19,6 +19,8 @@ Returns C<True>.
     say True.succ;                                    # True
     say False.succ;                                   # True
 
+succ is short for successor. In many languages, the succ function is used to return the "next" enum, also known as the successor. Bool is a special enum with only two values, C<False> and C<True>. When sorted, C<False> comes first, so C<True> is its successor. And since C<True> is the "highest" Bool enum, its own successor is also C<True>.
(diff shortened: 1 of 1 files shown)


jonathanstowe:
Looks good to me.

molecules:
Thanks!

AlexDaniel:
I thought that I wrote a comment… where did it go?
I liked everything except two things:
* “In many languages” – is it true that many languages actually have a succ function? How many? Is it a good way to start a sentence? Probably not.
* “to return the "next" enum” – “next enum” sounds a bit weird to me. Next enum value perhaps?

molecules:
Thanks @AlexDaniel. See in https://github.com/perl6/doc/commit/1b63b98eb737ae2491dfc245b3eff9e6a9eb1bd9.

PR #757, opened 2016-07-25
Correct output of permutations sub

0racle:


AlexDaniel:
Right. Also, one might argue that we should also include the parens, because they are present in the real output.

0racle:
I agree, but as they were omitted originally for reasons unknown to me, I left them out.

PR #758, opened 2016-07-25
Is this better? (note on $*SCHEDULER and threads)

molecules:
This was probably too much text to put into that cell anyway. Is this better?


AlexDaniel:
Perhaps also add “See note below” or something?

molecules:
@AlexDaniel Good call. Done.

PR #762, opened 2016-07-26
Add documentation for Callframe

zostay:
Fixed up the 5-to-6 caller note and then added documentation for CallFrame and added docs for package on Routine as well since that was needed and related.


dha:
Thank you! caller() was the function I needed to figure out the Perl 6 equivalent for that led directly to the writing of the 5to6-perl* docs. Glad to see full documentation on this. :-)

Issue #724, opened 2016-07-19
Make abbreviations of Language/* docs searchable


AlexDaniel:
These should be available:
- **IPC** → https://docs.perl6.org/language/ipc
- **MOP** →https://docs.perl6.org/language/mop

That's it, I think.

tommystanton:
Thanks, @AlexDaniel (I had mentioned "IPC" in #723)! :smile:

Altai-man:
Closed with https://github.com/perl6/doc/commit/5c268237055ad7d58ba3bfb6be161d5000cc5b93?

AlexDaniel:
@Altai-man yeah. Kinda. Looking at it, I realize that it would've been much better if it was pointing to [language/mop](https://docs.perl6.org/language/mop) directly instead of [language/mop#index-entry-MOP](https://docs.perl6.org/language/mop#index-entry-MOP).

Is there any easy way to do that?

gfldex:
`index-entry-` is added automatically to prevent clashes between index anchor ids and heading anchor ids.

gfldex:
Well, there are two special cases hardcoded in `template/search_template.js` already. And if you really really want to you could add more. That may have consequences for your immortal soul tho.

Altai-man:
@AlexDaniel according to what gfldex said, we can do that(even I can do that) by hardcoding. But hardcoding is obviously not cool and I don't have intentions to do that.
As a long-term solution we can, of course, remove `index-entry-` and then by ourselves look for duplicate ids by some test.
If you want, you can make hardcoding by yourself, but, as for me, I doubt it matters.

AlexDaniel:
¯\\\_(ツ)\_/¯

Of course it's not cool, but at the same time a link that does not point you to the start of a page (when it should do so) is less than awesome.

Should we just close this or wait for some better solution to come?

AlexDaniel:
For example, we can have a special file with alternative queries. There we can make it so that typing “MOP” is equivalent to typing Meta-Object Protocol. Who knows, maybe it is going to be very common?

Another solution is to add MOP to our glossary (notice that it is already linked from [MoarVM](https://docs.perl6.org/language/glossary#MoarVM_)), there we can provide a link to language/mop. This, however, will require additional mouse click, which is not cool.

Altai-man:
Hmmm.

I think, that this issue as it states "Make abbreviations searchable" is done(LTA, but still better that nothing to search).

[...]

PR #763, opened 2016-07-26
Now we take predefined search values from outer file

Altai-man:
Plain text file because it's universal for all.

I've left js syntax here, because it's easy and we don't have to do anything else with source file.

I want opinions about - is it needed to change file to some format and then parse it in `htmlify`(if someone hates piece of js code in some outer file)? Maybe like this:
```
Category
Value
Url

Category
...
```
it is quite easy file format and I can implement this.

Ability to make links that point to just pages, not anchors, is quite good for search, for example, see https://github.com/perl6/doc/issues/724


zoffixznet:
Personally I'm :-1: on having a human-generated list of search terms. Content changes and pages get deleted. This PR introduces another maintenance burden that we need to keep `predefined-search-items` up to date and accurate. 

Is there no way to use the `X<>` POD marker to generate these search terms?

Altai-man:
@zoffixznet, afaik - no. But I'm not a pod6 magician, of course, so I can be wrong.

The thing is, we already have two hardcoded entries in js-code. I just provided an easier way to update them.
cast @AlexDaniel, @gfldex.

zoffixznet:
> The thing is, we already have two hardcoded entries in js-code. I just provided an easier way to update them.

Ah, my bad. I missed that. It'd be best if we didn't have those, but if there's no way then there's no way.

AlexDaniel:
I like it.

We can have a discussion about whether we should have manually defined items or not, but we already do, so… Basically I see no reasons not to merge.

The only thing is that people will now be tempted to add search items manually (given that now it is slightly easier). But that's OK, once there is a way to get rid of this file we will do it, but right now let's improve the search.

PR #767, opened 2016-07-27
Css-based TOC animation

Altai-man:
Smooth css animation for TOC. This animation isn't based on (max-)height. The only JS part needed is to change css classes, should work with most browsers.

### DIFF
--- a/assets/sass/style.scss
+++ b/assets/sass/style.scss
@@ -554,3 +554,28 @@ div.highlight span.c-Singleline, div.highlight span.c {
         }
     }
 }
+
+/* hide TOC */
+.hiddenTOC {
+  font-size: 0;
+  margin: 0;
+  opacity: 0;
+  padding: 0;
+  /* fade out, then shrink */
+  -webkit-transition: opacity .25s, font-size .5s .25s, margin .5s .25s, padding .5s .25s;
+  -moz-transition: opacity .25s, font-size .5s .25s, margin .5s .25s, padding .5s .25s;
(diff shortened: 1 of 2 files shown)


Altai-man:
Fixes https://github.com/perl6/doc/issues/686

Altai-man:
We still need two classes for animation, since if we remove one, hiding or showing will be without animation transition.

AlexDaniel:
But, but… it does not work at all! It seems like .css file was not generated or something.

Altai-man:
It was fixed. I thought the server builds sass-files by itself.

zoffixznet:
@AlexDaniel which browser?

In my Firefox the failure modes are:

1. The text disappears, but the height of the table is still there, so the point of hiding the ToC is gone; I still have to scroll (this is on small-width view, where the ToC is at the top of the page)
2. After I hide the ToC, on page load the ToC flashes in shown position and then animates to an off position. It should start off in off position
3. FWIW, the animation no longer works on my phone in Opera browser. It's not one of our supported browsers, I know, but still... :crying_cat_face:

zoffixznet:
> It was fixed. I thought the server builds sass-files by itself.

Not yet. It just uses the generated `html/css/style.css` file.

coke:
I have the same issue as @zoffixznet - the text hides, but the vertical space it consumes is still there, and the default on page reload is wrong.

Please revert this for now.

Altai-man:
Reverted. I suppose, we should just remove animation at all then.

Issue #761, opened 2016-07-26
Let's not mangle the output


AlexDaniel:
As mentioned in https://github.com/perl6/doc/pull/757#issuecomment-234834178, such examples are present in ``doc/Type/List.pod6``, but it probably exists in other files as well.

I think that if the example has “say” in it, then we should provide non-mangled output.

Sure, parens are not a big deal, but still people may start asking questions like “is it up to date?”.

So what do you think? Should we always provide the output as is, or maybe changing it a little bit improves readability?

zoffixznet:
As-is is best IMO. Just run the code and copy-paste the output into docs.

This is especially important for new programmers who may think they're doing something wrong when they don't see the exact output.

gfldex:
Only truncate (long exceptions, stack trace, etc.), never change the output. At some point we might want to have a tool that EVALs all the examples and compares the output. The less false positives we get in that situation the better.

coke:
As is - we eventually want to be able to have tests where we can run the sample code and make sure the output matches.

AlexDaniel:
Thank you for your prompt comments.

I think that I will make one pass through as many examples as I can and then we would be able to say that this issue is solved (if anybody else is willing to do this then please say so). That being said, if you stumble upon some example that has slightly different output, please fix it.

Altai-man:
@AlexDaniel, I still have some free time, but I'm not sure. Is there are any pattern to search? Just code can be detected with 4 spaces, a result after `#` character also can be detected. Words like `Examples:` and `>` symbol after four spaces also can be code snippet. Is it all?
`grep -r "    #" doc/ | nl` gives 2018 lines, all other variants just 15-20.

AlexDaniel:
@Altai-man well, I was planning to use grep to find code snippets, but didn't plan to automate anything else. If you check two examples per minute, then it's about 120 examples per hour. Meaning that you can manage to do the whole thing during one evening.

Altai-man:
@AlexDaniel, I asked only about what patterns can be used to find examples, I didn't plan to automate anything.
If that's all, I'll prepare PR later, so you can do something else.

Altai-man:
Pull request is prepared.

Also, while digging into the examples, I noticed that we don't have single result style. Just now we have three different variants:
`#-> result`
`# result`
`# OUTPUT: «result␤»`
It's a bit inconsistent. Maybe we also should choose One True Variant and make a conversion?
If we want to test our output against samples output, single format will be helpful.

[...]

PR #775, opened 2016-07-28
Almost all example outputs were changed to literal versions

Altai-man:
See https://github.com/perl6/doc/issues/761 for details.

### DIFF
--- a/doc/Type/List.pod6
+++ b/doc/Type/List.pod6
@@ -245,7 +245,7 @@ these invocations in a result list.
 
 Unlike C<map> it flattens non-itemized lists and arrays, so
 
-    say ((1, 2), <a b>).flatmap(&uc).join('|');     # 1|2|A|B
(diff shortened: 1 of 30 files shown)


doc/Type/List.pod6
```diff
@@ -245,7 +245,7 @@ these invocations in a result list.
 
 Unlike C<map> it flattens non-itemized lists and arrays, so
 
-    say ((1, 2), <a b>).flatmap(&uc).join('|');     # 1|2|A|B
+    say ((1, 2), <a b>).flatmap(&uc).join('|');     # 1 2|A B
```
zoffixznet: While unrelated to the topic of this PR, it seems this example isn't the best one. The output shows no flattening.

Altai-man: Then someone should fix this exact line, obviously. New issue, perhaps? I can't make up a good example right now.

AlexDaniel:
Great job! Awesome.

Issue #746, opened 2016-07-22
Add a test for non-breaking space in Perl 6


AlexDaniel:
See this: https://github.com/perl6/doc/commit/d8d26e291901087f771ae59c4189fe2fcca3d530

Given that some tests are now optional, we can freely do stuff like this.

gfldex:
Please keep in mind that some rancid What Ever Online (TM) used by Online "Journalists" may turn Perl 6 into Perl&nbsp;6 in articles or even a heading. You may actually make the situation worse be being conservative while receiving and liberal while sending.

coke:
This is now available in 'make xtest', as a failing test.

coke:
Perl 6's take on unicode has always been to push the envelope and encourage others to improve their handling as well.

AlexDaniel:
Anybody willing to work on making it pass?

Altai-man:
@AlexDaniel, I already did it locally(just `sed`), but this test reacts not only to `Perl 6`, but to `Perl 5`. And, generally, `Perl $any-digit`. Do we want `Perl 5` to be non-breakable too?

PR #780, opened 2016-07-30
Now all entries of `Perl 6|5` are written with non-breaking space

Altai-man:
For https://github.com/perl6/doc/issues/746


AlexDaniel:
Some discussion: http://irclog.perlgeek.de/perl6/2016-07-30#i_12939655

Issue #752, opened 2016-07-24
Pragma(s) missing from docs


MasterDuke17:
It's not in the index, but if you do a site search the word 'pragma' occurs several times. I feel it should have its own section with explanations of the various pragmas, e.g., fatal, v6, variables.

----

New page: https://docs.perl6.org/language/pragmas

**⚠ feel free to edit this list**

- [x] use v6 ([testing#Writing_tests](https://docs.perl6.org/language/testing#Writing_tests))
- [x]  use lib ([modules#Finding_Modules](https://docs.perl6.org/language/modules#Finding_Modules))
- [x]  use variables ([variables#Default_Defined_Variables_Pragma](https://docs.perl6.org/language/variables#Default_Defined_Variables_Pragma))
- [x]  use fatal
- [x]  use soft ([functions#Re-dispatching](https://docs.perl6.org/language/functions#Re-dispatching))
- [x]  use MONKEY-TYPING ([augment](https://docs.perl6.org/syntax/augment))
- [x]  use MONKEY-SEE-NO-EVAL ([EVAL](https://docs.perl6.org/routine/EVAL))
- [x]  use precompilation (see #798)
- [x]  use newline (see #797)
- [x]  use trace (see #800)
- [x]  use strict (see #799)
- [x]  use nqp (see #796)
- [x]  use worries (see #801)
- [x]  use internals (see #809)
- [x]  use invocant (NYI) (see #810)
- [x]  use parameters (NYI) (see #811)
- [x]  use MONKEY-(BRAINS, GUTS, BUSINESS, TRAP, SHINE, WRENCH, BARS) (reserved, but I believe NYI)
- [x]  use experimental
- [x] ↑ and it's experiments (see #812)

AlexDaniel:
Let's create a list of all pragmas?

**⚠ feel free to edit this list**
* use v6 ([testing#Writing_tests](https://docs.perl6.org/language/testing#Writing_tests))
* use lib ([modules#Finding_Modules](https://docs.perl6.org/language/modules#Finding_Modules))
* use variables ([variables#Default_Defined_Variables_Pragma](https://docs.perl6.org/language/variables#Default_Defined_Variables_Pragma))
* use soft ([functions#Re-dispatching](https://docs.perl6.org/language/functions#Re-dispatching))
* use MONKEY-TYPING ([augment](https://docs.perl6.org/syntax/augment))
[...]

PR #793, opened 2016-08-03
try index method per zoffix's hint

tbrowder:


tbrowder:
it works--closed issue #752

Issue #794, opened 2016-08-03
"Usage" section seems useless to reader


Altai-man:
In some documentation files we have "Usage" section, like this:
```
Defined as:

    multi method fmt(Pair:D:) returns Str:D {}

Usage:

    PAIR.fmt(FORMATSTRING)
```

Judging by grep, we have about 300 occurrences of such "examples". But, eh, is it really useful?

If we explain some method or subroutine, I think, anyone, who reads documentation, do know how to apply subroutines and call methods. For a method it's just `value.method-name(maybe-some-args)` - this formula doesn't change with every section, we just repeat useless stuff. We don't have any output example to be really useful for anyone in such sections, too. If we apply a function, it'll almost always(yes, perl 6 has many ways to call a subroutine, but that's not the case) be `sub-name(value, some-other-args)`.
Is it me or this "Usage" sections are a bit redundant?

Secondly, because we have "fake" arguments there, like "PAIR" or "FORMATSTRING', it is impossible to check them with a compiler. I think, if we want to really show some usage, we need to rewrite all such strings as a proper examples, if we not, we need to remove such sections.

jonathanstowe:
I'm more with having the actual signature (pretty much as in the declaration,) and then if necessary some example (or examples,) that show it being used with some example output.

zoffixznet:
IIRC @dha started adding those `Usage` sections but then someone raised concerns similar to yours and the work stopped.

They do seem redundant, though personally I'd be OK if those were replaced with actual usage; that is actual code example, rather than `FORMATSTRING`, say `:foo<bar>.fmt: '%s: %s'`

coke:
I was the one who raised that concern originally.

I believe they are intended to address the issue that to truly understand the signature presented in a perl 6 routine, you might have to understand a lot of extra syntax; The simplified approach is supposed to make it easier to understand.

My concern is that we already have to explain all the details of how signatures work; now we may also have to explain all the details about the USAGE blocks work.

I am fine with going all out with USAGES where needed, but I think we need to document the docs in that case.

If we end up just using function signatures, then we should have quick links/popups to "what this signature means" that are automatically generated.

The current situation with some USAGES and some not is confusing.

Altai-man:
I believe that actual working examples will be more useful than just repeating of `Type.method-name` every time.
Signatures can really be strange for a novice(with all this `:D:`, slurps, etc), but actual example provides an ability to copy-paste code just from the docs and play with it a bit, to understand what is where.
[...]

PR #814, opened 2016-08-06
New pod config option: 'signature'

Altai-man:
See https://github.com/perl6/doc/issues/794#issuecomment-237981574

Firstly, cast @gfldex, since he is the main inventor of our doc-compiling project.

Since we need to compile signatures(errors occur there too) we need some method/routine/sub body to compile it without error. The thing is, we don't want reader to see this ugly `{}` empty method body.
Because of that, using @coke advice, I've implemented this option. If you write it like this:
```
=begin code :signature
    method some(Bar)
=end code
```
or even like that
```
=begin code :signature
    method some1(Bar)
    method some2(Bar)
=end code
```
empty brackets will be added automatically(in second case, for every string respectively).

My English is less than awesome(I guess, it really is), so I'll be really grateful if someone will fix up my bad-written description at the `CONTRIBUTION.md` page.

It can be merged if @gfldex approves.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,6 +97,16 @@ To export examples from all .pod6-files use `make extract-examples`. To run
 individual tests pick the right .p6-file from `examples/` as a parameter to
 `perl6`.
 
+### Signatures
+
+Signatures are part of actual code and need to be tested too. But without the body every
+signature will cause an error during the test. Use the pod-config option `signature` to
+automatically add an empty body.
+
+    =begin code :signature
+        method new(*@codes)
+    =end code
+
 ### Skipping tests
 
 Some examples fail with compile time exceptions and would interrupt the test
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -20,7 +20,9 @@ List or a list-like type.
 
 Defined as:
 
-    multi method ACCEPTS(Any:D: Mu $other) {}
+=begin code :signature
+    multi method ACCEPTS(Any:D: Mu $other)
+=end code
 
 Usage:
 
@@ -36,7 +38,9 @@ Many built-in types override this for more specific comparisons
 
 Defined as:
 
-    method any() returns Junction:D {}
+=begin code :signature
+    method any() returns Junction:D
+=end code
 
 Usage:
 
@@ -54,7 +58,9 @@ C<any>-L<Junction|/type/Junction> from it.
 
 Defined as:
 
-    method all() returns Junction:D {}
+=begin code :signature
+    method all() returns Junction:D
+=end code
 
 Usage:
 
@@ -72,7 +78,9 @@ C<all>-L<Junction|/type/Junction> from it.
 
 Defined as:
 
-    method one() returns Junction:D {}
+=begin code :signature
+    method one() returns Junction:D
+=end code
 
 Usage:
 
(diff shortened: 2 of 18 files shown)


Altai-man:
It sure feels a bit wordy, but solution with "Let's check it with regexes and will hope it didn't break somewhere" or something similar is not cool too.

AlexDaniel:
What about changing this:
```perl6
$content.subst("\n", "\{\} \n", :g)
```
to something like this (untested):
```perl6
$content.subst(/ [ ‘)’ | returns \s+ \S+ ] <( $$ /, " \{\}", :g)
```
This way it should support multi-line signatures as well. Like the one on [Str#split](https://github.com/perl6/doc/blob/master/doc/Type/Str.pod6#L177)

Altai-man:
@AlexDaniel my code was tested on all(updated) pages that start with "A" character. I'll try your variant, but it seems a bit more obscure, than mine(or not). If it passes, I'll likely update my pr.
But I'm not completely sure about what to do with multi-line signatures. Maybe this behavior will be changed in the future.

tbrowder:
On Friday, August 5, 2016, Altai-man <notifications@github.com> wrote:
...
>
> Because of that, using @coke <https://github.com/coke> advice, I've
> implemented this option. If you write it like this:
>
> =begin code :signature
>
> I think you need to document the new config option in S26.  And how about
the Perl 6 grammar and actions in rakudo?

Altai-man:
@tbrowder, I believe that Rakudo and Perl 6 grammar itself have nothing to do with how exactly we "test" our documentation, it's a separate thing. It is a custom option that is useful for the documentation testing, but it's definitely not mandatory. S26 only defines a presentation. It doesn't say anything about testing the examples. Please, correct me if I'm wrong here.
This test can be changed anytime if needed, hence I don't think we should make it more official then we already did.

For example, someone can implement an another way to test the documentation and just ignore this config option.

If you'll argue that we need to replace explicit signature marking with implicit signature guessing(like 'If there a word `method|routine|sub|etc` after four spaces and then symbols, but not the {} ones', you can be right. Then we need to revert this and rewrite our examples-extracting code a bit.

I'm not a S26 specialist, but by previously added `:skip-test` option I assumed, that custom config options are not something scary. Am I wrong?

Altai-man:
Also, after your comment I thought a lot about implicit detection and now it looks for me as a really better solution than explicit marking.
If we revert this PR and write a method detection, it'll automatically(i.e. without any changes) make all our signatures correct, without adding of two lines for every signature(and we have a lot of them).

tbrowder:
On Sunday, August 7, 2016, Altai-man <notifications@github.com> wrote:

> Also, after your comment I thought a lot about implicit detection and now
> it looks for me as a really better solution than explicit marking.
>
I'm certainly not an S26 expert either, but I have been dabbling in the
guts of rakudo pod handling and wanted to warn about possible side effects
from defining a pod attribute outside rakudo.

> If we revert this PR and write a method detection, it'll
> automatically(i.e. without any changes) make all our signatures correct,
> without adding of two lines for every signature(and we have a lot of them).
>
I don't pretend to understand completely, but that sounds better for the
moment. However, I think the more we customize the pod cleanup for the html
output we shouldn't forget that, eventually, we need to convert pod to
other output formats.

>

Altai-man:
@tbrowder in fact, the pod testing doesn't affect the html or any other output format. But yes, you're totally correct.
I'm going to revert this and implement signature detection.

PR #827, opened 2016-08-08
Add a missing semicolon

titsuki:


AlexDaniel:
But a semicolon is not required there, right?

I don't mind it being added, but still I wonder what is the rationale. Like, should we **always** add semicolons?

titsuki:
@AlexDaniel 
> But a semicolon is not required there, right?

Yes, you're right.
I have two reasons to add semicolons.

1) For keeping consistency
In this example, the above one has a semicolon at the end of its line as the following.
```
dd %h<oranges bananas>;
```

Moreover the other examples in this page also have a semicolon.

2) For running code snippet easily
The readers of this document sometimes copy, paste and run its codes. 
Perhaps they modify the pasted code in accordance with their usage before the execution.
If there is a semicolon, they don't need to pay attention whether the line pasted the example code is the end of the program or not.

> should we always add semicolons?

Sorry, I'm not sure... but one of the Perl 6 motto is "there's more than one way to do it".

PR #842, opened 2016-08-19
Add an index for ;;

### LINKED ISSUE #841: ;; is not searchable
In roast, `;;` is used here
https://github.com/perl6/roast/blob/master/S06-multi/syntax.t#L38-L53

In the perl 6 document, the description of the `;;` is in the following page
https://docs.perl6.org/type/Signature#Long_Names

titsuki:
Fix #841

### DIFF
--- a/doc/Type/Signature.pod6
+++ b/doc/Type/Signature.pod6
@@ -451,7 +451,7 @@ X<|Long Names>
 =head2 Long Names
 
 To exclude certain parameters from being considered in multiple dispatch,
-separate them with a double semi-colon.
+separate them with a double semi-colon X<C<;;>|;;>.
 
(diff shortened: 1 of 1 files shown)


Altai-man:
Thanks for the noticing.
I have suspicious that pointing straight to the header will be better here. I'll do a quick test now...

titsuki:
@Altai-man hmm... I think so too, but we have a CI environment and there's no restrictions not to leverage it.

titsuki:
ok I'll merge this PR :)

Altai-man:
If it works then I'm okay, but the style is somewhat different comparing to two nearest paragraphs. Anyway, I can't really work on the vacation, so let it be.
Thanks for your contribution!

titsuki:
oh... It seems broken
https://docs.perl6.org/type/Signature#Long_Names

PR #856, opened 2016-08-23
Add more details of function sprintf formatting

tbrowder:


Altai-man:
In case you are interested in `sprintf` documentation, please, consider checking of all examples and sections.
As far as I can see, section of `size`(https://docs.perl6.org/type/Str#size) doesn't work currently:
```
> sprintf '%h',  1;
Directive h is not valid in sprintf format sequence %h
  in any  at /home/sena/.rakudobrew/moar-nom/install/share/perl6/runtime/CORE.setting.moarvm line 1
  in any panic at /home/sena/.rakudobrew/moar-nom/install/share/nqp/lib/NQPHLL.moarvm line 1
  in any statement at /home/sena/.rakudobrew/moar-nom/install/share/nqp/lib/NQPHLL.moarvm line 1
  in any TOP at /home/sena/.rakudobrew/moar-nom/install/share/nqp/lib/NQPHLL.moarvm line 1
```
This table seems as a copy-paste from C function man page.

Accordingly to the https://design.perl6.org/S32/Str.html#sprintf and https://github.com/perl6/roast/blob/1f2ad001a9b605b4e41de7f5e3a820ae8dbd7a3d/S32-str/sprintf.t we don't have such directives.

tbrowder:
On Tue, Aug 23, 2016 at 1:10 PM, Altai-man <notifications@github.com> wrote:
> In case you are interested in sprintf documentation, please, consider
> checking of all examples and sections.
> As far as I can see, section of size(https://docs.perl6.org/type/Str#size)
> doesn't work currently:
...

I was going to check all before I added the info, but decided to get
it on the line first.   I do plan to check it all.

Thanks for the reality check.

Issue #861, opened 2016-08-26
method sum() (Range) is not documented


titsuki:
`method sum(Range:D:)` is tested here
https://github.com/perl6/roast/blob/master/S03-operators/range-int.t#L486-L514

titsuki:
`method sum()` is here
https://github.com/rakudo/rakudo/blob/nom/src/core/Range.pm#L607-L611

Altai-man:
Closed by https://github.com/perl6/doc/pull/863#issuecomment-242974415

PR #863, opened 2016-08-26
Added method sum() to Range pod

ahalbert:
#861

### DIFF
--- a/doc/Type/Range.pod6
+++ b/doc/Type/Range.pod6
@@ -181,4 +181,12 @@ by not actually generating the list if it is not necessary.
 Performs the same function as C<Range.list.roll>, but attempts to optimize
 by not actually generating the list if it is not necessary.
 
+=head2 method sum
+
+    multi method sum() returns Numeric:D
+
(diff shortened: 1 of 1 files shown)


Altai-man:
Seems good for me. Can you include the exact type of exception in the description, please? It can be useful.

ahalbert:
The exception is thrown in NQP, so would that be documented in Perl6? 

The error:
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in any call_with_capture at gen/moar/m-Metamodel.nqp line 3865
  in block <unit> at <unknown file> line 1

jonathanstowe:
Well it's ```X::Str::Numeric``` that is thrown and it is defined in the Exception.pm so yes arguably it should be documented.  I'm minded to merge this and then fix that immediately afterward.

ahalbert:
I've added the missing exception. It is rebased into the last commit.

jonathanstowe:
If I get a minute I'll add a document for the exception itself.

Issue #867, opened 2016-08-29
sub run needs a good example with real args


tbrowder:

ahalbert:
Proc has an example using run.

```
my $proc = run 'echo', 'Hallo world', :out;
my $captured-output = $proc.out.slurp-rest;
say "Output was $captured-output.perl()";       # Output was "Hallo world\n"
```

I'm thinking of using this example for routine run.

tbrowder:
closed with pr 868

PR #868, opened 2016-08-29
Added example code of how to use routine run.

ahalbert:
#867


AlexDaniel:
But the next line says:
```
See L<Proc|/type/Proc> for more details, for example on how to capture output.
```
Perhaps we should delete ``for example on how to capture output`` part, given that your pull request adds this example?

ahalbert:
I've removed the line.

Issue #822, opened 2016-08-08
Mark external links


AlexDaniel:
External links should look differently.

This is actually pretty common. For example, Wikipedia does that.

Typically it is implemented by adding an icon (I'd recommend svg). Maybe a unicode arrow will do (like ↗), but using unicode characters as icons is usually not the best idea.

This is also related to #710

ahalbert:
Would this be done in the code for building the documentation, or would we have to mark each link in the pod with the icon?

AlexDaniel:
Good question. Well, for the web version of the docs we want such links to have a css class… but I am not sure about everything else.

ahalbert:
I realized in the POD, it's obvious that it's an external link. So we may only want to do this for formats that obscure the link destination.

coke:
I think this is for the HTML site only; note that the style appears when running the site locally (once I installed a few Mojo modules) with 'make run', but not on the public site.

coke:
This was fixed with 85de0d2

PR #869, opened 2016-08-30
Created external link SVG

ahalbert:
#822 

I added a css class for links that do not match *.perl6.org . It adds the SVG file here at the end: 
https://commons.wikimedia.org/wiki/File:External-link-04-bold-12x12.svg . I'm not married to this file, but I know we can distribute it freely.

I'm a little concerned that I used this URL in the scss.  
`../images/LinkExternal12x12.svg`

On my local deploy of the documentation web server, this was rewritten to /css/images/LinkExternal12x12.svg. I'm not familiar enough with sass to know why this happened.

### DIFF
--- a/assets/sass/style.scss
+++ b/assets/sass/style.scss
@@ -562,3 +562,7 @@ div.highlight span.c-Singleline, div.highlight span.c {
         }
     }
 }
+
+a[href*="//"]:not([href*="perl6.org"])::after {
+    content: url(../images/LinkExternal12x12.svg);
+}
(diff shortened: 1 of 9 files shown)


AlexDaniel:
> This file is licensed under the Creative Commons Attribution-Share Alike 4.0 International license.

This image requires at least attribution. Also, does the ``share alike`` part of it plays any role in our case?

ahalbert:
We do have to release the image under the same license. CC-Share-Alike

ahalbert:
Closing after screwing up in git. Will come back to fix.

PR #870, opened 2016-08-31
corrections, improvements, notes modified or added, unimplemented fea…

tbrowder:
…tures marked NYI (after testing on late model rakudo)

### DIFF
--- a/doc/Type/Str.pod6
+++ b/doc/Type/Str.pod6
@@ -340,9 +340,8 @@ Directives guide the use (if any) of the arguments. When a directive
 passed is to be formatted into the string to be created.
 
 NOTE: The information below is for a fully functioning C<sprintf>
-implementation which hasn't been achieved yet due, in part, to Rakudo
-issues with native uint64 types. Formats or features not yet
-implemented are marked NYI.
+implementation which hasn't been achieved yet. Formats or features not
+yet implemented are marked NYI.
 
 The directives are:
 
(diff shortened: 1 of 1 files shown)


jonathanstowe:
In order to not lose the corrections I'll go with merge but I'm wondering if the NYI parts should actually just be removed or commented out or something

tbrowder:
Altai man seemed to want NYI, and I agree.  Maybe there is a better way to do it, though.  I guess it wouldn't hurt to look at the appropriate spec and maybe get jnthn or moritz to take a look.

Altai-man:
Indeed, this PR really has corrections and clarifications, so it's good for me. But I'm not the main maintainer(not even the most productive contributor) here, so it's just my opinion.

About the spec, https://design.perl6.org/S32/Str.html#sprintf says "mostly identical to the C library sprintf function" and describes some parts of it. The grammar there(and https://github.com/perl6/nqp/blob/b403469d0759a3264e847c926e5bc7e6ca732c45/src/HLL/sprintf.nqp) can give us thought that modifiers weren't planned. Although, they are described here.
We have an exception for unsupported directives, but `%ll` throws `unknown sequence` instead of it.

To be sure, we need a clarification about "how exactly is it mostly identical". If modifiers would be implemented after all, it is good as it now. If they wouldn't be implemented, we need to just remove this part of the description.

tbrowder:
I agree whole-heartedly. Much of the C sprintf (and friends) modifiers are required because of strict typing and catching errors.  Perl 6 shouldn't really need such.  

And the '$' arg selections are, IMHO, not useful at all.  In my over 30 years of C/C++ experience I haver never used it and I don't ever remember seeing them used.

PR #877, opened 2016-09-01
Added external link SVG image.

ahalbert:
#822 
I added the credits per CC guidelines for attribution. I wanted it to be in the HTML, so I added the credits to a comment in the page footer, rather than the CREDITS file


zoffixznet:
Thanks a lot!

PR #902, opened 2016-09-11
Correct info about closing an async socket

zostay:


zostay:
See [RT#129244](https://rt.perl.org/Ticket/Display.html?id=129244) for some additional discussion I had with myself on the subject.

Issue #903, opened 2016-09-12
Explain the differences between 4 different ways to use return type constraints


AlexDaniel:
``` perl6
       sub foo(--> Int)      {}; say &foo.returns # (Int)
       sub foo() returns Int {}; say &foo.returns # (Int)
       sub foo() of Int      {}; say &foo.returns # (Int)
my Int sub foo()             {}; say &foo.returns # (Int)
```

PR #908, opened 2016-09-15
Added documetation on function return types.

ahalbert:
#903


jonathanstowe:
I'm going to merge this but it does need a small change, that is that it is only a "compilation error" if the return value is known at compile time.  This probably needs to be expanded.

PR #936, opened 2016-10-04
Issue #926 produce method is not documented

ggoebel:
Add documentation for produce based on reduce (#926)


doc/Type/List.pod6
```diff
@@ -709,7 +709,7 @@ meta-operator provides a syntactic shortcut:
 Since C<reduce> is an implicit loop, it responds to C<next>, C<last> and
 C<redo> statements inside C<&with>:
 
-    say (2,3,4,5).reduce: { last if $^a > 7; $a + $^b }; # says 9
+    say (2,3,4,5).reduce: { last if $^a > 7; $^a + $^b }; # says 9
```
AlexDaniel: What's the point of this change?

PR #937, opened 2016-10-05
Fix #913: Add documentation for shell and command

### LINKED ISSUE #913: Proc.command undocumented
`Proc` [recently received](https://github.com/rakudo/rakudo/commit/c4fd9f515d) `@.command` attribute that contains the command and its args that the Proc was run with.

If it's not in roast, it needs to be added there as well.

zostay:
Also correct a few other tidbits.


gfldex:
ENOSPEC filed as https://github.com/perl6/roast/issues/166

PR #946, opened 2016-10-07
A few improvements

fluca1978:
I believe these little changes could make the object documentation more clear.
I've also fixed what I believe is a miswritten header.


tbrowder:
looks good to me also&mtag;merged

PR #961, opened 2016-10-11
A few enhancements on 5-to-6-operators

fluca1978:
Reading the page about the 5-to-6 operators I made a few corrections that hopefully make the page more clear.

### DIFF
--- a/doc/Language/5to6-perlop.pod6
+++ b/doc/Language/5to6-perlop.pod6
@@ -71,7 +71,7 @@ Numeric type.
 
 Unary <\> is no more. If you really want to take a reference to an existing
 named variable, you can use item context, like so: C<$aref = item(@array)>
-You can get a reference to a named subroutine by using the C<&> sigil: <$sref
+You can get a reference to a named subroutine by using the C<&> sigil: C<$sref
 = &foo> Anonymous arrays, hashes, and subs return their references during
 creation.
 
--- a/doc/Language/5to6-perlop.pod6
+++ b/doc/Language/5to6-perlop.pod6
@@ -11,10 +11,10 @@ equivalents with notes on variations between them where necessary.
 
 =head1 NOTE
 
-I will I<not> be explaining the operators in detail. This document is an
-attempt to guide you from the operators in Perl 5's perlop document to
+This documentI  I<does not> explains the operators in detail.
+This document is an attempt to guide you from the operators in Perl 5's perlop document to
 their equivalents in Perl 6. For full documentation on the Perl 6
-equivalents, please see the Perl 6 documentation.
+equivalents, please see the L<Perl 6 documentation|/language/operators>.
 
 =head2 Operator Precedence and Associativity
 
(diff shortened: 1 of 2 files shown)


fluca1978:
Please note that I've added also 41a0f94 as a possible substitution of the TODO in the type system for the class concept.
If this is not fine, we can split it into two separate pull requests.

jonathanstowe:
I think that's fine, it could probably do with a link to the https://docs.perl6.org/language/classtut or something but we can fix that later :)

jonathanstowe:
All done! Thanks!

doc/Language/5to6-perlop.pod6
```diff
@@ -11,15 +11,15 @@ equivalents with notes on variations between them where necessary.
...
+This document is an attempt to guide you from the operators in Perl 5's perlop document to
 their equivalents in Perl 6. For full documentation on the Perl 6
-equivalents, please see the Perl 6 documentation.
+equivalents, please see the L<Perl 6 documentation|/language/operators>.
 
 =head2 Operator Precedence and Associativity
 
 The operator precedence table is somewhat different in Perl 6 than it is in
-Perl 5, so I will not detail it here. If you need to know the precedence and
+Perl 5, so it will not be detail  here. If you need to know the precedence and
```
AlexDaniel: “not be detail”?

doc/Language/5to6-perlop.pod6
```diff
@@ -161,10 +161,10 @@ In list context, C<..> operates as the range operator and should not
 need to be changed. That said, there are exclusionary range operators
 that may be useful. These are:
 
-    infix  ..^  Excludes endpoint. C<1..^5; # 1 2 3 4 >
-    infix  ^..  Excludes startpoint. C<1^..5; # 2 3 4 5>
-    infix  ^..^ Excludes both start- and endpoint. C<1^..^5; # 2 3 4>
-    prefix ^    From 0, excludes endpoint. C<^5; # 0 1 2 3 4>
+    infix  ..^  Excludes endpoint. C<1..^5>; # 1 2 3 4 >
```
AlexDaniel: hmmm, what is happening here?
```
C<1..^5>; # 1 2 3 4 >
```
So we have two closing ``>`` and one ``<``.

fluca1978: Correct, shame on me I didn't look at the ending bracket.
I suspect the solution is to remove the C<> inline block at all. 
Is it possible to cherry pick this one? <https://github.com/perl6/doc/commit/a19a6b775f4c3391bb14280d891e8a7ecde4eed3>

doc/Language/5to6-perlop.pod6
b2gills: How about:

> ... than it is in Perl 5, and will not be discussed in detail here.

dha: It could just be changed to "detailed", but I think b2gills' version is probably best.

PR #962, opened 2016-10-11
Link to the classtut in the class explaination.

fluca1978:
From suggestion via https://github.com/perl6/doc/pull/961


jonathanstowe:
Nice one, cheers!

PR #963, opened 2016-10-11
Explain twigils to access members and methods within a class.

fluca1978:
I'm happier with this explaination of the twigils to access attributes or methods from the inside/outside of a class.
I believe the previous explaination from my pull request could lead to misunderstanding about twigils.

### DIFF
--- a/doc/Language/typesystem.pod6
+++ b/doc/Language/typesystem.pod6
@@ -80,8 +80,15 @@ can keep an internal state not shared with other instances of the same class.
 The behavioural part of each object is defined in the class using I<methods>,
 while the internal state is provided via I<attributes> (variable scoped within the
 class).
-Both methods and attributes (also called members) are accessed via the C<$.> twigil or,
-in the case they are defined as I<private> via the C<$!>.
+Within the class code, attributes (also called members) can always be accessed (set or get)
+via a twigil made by the attribute sigil and a C<!> (e.g., C<$!>, C<@!>, C<%!>), even
+if they have been declared as private. To access an attribute in read only mode (get)
+within the class code use a twigil made by the variable sigil and a C<.> (e.g., C<$.>, C<@.>, C<%.>).
+Methods can be accessed (i.e., invoked) within the class code  via the C<self.>
+derefencing or C<self!> one if declared as private.
+On the instance (i.e., outside the class code) both public attributes and methods
+are accessed via the C<.> operator applied to the object instance variable.
+
 
 For more information about classes see the L<classtut|/language/classtut> tutorial.
 


zoffixznet:
IMO, this muddies up the relationship between `!` and `.` twigils and doesn't explain *why* you can use `!` instead of `.`. There are also some inaccuracies in the text. The whole thing can be explained simpler:

*) `has`, together with the `$!`, `@!`, `%!`, and `&!` twigils, declares private class attributes
*) You can automatically generate accessors for those attributes by using the `$.`, `@.`, `%.`, and `&.` twigils instead.
*) `$.attr` calls the auto-generated accessor method to access `$!attr`, so within the class, where `$!attr` is visible, it makes sense to use `$!attr` directly

As for inaccuracies:
*) The `.` twigils do not give you read-only mode. They're just accessors for `!` attributes and they can be read-write if they've been declared with `is rw` trait
*) Methods and accessors to class variables (not attributes) can also be called on the type object, so you don't need an instance.

I've not read the context of the text, but there's also such a thing as class variables:
*) `my` (or `our`), together with the `$.`, `@.`, `%.`, and `&.` twigils declare class variables (which are shared among instances), but same as with the attributes, the `.` in the twigil indicates that an accessor method is automatically generated.
*) You can't use `my` (or `our`) with `!` twigils and can't access these class variables with the `!` twigils, so you always have to use the accessor method: the `.` twigil.

coke:
Note that 'derefencing' causes this PR to fail 'make xtest'

doc/Language/typesystem.pod6
```diff
@@ -80,8 +80,15 @@ can keep an internal state not shared with other instances of the same class.
 The behavioural part of each object is defined in the class using I<methods>,
 while the internal state is provided via I<attributes> (variable scoped within the
 class).
-Both methods and attributes (also called members) are accessed via the C<$.> twigil or,
-in the case they are defined as I<private> via the C<$!>.
+Within the class code, attributes (also called members) can always be accessed (set or get)
+via a twigil made by the attribute sigil and a C<!> (e.g., C<$!>, C<@!>, C<%!>), even
```
VynceMontgomery: I feel like the docs mostly use "twigil" to refer only to the second symbol -- the symbol beTWeen the sigil and the variable name -- and this uses it to mean the TWo symbols together. We should standardize on one.

b2gills: I have thought as twigil as referring to just the secondary sigil.

Talking about it this way would mean we can talk about just `.` `!` `*` `?` changing the behaviour of various variables. `«$ @ % &» X~ «. ! * ?»`

ajs: I would start with the thing that the user can do, and then move on to the name of the thing. So, "Within the class code, attributes (also called members) can always be accessed via a notation such as C<$!attrname> where the C<!> is called a "twigil" and indicates that the reference is to an attribute, not a simple variable." ... or something like that.

coke:
I agree with @zoffixznet , let's reject this version. Thanks for the effort, @fluca1978 .

AlexDaniel:
I agree with most of the mentioned issues. Alright then, seems like we will have to reject this particular pull request. However, current text is indeed less than awesome and can be clarified, so feel free to submit another PR (or PRs) if you still want to improve it (comment by @zoffixznet should give a couple of good ideas on how to do that). Thanks for your effort.

Issue #960, opened 2016-10-10
Using » in place of .map is a trap


AlexDaniel:
There are actually two cases to cover.

I've seen it several times on `#perl6` when people tried to do this when they first learned aout `»`:

``` perl6
<a b c d>».say # Oops!
```

Well, okay, it works today, but the idea is that the execution order is not defined, so doing `».say` is probably not the best idea.

OK, I get it. But still, if I don't care about the execution order, I can use `»` anywhere I use `.map`, right?

Oops.

See this:

```
say (<a b c d>, <x y z>).map: *.elems # (4 3)     ← OK, that's what we want
say (<a b c d>, <x y z>)».elems       # (4 3)      YEAH, it works, great!

# but we may want to use something more
# complex than just a method call. For example this:
say (<a b c d>, <x y z>).map: *.elems × -1     # (-4 -3)
say (<a b c d>, <x y z>)».&(*.elems × -1)      # ((-1 -1 -1 -1) (-1 -1 -1)) – ??? Why?
say (<a b c d>, <x y z>).deepmap: *.elems × -1 # ((-1 -1 -1 -1) (-1 -1 -1)) – OK, so it's like deepmap in this case…
```

Perhaps see these conversations:
- https://irclog.perlgeek.de/perl6/2016-10-06#i_13343973
- https://irclog.perlgeek.de/perl6/2016-10-08#i_13362495
- https://irclog.perlgeek.de/perl6/2016-10-10#i_13374587

AlexDaniel:
By the way, we have it mentioned in [/operators](https://docs.perl6.org/language/operators#postfix_%C2%BB._/_postfix_%3E%3E.), but a separate section in traps is not going to hurt.

Issue #959, opened 2016-10-10
｢once｣ is a potential pitfall (and state variables are too)


AlexDaniel:
Speaking of [good ideas that are actually bad](#364), we should also mention ``once`` and state variables.

See this discussion: https://irclog.perlgeek.de/perl6/2016-10-08#i_13363806 (it is actually pretty long).

TL;DR that's how it is supposed to work. In other words, common expectation of users that “once” means once is a pitfall.

AlexDaniel:
Resolved in 7bd27cb

Issue #602, opened 2016-06-17
Add ｢@a.push: |@b｣ as a trap to avoid


AlexDaniel:
We should add it here: http://doc.perl6.org/language/traps

See this: http://irclog.perlgeek.de/perl6/2016-06-17#i_12685558

AlexDaniel:
[This](https://docs.perl6.org/language/traps.html#Argument_Count_Limit) looks good?

AlexDaniel:
This looks alright to me.

PR #964, opened 2016-10-11
Add ^ twigil to traps. Fixes #364

### LINKED ISSUE #364: List $^placeholders as one of the traps
http://doc.perl6.org/language/variables#The_^_Twigil

> Although it is possible to use nearly any valid identifier as a placeholder variable, it's recommended to use short names or ones that can be trivially understood in the correct order, to avoid surprise on behalf of the reader.

In other words, some people may be tempted to have meaningful placeholder names while expecting it to work in the order of appearance. This is definitely a potential [trap](https://doc.perl6.org/language/traps).

nicqrocks:
This adds some explanation as to how the ^ twigil placeholder variables work and how some people can run into troubles with them.

### DIFF
--- a/doc/Language/traps.pod6
+++ b/doc/Language/traps.pod6
@@ -451,19 +451,7 @@ throws. Explicitly sinking it inside the C<try> avoids the issue:
 
 Using the C<^> twigil can save a fair amount of time and space when writing out small blocks of code. As an example:
 
-	for ^8 -> $first, $second { say $first + $second;	}
-
-can be shortened to just
-
-	for ^8 { say $^first + $^second; }
-
-=head1 Using Shortcuts
-
-=head2 The ^ twigil
-
-Using the C<^> twigil can save a fair amount of time and space when writing out small blocks of code. As an example:
-
-	for ^8 -> $first, $second { say $first + $second;	}
+	for ^8 -> $first, $second { say $first + $second; }
 
 can be shortened to just
 
--- a/doc/Language/traps.pod6
+++ b/doc/Language/traps.pod6
@@ -451,11 +451,11 @@ throws. Explicitly sinking it inside the C<try> avoids the issue:
 
 Using the C<^> twigil can save a fair amount of time and space when writing out small blocks of code. As an example:
 
-	for ^8 -> $first, $second { say $first + $second; }
+	for 0..7 -> $first, $second { say $first + $second; }
 
 can be shortened to just
 
-	for ^8 { say $^first + $^second; }
+	for 0..7 { say $^first + $^second; }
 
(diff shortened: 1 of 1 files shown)


nicqrocks:
Does it look like it needs anything or has something left out?

MasterDuke17:
The travis complaint is about tabs, and the rest of docs don't rely on line-wrapping, but the actual content looks good.

b2gills:
The `^8` should be changed to something else so that a new Perl 6 programmer is less likely to get confused

( It initially confused [\[Coke\]](https://irclog.perlgeek.de/perl6/2016-10-11#i_13378023) for example )

coke:
(confusing) specifically, using the up to prefix operator (^) when we're already talking about the ^ twigil.

AlexDaniel:
Just change it to ``1..4``.

AlexDaniel:
Ah. Oops, there are two instances of using the range operator. Yeah… don't know about the first one. ``0..7`` anyone?

nicqrocks:
Yeah, I was thinking about that myself. I think 0..7 (and similar) would be a good change.

nicqrocks:
Ok, that should be good. Can you guys think of anything else?

AlexDaniel:
It's good. But:

> The trouble arises when a person wants to use more complex names for the variables.

``$^first`` and ``$^second`` are already too complex. I don't think that anything besides 1 character name is justified.

What about demonstrating the right example instead?

It will read like this:

```
Using the C<^> twigil can save a fair amount of time and space when writing out small blocks of code. As an example:

	for 0..7 -> $a, $b { say $a + $b; }

can be shortened to just

	for 0..7 { say $^a + $^b; }

The trouble arises when a person wants to use more complex names for the variables.

…
```

AlexDaniel:
Also, consider using ``1..4`` instead of ``0..3``. It will correspond to ``$^one $^two $^three $^four`` variables better.

AlexDaniel:
Well the output has to be changed as well then: ``# 1 3 2 0`` → ``# 2 4 3 1``.

Also, please replace tabs with spaces (so that travis does not complain).


Other than that it is mergeable.

nicqrocks:
Ok 👌

tommystanton:
@nicqrocks (@MasterDuke17, @AlexDaniel): If you have [Ack installed](http://beyondgrep.com/install/), you can find out where the tabs are by running:

```text
$ ack '\t' doc/Language/traps.pod6
```

AlexDaniel:
@coke perhaps [``tabs.t``](https://github.com/perl6/doc/blob/master/t/tabs.t#L20) should also mention line numbers?
… or maybe it should even fix it itself when run with some env variable (``FIX_WHITESPACE`` ?).

Personally I have no issues with any of that, as I am able to configure any text editor to display whitespace. But a more helpful ``tabs.t`` could help contributors a lot, I think.

AlexDaniel:
I think you did a great job.

There are other tickets related to traps: #602 #959 #960, perhaps consider working on these too.

nicqrocks:
Sure, I'll take a look! 😃

PR #965, opened 2016-10-12
Typo fixes in about.pod6 and predef search items

sjn:


Altai-man:
Seems as a good fixes, so I'll merge it. Thanks!

PR #968, opened 2016-10-13
Add 'once' Block to Traps #959

nicqrocks:
This was made in response to issue #959. Added  an explanation of why the 'once' block would run more than one time.


nicqrocks:
I added some more explanation as to why it happens and how to replicate the behavior, but I'm not sure about the wording. It feels kind of repetitive to me.
Any suggestions?

AlexDaniel:
Examples work fine. Let's merge and see if anybody objects :)

PR #974, opened 2016-10-18
Sigil typo in MixHash example

jalbertcruz:


timo:
I'm not sure this is actually wrong. When you use the % sigil, you need the := operator to bind it. When you use a $ sigil, you can just assign it. So since the example uses := it makes sense to assume it was meant to be that way.

jalbertcruz:
Oh!, yes, It is. But then the problem is in "$h{pi} = 1;" it must be "%h{pi} = 1;"
I am going to change the pull request.

zoffixznet:
Thanks a lot!

PR #978, opened 2016-10-20
Resolves #977: Correct callable args on splice

### LINKED ISSUE #977: The description of the second callable arg to splice is wrong
In `Array.splice`, it says the second argument, when `Callable`, is called with two arguments. It is actually just called with one.

zostay:


coke:
This PR fails "make xtest".

```
# Failed test 'doc/Type/Array.pod6 (line 169): Must not have any trailing whitespace.'

```

zoffixznet:
Thanks a lot!

(I'll fix the xtest issue manually from a local checkout)

PR #980, opened 2016-10-20
inline predefined-search-items

coke:
This removes the top level file predefined-search-items and inserts it where it's used.

It's all static text, so having it separated doesn't add anything to the process, and instead complicates it for no gain.


AlexDaniel:
IRC log: https://irclog.perlgeek.de/perl6/2016-10-20#i_13437231

Altai-man:
Can't see the irc log currently, but anyway. It was me who proposed/implemented this separating, when it was seen as a good(or, at least, real) way to deal with some items from https://github.com/perl6/doc/issues/728. But the issue stacked in the end for some reasons(lack of manpower, I suppose) and, because of this, `predefined-search-items` file was never expanded as it supposed.
Hence, can be merged without worries, judging by the PR size.

PR #987, opened 2016-10-26
fix typo in doc about how to fix typo in doc, doc

VynceMontgomery:
the first example "=head2" block was spelled as "d2" instead of "=head2" which probably did actually blow some minds... it at least caused a misfire in mine.

(fixes apparent mistake from https://github.com/perl6/doc/commit/fb947c098cd6e9cec6874f5d77d38cee08f6a7a1 )


Altai-man:
Oh, thanks! Emacs marco tends to do such things once in a while...

PR #988, opened 2016-10-26
minor wording fix

VynceMontgomery:
This method was referred to as the second method, which was confusing since the method before it was also referred to as the second method. Removed the ordinal; just use the name, for future-proofing.


Altai-man:
Thanks!

Issue #971, opened 2016-10-16
Another precedence trap: -1**5


AlexDaniel:
Sometimes I do this:

Code:
```perl6
say -1**2
```

Output:
```
-1
```

When I actually mean this:

```perl6
say (-1)**2
```

Output:
```
1
```

LLFourn:
is it intended that infix ** be tighter than prefix - ?

pierre-vigier:
It seems normal for me that ** is tigher that -, after a quick check, python, ruby and Octave have the same behavior as well

labster:
Assuming only real numbers, in normal mathematical notation:

y = -x²
z = -3²

Would you expect y to be positive or negative?  The same for z?  I would definitely expect a negative number for y.

AlexDaniel:
perhaps it means that unicode ops should have looser precedence? 🤷

labster:
Aren't all ops Unicode?

LLFourn:
@labster I think @AlexDaniel  means that the more mathematical notation of ² could be tighter while `**2` be looser. I expected `**` to be looser simply because it looks looser to me. What I expected is apparently not the norm so nvm me :)

VynceMontgomery:
(subject of issue is misleading, because answer is the same either way)

- x**2 should definitely square before negating. 
-5 ** 2 sure feels to me like it should negate before squaring. 
whitespace affecting precedence is a sketchy place to be, but Perl is supposed to DWIM and in this case, it doesn't feel like - is a unary operator, it feels like it's part of the number literal. YMMV, of course.

VynceMontgomery:
in fact, -2 is given as an example of an integer literal, and yet it appears not to be parsed that way. this seems ... wrong.

jnthn:
Indeed, negatives are not parsed as literals (with the exception of `-Inf`), so if something gives -2 as an example of an integer literal it probably warrants a clarification.

VynceMontgomery:
What reasoning (and evidence) is there for -Inf parsing as a literal and not as unary - applied to Inf?

jnthn:
Can't say much about reasoning. As for evidence, I checked it in the grammar; see https://github.com/rakudo/rakudo/blob/nom/src/Perl6/Grammar.nqp#L3386

b2gills:
Perhaps `-x ** y` should produce an error telling you to use parens to disambiguate.

pierre-vigier:
[...]

PR #991, opened 2016-10-26
Clarifications on numeric literals

VynceMontgomery:
Belief in negative numeric literals creates false expectations regarding precedence conflicts (see #971)

### DIFF
--- a/doc/Language/syntax.pod6
+++ b/doc/Language/syntax.pod6
@@ -316,7 +316,7 @@ number:
 Integers default to signed base-10, but you can use other bases. For details,
 see L<Int|/type/Int>.
 
-    -2
+    -2          # actually not a literal, but unary - operator applied to numeric literal 2
     12345
     0xBEEF      # base 16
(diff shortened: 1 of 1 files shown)


AlexDaniel:
This pull request is OK. However, let's keep it open until [this](https://irclog.perlgeek.de/perl6/2016-10-27#i_13473779) is resolved (or alternatively, open another issue for that).

jonathanstowe:
That's actually interesting that the symbol which I can't type and the Inf have different precedence.

AlexDaniel:
It is not interesting, it is sad.

VynceMontgomery:
Yeah, I'm gonna go with both interesting and sad. And implementation
specific and not necessarily correct by spec.

.

zoffixznet:
Thanks for the PR.

For -Inf: it's neither interesting nor sad, but a bug. It's an unwanted sideeffect of making -Inf work as literal in signatures and is to be fixed soon.

PR #1010, opened 2016-11-14
Show how to go from signal numbers to Signal objects to use with the signal method

samcv:
Also add link to enum type.


samcv:
I agree. I made those changes. I think it goes much better in its own paragraph.

PR #1012, opened 2016-11-17
Attempt to improve the documentation of routines zip, and roundrobin.

bazzaar:
I've built the html docs (post-edit's to doc/Types/List.pod6) - looks ok
please take a look at new doc for List routine 'zip', and List sub 'roundrobin'

not sure what else to type here, or what files to attach, if any ...


zoffixznet:
Thanks a lot!

PR #1018, opened 2016-11-21
I believe this emphasis was wrong

ZzZombo:
At first I got the impression I DO HAVE to take care with ranges.


zoffixznet:
Thanks a lot.

PR #1020, opened 2016-11-26
Remove duplicate done-testing text

mienaikage:
Block of text immediately above removed lines is identical.


AlexDaniel:
Yeah, indeed. Thank you!

PR #1022, opened 2016-11-27
Fix example that mixed up q:w and Q:w

dmaestro:
The original example had very strange behavior; not sure how to
document this:

    Q:w { [ ] \{ \} } eqv ('[', ']', '{', '}') # False

    > Q:w { [ ] \{ \} }.perl
    ("[", "]", "\\\{\\}")

### DIFF
--- a/doc/Language/quoting.pod6
+++ b/doc/Language/quoting.pod6
@@ -162,7 +162,8 @@ X<|qw word quote>
 
 =for code :allow<B L>
 B<qw|>! @ # $ % ^ & * \| < > B<|> eqv '! @ # $ % ^ & * | < >'.words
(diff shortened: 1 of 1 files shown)


zoffixznet:
The output of the original example makes sense to me: `Q` does not parse escapes, so the backslashes come out literally (the the stuff it was `eqv` with is incorrect).

dmaestro:
What didn't make sense to me was getting only three elements, instead of four. The braces are the same as the quote delimiter, so it kind of makes sense, but then the space gets thrown away (after?) the word splitting. Still seems odd ...

zoffixznet:
Yeah, it has since been reported as a bug: https://rt.perl.org/Ticket/Display.html?id=130205#ticket-history

PR #1034, opened 2016-12-04
Small bunch of changes to make examples compileable.

Altai-man:
It includes such types of fixes:
1)Missed semicolons;
2)Typos;
3)Endless code without :skip-test;
4)Convert flags table to actual table, not code.

Review is welcome.

### DIFF
--- a/doc/Type/Int.pod6
+++ b/doc/Type/Int.pod6
@@ -36,8 +36,9 @@ respectively:
 
 These notations allow you to use variables, too:
 
-    :16($ninety-nine) # 153
-    :100[99, $two, 3] # 990203
+    my Int $ininety-nine = 99;
+    :16($ninety-nine); # 153
+    :100[99, $two, 3]; # 990203
 
 =head1 Methods
 
--- a/doc/Type/WrapHandle.pod6
+++ b/doc/Type/WrapHandle.pod6
@@ -8,8 +8,8 @@ class WrapHandle { ... }
 
 C<WrapHandle> is created and returned by L<wrap|/type/Routine#method_wrap>. It's only use is to unwrap wrapped routines. Either call L<unwrap|/type/Routine#method_unwrap> on a routine object or call the method C<restore> on a C<WrapHandle> object.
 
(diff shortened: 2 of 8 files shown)


jonathanstowe:
Seems fairly uncontroversial to me :)

doc/Type/Int.pod6
```diff
@@ -36,8 +36,9 @@ respectively:
 
 These notations allow you to use variables, too:
 
-    :16($ninety-nine) # 153
-    :100[99, $two, 3] # 990203
+    my Int $ininety-nine = 99;
+    :16($ninety-nine); # 153
+    :100[99, $two, 3]; # 990203
```
AlexDaniel: but what is $two?

doc/Type/WrapHandle.pod6
```diff
@@ -8,8 +8,8 @@ class WrapHandle { ... }
 
 C<WrapHandle> is created and returned by L<wrap|/type/Routine#method_wrap>. It's only use is to unwrap wrapped routines. Either call L<unwrap|/type/Routine#method_unwrap> on a routine object or call the method C<restore> on a C<WrapHandle> object.
 
-    sub f(){ say 'f was called' }
-    my $wrap-handle = &f.wrap({ say 'before'; callsame; say 'after' })
+    sub f(){ say 'f was called' };
```
AlexDaniel: a semicolon after ``sub f(){ }`` ? Why?

doc/Type/Int.pod6
Altai-man: Should be added too! Thanks.

doc/Type/WrapHandle.pod6
Altai-man: Because we can't automatically compile and check this example otherwise?

doc/Type/Int.pod6
tommystanton: Looks like there's a small typo introduced in the first line addition here (`$ininety-nine`).

PR #1035, opened 2016-12-06
Consistent return types

Altai-man:
See https://github.com/perl6/doc/issues/1024#issuecomment-264660360(and the whole issue history) for the original discussion.

To make docs signatures more consistent and to make them more version-agnostic this PR changes all `returns` to `-->` form.

Discussion is welcome!

The key points:
1)We *don't* prohibit "returns" form anywhere, we just want to use single form in all signatures over the docs.
1.5)It is tested, so maintenance burden will be not so big.
2)We know 'returns" just reads nicely.
3)The otherwise change ("-->" to "returns") was rejected. Not the least reason is that "returns" can do less than "-->'.


moritz:
I'm not in favor of such tests, because they raise the barrier to contribution.

That said, I contribute very little to the docs these days, so don't let that stop you.

coke:
Note that we've had xt/ tests for some time now that we can run ad hoc, and don't break the build. (I wish there was a way to force them to run for PRs, however.)

Altai-man:
Yeah, you're totally right! Thanks for notice.

It seems that nobody is against currently, so I'll continue to work. I will move the test to xt/ and resolve conflicts tomorrow.

t/return-type.t
```diff
@@ -10,18 +10,13 @@ my @files;
...
     my @lines;
     my Int $line-no = 1;
     for $file.IO.lines -> $line {
-        if so $line ~~ /(method|sub) .+? '-->'/
-        && $line !~~ /'--> True'/
-        && $line !~~ /'--> False'/ {
-        @lines.push($line-no);
-        }
-    $line-no++;
+        if so $line ~~ /(multi|method|sub) .+? ')' ' '+? 'returns' ' '+? (<alnum>|':')+? $/ {
```
AlexDaniel: ``\s+`` instead of ``' '+?``?

AlexDaniel: Also, it won't catch things that are spread over several lines. Not sure if that's a big problem.

AlexDaniel:
To those reading the discussion: **we like all tests**. Anything you can automate will be appreciated.

It's just that we will put some tests in ``xt/`` to cause less troubles for contributors, but we are still in favor of all tests.

t/return-type.t
Altai-man: >\s+
Yeah, my bad.
About several lines. There are really a few such signatures, firstly, and it will be a noisy rule, secondly.

But if anyone feels ready to write one - welcome.

Altai-man:
The conflicts were solved, so we can merge this freely and I'll add a test in the master then.

PR #1038, opened 2016-12-07
Document that identifiers like `'` and `-` can't be used after Q q, or qq quoting constructs

samcv:


samcv:
Added `(…)` example and rephrased it so that I wasn't using `'` inside the `Q'…'`

samcv:
Is this more clear?

zoffixznet:
Thanks a lot.

Issue #1039, opened 2016-12-07
Syntax doc should advertise and describe unicode identifiers


ronaldxs:
The Syntax doc, [docs.perl6.org/language/syntax](https://docs.perl6.org/language/syntax), has a sentence about Unicode that says that Perl 6 is Unicode text but does seem to advertise applications of Unicode source, other than maybe the hyper operator.  I think it might be helpful for the description of identifiers to mention the use of alternative lettering/writing systems from Unicode as well as mathematical symbols as mentioned in a [Rosetta Code example](https://rosettacode.org/wiki/Unicode_variable_names#Perl_6).  It may also be useful to discuss the use of Unicode subscripts and superscripts and the exclusion of numeric Unicode subscripts and superscripts from identifiers (see discussions [IRC #perl6 2016-04-14](https://irclog.perlgeek.de/perl6/2016-04-14#i_12340102) and an [AlexDaniel gist](https://gist.github.com/AlexDaniel/c89bd2786f9b63f31e4c)).  My knowledge in this area is limited and there are likely to be other applications of Unicode source code that I have overlooked.

AlexDaniel:
What about https://docs.perl6.org/language/unicode_texas ?

ronaldxs:
See subsequent IRC discussion: [https://irclog.perlgeek.de/perl6/2016-12-07#i_13700922](https://irclog.perlgeek.de/perl6/2016-12-07#i_13700922).

jonathanstowe:
```
class वर्ग { }; say वर्ग.new
```

ronaldxs:
First cut at documentation update and opportunity for anyone who would like to comment before PR:

[https://github.com/ronaldxs/doc/commit/b7e4d5354cc1b4b2b99d59471044e72812af00ef](https://github.com/ronaldxs/doc/commit/b7e4d5354cc1b4b2b99d59471044e72812af00ef)

ronaldxs:
A partial resolution was merged but leaving open ftm while I chase down some loose ends.

[...]

PR #1042, opened 2016-12-09
Examples and ref. to further explanation for Unicode in Perl 6 syntax

ronaldxs:
Attempt to resolve Issue #1039

### DIFF
--- a/doc/Language/syntax.pod6
+++ b/doc/Language/syntax.pod6
@@ -18,7 +18,9 @@ errors and give good error messages.
 =head1 Lexical Conventions
 
 Perl 6 code is Unicode text, current implementations support UTF-8 as the
-input encoding.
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Something I have missed the previous time:

> Subscript numerals are TBD.

Usually we document what we have right now (there are some existing places where this has to be fixed, btw). Anyway, I think I can change it myself (don't want to go through the whole review thing again).

AlexDaniel:
Actually, forget what I've just said. We have no tests for this, so currently it is indeed TBD :(

(the only somewhat related test is [this one](https://github.com/perl6/roast/blob/da6fde44ab97cdcbb50bf181528f55671914612c/S15-literals/numbers.t#L58), but it is probably not relevant enough).

PR #1052, opened 2016-12-14
Document `is deafult` for Subs

### LINKED ISSUE #1040: ｢is default｣ for subs is not searchable
```
<babydrop> m: multi foo (Int) is default {say "here"}; multi foo (Int) {say "there"}; foo 42
<camelia> rakudo-moar a98724: OUTPUT«here␤»
<babydrop> m: multi foo (Int) {say "here"}; multi foo (Int) is default {say "there"}; foo 42
<camelia> rakudo-moar a98724: OUTPUT«there␤»
<babydrop> Neat.
<babydrop> TIL
<AlexDaniel> hmmm
<AlexDaniel> so what does it do exactly?
<moritz> it disambiguates ambiguous dispatches
```
(IRC log: https://irclog.perlgeek.de/perl6/2016-12-08#i_13703911)

nicqrocks:
Added explanations and examples to document what `is default` does for Subs. This is to fix #1040.

### DIFF
--- a/doc/Type/Sub.pod6
+++ b/doc/Type/Sub.pod6
@@ -61,4 +61,26 @@ to it (or even themselves) and we can apply traits to objects at runtime.
     &trait_mod:<is>(&bar, :foo);
     # OUTPUT«is foo called␤is foo called␤»
 
+There is a special trait for C<Sub>s called C<is default>. This trait is
+designed as a way to disambiguate C<multi> function calls that would normally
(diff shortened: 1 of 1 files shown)


AlexDaniel:
This won't make it searchable, right?

nicqrocks:
I'm really not sure if it will or not, as I am unfamiliar with how it
works. I am guessing that it will not be searchable.

Issue #999, opened 2016-10-29
implement some kind of fuzzy matching for search


timo:
for example, ideally "startswith" and "starts_with" both should find "starts-with".

coke:
The PR for was merged (prematurely?) - there is a big performance slowdown for searches with small # of characters.

One possible fix is to only use getEditDistance() if the number of input characters is (e.g.) 3 or more - for this to work, we need to find what the default search algorithm was.

I think it's this:

https://github.com/jquery/jquery-ui/blob/master/ui/widgets/autocomplete.js#L638

coke:
Done. We now have fuzzy search with OK performance.

PR #1059, opened 2016-12-16
Coke/levenshtein

coke:
For review only at this point. Need to insure that, unlike the last version, this doesn't have any negative impacts on performance.


coke:
For issue #999

PR #1061, opened 2016-12-19
Use Github/Atom highlighter for <15 minutes builds and better highlighting

samcv:
To use, run `make init-highlights`

then run `make html-highlights`

~15 minute build with my old x220 laptop!!!

P.S. you will need node to use this.
P.P.S. node may spit out some error messages at the end of installing, but everything is probably fine. Just ignore it.

If in doubt that it installed properly, run `./highlights/highlight-filename-from-stdin.coffee` then type `./htmlify.p6` and then it should print out a bunch of html.

P.P.P.S. Make sure to clear your cache. If the code blocks are there but don't see anything highlighted, it probably has the old stylesheet cached.


samcv:
If somebody can check this out and let me know what you think ++

PR #1064, opened 2016-12-20
Style cleanup.

antquinonez:
General style cleanup; added contractions.

### DIFF
--- a/doc/Language/objects.pod6
+++ b/doc/Language/objects.pod6
@@ -307,7 +307,7 @@ Private methods are not inherited by subclasses.
 
 =head2 Submethods
 
-A submethod is a public method that is not inherited to subclasses. The name
+A submethod is a public method that is not inherited by subclasses. The name
 stems from the fact that they are semantically similar to subroutines.
 
 Submethods are useful for object construction and destruction tasks, as well
@@ -927,7 +927,7 @@ of an object, you could write:
 object system, where it's not necessary to create a new meta object for
 every type).
 
-There's a shortcut to prevent from using the same object twice:
+There's a shortcut to keep from using the same object twice:
 
     say 1.^name;                        # Int
     # same as
--- a/doc/Language/objects.pod6
+++ b/doc/Language/objects.pod6
@@ -4,8 +4,9 @@
 
 =SUBTITLE Object Orientation in Perl 6
 
-Perl 6 is an object oriented language at its core, although you can write
-programs in other programming styles.
+Perl 6 provides strong support for object oriented programming. Although 
+Perl 6 allows programmers to program in multiple paradigms, object oriented 
+programming is at the heart of the language.
 
 Perl 6 comes with a wealth of predefined types, which can be classified in
 two categories: normal and I<native> types.
@@ -26,21 +27,21 @@ L<say> "abc"B<.L<uc>>;
 # ABC
 
 This calls the L<C<uc>|uc> method on C<"abc">, which is an object of type
-L<C<Str>|Str>. To supply arguments to a method, add parentheses after the
-method, with arguments inside parentheses:
+L<C<Str>|Str>. To supply arguments to the method, add arguments inside parentheses
+after the method.
 
 =for code :allow<B L>
 my $formatted-text = "Fourscore and seven years ago...".L<indent>B<(8)>;
 
 C<$formatted-text> now contains the above text, but indented 8 spaces.
 
-Separate multiple arguments with commas:
+Multiple arguments are separated by commas:
 
 =for code :allow<B L>
 my @words = "Abe", "Lincoln";
 @words.L<push>("said"B<,> $formatted-text.L<comb>(L</\w+/|/language/regexes>));
 
-Alternately, separate the method and the argument list with a colon:
+Multiple arguments can be specified by separating the argument list with a colon:
 
 =for code :allow<B L>
 say @words.L<join>: '--';
@@ -74,7 +75,7 @@ All objects derive from C<Mu>.
 
 =head2 Type Objects
 
-Types are objects and you can get the I<type object> by writing its name:
+Types themselves are objects and you can get the I<type object> by writing its name:
 
     my $int-type-obj = Int;
 
@@ -140,7 +141,7 @@ another class.
 =head2 Attributes
 X<|Attribute> X<|Property> X<|Member> X<|Slot>
 
-Attributes are variables that exist per object basis. They are where
+Attributes are variables that exist per instance of a class. They are where
 the state of an object is stored. In Perl 6, all attributes are private.
 They are typically declared using the C<has> declarator and the C<!> twigil.
 
@@ -177,8 +178,8 @@ Now, after a C<Journey> object is created, its C<.origin>, C<.destination>,
 and C<.notes> will all be accessible from outside the class, but only
 C<.notes> can be modified.
 
-The above allows creation of C<Journey> objects without origin or
-destination values. To prevent this, provide default values or make
+If an object is instantiated without certain attributes, such as origin or
+destination, we may not get the desired result . To prevent this, provide default values or make
 sure that an attribute is set on object creation by marking an attribute with an L<is required> trait.
 
     class Journey {
@@ -246,8 +247,8 @@ declares a C<!> twigil and generates an accessor method.
 Looking at the code above,
 there is a subtle but important difference between using C<$!origin>
 and C<$.origin> in the method C<describe>. The first is always a lookup of the 
-attribute. It's cheap, and you know that it's the attribute
-declared in this class. C<$.origin> is a method call and thus may
+attribute. The first is an inexpensive and obvious lookup of the attribute.
+C<$.origin> is a method call and thus may
 be overridden in a subclass. Only use C<$.origin> if you want to
 allow overriding.
 
@@ -264,10 +265,10 @@ Method names can be resolved at runtime with the C<.""> operator.
 
 =head2 self
 
-Inside a method, the term C<self> is available, which is bound to the
-invocant, the object that the method was called on. C<self> can be
-used to call further methods on the invocant. Within methods, something like
-C<$.origin> is the same thing as C<self.origin>; however, the colon-syntax for
+Inside a method, the term C<self> is available, which is bound to the invocant 
+(the object the method was called on). C<self> can be
+used to call further methods on the invocant. Within methods, C<$.origin> works 
+the same as C<self.origin>, however the colon-syntax for
 method arguments is only supported for method calls using self, not the shortcut.
 
 Note that if the relevant methods C<bless>, C<CREATE>, C<BUILDALL> of
@@ -288,11 +289,10 @@ attributes in initializers.
 
 =head2 Private Methods
 
-Methods that are not going to be called or are not callable from anywhere outside
-the defining class are declared with an exclamation mark C<!> before the method
-name; such methods are private in the sense they are not visible
-from outside the class that declares them.
-Private methods are invoked with an exclamation mark instead of a dot:
+Methods with an exclamation mark C<!> before the method name are not callable 
+from anywhere outside the defining class; such methods are private in the sense
+that they are not visible from outside the class that declares them. Private 
+methods are invoked with an exclamation mark instead of a dot:
 
     method !do-something-private($x) {
         ...
@@ -307,7 +307,7 @@ Private methods are not inherited by subclasses.
 
 =head2 Submethods
 
-A submethod is a public method that is not inherited by subclasses. The name
+Submethods are public methods that are not inherited by subclasses. The name
 stems from the fact that they are semantically similar to subroutines.
 
 Submethods are useful for object construction and destruction tasks, as well
@@ -436,9 +436,9 @@ This outputs:
 C<Mu.new> calls method L<bless> on its invocant, passing all the named
 arguments. C<bless> creates the new object and then calls method C<BUILDALL>
 on it. C<BUILDALL> walks all subclasses in reverse method resolution order
-(i.e. from L<Mu> to most derived classes) and in each class checks for
-existence of a method named C<BUILD>. If it exists, it's called, again
-passing all named arguments from method C<new> to it. If not, the public
+(i.e. from L<Mu> to most derived classes) and in each class checks for the 
+existence of a method named C<BUILD>. If the method exists, the method is 
+called with all the named arguments from the C<new> method. If not, the public
 attributes from this class are initialized from named arguments of the same
 name. In either case, if neither C<BUILD> nor the default mechanism has
 initialized the attribute, default values are applied (the C<2 * $!x> in the
@@ -528,8 +528,8 @@ class MyClass {
 }
 =end code
 
-If after object construction, you need to check things or modify attributes, writing a submethod
-C<TWEAK> is a good approach (available in v6.d / Rakudo 2016.11):
+The C<TWEAK> method allows you to check things or modify attributes after 
+object construction:
 
 =begin code
 class RectangleWithCachedArea {
--- a/doc/Language/objects.pod6
+++ b/doc/Language/objects.pod6
@@ -4,8 +4,8 @@
 
 =SUBTITLE Object Orientation in Perl 6
 
-Perl 6 is an object oriented language at its core, even though it allows you
-to write programs in other programming styles.
+Perl 6 is an object oriented language at its core, even though you can write
+programs in other programming styles.
 
 Perl 6 comes with a wealth of predefined types, which can be classified in
 two categories: normal and I<native> types.
@@ -19,30 +19,29 @@ object. That includes literals, types (type objects), code and containers.
 
 =head1 Using Objects
 
-You can use objects by calling methods on them. To call a method on an
-expression, add a dot, followed by the method name:
+You can call methods on objects. To call a method on an expression, add a dot,
+followed by the method:
 
 =for code :allow<B L>
 L<say> "abc"B<.L<uc>>;
 # ABC
 
 This calls the L<C<uc>|uc> method on C<"abc">, which is an object of type
-L<C<Str>|Str>. If you need to supply arguments to the method, add
-parentheses after the method name with the arguments inside the parentheses:
+L<C<Str>|Str>. To supply arguments to a method, add parentheses after the
+method, with arguments inside parentheses:
 
 =for code :allow<B L>
 my $formatted-text = "Fourscore and seven years ago...".L<indent>B<(8)>;
 
 C<$formatted-text> now contains the above text, but indented 8 spaces.
 
-Multiple arguments simply require a comma between them:
+Separate multiple arguments with commas:
 
 =for code :allow<B L>
 my @words = "Abe", "Lincoln";
 @words.L<push>("said"B<,> $formatted-text.L<comb>(L</\w+/|/language/regexes>));
 
-Another method call syntax separates the method name and the argument list
-with a colon:
+Alternately, separate the method and the argument list with a colon:
 
 =for code :allow<B L>
 say @words.L<join>: '--';
@@ -67,17 +66,16 @@ objects are used:
 =for code :allow<L>
 $*IN.L<nl-in> = "\r\n";
 
-Here, we are calling method C<nl-in> on the C<$*IN> object,
-without arguments, and assign to the container it returned with the
+Here, we call method C<nl-in> on the C<$*IN> object, without arguments, 
+and assign to the container it returned with the
 L<C<=>|=> operator.
 
 All objects support methods from class L<Mu>, which is the root of the
-type hierarchy. In other words, all objects are derived from C<Mu>.
+type hierarchy. All objects derive from C<Mu>.
 
 =head2 Type Objects
 
-Types themselves are objects and you can get the I<type object> simply by
-writing its name:
+Types are objects and you can get the I<type object> by writing its name:
 
     my $int-type-obj = Int;
 
@@ -143,7 +141,7 @@ another class.
 =head2 Attributes
 X<|Attribute> X<|Property> X<|Member> X<|Slot>
 
-Attributes are variables that exist per instance of a class. They are where
+Attributes are variables that exist per object basis. They are where
 the state of an object is stored. In Perl 6, all attributes are private.
 They are typically declared using the C<has> declarator and the C<!> twigil.
 
@@ -180,9 +178,9 @@ Now, after a C<Journey> object is created, its C<.origin>, C<.destination>,
 and C<.notes> will all be accessible from outside the class, but only
 C<.notes> can be modified.
 
-The above allows creation of C<Journey> objects without an origin or
-destination. To change this you can provide a default value or make
-sure that it is provided by marking it with L<is required>.
+The above allows creation of C<Journey> objects without origin or
+destination values. To prevent this, provide default values or make
+sure that an attribute is set on object creation by marking an attribute with an L<is required> trait.
 
     class Journey {
         # error if origin is not provided
@@ -243,15 +241,15 @@ Methods are declared with the C<method> keyword inside a class body.
 
 A method can have a signature, just like a subroutine. Attributes can be
 used in methods and can always be used with the C<!> twigil, even if they
-are declared with the C<.> twigil. This is because really, the C<.> twigil
-declares a C<!> twigil and then additionally generates an accessor method.
+are declared with the C<.> twigil. This is because the C<.> twigil
+declares a C<!> twigil and generates an accessor method.
 
 Looking at the code above,
 there is a subtle but important difference between using C<$!origin>
-and C<$.origin> in the method C<describe>. The first is always a simple
-lookup of the attribute. It is cheap, and you know that it is the attribute
-declared in this class. The latter is really a method call and thus it may
-be overridden in a subclass. Only use C<$.origin> if you explicitly want to
+and C<$.origin> in the method C<describe>. The first is always a lookup of the 
+attribute. It's cheap, and you know that it's the attribute
+declared in this class. C<$.origin> is a method call and thus may
+be overridden in a subclass. Only use C<$.origin> if you want to
 allow overriding.
 
 Unlike Subroutines, additional named arguments will not produce compile time or
@@ -268,13 +266,13 @@ Method names can be resolved at runtime with the C<.""> operator.
 =head2 self
 
 Inside a method, the term C<self> is available, which is bound to the
-invocant, i.e.  the object that the method was called on. C<self> can be
+invocant, the object that the method was called on. C<self> can be
 used to call further methods on the invocant. Within methods, something like
-C<$.origin> is the same thing as C<self.origin>, however the colon-syntax for
+C<$.origin> is the same thing as C<self.origin>; however, the colon-syntax for
 method arguments is only supported for method calls using self, not the shortcut.
 
-Please note that if the relevant methods C<bless>, C<CREATE>, C<BUILDALL> of
-L<Mu> are not overloaded C<self> will point to the type object in those
+Note that if the relevant methods C<bless>, C<CREATE>, C<BUILDALL> of
+L<Mu> are not overloaded, C<self> will point to the type object in those
 methods.
 
 Since user defined constructors are called on type objects, C<self> will point
@@ -291,9 +289,9 @@ attributes in initializers.
 
 =head2 Private Methods
 
-Methods that are not going to be called or callable from anywhere outside
+Methods that are not going to be called or are not callable from anywhere outside
 the defining class are declared with an exclamation mark C<!> before the method
-name; such methods are called private in the sense they are not visible
+name; such methods are private in the sense they are not visible
 from outside the class that declares them.
 Private methods are invoked with an exclamation mark instead of a dot:
 
@@ -317,7 +315,7 @@ Submethods are useful for object construction and destruction tasks, as well
 as for tasks that are so specific to a certain type that subtypes must
 certainly override them.
 
-For example the L<default method new|/type/Mu#method new> calls submethod
+For example, the L<default method new|/type/Mu#method new> calls submethod
 C<BUILD> on each class in an L<inheritance|#Inheritance> chain:
 
 =begin code
@@ -440,7 +438,7 @@ C<Mu.new> calls method L<bless> on its invocant, passing all the named
 arguments. C<bless> creates the new object and then calls method C<BUILDALL>
 on it. C<BUILDALL> walks all subclasses in reverse method resolution order
 (i.e. from L<Mu> to most derived classes) and in each class checks for
-existence of a method named C<BUILD>. If it exists, it is called, again
+existence of a method named C<BUILD>. If it exists, it's called, again
 passing all named arguments from method C<new> to it. If not, the public
 attributes from this class are initialized from named arguments of the same
 name. In either case, if neither C<BUILD> nor the default mechanism has
@@ -531,8 +529,7 @@ class MyClass {
 }
 =end code
 
-If all you need to do is checking things after the actual construction,
-or modify attributes after the object construction, writing a submethod
+If after object construction, you need to check things or modify attributes, writing a submethod
 C<TWEAK> is a good approach (available in v6.d / Rakudo 2016.11):
 
 =begin code
@@ -550,14 +547,14 @@ say RectangleWithCachedArea.new( x2 => 5, x1 => 1, y2 => 1, y1 => 0).area;
 =head2 Object Cloning
 
 The C<Mu> parent class, from which all classes inherit, supplies a
-method named L<clone> which is somewhat magical, in that it can copy
-values from an instance's private attributes to create a new instance.
-This cloning is shallow, in that it only binds attributes to
-the same values they contained in the original instance; it does
-not make copies of those contained values.
+method named L<clone>, which is somewhat magical in that it can copy
+values from an object's private attributes to create a new object.
+This cloning is shallow since it only binds attributes to
+the same values contained in the original object; it does
+not make copies of those values.
 
-As with C<new>, one may provide initial values for public attributes.
-These override values received from the original instance.  (See
+As with C<new>, public attributes can be set to initial values. 
+These override values received from the original object.  (See
 the documentation for Mu's L<clone> for an example.)
 
 Note that since C<clone> is not a C<submethod>, a class which provides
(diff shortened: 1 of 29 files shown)


AlexDaniel:
``It is``→``It's`` – do we want that? I don't mind, but what others think about it?

doc/Language/objects.pod6
```diff
@@ -4,8 +4,8 @@
 
 =SUBTITLE Object Orientation in Perl 6
 
-Perl 6 is an object oriented language at its core, even though it allows you
-to write programs in other programming styles.
+Perl 6 is an object oriented language at its core, although you can write
```
samcv: If you want to keep this sentence as is, it should be “Although Perl 6 is an object oriented language at its core, you can write programs in other programming styles”.

It would be much better if this had two sentences about Perl 6 being an object oriented language, and then another after that says that you are allowed to write code in other programming styles as well.

doc/Language/objects.pod6
```diff
@@ -19,30 +19,28 @@ object. That includes literals, types (type objects), code and containers.
...
 
 =for code :allow<B L>
 L<say> "abc"B<.L<uc>>;
 # ABC
 
 This calls the L<C<uc>|uc> method on C<"abc">, which is an object of type
-L<C<Str>|Str>. If you need to supply arguments to the method, add
-parentheses after the method name with the arguments inside the parentheses:
+L<C<Str>|Str>. To supply arguments to a method, add parentheses after the
+method, with arguments inside parentheses:
```
samcv: Would be better without the comma, changing it to ‘the arguments’. This makes it match the other words that have ‘a’, and ‘the’ in the sentence.
 “add the parentheses after the method with the arguments inside parenthesis”.

doc/Language/objects.pod6
```diff
@@ -19,30 +19,28 @@ object. That includes literals, types (type objects), code and containers.
...
 
 =for code :allow<B L>
 my $formatted-text = "Fourscore and seven years ago...".L<indent>B<(8)>;
 
 C<$formatted-text> now contains the above text, but indented 8 spaces.
 
-Multiple arguments simply require a comma between them:
+Separate multiple arguments with commas:
```
samcv: Would be better as: “Multiple arguments are separated by commas”.

doc/Language/objects.pod6
```diff
@@ -19,30 +19,28 @@ object. That includes literals, types (type objects), code and containers.
...
 
 =for code :allow<B L>
 my @words = "Abe", "Lincoln";
 @words.L<push>("said"B<,> $formatted-text.L<comb>(L</\w+/|/language/regexes>));
 
-Another method call syntax separates the method name and the argument list
-with a colon:
+Alternately, separate the method and the argument list with a colon:
```
samcv: Would be nice to have this the same as above, making it start with “Multiple arguments…”.

doc/Language/objects.pod6
```diff
@@ -67,17 +65,16 @@ objects are used:
...
-All objects support methods from class L<Mu>, which is the root of the
-type hierarchy. In other words, all objects are derived from C<Mu>.
+All objects support methods from class L<Mu>, which is the type hierarchy root. 
+All objects derive from C<Mu>.
 
 =head2 Type Objects
 
-Types themselves are objects and you can get the I<type object> simply by
-writing its name:
+Types are objects and you can get the I<type object> by writing its name:
```
samcv: I liked it better as “Types themselves”, but otherwise the rest of the sentence is an improvement.

doc/Language/objects.pod6
```diff
@@ -138,7 +135,7 @@ another class.
 =head2 Attributes
 X<|Attribute> X<|Property> X<|Member> X<|Slot>
 
-Attributes are variables that exist per instance of a class. They are where
+Attributes are variables that exist per object basis. They are where
```
samcv: I don't like this change. You could make it “per instance of a class or object” though.

doc/Language/objects.pod6
```diff
@@ -175,9 +172,9 @@ Now, after a C<Journey> object is created, its C<.origin>, C<.destination>,
 and C<.notes> will all be accessible from outside the class, but only
 C<.notes> can be modified.
 
-The above allows creation of C<Journey> objects without an origin or
-destination. To change this you can provide a default value or make
-sure that it is provided by marking it with L<is required>.
+The above allows creation of C<Journey> objects without origin or
```
samcv: This paragraph should start with “Marking an attribute with” and end with a sentence to the effect of “This will prevent…".

doc/Language/objects.pod6
```diff
@@ -263,13 +260,13 @@ Method names can be resolved at runtime with the C<.""> operator.
 =head2 self
 
 Inside a method, the term C<self> is available, which is bound to the
-invocant, i.e.  the object that the method was called on. C<self> can be
+invocant, the object that the method was called on. C<self> can be
```
samcv: This change removes “i.e.” but doesn't replace it with anything with the same meaning (in other words). Should leave it as before or change the rest of the sentence to be written without it.

doc/Language/objects.pod6
```diff
@@ -263,13 +260,13 @@ Method names can be resolved at runtime with the C<.""> operator.
...
 used to call further methods on the invocant. Within methods, something like
-C<$.origin> is the same thing as C<self.origin>, however the colon-syntax for
+C<$.origin> is the same thing as C<self.origin>; however, the colon-syntax for
```
samcv: This was better before. It would be great if the sentence didn't use the words “something like” or “same thing”.

doc/Language/objects.pod6
```diff
@@ -263,13 +260,13 @@ Method names can be resolved at runtime with the C<.""> operator.
...
 method arguments is only supported for method calls using self, not the shortcut.
 
-Please note that if the relevant methods C<bless>, C<CREATE>, C<BUILDALL> of
-L<Mu> are not overloaded C<self> will point to the type object in those
+Note that if the relevant methods C<bless>, C<CREATE>, C<BUILDALL> of
```
samcv: I liked it better with ‘please’, but I think this is just personal opinion. Ultimately, I am fine either way.

doc/Language/objects.pod6
```diff
@@ -286,9 +283,9 @@ attributes in initializers.
 
 =head2 Private Methods
 
-Methods that are not going to be called or callable from anywhere outside
+Methods that are not going to be called or are not callable from anywhere outside
```
samcv: We shouldn't say “Methods that are not going to be called”. Should begin with: “Methods with the exclamation mark…”. The the next sentence should, and leave out the part about the objects not being called. Leave out the “not going to be called” part, since it's the methods that are private, they aren't private just because they are never called.

doc/Language/objects.pod6
```diff
@@ -305,14 +302,14 @@ Private methods are not inherited by subclasses.
 
 =head2 Submethods
 
-A submethod is a public method that is not inherited to subclasses. The name
+A submethod is a public method that is not inherited by subclasses. The name
```
samcv: Should be: “Submethods are public methods…”

doc/Language/objects.pod6
```diff
@@ -435,7 +432,7 @@ C<Mu.new> calls method L<bless> on its invocant, passing all the named
 arguments. C<bless> creates the new object and then calls method C<BUILDALL>
 on it. C<BUILDALL> walks all subclasses in reverse method resolution order
 (i.e. from L<Mu> to most derived classes) and in each class checks for
-existence of a method named C<BUILD>. If it exists, it is called, again
+existence of a method named C<BUILD>. If it exists, it's called, again
```
samcv: Again, this should not use the word ‘it’, at all. It's vague and especially we should not use it twice in the same sentence.

doc/Language/objects.pod6
```diff
@@ -447,10 +444,10 @@ After the C<BUILD> methods have been called, methods named C<TWEAK> are
 called, if they exist, again with all the named arguments that were passed
 to C<new>.
 
-Due to the nature of the default behavior of C<BUILDALL> and custom C<BUILD>
+Due to the default behavior of C<BUILDALL> and C<BUILD>
```
samcv: This is a run on sentence and should also not start with the word “Due”. You could start it with “The default behavior” which would be much better. Extra points if you break it into more than one sentence.

doc/Language/objects.pod6
```diff
@@ -526,8 +523,7 @@ class MyClass {
 }
 =end code
 
-If all you need to do is checking things after the actual construction,
-or modify attributes after the object construction, writing a submethod
+If after object construction, you need to check things or modify attributes, writing a submethod
```
samcv: Should not start with “If…”. Would be great to start with “The TWEAK submethod…”. A sentence after it can say “If after object construction”.

doc/Language/variables.pod6
```diff
@@ -891,7 +890,7 @@ my Int $i = 1; # that works
 { use variables :_; my Int $i; } # switch it off in this block
 =end code
 
-Please note that assigning L<Nil|/type/Nil> will revert the variable to it's
+Note that assigning L<Nil|/type/Nil> will revert the variable to it's
```
samcv: This should be “its” not “it’s”

doc/Language/variables.pod6
```diff
@@ -528,8 +527,8 @@ variables.
     say [$a, $b, $c].perl;
     # OUTPUT«["a", "b", Int]␤»
 
-To destructure a list into a single value create a list literal with one value
-by using C<($var,)>. When used with a variable declarator just providing
+To destructure a list into a single value, create a list literal with one value
```
samcv: Should say create a list with one element, not one value.

doc/Language/variables.pod6
```diff
@@ -343,9 +342,9 @@ sub say-it()  { say $^a; } # invalid
 -> $x, $y, $x { say $^a; } # invalid
 =end code
 
-Placeholder variables syntactically cannot have any type constraints. Be
-also aware that one cannot have placeholder variables with a single
-upper-case letter. This is disallowed in favor of being to able to catch
+Placeholder variables syntactically cannot have any type constraints.
+Placeholder variabless cannot have a single
```
samcv: Misspelled. Also, either don't start both sentences with “Placeholder variables” or add an “also” to the second one.

doc/Language/variables.pod6
```diff
@@ -328,7 +327,7 @@ to that block. This is because the placeholder variables are sorted in Unicode
 order. If you have self-declared a parameter using C<$^a> once, you may refer
 to it using only C<$a> thereafter.
 
-Although it is possible to use nearly any valid identifier as a placeholder
+Although it's possible to use nearly any valid identifier as a placeholder
```
samcv: I don't like this contraction.  The one later on in the sentence is okay though. This is because you are not referring to something previously referenced using the word “it”, you are using it to refer to something later on in the same sentence (and it's also not at the beginning of a sentence).

doc/Language/variables.pod6
```diff
@@ -293,8 +292,8 @@ generated for you because you used the C<.> twigil when the attributes were
 declared.  Note, however, that subclasses may override those methods. If you
 don't want this to happen, use C<$!x> and C<$!y> instead.
 
-The fact that the C<.> twigil just does a method call also implies that the
-following is possible too:
+The fact that the C<.> twigil does a method call implies that the
+following is possible, too:
```
samcv: Should be “is also possible:” with no comma.

doc/Language/variables.pod6
```diff
@@ -265,12 +264,12 @@ X<|$?>
 
 Compile-time variables may be addressed via the C<?> twigil. They are known
 to the compiler and may not be modified after being compiled in. A popular
-example for this is:
+example is this:
```
samcv: This was better before. Use the old version or bonus points for rewriting it to form a complete sentence.

You shouldn't say “A popular example is this.” but you could say “Here is a popular example of/for this.”). It should be noted I don't like that either… So either leave it as before or somehow make it more complete.

doc/Language/variables.pod6
```diff
@@ -222,7 +221,7 @@ old variable as we did with C<$*dynamic2>.
 The dynamic variables differ from other variable types in that referring
 to an undeclared dynamic variable is not a compile time error but a
 runtime L<failure|/type/Failure>, so a dynamic variable can be used
-undeclared as long as it is checked for definedness or used in a
```
samcv: This sentence is way too long. I don't like the contraction either. Bonus points for splitting it into two sentences.

doc/Language/variables.pod6
```diff
@@ -156,7 +155,7 @@ something on as-is:
 =head1 Twigils
 X<|Twigil>
 
-Twigils influence the scoping of a variable. Please be aware that twigils
+Twigils influence the scoping of a variable. Twigils
```
samcv: Good.

doc/Language/variables.pod6
```diff
@@ -129,13 +128,13 @@ See L<operators|/language/operators> for more details on precedence.
...
 
-It is possible to create "variables" in Perl 6 that do not have sigils:
+It's possible to create "variables" in Perl 6 that do not have sigils:
 
     my \degrees = pi / 180;
     my \θ       = 15 * degrees;
 
-Note, however, that these do not create L<containers|/language/containers>. That means C<degrees>
-and C<θ> above actually directly represent C<Num>s. To illustrate, try
+Note that these do not create L<containers|/language/containers>. That means C<degrees>
```
samcv: Change it from “That means” to “This means”.

antquinonez:
Thanks for all the feedback. My main goal in these edits was not to change or add to the substance of the the documentation; just to clean it up. I'll address all of your points and make some suggestions, take next steps.

doc/Language/objects.pod6
antquinonez: This section is about object oriented programming, so the less said about other paradigms, the better. How about:
Among the many computational models that Perl 6 supports, is object oriented programming.  Support for this style of programming is as the core of the language.

doc/Language/objects.pod6
antquinonez: How about:
To supply arguments to the method, add arguments inside parentheses after the method.

doc/Language/objects.pod6
antquinonez: I saw this as an opportunity to do away with the passive voice. I'll make your change.

doc/Language/objects.pod6
antquinonez: I'm struggling with putting together this sentence....
Multiple arguments can also be passed using another syntax:  the method, followed by a colon and  then the argument list.

Thoughts?

doc/Language/objects.pod6
antquinonez: How about:
Types themselves are objects and you can get the I<type object> by writing its name:

doc/Language/objects.pod6
antquinonez: So:
Submethods are public methods that are not inherited by subclasses.

doc/Language/variables.pod6
antquinonez: Woops.
Note that assigning L<Nil|/type/Nil> will revert the variable to its...

doc/Language/variables.pod6
antquinonez: Okay.

doc/Language/variables.pod6
antquinonez: Okay.
The fact that the C<.> twigil does a method call implies that the
following is also possible:

doc/Language/variables.pod6
antquinonez: So, no universal search and replace...okay.

doc/Language/variables.pod6
antquinonez: So:
To destructure a list into a single value, create a list literal with one element.

doc/Language/variables.pod6
```diff
@@ -585,7 +584,7 @@ Therefore, the subroutine
 
     say a for 1..6;
 
-will continue to increment C<$l> and append it to C<@x> each time it is
+will continue to increment C<$l> and append it to C<@x> each time it's
```
antquinonez: Okay.

doc/Language/objects.pod6
samcv: How about this? What do you think.

Perl 6 language is an object oriented language at its core. Although Perl 6 allows programmers to program in multiple paradigms, support for objects is at the heart of the language.

doc/Language/objects.pod6
samcv: Perfect

doc/Language/variables.pod6
```diff
@@ -612,7 +611,8 @@ As with C<my>, declaring multiple C<state> variables must be placed
 in parentheses and for declaring a single variable, parentheses may
 be omitted.
 
-Please note that many operators come with implicit binding, what will lead to actions at a distance. Use C<.clone> or coercion to create a new container that can be bound to.
+Many operators come with implicit binding that would lead to actions at a distance.
```
antquinonez: So:
Many operators come with implicit binding which can lead to actions at a distance.

doc/Language/variables.pod6
```diff
@@ -623,8 +623,7 @@ Please note that many operators come with implicit binding, what will lead to ac
     f for 1..3;
     dd @a; # «Array $var = $[:k1(3), :k2(3), :k3(3)]»
 
-All state variables are shared between threads. The result can be undesired or
-dangerous.
+All state variables are shared between threads. The result can be unexpected.
```
antquinonez: Okay.

doc/Language/objects.pod6
samcv: How is this?
Multiple arguments can be specified by separating the argument list with a colon:

doc/Language/objects.pod6
samcv: Perfect

doc/Language/objects.pod6
samcv: Looking at this again it seems that "Note" is fairly consistently used in the middle of paragraphs but Please note is more often used to start a paragraph. I think that seems reasonable? Thoughts?

doc/Language/variables.pod6
```diff
@@ -637,20 +636,20 @@ dangerous.
 
 =head3 The C<$> Variable
 
-As well as explicitly declared named state variables C<$> can be used
+Beside explicitly declared named state variables, C<$> can be used
```
antquinonez: You're right.
How about:
In addition to explicitly declared named state variables, C<$> can be used...

doc/Language/objects.pod6
samcv: Great

doc/Language/variables.pod6
```diff
@@ -659,7 +658,7 @@ perl6 -e '{ say ++$; say $++  } for ^5'
 =end code
 
 If you need to refer to the value of C<$> more than once within the
-scope it should be copied to a new variable.
+scope, it should be copied to a new variable.
```
antquinonez: How about:
To refer to the value of C<$> more than once within a scope, copy the value to a new variable.

doc/Language/variables.pod6
samcv: Yes

doc/Language/variables.pod6
samcv: Good

doc/Language/variables.pod6
samcv: Great

doc/Language/variables.pod6
```diff
@@ -691,7 +690,7 @@ provided.
 
 =head3 The C<@> Variable
 
-In a similar manner to the C<$> variable there is also a L<Positional>
+Similar to the C<$> variable, there's also a L<Positional>
```
antquinonez: So:
Similar to the C<$> variable, there is also a L<Positional>...

samcv: Sounds good

doc/Language/objects.pod6
antquinonez: Yeah. That was horrible. How about:
Attributes are variables that exist per class (are class attributes) or per instance of a class (are instance or object attributes).

doc/Language/variables.pod6
samcv: You can change this too if you want. In my opinion we shouldn't be turning “it is” to "it's” to make the sentence flow better. It is better rewritten to be more specific or not have to do that at all. Just making the distinction. You can change that to “it is” as well.

I think I incorrectly emphasized the wrong thing in that comment.

doc/Language/objects.pod6
antquinonez: So:

C<Mu.new> calls method L<bless> on its invocant, passing all the named
arguments. C<bless> creates the new object and then calls method C<BUILDALL>
on it. C<BUILDALL> walks all subclasses in reverse method resolution order
(i.e. from L<Mu> to most derived classes) **and in each class checks for the
existence of a method named C<BUILD>. If the method exists, the method is called with all the named arguments from the C<new> method.**

doc/Language/variables.pod6
samcv: I think this is good:

If you need to use the value of $ more than once in a scope, it should be copied to a new variable.

More emphasizes you are copying the value not refering to the same variable

doc/Language/objects.pod6
antquinonez: How about:
The default behavior for C<BUILDALL> and C<BUILD> is to allow for named arguments in a C<new> method to correspond directly to the public attributes of any of the classes in a Method Resolution Order (MRO) or to any named parameter of any C<BUILD> submethod.  

from

Due to the default behavior of C<BUILDALL> and C<BUILD>
submethods, named arguments to the constructor C<new> derived from C<Mu> can
correspond directly to public attributes of any of the classes in the method
resolution order, or to any named parameter of any C<BUILD> submethod.

doc/Language/objects.pod6
samcv: I like that except classes are objects, so that's a little weird.

Attributes are variables that exist per instance of a class.

Maybe that?

doc/Language/objects.pod6
antquinonez: How about:
Inside a method, the term C<self> is available, which is bound to the
invocant ( the object that the method was called on). 

[I hate to think that I'm not getting it...]

doc/Language/objects.pod6
samcv: The C< > were getting lost in the markdown, so added some ` to your comment.

How is this?
`Named arguments in a C<new> method call correspond directly to public attributes in dependent classes or named parameters in C<BUILD> or C<BUILDALL> submethods.  These are applied in in a Method Resolution Order (MRO).`

doc/Language/variables.pod6
antquinonez: How about:
Placeholder variables cannot have type constraints or a variable name with a single upper-case letter (this is disallowed to enable catching some Perl 5-isms).

samcv: Good

doc/Language/objects.pod6
samcv: Is this okay: `Inside a method, the term C<self> is available, which is bound to the invocant (the object the method was called on).`

doc/Language/objects.pod6
antquinonez: I'm thinking we don't want to say that it's primarily or especially an object oriented  language. It's also a functional programming language and a procedural and...

Maybe: 

Perl 6 provides strong support for object oriented programming. Although Perl 6 allows programmers to program in multiple paradigms, object oriented programming is at the heart of the language.

doc/Language/objects.pod6
antquinonez: The 'please' seems so polite...I'll restore the pleases, if you like.

doc/Language/objects.pod6
samcv: I like that :smile: 

Go with that.

doc/Language/objects.pod6
samcv: Okay, if not then just remove all of them so we are consistent.

doc/Language/objects.pod6
antquinonez: Okay.

doc/Language/variables.pod6
antquinonez: I'm not getting this, unless you're simply saying, find a way not to use it is (or it's) at the beginning of the sentence.

doc/Language/variables.pod6
antquinonez: Okay.

antquinonez:
I'll make the updates in my forked copy and mark items as done, here. When I'm done/done, I'll add a comment after this one.

doc/Language/objects.pod6
antquinonez: done.

doc/Language/objects.pod6
antquinonez: done

doc/Language/objects.pod6
antquinonez: done

doc/Language/objects.pod6
antquinonez: done

doc/Language/objects.pod6
antquinonez: Your suggestion reverts the change. I'll do that.

antquinonez: done

doc/Language/objects.pod6
antquinonez: How about:
`If an object is instantiated without certain attributes, such as origin or destination, above, we may not get the desired result. To prevent this, provide default values or make sure that an attribute is set on object creation by marking an attribute with an L<is required> trait.`

doc/Language/objects.pod6
```diff
@@ -238,15 +235,15 @@ Methods are declared with the C<method> keyword inside a class body.
...
+declares a C<!> twigil and generates an accessor method.
 
 Looking at the code above,
 there is a subtle but important difference between using C<$!origin>
-and C<$.origin> in the method C<describe>. The first is always a simple
-lookup of the attribute. It is cheap, and you know that it is the attribute
-declared in this class. The latter is really a method call and thus it may
-be overridden in a subclass. Only use C<$.origin> if you explicitly want to
+and C<$.origin> in the method C<describe>. The first is always a lookup of the 
+attribute. It's cheap, and you know that it's the attribute
```
antquinonez: done

doc/Language/objects.pod6
antquinonez: done

doc/Language/objects.pod6
antquinonez: How about:
`Within methods, C<$.origin> works the same as C<self.origin>, however the colon-syntax for ...`

doc/Language/objects.pod6
antquinonez: How about:
`Methods with an exclamation mark C<!> before the method
name are not callable from anywhere outside the defining class;
such methods are private in the sense they are not visible
from outside the class that declares them.
Private methods are invoked with an exclamation mark instead of a dot:`

doc/Language/objects.pod6
antquinonez: done

doc/Language/objects.pod6
antquinonez: done

doc/Language/objects.pod6
antquinonez: How about:

`The C<TWEAK> method allows you to check things or modify attributes after object construction:`

from 

`If after object construction, you need to check things or modify attributes, writing a submethod
C<TWEAK> is a good approach (available in v6.d / Rakudo 2016.11):`

doc/Language/variables.pod6
antquinonez: done

doc/Language/variables.pod6
antquinonez: done

doc/Language/variables.pod6
antquinonez: done. no change.

doc/Language/variables.pod6
antquinonez: How about:
`Dynamic variables are checked at runtime, as opposed to compile time. If your code refers to an undeclared dynamic variable at runtime, you will get a runtime L<failure|/type/Failure>, unless the variable is used in a boolean context or tested for definedness before being used for something else.`

doc/Language/variables.pod6
antquinonez: How about, removing the "A popular example...." and just giving the example?

doc/Language/variables.pod6
antquinonez: done

doc/Language/variables.pod6
antquinonez: done

doc/Language/variables.pod6
antquinonez: No period at the end of element.
The full sentence is
`To destructure a list into a single value, create a list literal with one element
by using C<($var,)>.

done`

doc/Language/variables.pod6
```diff
@@ -557,7 +556,7 @@ examples.
 The C<anon> declarator prevents a symbol from getting installed in the lexical
 scope, the method table and everywhere else.
 
-For example you can use it to declare subroutines which know their own name,
+For example, you can use it to declare subroutines which know their own name,
```
antquinonez: done. no change.

doc/Language/variables.pod6
antquinonez: done

doc/Language/variables.pod6
antquinonez: done

doc/Language/variables.pod6
```diff
@@ -637,20 +636,20 @@ dangerous.
...
 
-As well as explicitly declared named state variables C<$> can be used
+Beside explicitly declared named state variables, C<$> can be used
 as an anonymous state variable without an explicit C<state> declaration.
 
     say "1-a 2-b 3-c".subst(:g, /\d/, {<one two three>[$++]});
     # OUTPUT«one-a two-b three-c␤»
 
-Furthermore, state variables are not required to exist in subroutines. You
+Furthermore, state variables are not required in subroutines. You
```
antquinonez: This section of the documentation is not about declaring static variables outside of subroutines. It's about how...(a few lines later):
`Each reference to C<$> within a lexical scope, in effect, is a separate
variable.`

Your Perl 6 is stronger than mine, so let me know...

doc/Language/variables.pod6
```diff
@@ -637,20 +636,20 @@ dangerous.
...
 could, for example, use C<$> in a one-liner to number the lines in a file.
 
 =begin code :skip-test
 perl6 -ne 'say ++$ ~ " $_"' example.txt
 =end code
 
-Each reference to C<$> within a lexical scope in effect is a separate
+Each reference to C<$> within a lexical scope, in effect, is a separate
```
antquinonez: done

doc/Language/variables.pod6
antquinonez: done

doc/Language/variables.pod6
antquinonez: done

finanalyst:
There is a quality of courtesy in the Perl6 world that is quite 
uncommon, and I do not think that it should be discarded lightly.

Consequently, I would think that the 'please's should be kept in. After 
the removal of 'please's will come a push for a more formal language, 
and then the elimination of quirks. The documentation will become boring 
and ultimately less easy  to read, even though the original intention 
may have been clarity.

Regards,

Richard

finanalyst:
I think this change is wrong.

In the original version, Perl6 is the subject of the first clause and 
the second.

In the updated version, Perl6 remains the subject of the first clause, 
but the subject of the second is 'you'. This unbalances the sentence.

antquinonez:
@finanalyst Regarding my proposal to remove 'please' from the documentation, my intent was to trim.  I like the personal touches in technical documentation, the little jokes, and especially the informality. We should ask if something is necessary or useful or appreciated by the busy programmer...

finanalyst:
What you say is true about 'trimming', but 'please' is one of those 
words, such as 'and' and 'the' which tend to be filtered out of any 
normal reading.

The subliminal effect remains, which is of courtesy and an attention to 
the user.

antquinonez:
@samcv , are you interested in applying the changes you're 100% in agreement with and tossing out the rest?

samcv:
@antquinonez yes I am. Are you able to rebase the PR so it can be mergeable? I will sort through everything after if you just resolve the conflicts.

antquinonez:
Okay. Did  the rebase (I'm a little iffy that I did it right), but hopefully some correct stuff came out the other end.

tommystanton:
Hmm, I'm not sure why GitHub is showing that 250+ commits will be merged.  Testing locally, it looks to be less than 20 would be merged (using git-sh here in the CLI):

```text
$ git remote -v
antquinonez	https://github.com/antquinonez/doc.git (fetch)
antquinonez	https://github.com/antquinonez/doc.git (push)
upstream	https://github.com/perl6/doc (fetch)
upstream	https://github.com/perl6/doc (push)
$ git branch master-antquinonez antquinonez/master
$ git branch -vv
  master             a50900f [upstream/master] Merge pull request #1150 from ronaldxs/grammar_tutorial-grouping
* master-antquinonez 644c3f8 [antquinonez/master] fixed some merge issues
$ git log --pretty=oneline --abbrev-commit master..master-antquinonez
644c3f8 fixed some merge issues
e892ded Implemented more suggestions to pull request.
8c073d3 Made changes post feedback.
6a45e52 Mostly style updates.
ebfa8a5 Update variables.pod6
fd692d0 style cleanup.
905e06d Update objects.pod6
b847a6a Update objects.pod6
497e4dd Update objects.pod6
50a8441 Implemented more suggestions to pull request.
0d42f42 Made changes post feedback.
9e443d1 Merge pull request #1 from antquinonez/antquinonez-patch-1
1a2820d Update variables.pod6
5c93847 Mostly style updates.
a4b1e45 style cleanup.
8ecb6b1 Update objects.pod6
21d4335 Update objects.pod6
a802d56 Update objects.pod6
```

antquinonez:
I think I did some very bad things, here. I'm going to close this pull request and get my act together. Sorry guys.

PR #1076, opened 2016-12-24
Document item

### LINKED ISSUE #626: item context not documented
Item context is listed among operators in [S03](https://github.com/perl6/specs/blob/master/S03-operators.pod), but does not appear to be in the Language/operators documentation, nor listed on the docs site in the list of Routines.

The phrase "item context" is mentioned in only two places and not well explained in either:

```
doc/Language/5to6-perlop.pod
73:named variable, you can use item context, like so: C<$aref = item(@array)>

doc/Language/faq.pod
300:on an expression, and item context (not flattening) with C<$( ... )>
```

titsuki:
Fix #626

See also:
https://github.com/perl6/roast/blob/master/S03-operators/context.t


titsuki:
It seems that there aren't any objections, so I'll merge this.

zoffixznet:
Sorry, I don't have notifications for doc repo, so didn't see it until the IRC bot announced it...

Would you, please, rewrite the examples using `.perl` routine? `dd` is not a standard Perl 6 routine and should be avoided in examples, since the doc website is for the Perl 6 language and not just the Rakudo implementation (that provides `dd`).

titsuki:
@zoffixznet
Thanks for your reply !
I'll do it soon.

PR #1100, opened 2017-01-01
Edit headers to 'is default' sections

nicqrocks:
This adds a header to the 'is default' section of the Sub type so that it is indexed. The 'is default' section for variables has also been changed a bit so that it is in the same format as the Sub one.


nicqrocks:
Oops, I forgot the 2! 😦
Thank you for catching that.

nicqrocks:
Ok, that should fix it.

Altai-man:
Seems good enough for me. Thanks for your contribution, @nicqrocks!

PR #1109, opened 2017-01-02
Document the undefined sort order of standard collections

toolforger:
Also, removed a WRONG marker: it is not wrong to output Bag
elements in an undefined order, i.e. if the user didn't care
about output order.

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -596,7 +596,7 @@ X<Hyper method call operator>. Will call a method on all elements of a C<List> o
 Take care to avoid a common mistake of expecting side-effects to occur in order. The following
 C<say> is B<not> guaranteed to produce the output in order:
 
-    @a».say;  # WRONG! Could produce a␤b␤c␤ or c␤b␤a␤ or any other order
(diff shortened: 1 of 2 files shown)


doc/Language/operators.pod6
```diff
@@ -596,7 +596,7 @@ X<Hyper method call operator>. Will call a method on all elements of a C<List> o
 Take care to avoid a common mistake of expecting side-effects to occur in order. The following
 C<say> is B<not> guaranteed to produce the output in order:
 
-    @a».say;  # WRONG! Could produce a␤b␤c␤ or c␤b␤a␤ or any other order
```
AlexDaniel: But @a is not a Bag?

Otherwise looks like a great start.

zoffixznet:
@toolforger any update on the changes request above?

zoffixznet:
Closing due to lack of response. @toolforger please ping us if you come back to it.

zoffixznet:
P.S.: my comments on the PR: the text is very wordy: "Baggies and Mixies don't guarantee order" would be my summary of it. And the `TODO:` bit can probably be a comment rather than something exposed to the reader.

PR #1131, opened 2017-01-11
Don't generate .png typegraph files

### LINKED ISSUE #1063: consider not generating PNGs.
http://caniuse.com/#feat=svg

I have no problem supporting the type graphs only on the supported browser lists.

Less to do when building the docs, the better. Please comment below if you want to vote on this issue.

coke:
For consideration.

Untested, as 'make htmlify' is insanely slow on my particular machine at the moment. Should resolve #1063 if it's accepted.

### DIFF
--- a/htmlify.p6
+++ b/htmlify.p6
@@ -671,12 +671,6 @@ sub write-type-graph-images(:$force, :$parallel) {
             @type-graph-images = ();
         }
 
-        @type-graph-images.push: $viz.to-file("html/images/type-graph-{$type}.png", format => 'png', size => '8,3');
-        if @type-graph-images %% $parallel {
-            await(@type-graph-images);
-            @type-graph-images = ();
-        }
-
         print '.';
 
         LAST await(@type-graph-images);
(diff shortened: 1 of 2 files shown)


coke:
I'll go ahead and merge this in a week with no objection.

AlexDaniel:
> Code removed is code that is doesn't have to be maintained.

So the pull request removes 14 lines of code and one feature that could be useful to some. The justification is that it is slow on your machine.

Well, then just implement ``--light`` option that doesn't generate png graphs, or maybe even **all** graphs, so that it is faster on your machine. This would be useful for other purposes too.

👎

coke:
Could be, not is.

coke:
BTW, the justification is not only that "it is slow on my machine" (the doc build is VERY slow, and this is a drop in the bucket). It's extra time spent on every build, and we don't need two ways to do something when one suffices. Any extra code you have is code that you has to be maintained.

http://caniuse.com/#feat=svg has a listing of all the browsers that support SVG.

I intended the discussion to be centered around whether or not that browser listing is recent enough to drop the PNGs, my assertion being that it is.

coke:
Feedback on https://irclog.perlgeek.de/perl6/2017-01-23#i_13975600 seems to generally be: these were useful once but are no longer.

zoffixznet:
> So the pull request removes [...] feature that could be useful to some [...] just implement `--light` option

I see no reason to add, document, and support an option on a sole basis of "could be useful to some." Especially considering the ease of converting the SVGs to PNGs should anyone actually need the latter (`convert` does a good job even without any additional settings).

> we don't need two ways to do something

Or to waste bandwidth on bots crawling those PNGs (yes, I heard of robots.txt)

> whether or not that browser listing is recent enough

It is. SVGs been supported for ages. The only browsers I recall giving trouble are earlier IEs that are no longer even supported by Microsoft in any way. To corroborate, here's a [SO question](http://stackoverflow.com/questions/8926285/svg-browser-support) from **5 years ago** with the answer that all browsers supported it for years.

> I'll go ahead and merge this in a week with no objection.

:+1:

coke:
Based on IRC feedback and zoffix's last comment, merging.

Issue #776, opened 2016-07-28
Single style for output of examples


Altai-man:
See https://github.com/perl6/doc/issues/761#issuecomment-236023547

TL;DR: right now we don't have a single "official" output format for our examples, and everyone who writes documentation uses their own style. It's bad if we want to test our examples result automatically in the future. Our current formatting variants:
- `#-> result`
- `# result`
- `# OUTPUT: «result␤»`
- `# OUTPUT: result`
- ``` perl6
  # OUTPUT:
  # result1
  # result2
  # result3
  # result4
  ```

```

Please, if you want to propose your own "style" - do it with a comment or just edit this message.
We need opinions about what style is best.

This issue is marked as `big`, because after selection we also need to rewrite all output results that don't match new "official" format line.
```

zoffixznet:
I like using the following, as the this way multi-line output aligns well and it's clear that it's output.

```
# OUTPUT:
# result
```

I'm against using `# result`, because it'd conflict with regular comments, especially when we do automated output check and I'm against `OUTPUT: «result␤»` because it's hard to type the fancy-pants symbols (I don't even know how to type them on my phone at all).

gfldex:
You don't type the result at all. You copy-pasta them from camelia into the docs because you tested the example against camelia to make sure they hold true.

The OUTPUT: «result␤» form allows easy parsing, I do agree that ¶ would have been a better pick because more ppl will be familiar with that symbol. We could change that automatically for html-output.

I'm sorry but the "I don't know how to ... on my phone"-argument is to selfish to be relevant to hopefully millions of readers of docs.perl6.org.

AlexDaniel:
I agree that multi-line output should normally be on multiple lines. As for the ``«»``, these are usually typable on phones (and yeah, I think that the argument that you can't type something does not work so well in Perl 6 world). However, when I look at this:
```
OUTPUT: «result»
OUTPUT: result
```
It feels like «» are slightly redundant. It is parseable and it looks great.

So:
```perl6
say qw{1 2 3 4}; # OUTPUT: (1 2 3 4)
```

```perl6
.say for qw{1 2 3 4};
# OUTPUT:
# 1
# 2
# 3
# 4
```

↑ I see no problem here whatsoever

Comparing that to this:
```perl6
say qw{1 2 3 4}; # OUTPUT: «(1 2 3 4)»
```
↑ Meh, «» look a bit weird.

Altai-man:
[...]

PR #1165, opened 2017-01-26
Make comment style the same in Type/ directory

Altai-man:
**General description**

[Here](https://github.com/perl6/doc/issues/776) was raised a question about different styles of output of the examples. Soon it led to efforts to make examples extracted and executed automatically(with @gfldex++ support) and thoughts about automatic checking of output correctness(not yet).

So this PR introduces a migration of Type/ directory files to a style that was proposed by @coke, simply saying:
- say 3; # OUTPUT: «3␤»
- 2+5; # RESULT: «7»
- 2+5i; # this is a complex number

**Why is single style important**

* Inconsistency is bad for the reader. We have `#!>` in some places, `OUTPUT«`, `# prints` and others without any logic between usages.
* We cannot adequately write a tool to automate our examples testing without some machine-easy recognizable style.

**Pros**

* With unified style we can check our examples(it is important - the wrong examples are tedious to catch by hands and it confuses reader).
* With unified style we can easily convert it to other styles automatically.

**Cons**

* While working on it, I noticed that in some rare places(Str.pod6, Date-related pod6 files, for example) this style of output is... A bit noisy to read. But it can be solved by simple "stripping out" this "OUTPUT|RESULT" things on pod6-to-html stage. Or this style can be changed automatically to something else.
* It somewhat makes contribution harder. If we have a single style, we need to document it in CONTRIBUTION.md and stick to it.

**Additional questions**

* I don't know what to do with Language/ folder. Theoretically, we need to test it too, but it isn't even compileable(at this moment), so I didn't touch it. I also have concerns about "Is this work needed at all?"-part.
* ~~There are conflicts now, but I still have some time and I will resolve them if this PR will be accepted as helpful.~~ I can do some additional work if needed.

**Additional points**

* I've reviewed it carefully and it(hopefully) doesn't introduce any typo. Instead, I've fixed a few.
* It passes `make test` and `make xtest`.
* I know that I'm leaving my "trace" all over the docs, 79 files were changed after all. Can't be helped.

### DIFF
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -41,8 +41,8 @@ Defined as:
 Interprets the invocant as a list and creates an
 C<any>-L<Junction|/type/Junction> from it.
 
-    say so 2 == <1 2 3>.any;        # True
-    say so 5 == <1 2 3>.any;        # False
+    say so 2 == <1 2 3>.any;        # OUTPUT: «True␤»
+    say so 5 == <1 2 3>.any;        # OUTPUT: «False␤»
 
 =head2 method all
 
@@ -53,8 +53,8 @@ Defined as:
 Interprets the invocant as a list and creates an
 C<all>-L<Junction|/type/Junction> from it.
 
-    say so 1 < <2 3 4>.all;         # True
-    say so 3 < <2 3 4>.all;         # False
+    say so 1 < <2 3 4>.all;         # OUTPUT: «True␤»
+    say so 3 < <2 3 4>.all;         # OUTPUT: «False␤»
 
 =head2 method one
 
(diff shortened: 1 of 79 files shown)


Altai-man:
The conflicts are resolved.

AlexDaniel:
Conflicts have to be resolved again.

Altai-man:
@AlexDaniel, the new conflict is a minor one, it just needs to remove a "dead" section.
Commits are going quickly here and I don't think that it is wise to do a resolving commit before PR will be accepted/rejected, since after some time a new conflict can appear. It is best to remove conflicts in one last commit, imho(although, it can be a question of taste...).

Altai-man:
Accepted! I'll merge it now.

Also, @coke, what do you think about Language/ section? At a glance, it seems to have the same problem.
But a lot of examples there are "broken"(they are fine, but have throwing exceptions, absence of `skip-code` directives for perl code, etc, which makes things harder to deal with), so it means another big commit(or I will split it in chunks, but still) from me that will touch many files at once. Is it okay?

PR #1167, opened 2017-01-29
unicode quotes messes up syntax highlighting

WildYorkies:
I assume they were unicode. I'm sure it works in code, but it was messing up the rest of the syntax for that code block.

Also, the first " that was on the bottom of the line was invisible on the docs website anyway.

### DIFF
--- a/doc/Type/Sub.pod6
+++ b/doc/Type/Sub.pod6
@@ -16,14 +16,14 @@ operators separate the two parts by white space.
     say $s.WHAT;        # (Sub)
     say $s(2, 5);       # 7
 
-    sub postfix:<♥>($a){ say „I love $a!“ }
+    sub postfix:<♥>($a){ say "I love $a!" }
     42♥;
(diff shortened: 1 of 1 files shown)


zoffixznet:
This is the problem with highlighter, not the docs. I opened an Issue there: https://github.com/perl6/doc/pull/1167

samcv:
Was just about to comment. Should be an easy fix, will add those two to the list that get automatically generated in.

AlexDaniel:
> Also, the first " that was on the bottom of the line was invisible on the docs website anyway.

In other words, you do not have a font that can render this character. Works just fine here, not a doc issue. That being said, we have some pages associated with unicode:
* https://docs.perl6.org/language/unicode_entry
* https://docs.perl6.org/language/unicode_texas
* https://docs.perl6.org/language/unicode

Perhaps once you find how to resolve the issue on your machine, you can contribute the solution as a tip for one of the pages?

PR #1168, opened 2017-01-29
Adds definition for Repository

JJ:
it's not in the TODO list, but it simply wasn't there, so I said, why not...


zoffixznet:
Thanks.

JJ:
Thanks. I've seen a bunch of other terms I could have a go at. Is it better
to mark them as TODO and then proceed, or should I define them directly?

zoffixznet:
Just commit them directly.

JJ:
:+1:

PR #1171, opened 2017-01-31
added line breaks to grammar_tutorial

antquinonez:
In preparation for future edits.

### DIFF
--- a/doc/Language/grammar_tutorial.pod6
+++ b/doc/Language/grammar_tutorial.pod6
@@ -128,16 +128,19 @@ Let's suppose we'd like to parse a URL into the component parts that make up a
 RESTful request. Let's decide that we want the URLs to work like this:
 
 
-    =item1 The first part of the URI we'll call the "subject", like a part, or
-        a product, or a person.
-    =item1 The second part of the URI we'll call the "command", like standard
-        CRUD stuff (create, retrieve, update, or delete).
-    =item1 The third part of the URI will be arbitrary data. Perhaps the
-        specific ID we'll be working with, or a long list of data separated
-        by "/"'s.
-    =item1 When we get a URL, we'll want 1-3 above to be placed into a nice data
-        structure we can use without having to do all sorts of splitting, and
-        that can be easily altered in the future or expanded upon (or extended).
+    =item The first part of the URI we'll call the "subject", like a part, or
+          a product, or a person.
+
+    =item The second part of the URI we'll call the "command", like standard
+          CRUD stuff (create, retrieve, update, or delete).
+
+    =item The third part of the URI will be arbitrary data. Perhaps the
+          specific ID we'll be working with, or a long list of data separated
+          by "/"'s.
+
+    =item When we get a URL, we'll want 1-3 above to be placed into a nice data
+          structure we can use without having to do all sorts of splitting, and
+          that can be easily altered in the future or expanded upon (or extended).
 
 So if we got a URI on the server of "/product/update/7/notify" we would want our
 grammar to give us a nice $match object that has a "I<subject>" of "product", a
--- a/doc/Language/grammar_tutorial.pod6
+++ b/doc/Language/grammar_tutorial.pod6
@@ -27,10 +27,10 @@ language.
 
 =SUBTITLE The Broad Concept of Perl Grammars
 
-Regular Expressions (regex) work well to find patterns in strings and manipulate
-them. However, when you need to find multiple patterns at once, or need to
-combine patterns, or test for patterns that may surround strings, or other
-patterns - regular expressions alone are not adequate.
+Regular Expressions (regex) work well to find patterns in strings and
+manipulate them. However, when you need to find multiple patterns at once, or
+need to combine patterns, or test for patterns that may surround strings, or
+other patterns - regular expressions alone are not adequate.
 
 Grammars provide a way to define how you want to examine a string using regular
 expressions, and you can group these regular expressions together to provide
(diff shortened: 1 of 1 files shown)


doc/Language/grammar_tutorial.pod6
```diff
@@ -70,25 +104,48 @@ When a method (regex, token or rule) matches in the grammar, that matched string
...
-Let's suppose we'd like to parse a URL into the component parts that make up a RESTful request. Let's decide that we want the URLs to work like this:
+Let's suppose we'd like to parse a URL into the component parts that make up a
+RESTful request. Let's decide that we want the URLs to work like this:
 
 
-    =item1 The first part of the URI we'll call the "subject", like a part, or a product, or a person.
-    =item1 The second part of the URI we'll call the "command", like standard CRUD stuff (create, retrieve, update, or delete).
-    =item1 The third part of the URI will be arbitrary data. Perhaps the specific ID we'll be working with, or a long list of data separated by "/"'s.
-    =item1 When we get a URL, we'll want 1-3 above to be placed into a nice data structure we can use without having to do all sorts of splitting, and that can be easily altered in the future or expanded upon (or extended).
+    =item1 The first part of the URI we'll call the "subject", like a part, or
```
Altai-man: Why do we have `=item1` here instead of `=item`? In the rest of the docs we use the latter. And the indentation feels unnatural.

doc/Language/grammar_tutorial.pod6
```diff
@@ -27,41 +27,75 @@ language.
 
 =SUBTITLE The Broad Concept of Perl Grammars
 
-Regular Expressions (regex) work well to find patterns in strings and manipulate them. However, when you need to find multiple patterns at once, or need to combine patterns, or test for patterns that may surround strings, or other patterns - regular expressions alone are not adequate.
```
Altai-man: How did you do such a wrapping? I may be wrong, but, for example, my emacs gives me a different result.
For this line emacs "fill-region" does
```
Regular Expressions (regex) work well to find patterns in strings and
manipulate them. However, when you need to find multiple patterns at
once, or need to combine patterns, or test for patterns that may
surround strings, or other patterns - regular expressions alone are
not adequate.
```
but your result is
```
Regular Expressions (regex) work well to find patterns in strings and manipulate
them. However, when you need to find multiple patterns at once, or need to
combine patterns, or test for patterns that may surround strings, or other
patterns - regular expressions alone are not adequate.
```

Altai-man:
I'm not sure how paragraphs should fill in .pod6 files world, so I didn't request changes, just gave a comment here.

doc/Language/grammar_tutorial.pod6
antquinonez: Woops. I forgot to look into the indentation but forgot to do so. The item1s were all there in the original -- I didn't touch that .

doc/Language/grammar_tutorial.pod6
antquinonez: Used my eyeball and cut at the 80 char column limit, more or less.

doc/Language/grammar_tutorial.pod6
Altai-man: Yes, it is adaptiveoptics++ initial commit was, but I think it is better to cut `1` out and remove the indentation here. Perhaps in another pr.

doc/Language/grammar_tutorial.pod6
Altai-man: Sounds a bit unstable for me, but the initial intention of changes is good. What do you think about improving it with an automatic line-wrapper?
And maybe it will be good to write a test for too long lines... It can be written easily as a checker, and we have Text::Fill in our ecosystem to even fix that automatically in Perl 6. I'll ask tomorrow in #perl6 channel about how important it is.

doc/Language/grammar_tutorial.pod6
antquinonez: How about indentation that lines up like this:
item somehting, somehting
        something else

antquinonez:
Made some changes to the item list. Looks nicer to me. Thoughts?

doc/Language/grammar_tutorial.pod6
antquinonez: My thoughts are: you probably don't want to consistently re-wrap lines as you make edits. Messes with diff, doesn't it? I had an issue earlier with this. if you re-wrap lines, you have to make sure you don't do it to embedded code examples (I'm guessing) since you'll need to deal with comments and strings and other stuff that will break if you just move the hanging bits to the next line. I'm new to editing POD, so there may be already available solutions to this.

Altai-man: Opinion accepted! Anyway, even without tests and stuff, this needs to be done by a text formatter, not by hands. I can do it myself with emacs and note this pr in the commit message or you can send another commit here.

antquinonez: Please, let me do the Emacs text-formatter so I can become familiar with it. Are you okay with the way the 'item's were indented?

Altai-man: I'd rather remove indentation at all, to be honest. It seems redundant.
About emacs... This text editor is just my own preference and it can take a long(very long) way to get used to it. If you use linux or macos, you can do a conversion by using such command:
`fold -w 80 -s filename`, `filename` must be replaced with an actual name, of course, and use the output.
Actually, I've discovered that default emacs `fill-paragraph` does formatting in 72 characters, not 80, so it is not the best option too.

antquinonez: Okay. I'll run the linux command and un-indent the the 'item's.

antquinonez:
Ran the fold command and removed  the indents on the 'item's (did add line breaks between them, though). What do you think?

Altai-man:
Looks good enough for me. I'll merge it now and look again after update of the site. If something will be wrong, I fix it directly. Thanks!

PR #1176, opened 2017-02-02
Language section examples reworking (part two)

Altai-man:
* Make many examples compile by adding semicolons, adding actual test data, etc.
* Add `:skip-test` directive to many examples. For further explanation see https://github.com/perl6/doc/issues/776.
* Use `table` instead of plain code where appropriately;
* Reformatting of some old examples.

Sadly, this is not the end, just some preparation work; My further plan is:
* To reformat results in this section too, just as for Type/.
* To write an addition to the contribution guide about the output style.
* To close this big issue.

I wanted this to be merged quickly to prevent conflicts, since this work is slow and our commits to master are everyday thing. After that, I will be able to continue my work in a small chunks directly into the master to be closer to upstream.

### DIFF
--- a/doc/Language/about.pod6
+++ b/doc/Language/about.pod6
@@ -93,7 +93,7 @@ example, often contain the methods the class implements.
 Definitions must be in one of the following forms to be recognized as
 the start of a documentable named, say, Z.  First the code in the document source:
 
-=begin code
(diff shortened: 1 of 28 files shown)


gfldex:
travis seams to be stalled, merging anyway

PR #1181, opened 2017-02-03
Unify Language/ examples output style

Altai-man:
No more `#->`. No more `#=>`. No more `# OUTPUT<<`(all of these could be mixed even in a single file). Just our plain unified style described here: https://github.com/perl6/doc/issues/776#issuecomment-240106630
This is a final stage of formatting work, there will be (hopefully) no more such PRs.

### DIFF
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -30,7 +30,7 @@ A default constructor allows the setting of attributes for the created object:
     # Create a new Rectangle from two Points
     my $r = Rectangle.new(lower => Point.new(x => 0, y => 0), upper => Point.new(x => 10, y => 10));
 
-    say $r.area(); # -> 100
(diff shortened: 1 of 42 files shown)


Altai-man:
The PR somewhat grew today, hence the explanation:
* New fixes for the examples.
* (important) The actual test of compiled examples was added.

This test:
1)Extracts examples from documentation;
2)Tries to compile it;
3)If the exitcode is 0, then it's okay, otherwise changes was bad.
We skip unsuitable or infinite examples by `:skip-test` directive. **As for now, all documentation passes this test.**

Why do we need this? Okay, during my journey into the documentation, I've reviewed it entirely and found quite a lot of bugs, broken examples, mistakes in comments, etc. Documentation and code surely start to rot with time, but we can at least prevent this a little.

cast @AlexDaniel, @gfldex, @coke for reviewing, discussing, merging/rejecting.

PR #1193, opened 2017-02-12
add haskell-to-p6 file

WildYorkies:


doc/Type/Sub.pod6
```diff
@@ -16,14 +16,14 @@ operators separate the two parts by white space.
...
 
-    sub postfix:<♥>($a){ say „I love $a!“ }
+    sub postfix:<♥>($a){ say "I love $a!" }
     42♥;
     # OUTPUT: «I love 42!␤»
     sub postcircumfix:<⸨ ⸩>(Positional $a, Whatever){ say $a[0], '…', $a[*-1] }
     [1,2,3,4]⸨*⸩;
     # OUTPUT: «1…4␤»
     constant term:<♥> = "♥"; # We don't want to quote "love", do we?
-    sub circumfix:<α ω>($a){ say „$a is the beginning and the end.“ };
```
Altai-man: Why such changes?

doc/Language/haskell-to-p6.pod6
```diff
@@ -0,0 +1,414 @@
...
+
+    say test-animal Animal::Dog; # more explicit
+    say test-animal Horse;
+
+=head2 Type Aliases and Subsets
+
+In Haskell, you can alias an existing type to simply increase clarity of intent and re-use
+existing types.
+
+    alias Name = String
```
Altai-man: What is this `alias` thing? Did you mean `type Name = String`?

doc/Language/haskell-to-p6.pod6
```diff
@@ -0,0 +1,416 @@
...
+explain how Perl6 roles compare to Haskell typeclasses
+
+=head1 Functions
+
+=head2 Definitions and Signatures
+
+=Multiple Dispatch
+
+Haskell makes heavy use of multiple dispatch in function definitions.
+
```
Altai-man: It is indeed kinda multiple dispatch, but I think that "pattern matching" will suit here better for haskellers, wont it?

WildYorkies:
@Altai-man Good catches.

timo:
i'm not sure why there was a change to Sub.pod6 in between those other commits? it doesn't seem to belong to the haskell-to-p6 pull request, i think.

WildYorkies:
@timo It was leftover from a changed I wanted to make but never submitted. fixed now

doc/Language/haskell-to-p6.pod6
```diff
@@ -0,0 +1,420 @@
...
+      Just x  -> x
+      Nothing -> 0
+
+In Perl6, since type objects coexist with regular objects, we have the concept of Defined
+and Undefined objects. Plain type objects are undefined while instansiated objects are defined.
+
+    sub parse-int(Str $s --> Int) { ... }
+    
+    given parse-int($string) {
+      when * ~~ Int:D { * }
```
CIAvash: Shouldn't `{ * }` be `{ $_ }`?

Also, you can just write `when Int:D { $_ }` and `when Int:U { 0 }`.

timo: Maybe here would be a good place to put a note about the defined-or operator `//`? Or is that too tangential?

WildYorkies: @CIAvash Good call.

@timo I talked about orelse (which is basically defined-or). But I'm just trying to get this initial draft merged so others can contribute to it.

doc/Language/haskell-to-p6.pod6
```diff
@@ -0,0 +1,420 @@
...
+    
+The equivalent in Perl6 is the following.
+
+    my constant Name = Str;
+
+    sub full-name ( Name \first, Name \last --> Name ) { first ~ last }
+    
+It should be noted that in Perl6, one can also created a subset of an existing type.
+
+    subset Name of String where { *.chars < 20 };
```
CIAvash: This should be `subset Name of Str where *.chars < 20;`

Also a typo in the line above, "created" => "create"?

CIAvash:
I think `==>` and `<==` feed operators should be mentioned too. But of course they don't have to be part of this pull request.

WildYorkies:
@CIAvash Agreed. I have a placeholder for those under the heading "Currying" I believe.

doc/Language/haskell-to-p6.pod6
```diff
@@ -0,0 +1,420 @@
...
+    greeting :: String -> String
+    greeting  ""   = "Hello, World!"
+    greeting "bub" = "Hey bub."
+    greeting  name = "Hello, " ++ name ++ "!"
+    
+Perl6 does this as well! You just use the `multi` keyword to signify that it is a multiple dispatch
+function.
+
+    proto greeting ( Str   --> Str ) {*}
+    multi greeting ( ""    --> "Hello, World!" ) {}
```
awwaiid: Putting a constant string as the return type feels weird to me.

WildYorkies: I would imagine that it does, as many things do in P6 😄 

I'm not sure a debate has been had on the merits of this style or not. But I do think that having constraints on the return type in the signature brings benefits in regards to more exact contracts, no?

awwaiid: Yes, but that would be more like `multi greeting ( "" --> Str ) { "Hello, World!" }`. It is the combining of the value and the thus implicitly the type on the return value that I find odd, though it seems to be allowed. Funnily I don't feel that way on the parameter itself. Might be my own taste.

doc/Language/haskell-to-p6.pod6
```diff
@@ -0,0 +1,420 @@
...
+    
+The equivalent in Perl6 is the following.
+
+    my constant Name = Str;
+
+    sub full-name ( Name \first, Name \last --> Name ) { first ~ last }
+    
+It should be noted that in Perl6, one can also create a subset of an existing type.
+
+    subset Name of String where *.chars < 20;
```
CIAvash: You forgot to change `String` to `Str`.

doc/Language/haskell-to-p6.pod6
b2gills: This document is for people coming from Haskell.

As such it makes sense to use syntax that would seem more familiar to them.

It also means that no matter what code you put in the routine, it always returns the expected value.

Then there is the benefit that calling `.signature.returns` on the routine tells you exactly what it returns.

WildYorkies: How do you feel about `multi is-thing(Thing $ --> True) {}` ?

awwaiid: I'm fine leaving this as-is overall I suppose, the feedback that it tickled my senses is enough.

`--> True` is a fun one! It makes me feel conflicted :) . I guess in a case like that my wish would be for `multi is-thing(Thing $) { True }` to be the equivalent, so that the value and the type are statically known/inferred by the compiler.

Anyway -- is fine, I don't feel super strong about it, and it definitely _works_ the way you have it.

WildYorkies:
- [ ] Make sure this shows up on the initial Language page index on the website
- [ ] Make sure this is linked to on the "resources" of the website when it is good and ready

Issue #1194, opened 2017-02-13
xt/examples-compilation.t doesn't tell you what failed


coke:
A run just now shows: 

```
not ok 4 - Language/concurrency.p6 examples check isn't successful

# Failed test 'Language/concurrency.p6 examples check isn't successful'
# at xt/examples-compilation.t line 23
```

But some of the files have hundreds of lines.

The sanity check should be per example, and it should refer back to the line number in the original file.

Altai-man:
This issue is good as a wishlist, but for now it works in different fashion.

If someone introduces a new example, he can run tests and look, is it working or not(and if not - you don't want to push it in the repo!). So, the changes author needs to review is a small subset of the actual file.

But yes, it will be a huge improvement to simplify your everyday work.

coke:
Altai-man - if the test was clean and was part of t/, and run every time, sure. But as part of xt/ , it's very easy for it to get into a bad state, like it is now.

Altai-man:
Indeed. I'm not sure by what criteria tests are selected to go into t/ or xt/. As for me, examples compilation test process is somewhat slow, that's why I placed it into xt/.

Anyway, closer to the core of the issue...

I can improve the test to grab the compiler output and print a line which was a cause of an error. It won't be as neat as original line number, because we lost it after extraction, but it will be much easier to just find appropriate string in a bad file.

On the other hand, I can parse pod6 files directly and check every example - and give proper line number to the user, but it will be some kind of already existing `extract-examples.pl6`. We can ask gfldex, who wrote it originally, what was his purpose in file extracting instead of such checking.

coke:
t/ - things that should pass on every commit and can be considered to break the build.
xt/ - can fail without breaking the build, but should keep an eye on.

So, if someone's editor leaves some trailing white spaces, or they misspell a word, that's OK. I would actually argue that failing examples are bad and should break the build.

Altai-man:
As I got it, tabs in files can break the build, but wrong examples are okay?

coke:
As of today, yes.

Altai-man:
Okay.
Anyway, I can take the job to check examples everyday(warning: not a solution).
I can update the test to show broken input, as I wrote in first variant too.
I can also discuss how it should work in general with gfldex and write some proper solution too.

This is my variants about how we can solve it.

Altai-man:
Since I'm assigned, to print what exactly brought the compiler's wrath will be enough to close this?

coke:
[...]

PR #1207, opened 2017-02-20
Coke/examples

coke:
For #1194 

This is not yet intended to be merged, but to be considered for review.

Runtime for this test is definitely slower than before, but the output in case of failure is much more helpful.

The slowness is because this branch removes the caching of a combined example per test file, using instead a precompiled version of the source (same mechanism used by htmlify) and walking over that each time the test is run. We can add a similar kind (but more complicated given the split into chunks) of caching back if needed.


Altai-man:
Let's see. In general, I like your solution, but the performance is concerning me. On my machine timing for an old variant is
```
214.27user 8.74system 3:46.54elapsed 98%CPU (0avgtext+0avgdata 850816maxresident)k
0inputs+0outputs (2major+4330917minor)pagefaults 0swaps
```
and
```
695.34user 97.95system 13:20.38elapsed 99%CPU (0avgtext+0avgdata 611648maxresident)k
0inputs+0outputs (0major+44129286minor)pagefaults 0swaps
```
for a new.

I wonder, can we improve it with concurrency? We can run examples from a single file in batches of some size concurrently and afterwards concatenate the results in a correct order. Should I try it?

With `hyper` added on last function the results on my machine is
```
1079.25user 137.64system 5:56.23elapsed 341%CPU (0avgtext+0avgdata 774800maxresident)k
0inputs+0outputs (0major+44239988minor)pagefaults 0swaps
```

Also, we should count tests from 1, not from zero, I suppose.

coke:
In general, my preference is to not add concurrency to the individual tests, since we are managing that at the test file level with prove -j / TEST_JOBS

Altai-man:
In this case, I have no objections against this changes. :+1:

coke:
New version using EVAL (with a different wrapper script to avoid executing as much as possible) instead of 'perl6 -c' - much faster; I think the previous version was not running the code but was running the code "True" each time; So the current version has some test failures (need to at least be passing the test before merge.)

coke:
Fixed - I missed the pod formatting code support on the initial version (because I wasn't actually running the example code!). Now passes all tests and is very fast.

coke:
TImings with the latest on master and coke/example:

master: 

real	1m27.413s
user	1m13.813s
sys	0m6.849s

branch: 

real	0m28.238s
user	0m25.326s
sys	0m1.414s

coke:
I'm OK if this gets merged now.

PR #1209, opened 2017-02-20
Add a concrete example of CStruct initialization

### LINKED ISSUE #1186: NativeCall examples doesn't show how to initialize a CArray in CStruct
For initialization, CArray in CStruct requires binding operator like the following test:
https://github.com/rakudo/rakudo/blob/nom/t/04-nativecall/06-struct.t#L12-L29

However, it seems that there aren't any examples that shows the above solution.

See also:
https://docs.perl6.org/language/nativecall

titsuki:
Fix #1186

### DIFF
--- a/doc/Language/nativecall.pod6
+++ b/doc/Language/nativecall.pod6
@@ -286,7 +286,22 @@ type of a native function, the memory is not managed for you by the GC.
 
 NativeCall currently doesn't put object members in containers, so assigning new values
 to them (with =) doesn't work. Instead, you have to bind new values to the private
-members: C<$!struct-member := StructObj.new;>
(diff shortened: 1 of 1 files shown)


coke:
This includes a reference to IntStruct, which isn't a type in Perl 6.

titsuki:
@coke
Thanks for reviewing!
I used IntStruct as a user-defined NativeCall object like the below:
https://github.com/rakudo/rakudo/blob/nom/t/04-nativecall/06-struct.t#L41-L50

However, as you noticed, I should indicate that IntStruct is a used-defined object.

titsuki:
Fix by #1211

PR #1215, opened 2017-02-23
change else-if to given/when switch

WildYorkies:


zoffixznet:
Feels pointless to alias to $contents just to immediately use `given` to alias it to $_ you would have had with `for` by default

WildYorkies:
Can you combine given and for?

```perl6
given for @items {
  when Str { $_ }
  when Positional { $_ }
}
```

Just trying to make the original code more readable.

b2gills:
All `given` does is temporarily store its argument in `$_`, which `for` also does by default.

b2gills:
Also you didn't need to add the parameter to the `$_.map: ...`

And you could do the same improvement to that inner mapped block

And you can do an implicit `map` method call ( not that it makes it any clearer )

    when Positional {
        @chunks.append: .map: {
            when Pod::Block {
                *.&pod-gist
            }
            when Str {
                $_
            }
        }
    }

WildYorkies:
The code needs help.

If the item is a Pod::Block, it recurses. The rest of the patterns should follow that example.

`sub pod-gist` should be a multi that accepts either a Pod::Block, Str, or Positional.

b2gills:
That could lead to a cleaner implementation

PR #1220, opened 2017-02-26
reorganize "constraining return types"

WildYorkies:


tbrowder:
Why did you remove reason 3 for not using return?

AlexDaniel:
Not sure, but the text is moved to another section.

FWIW, this is the discussion for the whole thing: https://github.com/perl6/doc/issues/1024

WildYorkies:
@tbrowder 
>Why did you remove reason 3 for not using return?

Because it is relevant to `returns`, so I put it there. Along with more examples for each method of type checking return values.

WildYorkies:
Any requests before I merge this? @tbrowder would you like me to add the 3rd point back in just to reinforce it a bit more?

tbrowder:
I think I would add a parenthetical note to the sentence leading to the return examples (or an inline comment in the '-->' example) pointing to the summary of return types and the reason for preferring '-->'. That shouldn't get lost in the noise, but I can wait until after the merge to see if I still have the same opinion.

PR #1244, opened 2017-03-11
about.pod6

hankache:
* s/POD/Pod
* s/pygments/highlights
* Add link to https://docs.perl6.org/language/pod

### DIFF
--- a/doc/Language/about.pod6
+++ b/doc/Language/about.pod6
@@ -2,7 +2,7 @@
 
 =TITLE About the Docs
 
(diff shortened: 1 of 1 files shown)


zoffixznet:
FWIW, in the docs we always write `Perl 6` as `Perl 6` with a non-breaking space in between. There are even tests for that; I think they can be run with `make xtest`

hankache:
@zoffixznet It passes the tests. Am I missing something?

zoffixznet:
> It passes the tests.

OK. Then I guess GitHub is messing with the code...

PR #1253, opened 2017-03-20
Caveat about volatile cache state in Seq

zostay:
Discussed here: <https://irclog.perlgeek.de/perl6/2017-03-20#i_14295300>


zoffixznet:
Thanks.

PR #1255, opened 2017-03-21
Fix rendering problem

dwarring:
See https://irclog.perlgeek.de/perl6/2017-03-21#i_14298152


zoffixznet:
This doesn't fix anything (as you're rendering the same characters) and just deletes one of the examples for `E< >`.

The rendering issue is due to an apparent bug in Rakudo's handling of `E< >`

zoffixznet:
> The rendering issue is due to an apparent bug in Rakudo's/Pod::To::HTML's handling of E< >

Actually, it might be Pod::To::Big page as I see Pod::To::HTML renders them fine.

Issue #1263, opened 2017-04-01
infix:<.> is not documented


AlexDaniel:
https://irclog.perlgeek.de/perl6/2017-04-01#i_14361587

```
42     .      say   # OUTPUT: «42␤»
```

Slightly related: https://docs.perl6.org/routine/$FULL_STOP

zoffixznet:
Should be addressed by #1273

PR #1273, opened 2017-04-08
Document the infix method call operators

dmaestro:
This documents the behavior implemented in:
    https://github.com/rakudo/rakudo/commit/cb25b2f475f31335d77d018474482aceec2c74de
and
    https://github.com/rakudo/rakudo/commit/be141b8d9bddc9798db6f11f70cdb117b7309526

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -23,6 +23,7 @@ tightest to loosest:
     N  Autoincrement     ++ --
     R  Exponentiation    **
     L  Symbolic unary    ! + - ~ ? | || +^ ~^ ?^ ^
+    L  Dotty infix       .= .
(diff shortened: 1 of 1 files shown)


dmaestro:
This is to address Issue #1263

zoffixznet:
Thanks. Unsure the `.=` should be in the list. As far as language is concerned, it's just a metaop, same as `×=`, `~=`, and `÷=` innit?

dmaestro:
I agree there is uncertainty about classifying some of these operators. Trying to find a balance between interpreting the spec and the implementation ;-) is not easy.
Note that the "new" precedence classification is really (in the ops definition) the same precedence level as the above higher ("Symbolic Unary") precedence in the code. What they have in common is sensitivity to whitespace to act with higher or lower precedence, as shown in the examples.
While it was necessary to document some otherwise confusing behavior, I have no illusion that this should be the last word.

PR #1282, opened 2017-04-17
Capitalize the Exception method

briandfoy:
The example shows it capitalize and running it uncapitalized does nothing apparently (but certainly not what the example does).

### DIFF
--- a/doc/Type/Failure.pod6
+++ b/doc/Type/Failure.pod6
@@ -34,11 +34,11 @@ Returns C<True> for handled failures, C<False> otherwise.
 
     sub f() { fail }; my $v = f; say $v.handled; # OUTPUT: «False␤»
 
-=head2 method exception
+=head2 method Exception
 
 Defined as:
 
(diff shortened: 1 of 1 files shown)


timo:
this refers to the getter of the "exception" attribute; it's supposed to be lowercased.

Since Failure is Nil, calling any method (especially methods it doesn't have) on it will return self, so what happens in the example is:

A Failure object is created. Exception is called on it (but it might as well have been GarbleWharbleFarble instead). That method returns self. The result is sunk. The Failure gets thrown because sinking an unhandled failure throws the inner exception.

briandfoy:
Then, the example needs to change I think.

zoffixznet:
Fixed the example. Thanks.

zoffixznet:
> Exception is called on it (but it might as well have been GarbleWharbleFarble instead). That method returns self. The result is sunk.

That's incorrect. The `.exception` is an attribute for the `Failure` type that contains its exception. It doesn't get thrown just by calling exception.

PR #1288, opened 2017-04-24
5to6: Add caution about `exec` workarounds

treyharris:
"shell and exit" are poor substitutes for exec in some common
situations. Add language to 5to6-perlfunc to explain these situations
and suggest NativeCall if they pertain.

Also note that most Unixes have no single "exec" function exactly (Perl
5 automagically dispatches the right one for the situation).


zoffixznet:
Thanks.

PR #1294, opened 2017-04-29
Added a note about run/shell exceptions due to non-zero exits

CurtTilmes:
Per discussion on IRC and SO:

http://stackoverflow.com/questions/43199427/why-doesnt-perl-6s-try-handle-a-non-zero-exit-in-shell


zoffixznet:
Thanks.

PR #1296, opened 2017-05-05
Added multi-positional use of assuming

molecules:
Borrowed content from brian d foy's Stack Overflow post http://stackoverflow.com/a/43669837/215487


coke:
Default licensing of content on stack overflow means we probably should not be cutting and pasting it into our doc site.

zoffixznet:
Thanks.

zoffixznet:
Talked to coke on IRC; for anyone else: I moved the ref to SO to [license section of README](https://github.com/perl6/doc/commit/e8ae3b), so that all the licensing stuff is in one place and isn't part of actual content users learn from.

PR #1324, opened 2017-05-23
Invalid Python code in class-inheritance-example

jecki:
The lines
class Animal:
	def jump:
		print "I am jumping"

are invalid, because the self parameter is missing the definition of method "jump" and the print-statement requires parentheses in Python 3

correct code:

class Animal:
	def jump(self):
		print("I am jumping")

### DIFF
--- a/doc/Language/py-nutshell.pod6
+++ b/doc/Language/py-nutshell.pod6
@@ -548,8 +548,8 @@ Inheritance is done using C<is>:
 Python
 
     class Animal:
-        def jump:
-            print "I am jumping"
+	    def jump(self):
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks!

toolforger:
@jecki, your patch replaced some spaces with tabs. Now we have a mixed-indentation code, which is a Python Worst Practice - can you fix that? (I could fix that myself, but I'd have to pull the repo first.)

jecki:
sure! just done that.

toolforger:
Show me the commit and I'll review and merge :-)

Issue #1340, opened 2017-05-26
Don't use a non-twigil ^ to explain the twigil ^


briandfoy:
in [Variables](https://docs.perl6.org/language/variables)

> The ^ twigil declares a formal positional parameter to blocks or subroutines.

The example includes the use of a different `^`:

	for ^4 {
		say "$^b follows $^a";
	}

I think it would be less confusing to the newbie to see only one type of `^` in that example.

coke:
The merge 4 days ago fixed this.

PR #1342, opened 2017-05-27
Changes a bit examples and grammar

JJ:
Mainly for #1340 but while I was at it, I have formalized a bit some
more text. This example is more perl6-ish, using `reduce` which is
exclusive of Perl6


KrisShannon:
I'm not sure the example is better.  Because multiplication is associative the ordering distinction is not as explicit.  Maybe change it to:

~~~
my @powers-of-three = 1,3,9…100;
say reduce { $^b - $^a }, 0, |@powers-of-three;

# OUTPUT:
# 61

JJ:
Sure.

Issue #992, opened 2016-10-27
Cool part of List forms a nice landmine, document it all as traps


AlexDaniel:
``` perl6
my @a = <a b c d>;
say @a.contains(‘c’);    # True
say @a.index(‘a’);       # 0
say @a.starts-with(‘a’); # True
```

VynceMontgomery:
I don't understand what about this is a trap.

pmichaud:
Each of the examples given does exactly what I would expect it to do... I'm also not sure how this is a landmine/trap.

gfldex:
see: https://irclog.perlgeek.de/perl6-dev/2016-10-27#i_13476643

The majority opinion seams to be no-trap. However, we do a poor job to explain the dynamic nature of Perl 6 in general. We don't even got a section in the docs to put such explanations.

AlexDaniel:
Actually, the majority opinion is that it is a trap indeed, hence the whole discussion about ways to solve it. For example, the first comment here clearly indicates that.

pmichaud:
Just as there are operators that coerce their operands to a specific type (e.g., &infix:<+>, &infix:<~>, &infix:<?&>), so there are (Cool) methods that treat their invocants as a specific type.  For example, @array.flip means something very different from @array.reverse .  And we eliminated ".length" altogether because of its inherent ambiguity, moving instead to .elems, .chars, .codes, etc.

So, .index implies "coerce to a string".  I don't know when .contains and .starts-with were added to Perl 6, but it seems clear to me that those were intended to be "stringy methods", and if I use them I expect them to treat their invocant as if they are a string.  That's just the way Perl 6 is designed to work.

Moving List out of Cool strikes me as all sorts of wrong -- having Lists (Arrays) automatically coerce into string or numbers as needed has always struck me as being one of the fundamental strengths of Perl.

Pm

AlexDaniel:
@pmichaud yea, great, but as far as I know there is no ticket yet where this discussion would be appropriate.

[...]

PR #1412, opened 2017-07-07
Document some of the Cool traps

VynceMontgomery:
As discussed in #992 there are a few possible traps in Cool, especially with regard to Lists. Some of these behaviors may change, but while they work this way, let's document them.

### DIFF
--- a/doc/Language/traps.pod6
+++ b/doc/Language/traps.pod6
@@ -229,6 +229,91 @@ Instead, use an expression that produces a value when you want a value.
 =for code
     my $a = 2; say join ",", (+$a, ++$a); # OUTPUT: «2,3␤»
 
+=head1 Cool tricks
+
+Perl6 includes a L<Cool|/type/Cool> class, which provides some of the DWIM 
+behaviors we got used to in Perl5 by coercing arguments when necessary. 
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks.

coke:
Please make sure PRs pass xt/ before merging.

zoffixznet:
I thought we agreed xt/ passing would be made monthly, at rakudo release time.

coke:
I thought that was a one time thing to allow you to deal with the large number of failing tests as a result of the IO work.

zoffixznet:
No, it was [a much more recent discussion](https://irclog.perlgeek.de/perl6/2017-06-19#i_14757201) that started with a proposal to block all contributions unless they pass xt tests.

The relevant parts of of discussion that I thought meant we fix xt only once a month are these:

```
    <eveo> [...] I don't mind making it it part of the Rakudo release process to ensure the docs xt tests pass
    <[Coke]> [...] if they got fixed on the regular, I'd have nothing to complain about.
    <eveo> .in 24d Zoffix fix xt on docbuild
    <eveo> [Coke]: I'll fix then all the xt tests as part of rakudo's release process.
```

PR #1423, opened 2017-07-25
possible fix on reformatting the slurp example

faraco:
* on doc site
   - https://docs.perl6.org/routine/slurp
- Since I don't know if this really 'fixing' it, I made this as a PR. Please suggests me changes to actually deal 
with this. Thanks.

best regards,
faraco.

### DIFF
--- a/doc/Type/IO.pod6
+++ b/doc/Type/IO.pod6
@@ -272,7 +272,7 @@ Accepts C<:bin> and C<:enc> optional named parameters, with the same meaning
 as L<open()|/routine/open>.  The routine will C<fail> if the file does not
 exist, or is a directory.
 
-=for code :skip-test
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Does it pass ``make xtest``? I'm thinking about [this one](https://github.com/perl6/doc/blob/master/xt/examples-compilation.t).

faraco:
Should I add `:skip-test` attribute too?

AlexDaniel:
Well, yes, or change the examples so that they pass.

faraco:
Since my main goal is to fix the formatting first, I'll add the `:skip-test` for now.

AlexDaniel:
Well, we will see how it goes. Thanks!

PR #1435, opened 2017-08-07
Changed link in the data dump section of the FAQ

nkh:


AlexDaniel:
👍, but shouldn't it be “http://modules.perl6.org/s/dump” ? Both work, I don't know which one is preferred.

nkh:
I wouldn't know, I took the link I had in my browser.

AlexDaniel:
Right. It's probably OK. I was thinking about this commit https://github.com/zoffixznet/perl6-buggable/commit/36f1cd02808ac3c04defaa8ed11f4a7516b0763e and how it uses ``/s/`` for everything.

By the way, you removed ``X<|Data::Dumper (FAQ)>``. I think the intention was to let people who know perl 5 find this question/answer easier. Should we really remove it?

Now that I think about it, maybe the mention of rakudo-specific ``dd`` should actually say that it is not a Perl 6 feature, or at least link to the more [detailed description](https://docs.perl6.org/programs/01-debugging#index-entry-dd). Feel free to do it if you have time.

AlexDaniel:
@nkh whoops, I think this last commit was not supposed to be in this pull request. Github adds whatever is pushed into the branch to the PR, which is exactly what happened here (this way you can make corrections and add more commits after discussion). Maybe you want to create another [branch](https://git-scm.com/book/en/v2/Git-Branching-Branches-in-a-Nutshell) for your other contributions.

zoffixznet:
Thanks. Merged.

> 👍, but shouldn't it be http://modules.perl6.org/s/dump ? Both work, I don't know which one is preferred.

It's a bit muddy ATM: `/s/foo` won't accept stuff like `/` in it, but `?q=foo` will. Currently, it's so we could get a JSON API for free by just tacking on `/.json` at the end of the URL and using more liberal placeholders in routes will interfere with that. I think eventually we'll go that route, after defining the API a bit better.

PR #1437, opened 2017-08-07
if/elsif/else series of blocks as a special case for implied separator

molecules:
Does this look okay? Inspired by brian d foy's question (https://stackoverflow.com/questions/45520479/is-perl-6s-uncuddled-else-a-special-case-for-statement-separation).

### DIFF
--- a/doc/Language/syntax.pod6
+++ b/doc/Language/syntax.pod6
@@ -77,15 +77,14 @@ A Perl 6 program is a list of statements, separated by semicolons C<;>.
 A semicolon after the final statement (or after the final statement inside a
 block) is optional, though it's good form to include it.
 
-A closing curly brace followed by a newline character implies a statement
-separator, which is why you don't need to write a semicolon after an C<if>
-statement block.
+In general, a closing curly brace followed by a newline character implies a statement
+separator,  which is why you don't need to write a semicolon after an C<if> statement block.
 
 =begin code
 if True {
-    say "Hello";
+    put "Hello";
 }
-say "world";
+put "world";
 =end code
 
 Both semicolons are optional here, but leaving them out increases the chance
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Cool. Looks helpful.

But why ``put`` instead of ``say``?

molecules:
For this example, it doesn't make a difference, but I've been bitten by the difference a few times.

Compare `say 1 .. 5` to `put 1 .. 5`.

I feel that `put` is more predictable. It always outputs the `.Str` value, which is what I would expect. `say` outputs the `.gist` of a variable, which occasionally isn't what I would expect.

molecules:
I'm totally fine sticking with `say` in this instance, if that is considered preferred by the community.

AlexDaniel:
Well, the docs use ``say`` consistently. I think that yes, the example should use whatever the rest of the docs are using, unless there is some specific reason to use something else.

The .gist is supposed to be human-readable, and this is what we want almost everywhere in the docs. If .gist of something is less than awesome, then we should file it as a rakudobug. In case of ``..``, however, you get a range object, so the .gist of it seems to be a bit more transparent. But it probably depends on the way you look at it.

molecules:
Thanks! Updated `put`s to `say`s.

AlexDaniel:
Looks fine. By the way, ``put`` vs ``say`` is an interesting issue, there's a chance that we might want to replace a bunch (not all) of ``say``s in the docs with ``put``s. But I think that's a separate issue.

molecules:
I'm all for doing a `say`/`put` refactor, taking into account the few cases where the `.gist` really is desired.

AlexDaniel:
I wouldn't be so eager to say that there are *few* cases where ``.gist`` is desired, but it's true that some things would be better off with ``put``.

AlexDaniel:
Another thing to note: ``put`` is a [fairly new addition](https://github.com/rakudo/rakudo/commit/bcebf8e96ff7cd68cdc9f4ef9815c724544082e6), so in case of some older parts of the docs ``put`` wasn't considered as an option.

ronaldxs:
I tried to raise the issue of `put` vs `say` with @briandfoy after looking at a preview of Learning Perl 6.  The book appears to me to prefer `put` and use the two a bit interchangeably whereas this discussion seems to me to suggest that `say` might be a preferred default for introductory examples where there is no reason to expect other software to read the output.

AlexDaniel:
Your understanding is correct. I don't have access to Learning Perl 6 so can't comment on that.

When I say that docs should sometimes use ``put`` I'm mostly referring to examples like this (from [Proc](https://docs.perl6.org/type/Proc) page):
```perl6
my $p = run "cat", "-n", :in, :out;
$p.in.say: "Hello,\nworld!";
$p.in.close;
say $p.out.slurp: :close;
```
It does work… but realistically, how often would somebody need to pipe a gist of something into a process? This is exactly the situation where one is better off with ``put``. In the end, someone will use that example (maybe not knowing about ``put`` at all), and there we have a user using .gist when interfacing with other programs.

We can change that example quickly, but I'm interested in seeing a more structured effort to resolve the issue (not just the one snippet that jumped at me).

coke:
This is something that should be covered by the STYLE guide, and enforced by the existing editors. (Knowing which context the sample is in isn't something that's going to be easily automatable.)

AlexDaniel:
I created this issue so that we don't have to discuss it on a closed pull request: https://github.com/perl6/doc/issues/1441

PR #1439, opened 2017-08-07
Regarding separators in if/elsif/else blocks.

rafaelschipiura:
Trying to simplify the explanation to make it easier to understand.


zoffixznet:
Thanks.

PR #1440, opened 2017-08-07
if/elsif/else construct being one statement simplifies explanation

molecules:
Jonathan Worthington pointed out that the if/elsif/else block is really a single statement, which makes this easier to explain.


AlexDaniel:
Currently we have two pull requests for this. Another one: https://github.com/perl6/doc/pull/1439

AlexDaniel:
Does it mean that the other one is better? :)

PR #1450, opened 2017-08-15
Add missing title for but in list construction

book:


zoffixznet:
Thanks.

FWIW, you have commit rights to most repos under `perl6` org and can commit directly.

book:
I know, I wasn't sure what the polite way to do things was.

PR #1460, opened 2017-08-19
Add documentation of loose xor operator

dmaestro:


zoffixznet:
Perhaps it'd be better to describe it as simply a looser version of  [`^^` operator](https://docs.perl6.org/routine/$CIRCUMFLEX_ACCENT$CIRCUMFLEX_ACCENT), linking to it. That way the user learns two operators for the price of one.

dmaestro:
Good suggestion - I'll revise this, probably tomorrow.

coke:
Thanks

Issue #1471, opened 2017-08-24
A closing curly brace followed by a newline character doesn't imply anything.


briandfoy:
From the [syntax doc](https://docs.perl6.org/language/syntax#Separating_Statements):

> A closing curly brace followed by a newline character implies a statement separator, which is why you don't need to write a semicolon after an if statement block.

But this is not true. The closing curly in an uncuddled if-else won't imply a statement separator. The actual rule is that the end of a statement followed by vertical whitespace implies a statement separator. But, a statement can be anything that Perl 6 decides is a statement. 

See [Is Perl 6's uncuddled else a special case for statement separation?](https://stackoverflow.com/q/45520479/2766176)

AlexDaniel:
This is clarified in the same section:
> This doesn't happen in series of blocks that are part of the same if/elsif/else (or similar) construct because they are part of a single statement.

Isn't it good enough?

briandfoy:
It would be better without special cases.

b2gills:
I think of it more as the specialness of `else`/`elsif` that causes it to bind to the previous iffy statement, including ignoring/undoing the automatic statement separator. (but not ignoring an explicit statement separator)

After all it is special anyway, because it has to follow such a statement.

smls:
To be really pedantic, the sentence quoted from the doc also doesn't specify that the rule only applies to the **closing braces of blocks** in the **"main" Perl 6 slang**.

It doesn't apply inside the "string" slang:

``` Perl 6
dd "{1+1}
4";
# Prints "2\n4"
```

Nor inside the "regex" slang:

``` Perl 6
say "abcd" ~~ / . ** {1+1}
$ /;
# Prints ｢cd｣
```

Nor to curly braces that aren't the end delimiter of a block:

```
sub infix:<|}> { $^a + $^b };  say 10 |}
5;
# Prints 15
```

@b2gills:
> I think of it more as the specialness of else/elsif that causes it to bind to the previous iffy statement, including ignoring/undoing the automatic statement separator.

Agreed. As long as the promised Perl 6 macro support with a public API for traversing/manipulating Perl 6 AST hasn't arrived yet, there's nothing to say that an `if` / `else` pair is even *supposed* to be viewed as a single statement. (How Rakudo/NQP sees it internally, is merely an implementation detail.)

zoffixznet:
Thanks. I believe https://github.com/perl6/doc/commit/7bb1fec099ca12119575f300ee4ca3682bf662f8 addresses this Issue.

PR #1476, opened 2017-08-29
Implied separator rule: focus on ending blocks (instead of the curly brace itself)

molecules:
I've modified the explanation to focus more on the ending blocks than on the ending semicolon. This is to make improvements toward resolving (or at least mitigating) concerns with issue #1471. Also included an example of implied separator rule with a colon method call.


zoffixznet:
Thanks.

Issue #1477, opened 2017-08-29
Unnecessary style recommendation: semicolon after the last statement


AlexDaniel:
https://docs.perl6.org/language/syntax#Separating_Statements_with_Semicolons
> A semicolon after the final statement (or after the final statement inside a block) is optional, though it's good form to include it.

What is meant by “good” here? Why? Can't we just end the sentence on “is optional”, without giving unjustified recommendations?

coke:
This was probably meant as more of a style suggestion: same reason you'd end a multi-line list with a comma when you don't have to, to prevent yourself from adding code at the end and getting a compile error.

But that can go in a style guide. Feel free to modify this as described.

coke:
(Note: not OUR style guide. more like a "best practices" thing if someone wants to make one.)

smls:
FYI, moritz once started a "Well Thought-Out Practices for Perl 6" document here:
https://github.com/moritz/perl6-wtop/blob/master/practices.pod

AlexDaniel:
> same reason you'd end a multi-line list with a comma when you don't have to, to prevent yourself from adding code at the end and getting a compile error

It's not the same. Very often you *rely* on the returned value from the last statement. The fact that you can't add code at the end without modifying the line that returns stuff is a very good reason to always omit the semicolon after the last statement.

I've been using this style (always a trailing comma after a list, never a trailing semicolon after the last statement in a block) for a few months with a great success. There were a few times when it saved me.

Also, whether you use an explicit ``return`` or not makes no difference. With or without return, the last block is always special, no matter if you care about the returned value or not. If there is ever a well thought out style guide, I'd expect it to recommend no trailing semicolon.

zoffixznet:
👍 on removing that sentence. I too found it off-colour when I read it a few days ago. Personally, I leave off the colon deliberately, to indicate the statement's value is used as a return value.

AlexDaniel:
(Off-topic FWIW) What @zoffixznet just said sounds reasonable too, but I'm convinced that doing that makes this style decision less prominent when someone is reading (or contributing to) your code. I've noticed that in most cases the last line in any block (even if/else) is always somewhat special, so the amount of times I had to add a semicolon to something is probably 0. I can see how it can improve the readability if you have ``do`` in front of your if/else, but at the same time I don't see myself removing semicolons if I ever decide to slap a ``do`` on something. Strict binary style decisions make it easier for me, but YMMV.

molecules:
[...]

PR #1479, opened 2017-08-30
Cleaning up "Separating statements by semicolons"

molecules:
Trying to address #1477

### DIFF
--- a/doc/Language/syntax.pod6
+++ b/doc/Language/syntax.pod6
@@ -97,12 +97,17 @@ say "world"
 
 =head2 Implied Separator Rule (for statements ending in blocks)
 
-Complete statements ending in blocks can omit the trailing semicolon, if no
+Complete statements ending in bare blocks can omit the trailing semicolon, if no
 additional statements on the same line follow the block's closing curly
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Looks about right. 👍

molecules:
What is the difference between travis-ci/push and travis-ci/pr ? The config files for each were identical.

zoffixznet:
2 cents:

> However, if a block is a postcircumfix operator (e.g. indexing of a hash)

A postcircumfix operator is not [a block](https://docs.perl6.org/type/Block). It just uses the same curly braces.

molecules:
I'm not sure how else to explain it:

```
my %hash = '5' => 'A', '7' => 'B';

put %hash{
            { put 'inside hash index'; 1}() +
            4
        };
put %hash{ 28/4 };

# Outputs
# inside hash index
# A
# B

```

zoffixznet:
Perhaps explanation isn't needed? If the text talks about blocks it might just add more confusion to also start mentioning non-blocks.

zoffixznet:
👍 Thanks!

Issue #1436, opened 2017-08-07
Document the use of `.polymod` for base conversion.


smls:
In practice, I've found the [`polymod`](https://docs.perl6.org/routine/polymod) method to be good for two things:

1. Decomposing a quantity into a compound quantity with multiple units:  
  `my ($sec, $min, $hrs, $days, $weeks) = $seconds.polymod(60, 60, 24, 7);`  
  `my ($B, $KiB, $MiB, $GiB, $TiB) = $bytes.polymod(1024 xx 5);`

2. Getting the digits of a number in an arbitrary base:  
  `my @digits-in-base37 = $number.polymod(37 xx *);`

The documentation currently presents use-case (1), but not (2).

AlexDaniel:
Interesting idea. However:
```perl6
say +0.polymod(36 xx *) # OUTPUT: «0»
```
So, it's 0 digits, is it correct? Well… I don't know, depends on what you need I guess

“Arbitrary” base is also an interesting way to put it. It will hang if the “base” is 1, which is not something someone would have, but still. Then, it works for negative bases, but throws with base 0.

Too many edge cases I think, but the idea is very interesting.

smls:
I wouldn't call base 1 or negative bases "edge cases" so much as "out of scope".
AFAIK, standard positional numeral systems are restricted to integers bases greater or equal to 2, and `polymod` works fine for all of those.
While the concepts of "unary numeral system" and "negative base" do exist, they are defined separately, and should not be expected to be handled by the same expression that handles standard positional numerals.

You're right about `polymod` yielding an empty list for input `0`. When that is unwanted, the expression can be easily extended to:

    my @digits-in-base37 = $number.polymod(37 xx *) || 0;

However, in the cases where I've used it so far, that wasn't necessary, because the presence or absence of the zero doesn't affect the reverse operation:

    my $number = [+] @digits-in-base37 Z* (37 X** 0..*);

Altogether, I think `polymod` is currently the best solution for base conversion in Perl 6 when you either...

- Need to deal with **bases greater than 36** *(since the [`.base`](https://docs.perl6.org/routine/base) method doesn't support those)*.
- Need to know the **numeric value** of each digit *(since `.base` represents digits as alphanumeric characters in a string)*

That's why I suggested presenting this idiom in `polymod`'s documentation.

AlexDaniel:
See https://github.com/perl6/doc/pull/1487. Is that it? Or do we want some extra text explaining why it works?

[...]

PR #1487, opened 2017-09-02
Document the use of polymod for base conversion

AlexDaniel:
Attempt to resolve issue #1436.


zoffixznet:
Perhaps the example should go the extra mile and convert the digit indices to actual digits of the number?

```perl6
my @digits-in-base37 = flat 0..9, 'A'..'Z', '♥';
my $number = 9123607;
my $number-in-base37 = @digits-in-base37[ reverse $number.polymod: 37 xx * ].join;
say $number-in-base37; # OUTPUT: «4W4F♥␤»
```

rafaelschipiura:
I think one important detail missing it to remind people that the built-in base conversion only covers a limited number of bases.

Issue #1468, opened 2017-08-23
promote react style in Proc::Async


gfldex:
https://docs.perl6.org/type/Proc::Async
something along the lines of:
```
my $line-source = Proc::Async.new("perl6", "bin/line-source.p6", :w);

react {
    whenever $line-source.stdout -> $l is copy {
        say $l.chop;
    }
    whenever $line-source.start { done }
    whenever signal(SIGINT) {
        $line-source.close-stdin;
        exit 0
    }
    whenever Promise.in(2) { $line-source.put: 1 }
}
```

AlexDaniel:
Yes: https://irclog.perlgeek.de/perl6-dev/2017-07-17#i_14882806

AlexDaniel:
Also, maybe this:
```perl6
    whenever Promise.in(2) {
        whenever $line-source.put(1) {
            $line-source.close-stdin
        }
    }
```

AlexDaniel:
Also, maybe change ``whenever $line-source.stdout`` to ``whenever $line-source.stdout.lines``

coke:
Any chance we can get this as a PR?

AlexDaniel:
Pretty sure that this ticket is now resolved. I'll even it open for someone else to double-check.

rafaelschipiura:
Looks good.

PR #1489, opened 2017-09-02
Proc::Async example with ｢react｣

AlexDaniel:
Tackle issue #1468.

IMO in this case it is reasonable to expect that people will be
copying the example and reducing it to their needs, so I tried to
include everything that may be needed.

The logic for kill signals is questionable, I don't know if there's
anything better we can show.

**This is not mergeable because of [RT #132016](https://rt.perl.org/Ticket/Display.html?id=132016)**. Either wait for the
ticket to be resolved or get rid of ｢Supply.merge｣ and ｢Supply.zip｣.

### DIFF
--- a/doc/Type/Proc/Async.pod6
+++ b/doc/Type/Proc/Async.pod6
@@ -28,7 +28,7 @@ standard output and error handles, and optionally write to its standard input.
             say ‘stderr: ’, $_
         }
         whenever $proc.start {
-            say ‘Proc finished’;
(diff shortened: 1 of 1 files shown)


jonathanstowe:
Not a thing on this PR but It occurs to me that there probably isn't an example of  ````react```` with a variety of dissimilar supplies like this (or even more complex,) perhaps something fpr the concurrency page?

PR #1491, opened 2017-09-02
Fix examples indentation

Altai-man:
I am really not sure if we planned this, but looking on https://docs.perl6.org/language/traps it seems that code examples indentation is dead broken: examples are shifted to the right, while no one does that. It looks very strange for me. And even if it's still a convention(?), we had two spaces indentation there too.

Normalize indentation level for all examples on traps page.


AlexDaniel:
Err, yeah, I was simply copying surrounding examples without thinking about it too much.

Issue #1490, opened 2017-09-02
List.squish supports “with” parameter, but it is not documented


AlexDaniel:
```perl6
say [42, “42”].squish(with => &infix:<eq>) # (42)
```

PR #1493, opened 2017-09-02
Document `with` of `squish`

Altai-man:
Fixes https://github.com/perl6/doc/issues/1490

Is explanatory enough?


AlexDaniel:
That's good enough. Thanks!

PR #1498, opened 2017-09-02
Clear ambiguity regarding Int/Real rand()

### LINKED ISSUE #787: Int/Real.rand is ambiguous
At: https://docs.perl6.org/type/Int#routine_rand

The docs use the word "between" without clarifying the endpoint behavior. This should probably be made explicit. I went digging for the exact code, but ended up getting lost at the point that nqp doesn't actually reference where rand comes from. I'm assuming it's built into the behavior of Nums somewhere, but I was unable to find it immediately. My assumption would be that it's a real between 0 (inclusive) and n (non-inclusive).

zakame:
Specify whether the endpoints are included or not when selecting the
pseudo-random number.

References:

- https://github.com/perl6/roast/blob/master/S32-num/rand.t#L13-L30
- https://github.com/MoarVM/MoarVM/blob/master/3rdparty/tinymt/tinymt64.c#L97-L116

Fixes #787.

### DIFF
--- a/doc/Type/Real.pod6
+++ b/doc/Type/Real.pod6
@@ -21,9 +21,9 @@ Converts the number to a C<Rat> with the precision C<$epsilon>.
     sub term:<rand> (--> Num:D)
     method rand(Real:D: --> Real:D)
 
-Returns a pseudo-random number between zero and the number.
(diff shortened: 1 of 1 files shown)


Altai-man:
Looks good. Thanks a lot!

zakame:
Thanks! 🎉 🍕

PR #1499, opened 2017-09-02
tell what use is looking for, hint to where. closes #1454

### LINKED ISSUE #1454: be explicit that `use` looks for *.pm and *.pm6
At least in https://docs.perl6.org/language/modules, there may be more spots.

gfldex:


Altai-man:
Was just looking on that issue! Seems good, thanks.

PR #1501, opened 2017-09-02
Show how to put modules into default locations. closes #1500

### LINKED ISSUE #1500: we don't tell where modules are looked for by default
needs to be added at least to https://docs.perl6.org/language/modules#Finding_Modules

It's the job of zef to know exactly but we should mention that files will be put into $HOME.

gfldex:


Altai-man:
Thanks! In my opinion, it can be better with:
1)Removing of the indentation.
2)Indenting of =begin/=end blocks too(less pretty solution).
Otherwise unnecessary white space will be rendered, which is LTA.

gfldex:
indentation fixed

PR #1502, opened 2017-09-02
(Try to) clarify type smileys for .DEFINITE-ness

### LINKED ISSUE #1469: Type constraint smileys conflate .defined-ness with .DEFINITE-ness
The type smileys constrain based on whether an object is `.DEFINITE` (or "concrete" or a "type object" vs. "an instance"), but [the current description](https://docs.perl6.org/type/Signature#Constraining_Defined_and_Undefined_Values) describes the difference very loosely, and it's easy to misunderstand it to be saying that the smileys check for value of `.defined` instead.

(P.S.:  "type smileys" doesn't find any search results, even though [the docs do seem to use that term in a few places](https://www.google.com/search?q=site%3Adocs.perl6.org+smileys))

zakame:
Based on conversation with @zoffixznet, try to add more context (and
definiteness) to the current elaboration for the `:D`, `:U`, and `:_`
type constraints.  I realize this is also a good place here to show type
objects and instances, to support the concept of definiteness.

Fixes #1469.

### DIFF
--- a/doc/Type/Signature.pod6
+++ b/doc/Type/Signature.pod6
@@ -220,8 +220,10 @@ correct type.
     #     (Str:D $: *%_)»
     say limit-lines "a \n b", Int # Always returns the max number of lines
 
-In this case, we really only want to deal with defined strings. To do this, we
-use the C<:D> type constraint.
+In the code above, we really only want to deal with defined strings. To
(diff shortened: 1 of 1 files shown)


zoffixznet:
👍 Thanks!

zakame:
Thanks too! 🎉

Issue #1494, opened 2017-09-02
A lot of TODOs in Any


AlexDaniel:
https://docs.perl6.org/type/Any#method_kv

Altai-man:
Some table to keep up things:

- [ ] sum -  https://github.com/rakudo/rakudo/blob/nom/src/core/Any.pm#L469 but `my $a; $a.sum;` throws an exception. ???
- [x] keys
- [ ] flatmap - `$a.flatmap;` throws `Cannot resolve caller flatmap(Any: );`. It seems it is not even here in the first place. Delete?
- [ ] roll - TBD
- [ ] pick - TBD
- [ ] head - throws `Cannot resolve caller`. Delete?
- [ ] tail - throws `Cannot resolve caller`. Delete?
- [ ] skip - TBD
- [ ] prepend - TBD
- [ ] unshift - TBD
- [ ] first - TBD
- [ ] unique - TBD
- [ ] repeated - TBD
- [ ] reduce - TBD
- [ ] permutations - TBD
- [ ] categorize - TBD
- [ ] classify - TBD
- [ ] produce - TBD
- [ ] rotor - TBD
- [ ] pairs - TBD
- [ ] antipairs - TBD
- [ ] kv - TBD
- [ ] tree - TBD
- [ ] nl-out - TBD
- [ ] invert - TBD
- [ ] combinations - TBD
- [ ] print-nl - TBD
- [ ] iterator - TBD
- [ ] grep - TBD
- [ ] match - TBD
- [ ] append - TBD
- [ ] join - TBD
- [ ] values - TBD
- [ ] collate - TBD
- [ ] batch - TBD
- [ ] cache - TBD

Altai-man:
`head` and `tail` on Any should be removed, I guess, because they need `Any:D`, not `Any:U`.

Altai-man:
`print-nl` is broken, so delete it for now(?). See http://colabti.org/irclogger/irclogger_log/perl6?date=2017-09-02#l1233

AlexDaniel:
@Altai-man what's the current status of this?

Altai-man:
@AlexDaniel, hi.

Well, I indeed did some small effort towards this issue closing, perhaps not too much though, but as for now the things are like that:
1)Unfortunately, right now I don't have enough time&knowledge to explain `Any` business...
2)...as seen in https://github.com/perl6/doc/issues/1731 - a nice catch, by the way.
[...]

Issue #1189, opened 2017-02-11
Unify differences between `multi` and `multi sub`


trosel:
Some code is like:

`multi sub hello($name) {say "hi $name"}`

Other code is like:

`multi hello($name) {say "hi $name"}`

I'm not sure what is going on here, but this is something that should be decided upon. I think `multi sub` makes sense because you always clarify `multi method` in classes. And yet, I've never seen a `proto sub`. I've only seen `proto` by itself.

Thoughts?

zoffixznet:
It defaults to `sub`, so you can omit it. In core, though, we always add it anyway. Makes it easier to `grep 'sub blah'` the source.

Altai-man:
Judging by
```
> grep -I -r 'multi sub ' doc/ | wc -l
278
> grep -I -r 'multi [^sm]' doc/ | wc -l
50
```
plain multi case is less common. For the reasons of consistency we can edit such cases to full form. It is quite easy job too, if you have a good editor and some time on board.

WildYorkies:
I think consistency in the docs is a worthwhile goal

gfldex:
I do not agree. You don't learn how to deal with complexity by reading code that does it's best to hide it. If we prefer one style we should prefer the shorter version because lazyness is a programmers virtue.

WildYorkies:
@gfldex Whether the docs show `multi sub` or just `multi`, I still think consistency should be the goal.

Programmers can learn complexity by reading real world code / books / tutorials. The docs should feel as if they're all written by one person, no? Hence unifying the comment styles, etc.

awwaiid:
So how about unifying on `multi` alone, only using `multi sub` long-hand when explaining that it is long-hand.

WildYorkies:
I just read in the docs that one reason why `-->` is preferred over `returns` is for consistency in the docs.

So someone at some time deemed consistency in documentation worthwhile. Whoever wrote that 😸

Altai-man:
@WildYorkies, you can read https://github.com/perl6/doc/issues/1024#issuecomment-264660360 - it explains the choosing of "-->" and there are more reasons: It is more universal(you can use it for constant values) and there are no plans about deprecation, TimToady likes it, etc. The consistency made only one thing in that issue: the issue was opened because of that.

If the docs don't clearly say about possible deprecation and constant values, it is bad. Though, I take a peek now and as for me, it is still explained.

[...]

PR #1503, opened 2017-09-02
Attempt to document most of `Any` methods

Altai-man:
See https://github.com/perl6/doc/issues/1494 for details.

Help is needed with:
- [ ] Confirm removal of "broken" methods.
- [ ] Check my understanding of Any's semantics to ensure there are no lies in this text.
- [x] Help with some methods definition, such as: `collate` and `cache`.

### DIFF
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -789,7 +789,7 @@ Based on C<$matcher> value can be either C<((Any))> or empty List.
 
     my $a;
     say $a.grep({ True }); # OUTPUT: «((Any))␤»
-    say $a.grep({ $_ });   # OUTPUT: «(␤»
+    say $a.grep({ $_ });   # OUTPUT: «()␤»
 
 =head2 method append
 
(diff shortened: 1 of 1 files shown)


AlexDaniel:
I've separated the third point into https://github.com/perl6/doc/issues/1507 so that this pull request is more manageable.

doc/Type/Any.pod6
```diff
@@ -111,7 +111,7 @@ into the newly created Array.
 
 Defined as:
 
-    multi sub    reverse(*@list  --> Seq:D)
+    multi        reverse(*@list  --> Seq:D)
     multi method reverse(List:D: --> Seq:D)
```
moritz: While the "sub" is redundant, I prefer to have it in there, just because not everybody knows that "multi" on its own implies "sub".

Altai-man: See https://github.com/perl6/doc/issues/1189

Altai-man:
Also, `reduce` and `produce` on Any need an explanation too.

Issue #1121, opened 2017-01-05
Document use of method invocant to distinguish between class and object methods


ronaldxs:
See also perl6/roast [#217](https://github.com/perl6/roast/issues/217) and [IRC](https://irclog.perlgeek.de/perl6/2017-01-05#i_13862614).

The technique is sometimes used by examples in the docs but not really described in either of [Classes and Objects](https://docs.perl6.org/language/classtut) or [Object Orientation](https://docs.perl6.org/language/objects).  It would seem helpful to have it discussed directly someplace where someone might look for it.

```Perl6
class C {
    multi method f(::?CLASS:U:){say "class method"}
    multi method f(::?CLASS:D:){say "object method"}
}
C.f;       # says class method
C.new.f;   # says object method
```

zakame:
Hi, I think I can help out on this.  I suppose this would be elaborated more on [Object Orientation](https://docs.perl6.org/language/objects), to keep the class tutorial simple?

zoffixznet:
> I suppose this would be elaborated more on Object Orientation, to keep the class tutorial simple?

Sounds good. Perhaps in the [Methods](https://docs.perl6.org/language/objects#Methods) section?

zakame:
Yep, that section sounds good.  And since `multi` is shown here, I suppose linking to [declarator `multi`](https://docs.perl6.org/syntax/multi) might be needed (since there's no example of multi-methods prior to this yet...)

ronaldxs:
Move to "Class and Instance Methods" section after Classes/self.

PR discussion: https://github.com/perl6/doc/pull/1495#issuecomment-326747280

zakame:
@ronaldxs I have the change to add header (before `Methods` and after `self`,) is that fine?  If so, I can commit it directly.

ronaldxs:
planning to close soon if no objection ...

PR #1504, opened 2017-09-02
Further improve docs for methods + self

zakame:
- Make discussion on class/instance methods into its own Section
- Add example code for `self` especially to illustrate its behavior with
  different invocants

Thanks to @ronaldxs for further comments :+1:

References #1121 (hopefully fixes too :joy_cat:)


zakame:
Also looking for more examples for `self` 👍 Suggestions welcome!

zakame:
Thanks again! 🎉

PR #1506, opened 2017-09-02
Modify Proxy Example To More Closely Match Text

jstuder-gh:
The accompanying text seems to indicate that the modification of the
stored value would occur at retrieval.

Text-'To create a container that returns twice of what was stored in it'

The example showcases the capabilities of Proxy in the same way, but
hopefully is a bit more clear.


AlexDaniel:
Thank you! 🍕

PR #1509, opened 2017-09-02
better example in Iterator, closes #1508

### LINKED ISSUE #1508: better example in Iterator useful
```
class SkippingArray is Array {
    method iterator {
        class :: does Iterator {
            has $.index is rw = 0;
            has $.array is required;
            method pull-one {
                $.index++ while !$.array.AT-POS($.index).defined && $.array.elems > $.index;
                $.array.elems > $.index ?? $.array.AT-POS($.index++) !! IterationEnd
            }
        }.new(array => self)
    }
}

my @a := SkippingArray.new;

@a.append: 1, Any, 3, Int, 5, Mu, Mu, 8;

for @a -> $a, $b {
    say [$a, $b];
};

```

gfldex:


gfldex:
fixed as per jnths suggestion: https://github.com/perl6/doc/issues/1508#issuecomment-326774764

0racle:
The paragraph above the SkippingArray example seems to be referencing the old example...

_For example, the following `Seq` will not cause the `die` to be called..._

No such `die` exists in the new example.

Issue #1082, opened 2016-12-25
Docs on “my” should mention dynamic variables


briandfoy:
The [documentation for my](https://docs.perl6.org/syntax/my) starts with:

> Declaring a variable with my gives it lexical scope.

But this isn't true in the sense that people expect since it does something different with dynamic variables (those with the * twigil). [S04](https://design.perl6.org/S04.html#___top) says:

> We use the phrase "lexical scoping" in its industry-standard meaning to indicate those blocks that surround the current textual location.

And later, 

> Except for such formal parameter declarations, all lexically scoped declarations are visible from the point of declaration to the end of the enclosing block. Period. Lexicals may not "leak" from a block to any other external scope 

But, that's not what `my` does with dynamic variables, say, as in Moritz's [Testing say](https://perlgeek.de/blog-en/perl-6) article) where he temporarily replaces `$*OUT`'s value for a different scope that doesn't enclose it: 

	my $output = do {
		my $*OUT = OutputCapture.new;
		doublespeak(42);
		$*OUT.captured;
	};

There's a cryptic note at the end of the my docs that seems to contradict the rest of the documentation so far:

> To make new-location() print nowhere, make $location a dynamic variable using the * twigil.

That is, if `$location` were `$*location`, it's not a lexical variable anymore despite the `my`.

The example for the [* Twigil](https://docs.perl6.org/language/variables#The_*_Twigil) shows the same thing. A dynamic variable declared with `my` isn't lexical in the sense that S04 defines or that people will expect.

jonathanstowe:
I'm not sure, the effect of  a "my dynamic variable" is lexical, The difference with a dynamic variable is that you get a run-time failure if you try to refer to one if it isn't defined, rather than compile time.

```perl6
sub foo() {
    say $*FOO || "boop";
}

foo();

my $*FOO = "blar";

foo();

```

The difference is that unless it has been introduced globally say with ```PROCESS::<$FOO>``` you will need to introduce ir with ```my````  or get the compile time failure.

I think the difficulty is that dynamic variables are a different category of thing that may have a lexical scope, not just a normal variable.

jnthn:
The `*` twigil means "can be resolved dynamically". The `my` is talking about the storage, and `my $*a` is indeed stored just the same as `my $a`. Note that `our $*var` is also completely valid with the usual meaning of `our` applying:

```
> sub foo() { say $*a }
sub foo () { #`(Sub|91195816) ... }
[...]

PR #1522, opened 2017-09-03
“my” docs mentioning dynamic variables

rafaelschipiura:
Explains the relationship between lexical/package and normal/dynamic scope a bit more.

Asked in  #1082

### DIFF
--- a/doc/Language/variables.pod6
+++ b/doc/Language/variables.pod6
@@ -228,7 +228,7 @@ boolean context before using it for anything else:
 
     =begin code
     sub foo() {
-        $*FOO // 'foo';u can send a pull request.
+        $*FOO // 'foo';
     }
 
     say foo; # OUTPUT: «foo␤»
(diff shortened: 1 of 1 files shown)


doc/Language/variables.pod6
```diff
@@ -491,7 +492,7 @@ new-location; # OUTPUT: «outside␤»
 =end code
 
 To make C<new-location()> print C<nowhere>, make C<$location> a dynamic
-variable using L<the * twigil|#The_*_Twigil>.
+variable using L<the * twigil|#The_*_Twigil>. This twigil makes the compiler look up the symbol in the calling scope instead of the outer scope which is the case otherwise.
```
moritz: The compiler always looks in the current scope first. Looking into outer/caller's scopes is always a fallback. This description makes it sound as if only the caller's scope is searched.

zoffixznet:
👍 Thanks!

Issue #1212, opened 2017-02-21
"it would be more useful to link to the non-raw pages on github ..."


raiph:
**TL;DR:**

> moritz: "it would be more useful to link to the non-raw pages on github ..."

> paultcochrane: "To be honest, I think so too. ..."

Aiui it's an easy change; shall we do it?

----

When fixing typos and the like in other projects shared on github I've generally been able to just click a pencil icon, make an edit, and click the save button and then my changes will appear live in a short while (or, if I don't have commit rights, I click another few buttons to create a PR). This encourages me to make edits.

In an old closed issue the last two comments were Moritz's ["it would be more useful to link to the non-raw pages on github"](https://github.com/perl6/doc/issues/53#issuecomment-76188572) and Paul's agreement: "To be honest, I think so too."

There are a variety of problems of course. Paul mentions one. Other factors to consider include that it might be considered better to push folk to read the contributing doc before they do a PR.

Based on my current understanding that this is an easy-to-apply change, I agree with Moritz and Paul that, all things considered:

> it would be more useful to link to the non-raw pages on github

coke:
To my way of thinking, if someone has commit bits, it is not that unreasonable for them to use a local checkout to fix the issue, or to be able to search the repo to find the bug.

IMO, this would be nice to have, but it's a lot of work for little payoff.

raiph:
> a lot of work

I had thought this was a one line change. If it's more complicated than just altering the footer link as suggested by Moritz and Paul then I think I've misunderstood what they were suggesting and we can just close this issue. Feel free to do so.

> for little payoff

You may be underestimating the value of making things easy for newbies.

I've fixed stuff from typos to writing the original of [a doc page](https://docs.perl6.org/language/performance) by sticking to use of the Github UI. Perhaps I'm unique but I instead think there's a class of newbies who are willing to contribute to Perl 6 docs but prefer to use the web UI and avoid learning the git CLI.

> if someone has commit bits

Having or not having a commit bit made zero difference to my process.

If you prefer, just assume this ticket is about someone who does *not* have a commit bit. And if your answer to that is that everyone gets a commit bit, then please remember that giving someone a commit bit doesn't actually change what that person knows.

jbolden1517:
[...]

Issue #53, opened 2015-02-25
Add Getting Started section to the README


mkrull:
As someone new to Perl 6 I found some documentation missing and in the processes of getting familiar with certain undocumented or poorly documented packages it seems natural for me to contribute docs. Especially from the unspoiled 'what would a new user need' perspective.

But even though I am not new to programming, not new to documenting code or writing docs in general and not even new to writing POD I am having a hard time to figure out where to actually start contributing.

I am pretty sure a quick getting started guide could help a great deal to get drive-by contributions to the docs.

pmichaud:
How should the getting started guide proposed here differ from http://perl6.org/getting-started/ ?

Or, put another way, is http://perl6.org/getting-started/ a good start on this, or is it missing something fundamental (and if so, what)?

mkrull:
I was thinking more of a getting started guide for this repo. A quick pointer to how documentation works to make it easier to contribute pieces of documentation quickly and still correctly.

uvtc:
The readme here already has a "Help Wanted!" section. In that section should be a link to the new CONTRIBUTING.md doc. That CONTRIBUTING.md doc is still quite new I think --- maybe it should contain additional info such as how often the docs are pulled from this repo, processed, and uploaded to doc.perl6.org.

paultcochrane:
Any help is greatly appreciated.  Thanks for your interest!

Moritz has been the driving force behind the documentation effort, so it's probably a good idea to ask him for tips about where to begin. In general, you can simply ping someone on the #perl6 channel on freenode.net and you'll get pointed in the right direction.

Things which occur to me right now which could be done include:
  - clone the repo and grep/ack for TODO items in the POD; the POD is under the lib directory (and subdirs).  If you find something which you think you could document, write it up and add it as a PR.  Make sure to also cross-check with the perl6/specs repo to see if there is some information about that topic there already which could be massaged for good use in per6/doc.
  - simply read the docs. :-) If you find anything which you don't understand, or doesn't seem clear, or could be added, or typos, or whatever, make the relevant change(s) and submit a PR.  Enough good PRs and you'll get a commit bit and you can really let loose ;-)
  - check the current docs for examples and anywhere you think that a concept would be better explained through an example, add the code and submit a PR
  - to build the documentation locally you'll need to install Pod::To::HTML via panda and then run the htmlify.p6 script in the perl6/doc base directory.
  - some internal links don't always work, or might not be linked to the correct document.  Again, correct the POD and submit a PR.
[...]

PR #1533, opened 2017-09-04
Link to non-raw GitHub page, with edit link

zakame:
Let the pod6 docs link to their non-raw GitHub pages, so visitors can
see and maybe help out on improving some more.  And for convenience,
also provide an edit link.

There's probably some more improvement to do on the footer as well; will
further work on separate commits/PRs.

### DIFF
--- a/lib/Pod/Htmlify.pm6
+++ b/lib/Pod/Htmlify.pm6
@@ -93,14 +93,18 @@ sub footer-html($pod-path) is export {
     my $footer = slurp 'template/footer.html';
     $footer.subst-mutate(/DATETIME/, ~DateTime.now.utc.truncated-to('seconds'));
     my $pod-url;
+    my $edit-url;
     my $gh-link = q[<a href='https://github.com/perl6/doc'>perl6/doc on GitHub</a>];
     if $pod-path eq "unknown" {
         $pod-url = "the sources at $gh-link";
+        $edit-url = "";
     }
     else {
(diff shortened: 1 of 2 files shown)


rafaelschipiura:
There's still the issue that Github can't render Pod6: https://github.com/github/markup/issues/907

zakame:
@rafaelschipiura ouch, that's a bit annoying, but not too bad that we can't view the raw pod6.  I probably should have made it clear in the original PR text that I mean changing the *link* to a page with the GitHub amenities, rather than to a direct raw text of the pod6.

OTOH, how can I help getting Pod6 rendering to happen?

AlexDaniel:
This ticket is related to https://github.com/perl6/doc/issues/1212.

For github pod parsing see https://github.com/perl6/doc/issues/167.

rafaelschipiura:
@zakame I do think it would be better to link to the decorated Github page instead of the raw text, so I agree with merging this. But it doesn't fix #1212, since that is requesting a link to the composed page.

zakame:
>But it doesn't fix #1212, since that is requesting a link to the composed page.

@rafaelschipiura that's probably splitting hairs since #1212 doesn't explicitly say about linking to a fully composed/rendered page, only that we can get to a page where we can proceed to edit (see further context in https://github.com/perl6/doc/issues/53#issuecomment-76188572.)  Either way though, I think the core problem (to fix a link to point somewhere we can proceed to do further action later) is hopefully solved by this PR.

zakame:
Thanks @AlexDaniel ! :tada:

Issue #1536, opened 2017-09-05
Mention importance of closing handles in 5to6 open docs


jnthn:
Here:

https://docs.perl6.org/language/5to6-perlfunc#open

It would be good to highlight that handles to not get closed at scope exit. It's mentioned in the open docs already, but it would be good to call it out as a significant difference between 5 and 6.

rafaelschipiura:
Solved by https://github.com/perl6/doc/pull/1537

PR #1537, opened 2017-09-05
Mention importance of closing handles in 5to6

rafaelschipiura:
#1536


zoffixznet:
Personally, I wouldn't mention the `LEAVE` phaser here. If someone's reading 5to6 docs, it'd be unfamiliar territory, and `LEAVE` has caveats, such as still being called even when a routine is called with wrong args.

Perhaps the prose here should just refer to https://docs.perl6.org/type/IO::Handle#routine_close for various ways of closing handles?

rafaelschipiura:
Just putting C<close> is enough?

zoffixznet:
> Just putting C is enough?

`C<>` just styles it as code. You need `L<close>` to make it a link to close routine.

zoffixznet:
👍 Thanks!

PR #1540, opened 2017-09-06
Does this resolve #1538?

### LINKED ISSUE #1538: In a then callback, the Promise is always true
Here:

https://github.com/perl6/doc/blob/master/doc/Type/IO/Socket/Async.pod6#L33

There's:

    await IO::Socket::Async.connect('localhost', 3333).then( -> $promise {
        if $promise.status {

But `status` will be non-zero when the `Promise` is either kept or broken - which it must be if the `then` has been called. Note that checking if the status is `Kept` is also a bad thing to advise, as it means the exception will be silently thrown away. Just removing the `if` would do it, then the exception will be raised on accessing `.result`.

molecules:
Does simply removing this `if` statement resolve #1538?

### DIFF
--- a/doc/Type/IO/Socket/Async.pod6
+++ b/doc/Type/IO/Socket/Async.pod6
@@ -34,7 +34,7 @@ await IO::Socket::Async.connect('localhost', 3333).then( -> $promise {
         .print('Hello, Perl 6');
         react {
             whenever .Supply() -> $v {
-               $v.say;
+                $v.say;
                 done;
             }
         }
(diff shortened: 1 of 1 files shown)


doc/Type/IO/Socket/Async.pod6
```diff
@@ -30,17 +30,15 @@ And a client that connects to it, and prints out what the server answers:
...
-            react {
-                whenever .Supply() -> $v {
-                    $v.say;
-                    done;
-                }
+    given $promise.result {
+        .print('Hello, Perl 6');
+        react {
+            whenever .Supply() -> $v {
+               $v.say;
```
jnthn: Seems there's a 3-space indent here instead of 4.

jnthn:
It would resolve it, just one tiny detail.

PR #1547, opened 2017-09-10
Better example for IO::Socket::Async, closes #1546

### LINKED ISSUE #1546: better example for IO::Socket::Async

gfldex:


doc/Type/IO/Socket/Async.pod6
```diff
@@ -12,13 +12,26 @@ server and the client side.
 Here is the equivalent example to that in L<IO::Socket::INET> of a simple
 echo server that listens on port 3333:
```
zakame: needs updating here too 👍

doc/Type/IO/Socket/Async.pod6
```diff
@@ -12,13 +12,26 @@ server and the client side.
...
-            await $conn.write: $buf
+=begin code :skip-test<IO>
+loop {
+    react {
+        whenever IO::Socket::Async.listen('0.0.0.0', 8080) -> $conn {
+            react { 
+                whenever $conn.Supply.lines {
+                    say .Str;
+
+                    { await $conn.print(.Str ~ "\n") } for
```
AlexDaniel: Hm, I wonder if this is actually needed, or do we only have to await for the last print?

doc/Type/IO/Socket/Async.pod6
```diff
@@ -12,13 +12,26 @@ server and the client side.
...
+                    say .Str;
+
+                    { await $conn.print(.Str ~ "\n") } for 
+                    "HTTP/1.1 200 OK",
+                    "Content-Type: text/html; charset=UTF-8",
+                    "Content-Encoding: UTF-8",
+                    "",
+                    "<html><body>Hello World!</body></html>";
+
+                    $conn.close;
```
AlexDaniel: Generally, having the react/whenever block suggests that you might have several whenevers, but in this particular case we are showing a less than awesome precedent of ``await``-ing inside the whenever and therefore not letting other whenevers do the job (there are none in this case, but as an example it is less than awesome).

gfldex:
Example updated to be async proper and with better error handling.

zoffixznet:
Is this now mergeable?

jonathanstowe:
Looks reasonable to me

zoffixznet:
👍 Thanks!

PR #1562, opened 2017-09-17
It seems we have sub form of `Str`

W4anD0eR96:
It seems we have sub form of `Str`
`say Str(Mu.new()).perl; # OUTPUT: "Mu<140381106429184>"`
`say Str(42).perl; # OUTPUT: "42"`

But `say Str(Mu); # OUTPUT: (Str(Mu))`, which perform different from `Mu.Str`

### DIFF
--- a/doc/Type/Mu.pod6
+++ b/doc/Type/Mu.pod6
@@ -104,7 +104,8 @@ public attributes:
 
 =head2 method Str
 
-    multi method Str(--> Str)
+    multi sub    Str(Mu --> Str)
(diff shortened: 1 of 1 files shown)


jnthn:
They aren't actually subs. An expression `TypeName(value)` parses as a coercion. It's a parse-time distinction based on `TypeName` being known type. It then will end up calling the `Str` method on `value`. So documenting it as a `Str` sub existing is incorrect; this is a general mechanism that applies to all types.

PR #1571, opened 2017-09-21
Document :Perl5 adverb

robertlemmen:


doc/Language/regexes.pod6
```diff
@@ -1273,6 +1273,18 @@ Grammars provide an easy way to override what C«<.ws>» matches:
 When parsing file formats where some whitespace (for example, vertical
 whitespace) is significant, it's advisable to override C<ws>.
 
+=head3 Perl 5 compatibility adverb
+
+The B<C<:Perl5>> or B<C<:P5>> adverb switch the Regex parsing and matching
+back to the way Perl 5 regexes did behave:
```
MasterDuke17: "back to the way" and "did behave" seem to suggest Perl 5 is old/doesn't exist anymore. I'd suggest changing to something like "...switch the Regex parsing and matching to the way Perl 5 regexes behave:"

PR #1574, opened 2017-09-23
note multi hiding via non-multi in lexical scope

### LINKED ISSUE #1570: improve MMD
it would have helped me to know that the dispatcher will ignore candidates in any outer scope if there is a non-multi in the given scope

example in https://gist.github.com/ed62484b162e7d7d06376e2f42deefb2

gfldex:
closes #1570


zoffixznet:
👍 Thanks!

Issue #1584, opened 2017-09-29
[RFC] Rename "Texas" term to "ASCII"


zoffixznet:
Started from [RT#132179](https://rt.perl.org/Ticket/Display.html?id=132179)

Along with avoiding whatever negative feelings anyone might have about the [stereotype](http://allthetropes.wikia.com/wiki/Everything_Is_Big_in_Texas) this joke term is based on, calling ASCII-alternative ops is self-explanatory for anyone unfamiliar with our jargon.

So, seems like a two-birds-with-one-commit type of change. Anyone 👍 ?

jonathanstowe:
I've never been that keen on the term to be honest.
As an aside I discovered that none of the developers on my team that use Linux had set up the compose key so I reverted a bunch of '⇒' in some Scala code.

AlexDaniel:
Yeah, that sounds about right. I don't think we win anything by using the word “texas”. As a non-American, I wasn't even aware of the whole “Everything Is Big in Texas” thing, and perl 6 is probably the only reason I know about it…

> As an aside I discovered that none of the developers on my team that use Linux had set up the compose key so I reverted a bunch of '⇒' in some Scala code.

Let's stay on topic. This has nothing to do with this issue.

Altai-man:
I cannot really say whether someone can be offended with a word "Texas" or not(on this level it can be like "What if someone has a negative feelings toward word "ASCII", we should avoid that too"), but it is certainly more technically correct term, indeed.

AlexDaniel:
This is a RFC, so please leave it open for just a little bit in case someone wants to leave a comment. That said, we no longer have “Texas” mentions after #1585.

By the way, can we somehow redirect ``unicode_texas`` to ``unicode_ascii``? Because otherwise there are some broken links now, like [here](https://www.learningperl6.com/2016/11/22/quick-tip-7-texas-and-unicode-things/).

zoffixznet:
> By the way, can we somehow redirect unicode_texas to unicode_ascii?

[Done](https://github.com/perl6/doc/commit/b15d8de5a4bd4b90dbdd5c8cb99b0aea0115fbe7)

tbrowder:
considered satisfied by PR #1585

PR #1585, opened 2017-09-29
changes required by changing 'Texas' to 'ASCII'

tbrowder:
I'll merge after (1) a good doc build and (2) no one objects.


AlexDaniel:
> after no one objects

We will not be waiting for every person in existence to express their agreement. If someone feels very strong about reverting this, then they can revert. Please feel free to merge.

tbrowder:
addresses #1584

PR #1590, opened 2017-10-07
Can find implementation of Slang in rakudo

W4anD0eR96:
https://github.com/rakudo/rakudo/blob/8ff76b5962ed34caa7b66b3f95644410fdb9b17e/src/core/Slang.pm


Tyil:
Thanks

PR #1598, opened 2017-10-10
File test adverb forms with links to method forms

molecules:
Also intended to fix broken link to https://docs.perl6.org/type/IO::Path#File_test_operators from https://docs.perl6.org/language/5to6-perlfunc#Filetests. Much was copied from https://docs.perl6.org/type/IO::Path#File_test_operators.


Tyil:
Thanks!

Issue #1588, opened 2017-10-05
Mention ｢my %h{Int}｣ syntax when talking about non-string keys


AlexDaniel:
See https://docs.perl6.org/type/Hash#Non-string_keys

There's a little useful hint on https://docs.perl6.org/type/Hash#method_keyof:
```
my %h2{Str} = 'oranges' => 7;            # (keys must be of type Str)
```
But it's hard to find it from “Non-string keys” section.

BenGoldberg1:
It might also be worthwhile to also mention hashes with no restriction on keys `my %h3{Any};`

W4anD0eR96:
I think this mention some of it: https://docs.perl6.org/type/Hash#Object_hashes_and_type_constraints
while it gets a wrong hierarchy in the doc -- it is below `Looping over hash keys and values'

W4anD0eR96:
@AlexDaniel #1602 may fix it, and this issue can be closed

AlexDaniel:
I think [it](https://docs.perl6.org/type/Hash#Non-string_keys_(object_hash)) is a bit convoluted and hard to read, but indeed this issue is resolved. I'll leave this open for just a little bit in case someone has any thoughts on how to reword it so that it reads better, but otherwise this can be closed.

W4anD0eR96:
@AlexDaniel Yes, maybe it can be rewritten using more simple examples.

PR #1602, opened 2017-10-11
Try to fix intro of non-string keys of Hash

W4anD0eR96:
Mentioned in #1588 .
We do have a intro of non-string keys at https://docs.perl6.org/type/Hash#index-entry-object_hash
but obviously its position is wrong.
So I try to rearrange the text and make a PR, but its readability should be improved.
Not a native English speaker, I ask for your help to make the doc read more smoothly.

### DIFF
--- a/doc/Type/Hash.pod6
+++ b/doc/Type/Hash.pod6
@@ -113,6 +113,8 @@ You can assign to multiple keys at the same time with a slice.
 
 =head2 Non-string keys
 
+X<|non-string keys>X<|:{}>
+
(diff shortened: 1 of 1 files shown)


W4anD0eR96:
@Tyil I think now the PR is ready to merge, plz check when you are free

I simply cut the section **object hash** and paste to the new position
Then I rewrite the first paragraph to make the doc read more smoothly
Because I find somebody call it object hash( #1604 ) while somebody call it non-string hash ( #1588 ), I set the section title as non-string keys (object hash) and make both of them searchable.

Tyil:
Thanks!

PR #1612, opened 2017-10-17
Add hook to "Parameterized Roles"

W4anD0eR96:
I think it is worth to be searchable


zoffixznet:
I sent you an invite to perl6 GitHub org. You can accept it here: https://github.com/perl6/

That gives you commit bit to most of our repos, so you can merge this PR yourself and commit directly in the future.

PR #1614, opened 2017-10-18
Remove link to outdated META6 spec

Tyil:
After discussion in #perl6, the consensus seems to be that the link referenced in the docs does not resemble the META6 spec. To avoid confusion, the link should be removed.

### DIFF
--- a/doc/Language/modules.pod6
+++ b/doc/Language/modules.pod6
@@ -562,7 +562,7 @@ lib
     The L<Test::META module | https://github.com/jonathanstowe/Test-META/>
     can help you check the correctness of the META.info file.
 
-    There are more fields described in the L<META specification |
+    There are more fields described in the L<META design documents |
     https://design.perl6.org/S22.html#META6.json>, but not all of these are
     implemented by existing package managers. Hence you should stick to the
     fields described in the above example block to ensure compatability with
--- a/doc/Language/modules.pod6
+++ b/doc/Language/modules.pod6
@@ -565,7 +565,7 @@ lib
     There are more fields described in the L<META design documents |
     https://design.perl6.org/S22.html#META6.json>, but not all of these are
     implemented by existing package managers. Hence you should stick to the
-    fields described in the above example block to ensure compatability with
+    fields described in the above example block to ensure compatibility with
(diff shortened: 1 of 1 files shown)


samcv:
More accurately you should probably mention that the design docs may better describe the different fields, but you should stick to the fields that are shown in S22 for maximum compatibility with package managers, since not all things speced in S22 are currently implemented.

There is further explaination in S22 that can be helpful, as long as it isn't assumed all fields described in S22 are implemented.

Tyil:
That could be a cleaner way to go about this.

I was reading up on the spec to make `App::Cpan6` adhere to this spec. For this, I could make use of the `meta-version` referenced in S22. A value of `1` can indicate the current set of implemented fields in `zef`, with `2` indicating the "full spec" as shown in S22.

I'm not sure if this would be agreeable with the other devs, though. I might ask for information in `#perl6-toolchain`.

tbrowder:
On Wed, Oct 18, 2017 at 16:19 Patrick Spek <notifications@github.com> wrote:
...

> I'm not sure if this would be agreeable with the other devs, though. I
> might ask for information in #perl6-toolchain.
>

I’m in favor of anything that tightens the requirements for a well-behaved
module.

niner:
S22 is currently a mixture. Apparently despite my efforts, parts of it are still outdated. Other parts are simply not yet implemented by our toolchain, but are very much our plan for it. I very much like samcv's suggestion.

Tyil:
There seems to be little interest to improve the underlying issue I had here, so I'll just go with the suggestion from @samcv.

doc/Language/modules.pod6
```diff
@@ -562,6 +562,12 @@ lib
     The L<Test::META module | https://github.com/jonathanstowe/Test-META/>
     can help you check the correctness of the META.info file.
 
+    There are more fields described in the L<META specification |
+    https://design.perl6.org/S22.html#META6.json>, but not all of these are
+    implemented by existing package managers. Hence you should stick to the
+    fields described in the above example block to ensure compatability with
```
jonathanstowe: compatibility

PR #1619, opened 2017-10-24
Document List cmp infix and sorting by multiple values (via List)

jstuder-gh:
This is a powerful way of sorting, but it's not currently apparent that
it can be done this way in the documentation.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -125,6 +125,9 @@ E: ilmari.vacklin@reaktor.fi
 N: isBEKaml
 E: nastavs@gmail.com
 
+N: Jeremy Studer
+E: studerj1.mail@gmail.com
+
 N: Jimmy Zhuo
 E: zhuomingliang@yahoo.com.cn
 
--- a/doc/Type/List.pod6
+++ b/doc/Type/List.pod6
@@ -690,7 +690,7 @@ Defined as:
     multi method sort(List:D:      --> Seq:D)
     multi method sort(List:D: &custom-routine-to-use  --> Seq:D)
 
-Sorts the list, smallest element first. By default C<< infix:<cmp> >>
+Sorts the list, smallest element first. By default L«C<< infix:<cmp> >>|/routine/cmp»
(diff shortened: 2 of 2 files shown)


jstuder-gh:
I'm hoping to get some feedback on the style of this. For instance the %()'s in the sort example. Is it better to use that or the curly braces? I like the % prefixed parens as it can't be confused for a Block, but I don't see it as much in Perl6 examples. TIMTOWTDI is great, but I suppose it could be confusing or off-putting to the Perl newcomer reading the docs if the syntax used in the examples is inconsistent.

Also, regarding the .<KEY> notation, is this intentional? I've actually never seen a hash value accessed in this way but was experimenting. I was wondering if since the dot prefix allows for method calls on the topic variable, if it would work for hash access too and it turns out it does. I kind of like this notation as it is easier to type but I don't want to start any bad habits (or start documenting bad habits).

Tyil:
I personally prefer the `%()` style as well, it makes intention clear and is less confusing, as `{}` are also used for a few other things. I don't think there's anything wrong using it in the docs either.

On the `.` notation, I am not sure. A more experienced Perl monk might be able to give a good answer here.

bduggan:
> Also, regarding the . notation, is this intentional? I've actually never seen a hash value accessed in this way but was experimenting. I was wondering if since the dot prefix allows for method calls on the topic variable, if it would work for hash access too and it turns out it does. I kind of like this notation as it is easier to type but I don't want to start any bad habits (or start documenting bad habits).

There are a lot of occurrences of this syntax in the test suite (I found many by grepping for `.<`), so it is okay.  For instance:  [here](https://github.com/perl6/roast/blob/master/S04-statements/given.t#L223) and [here](https://github.com/perl6/roast/blob/master/S29-conversions/hash.t#L25).

rafaelschipiura:
Yes, indexing into the topic with a dot is fine.

jstuder-gh:
Thanks everyone for the clarification.

Issue #1627, opened 2017-10-25
Intro example in IO::Socket::Async doesn't work


zoffixznet:
The example of server/client at the top of https://docs.perl6.org/type/IO::Socket::Async doesn't work.

Playing with the code a bit, looks like the `.print` in the client needs a newline added to the end, 'cause server is waiting for lines. Also, the server sends several lines, but the client calls `done` after just receiving the first line.

P.S.: not sure why the `try` is needed in the server. Also, why not just do one print with a heredoc.

P.P.S: perhaps it makes more sense to make *one* program a user can start with both server and client. Just shove the server into a separate Promise and wait a bit for it to start up before connecting with a client or maybe even add some Promise that signals when the server started up and when the client can start connecting.

PR #1628, opened 2017-10-25
Fixed server and client examples for IO::Socket::Async

7ojo:
I'm little bit familiar about using this IO::Socket::Async so I thought to try to resolve this #1627 issue.

I also read about suggestion to put server+client examples to one, but I think is not so good idea for newcomers. Maybe.


zoffixznet:
👍 Thanks!

PR #1633, opened 2017-10-26
Added Dockerfile and instructions to CONTRIBUTING.md

7ojo:
Added Docker way to get tools up and running for contributing


zoffixznet:
@7ojo FWIW, you should have a commit bit that lets you commit to most of our repos directly.

7ojo:
@zoffixznet Thanks for comment :) Yep, I have the commit bit, but was too shy to commit directly on these. Until next time then!

PR #1635, opened 2017-10-28
Add example "Add actions directly", review needed

W4anD0eR96:
Since this tutorial shows actions in grammar only by associating a action object, I add an example to introduce write actions directly.

As I am a nonnative, maybe there is some improper expression in the text.

Thanks for your review!

### DIFF
--- a/doc/Language/grammar_tutorial.pod6
+++ b/doc/Language/grammar_tutorial.pod6
@@ -567,6 +567,41 @@ Here's the final code:
     }
     =end code
 
+=head2 Add actions directly
+
+Above we see how to associate grammars with a action objects and perform
+actions on the match object. However, when we want to deal with the match
+object, that isn't the only way. See the example below:
+
+    =begin code
+    grammar G {
+      rule TOP { <function-define> }
+      rule function-define {
+        'sub' <identifier> {
+          say "func " ~ $<identifier>.made;
+          make $<identifier>.made;
+        } '(' <parameter> ')' '{' '}' {
+          say "end " ~ $/.made;
+        }
+      }
+      token identifier { \w+ { make ~$/; } }
+      token parameter { \w+ { say "param " ~ $/; } }
+    }
+
+    G.parse('sub f ( a ) { }');
+    # OUTPUT: «func f␤param a␤end f␤»
+    =end code
+
+This example is a reduced portion of a parser. Let's focus more on the feature
+it shows.
+
+First, we can add actions inside the grammar ifself, and such actions are
+performed once the match arrive them. Note that action object's method will be
+performed after the whole regex item matched, while actions inside the grammar
+are more flexible. Second, it shows what C<make> really does, which is no more
+than a sugar of C<$/.made = ...>. And this trick introduces a way how we pass
+message within a regex item.
+
 Hopefully this has helped introduce you to the grammars in Perl 6 and shown you
 how grammars and grammar action classes work together. For more information,
 check out the more advanced L<Perl Grammar
--- a/doc/Language/grammar_tutorial.pod6
+++ b/doc/Language/grammar_tutorial.pod6
@@ -569,7 +569,7 @@ Here's the final code:
 
 =head2 Add actions directly
 
-Above we see how to associate grammars with a action objects and perform
+Above we see how to associate grammars with action objects and perform
 actions on the match object. However, when we want to deal with the match
 object, that isn't the only way. See the example below:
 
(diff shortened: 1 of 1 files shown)


doc/Language/grammar_tutorial.pod6
```diff
@@ -567,6 +567,41 @@ Here's the final code:
     }
     =end code
 
+=head2 Add actions directly
+
+Above we see how to associate grammars with a action objects and perform
```
Tyil: You can leave off the "a", so it becomes "with action objects".

doc/Language/grammar_tutorial.pod6
```diff
@@ -567,6 +567,41 @@ Here's the final code:
...
+
+    G.parse('sub f ( a ) { }');
+    # OUTPUT: «func f␤param a␤end f␤»
+    =end code
+
+This example is a reduced portion of a parser. Let's focus more on the feature
+it shows.
+
+First, we can add actions inside the grammar ifself, and such actions are
+performed once the match arrive them. Note that action object's method will be
```
Tyil: "once the match arrive them", might want to rewrite it as "arrives at them". I'm not sure if a match "arrives" at a certain action either.

doc/Language/grammar_tutorial.pod6
```diff
@@ -567,6 +567,41 @@ Here's the final code:
...
+performed after the whole regex item matched, while actions inside the grammar
+are more flexible. Second, it shows what C<make> really does, which is no more
+than a sugar of C<$/.made = ...>. And this trick introduces a way how we pass
+message within a regex item.
```
Tyil: "how we pass messages" or "how we pass a message".

doc/Language/grammar_tutorial.pod6
W4anD0eR96:
```
grammar G {
  rule TOP {
    | a { say "42" } b
    | a { say "12" } c
  }
}

G.parse("a c");
```
I want to show that both `say "42"` and `say "12"` are reached, as the example above.
How can I express this?

doc/Language/grammar_tutorial.pod6
```diff
@@ -567,6 +567,40 @@ Here's the final code:
...
+
+Above we see how to associate grammars with action objects and perform
+actions on the match object. However, when we want to deal with the match
+object, that isn't the only way. See the example below:
+
+    =begin code
+    grammar G {
+      rule TOP { <function-define> }
+      rule function-define {
+        'sub' <identifier> {
```
moritz: I really don't like the indentation of the code block here, as it suggests a relation to the previous regex atoms that isn't there.

Most regexes/grammars I've seen or written put the curly braces of multi-line blocks on separate lines, which makes them stand out as they should.

For single-line code blocks, it's ok to put the curlies on the same line as the code block itself:

```
    { say "end " ~ $/.made; }
````

File: doc/Language/grammar_tutorial.pod6 [hunk 5]
```diff
@@ -567,6 +567,40 @@ Here's the final code:
...
+
+    G.parse('sub f ( a ) { }');
+    # OUTPUT: «func f␤param a␤end f␤»
+    =end code
+
+This example is a reduced portion of a parser. Let's focus more on the feature
+it shows.
+
+First, we can add actions inside the grammar ifself, and such actions are
+performed once the match arrives at them. Note that action object's method will
```
moritz: Technically it's not the match that arrives at the actions, but rather the control flow of the regex. (The match is related to the position in the string, not directly to the position within the regex).

doc/Language/grammar_tutorial.pod6
W4anD0eR96: I try to rewrite it following the style at https://github.com/moritz/perl-6-regex-code/blob/master/chapter-09-grammars/08-symbol-table.p6

BTW, looking forward to your new book!

doc/Language/grammar_tutorial.pod6
W4anD0eR96: That is what I want to express

doc/Language/grammar_tutorial.pod6
geekosaur: "once their [corresponding] patterns are matched"?

doc/Language/grammar_tutorial.pod6
```diff
@@ -567,6 +567,42 @@ Here's the final code:
...
+    }
+
+    G.parse('sub f ( a ) { }');
+    # OUTPUT: «func f␤param a␤end f␤»
+    =end code
+
+This example is a reduced portion of a parser. Let's focus more on the feature
+it shows.
+
+First, we can add actions inside the grammar ifself, and such actions are
```
Tyil: "ifself" → "itself"

PR #1636, opened 2017-10-29
Modify Label Doc and Examples

jstuder-gh:
The largest of the changes here is the documenting of the redo method on
Label. Also made some modifications for the existing label examples.

Changed the example for the next method so that it's output differs from
that of the last example. With the change, the numbers 5 through 10 are
printed, which makes it more clear that the loop has iterated a total of
ten times but only reached the print statement during 6 of those.

Also, added the output to the example, as I believe that is clearer and
allows better visualization of what occurred. It also is more consistent with
the example-then-output style used throughout the docs.

These example changes are largely subjective, but I think they are for the allow
for better clarity / consistency.

### DIFF
--- a/doc/Type/Label.pod6
+++ b/doc/Type/Label.pod6
@@ -41,6 +41,27 @@ Begin the next iteration of the loop associated with the label.
 
     # OUTPUT: «5 6 7 8 9 10 »
 
+=head2 method redo
+
+Defined as:
+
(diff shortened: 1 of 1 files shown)


zoffixznet:
👍 Thanks!

FWIW, you have a commit bit to most of our repos and can just commit directly.

PR #1637, opened 2017-10-29
Modify Dockerfile and Add Docker-related Make targets

jstuder-gh:
Made a number of modifications to the Dockerfile in the repo:

* Minimized the number of cache layers in final image (by reducing the
  number of Run commands)
  * This way Docker cannot reuse cache layers containing outdated
    package versions.
* Purge build dependencies and apt cache data.
  * These files do not appear in final image due to minimal intermediate
    cache layers.
  * The resulting image is over 100 MB smaller as a result.
* Update NodeJS to the latest stable release via the
  [n](https://github.com/tj/n) script.
* Added Test::META module and GNU aspell (used in tests).

Also added a .dockerignore file, so that files within the repo do not
get used in the build context.

Added Make targets that make the Docker containers a bit easier to work with.
Add info to CONTRIBUTING indicating how to use these targets.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+./*
+.sass-cache/
+.git/
+.gitignore
+.travis.yml
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -264,7 +264,7 @@ render the HTML documentation
 Now point your web browser to http://localhost:3000 to view the
 documentation.
 
-#### Using Docker
+#### Using Docker to build and view the documentation
 
 You can skip all the above and just build and view documentation with these simple commands (if you have docker already installed):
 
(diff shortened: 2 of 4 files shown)


jstuder-gh:
I didn't want to commit something like this to the repo without either some sort of review or the approval of other members and/or owners.

This greatly modifies the existing Dockerfile (what and why in description) in the repo and introduces a number of new targets to the Makefile.

moritz:
This looks reasonable to me, though I'm by no means a Docker expert.

jstuder-gh:
I'm no expert either. It's functional, but there may be better ways of doing things.

More than anything though, I just don't want to be committing things other members don't want or don't agree with. If the general consensus is people are okay with this, I will gladly commit it. Thanks for your feedback.

moritz:
@7ojo you committed the original Dockerfile. Any objections?

7ojo:
No objections! Seems fine to me 👍  Using Makefile simplifies things nicely if you're not so familiar with docker

PR #1648, opened 2017-11-03
Add Metamodel::EnumHOW page

### LINKED ISSUE #1642: Deadlink for Metamodel::EnumHOW class in the the Meta-Object Protocol Page
On the Meta-Object Protocol Page (https://docs.perl6.org/language/typesystem#enum and scroll down to Introspection), the link for Metamodel::EnumHOW doesn't exist. I think that either we remove the link or add documentation for Metamodel::EnumHOW.

bduggan:
This adds a page for MetaModel::EnumHOW.  Looking for feedback -- I took these from the source, though I don't think they are part of the spec, so not sure if it's appropriate to list them here?  (Or maybe there should be a note about them being rakudo-specific?)

Resolves #1642

### DIFF
--- a/doc/Type/Metamodel/EnumHOW.pod6
+++ b/doc/Type/Metamodel/EnumHOW.pod6
@@ -1,0 +1,74 @@
+=begin pod
+
+=TITLE class Metamodel::EnumHOW
+
+=SUBTITLE Metaobject representing a Perl 6 enum.
+
+    class Metamodel::EnumHOW
+        does Metamodel::Naming
+        does Metamodel::Documenting
+        does Metamodel::Stashing
+        does Metamodel::AttributeContainer
+        does Metamodel::MethodContainer
+        does Metamodel::MultiMethodContainer
+        does Metamodel::RoleContainer
+        does Metamodel::BaseType
+        does Metamodel::MROBasedMethodDispatch
+        does Metamodel::MROBasedTypeChecking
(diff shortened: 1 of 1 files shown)


jnthn:
I think it's OK to document the MOP, in general. It may not all be spec, but in reality it's one of the most stable parts of the Rakudo implementation. Future Perl 6 implementations will need to implement a MOP too, and whether we document the Rakudo one or not, people will discover and use it anyway, which will probably make it a de facto standard. That usage will in turn probably drive us to spec increasing amounts of it, to make sure we aren't breaking the ecosystem.

AlexDaniel:
All these underscores tho… :S

AlexDaniel:
Anyway, we already have [ClassHOW](https://docs.perl6.org/type/Metamodel::ClassHOW) so adding EnumHOW won't hurt (although both pages should probably have a warning of some sort).

geekosaur:
(Summary of discussion on IRC)

I would be very cautious about adding things to the spec that could tie you to a particular implementation. Documenting such is fine, but should include warnings that they are not part of the spec and are use-at-your-own-risk.

I mean, there are already modules in the ecosystem that dive into nqp; does that now become a spec candidate because of the potential breakage?

At some point in the future, a limited public API/spec for the MOP might be an idea; I personally would want to see a bit more experience and preferably other implementations before committing to one, though.

PR #1652, opened 2017-11-04
Document Regex Interpolation, jnthn++

W4anD0eR96:
I think it's worthy to be documented, because people will looking for a way to using patten given at the runtime.

But I'm not sure if this's the only way to interpolate patten(string) into regex.

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -1255,17 +1255,45 @@ list of predefined subrules is listed in
 L<S05-regex|https://design.perl6.org/S05.html#Predefined_Subrules> of design
 documents.
 
-=head1 X«Regex Interpolation|Regex Interpolation; <$pattern>»
+=head1 X<Regex Interpolation|regex, Regex Interpolation>
 
 If you want to build a regex using pattern given at the runtime, regex
 interpolation is what you looking for.
 
-    sub runtime-regex(Str $pattern --> Regex) { /<$pattern>/ }
+There are two ways you can interpolate a string into regex as a pattern,
+which are quite similar to interpolate it into a double-quote string.
+That is using C«<$pattern>» or C«<{$pattern.method}>».
 
-    my $regex = runtime-regex(get);  # INPUT : «\d+␤»
-    my $text = get;                  # INPUT : «42␤»
+    my $text = 'camelia';
+    my $pattern = '\w+';
+    my $pattern_ = 'ailemac';
+    say $text ~~ / <$pattern> /;          # OUTPUT: «｢camelia｣␤»
+    say $text ~~ / <{$pattern}> /;        # OUTPUT: «｢camelia｣␤»
+    say $text ~~ / <{$pattern_.flip}> /;  # OUTPUT: «｢camelia｣␤»
+    # say $text ~~ / <$pattern_.flip> /;  !!Compile Error!!
 
-    say $text ~~ $regex;             # OUTPUT: «｢42｣␤»
+Note that this syntax causes I<implicit EVAL>, that is a known trap.
+
+When your pattern is a lexical string, there are two more syntax helping you
+interpolate it.
+
+    my $text = 'camelia';
+    my $string = 'camelia';
+    my $string_ = 'ailemac';
+    say $text ~~ / $string /;                 # OUTPUT: «｢camelia｣␤»
+    say $text ~~ / $($string) /;              # OUTPUT: «｢camelia｣␤»
+    say $text ~~ / $($string_.flip) /;        # OUTPUT: «｢camelia｣␤»
+    say $text ~~ / $string_.flip /;           # OUTPUT: «Nil␤»
+    say 'ailemacxflip' ~~ / $string_.flip /;  # OUTPUT: «｢ailemacxflip｣␤»
+
+Because it interpolates a string lexically, it doesn't EVAL the string
+implicitly.
+
+    my $text = '42';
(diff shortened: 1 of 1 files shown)


doc/Language/regexes.pod6
```diff
@@ -1255,6 +1255,18 @@ list of predefined subrules is listed in
 L<S05-regex|https://design.perl6.org/S05.html#Predefined_Subrules> of design
 documents.
 
+=head1 X«Regex Interpolation|Regex Interpolation; <$patten>»
```
AlexDaniel: patten

doc/Language/regexes.pod6
```diff
@@ -1255,6 +1255,18 @@ list of predefined subrules is listed in
...
+
+If you want to build a regex using pattern given at the runtime, regex
+interpolation is what you looking for.
+
+    sub runtime-regex(Str $patten --> Regex) { /<$patten>/ }
```
AlexDaniel: patten

doc/Language/regexes.pod6
```diff
@@ -1255,6 +1255,18 @@ list of predefined subrules is listed in
...
+
+    my $regex = runtime-regex(get);  # INPUT  : «\d+␤»
+    my $text = get;                  # INPUT  : «42␤»
+
+    say $text ~~ $regex;             # OUTPUT : «｢42｣␤»
```
AlexDaniel: Perhaps don't use a space before the column (“OUTPUT:” instead of “OUTPUT :”). I think the idea was that this should be parseable so that we can use it to check the output, so better keep it consistent with other things. Also, normally you can copy-paste the output from ``camelia`` bot directly.

doc/Language/regexes.pod6
W4anD0eR96: not sure if you want to convert 'regex' into 'patten', plz check it.

AlexDaniel: well, isn't it supposed to be “patteRn”? :)

W4anD0eR96: see it, sorry :P

AlexDaniel:
IMO it's better to have fixed variables instead of calling ``get``. That would make it a bit easier to read.

Also, maybe a link to [this trap](https://docs.perl6.org/language/traps#%3C{$x}%3E_vs_$($x):_Implicit_EVAL) is appropriate?

Also, what's the difference between ``<$x>`` and ``<{$x}>``?

W4anD0eR96:
@AlexDaniel , the trap doc should be modified, because `$($x)` doesn't implicitly EVAL.

I will works on it now.

doc/Language/regexes.pod6
```diff
@@ -1255,6 +1255,47 @@ list of predefined subrules is listed in
 L<S05-regex|https://design.perl6.org/S05.html#Predefined_Subrules> of design
 documents.
 
+=head1 X<Regex Interpolation|regex, Regex Interpolation>
+
+If you want to build a regex using pattern given at the runtime, regex
```
MasterDuke17: Some grammar fixes and this sentence become:

If you want to build a regex using a pattern given at runtime, regex
interpolation is what you are looking for.

doc/Language/regexes.pod6
```diff
@@ -1255,6 +1255,47 @@ list of predefined subrules is listed in
...
+interpolation is what you looking for.
+
+There are two ways you can interpolate a string into regex as a pattern,
```
MasterDuke17: And this one becomes (also, there are three ways):

There are three ways you can interpolate a string into a regex as a pattern.That is using C«$pattern», C«<$pattern>», or C«<{$pattern}>».

doc/Language/regexes.pod6
```diff
@@ -1255,6 +1255,47 @@ list of predefined subrules is listed in
...
+    my $pattern_ = 'ailemac';
+    say $text ~~ / <$pattern> /;          # OUTPUT: «｢camelia｣␤»
+    say $text ~~ / <{$pattern}> /;        # OUTPUT: «｢camelia｣␤»
+    say $text ~~ / <{$pattern_.flip}> /;  # OUTPUT: «｢camelia｣␤»
+    # say $text ~~ / <$pattern_.flip> /;  !!Compile Error!!
+
+Note that this syntax causes L«implicit EVAL|
+/language/traps#<{$x}>_vs_$($x):_Implicit_EVAL», that is a known trap.
+
+When your pattern is a lexical string, there are two more syntax helping you
```
MasterDuke17: Oh, I see, this is where you talk about just interpolating as a string. I would also mention that is the string to interpolate is statically typed as a Str or str (i.e., my Str $string = 'camelia'), it's much faster.

W4anD0eR96:
```
my Str $pattern = '\d+';

say "42" ~~ /<$pattern>/;
``

Code above works too. So I think statically type `$pattern` as `Str` shows nothing.

Anyway, your other reviews are worthy. Thank you!

MasterDuke17: Yeah, typing it doesn't change the behavior if you're using <$pattern>, but if you're just using $pattern it's *much* faster (~10x).

W4anD0eR96: got it. sorry for misunderstanding

MasterDuke17: No worries, I'm glad we're adding some documentation about regex interpolation, it's a common question.

PR #1661, opened 2017-11-05
Add vim modeline at end of every file without one

jstuder-gh:
The settings in the modeline do the following:

* Highlight this file as Perl6
* Expand tabs into 4 spaces

This is helpful for vim users for highlighting an atypical file format
and creates formatting consistency with regards to tabs and spacing.

### DIFF
--- a/doc/404.pod6
+++ b/doc/404.pod6
@@ -12,3 +12,5 @@
 </p>
 
 =end Html
+
+# vim: expandtab softtabstop=4 shiftwidth=4 ft=perl6
--- a/doc/HomePage.pod6
+++ b/doc/HomePage.pod6
@@ -59,3 +59,5 @@ Documentation for the different but related <a href="https://www.perl.org/">Perl
 can be found on the <a href="http://perldoc.perl.org/">Perl 5 documentation website</a>.
 </p>
 =end Html
+
+# vim: expandtab softtabstop=4 shiftwidth=4 ft=perl6
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -1755,4 +1755,4 @@ be better than losing the information about a real need.
 
 =end comments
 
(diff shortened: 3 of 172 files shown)


jstuder-gh:
Some of the pod files in the repositories had vim modelines appended at the end while many did not. This commit will add them to all pod files. 

To those unfamiliar with vim, these modelines will tell vim that within the buffer for this file, use these settings. They will also override the defaults set in a user's vimrc and help to enforce a certain standard. The settings specified here:

* Expand tabs into 4 spaces.
* Shift width (using < or >) also shifts 4 spaces.
* This file is interpreted as Perl 6 (for syntax highlighting).

The modelines themselves are unobtrusive but I want to make sure that other users agree with these settings across before merging them in. I did not see any anything regarding tabs/spaces in the styleguide.

AlexDaniel:
What's the difference between ``shiftwidth=4`` and ``softtabstop=4``?

jstuder-gh:
softtabstop indicates to vim that the tab character be expanded to n spaces.

shiftwidth indicates that when shifting (via < or > keys in vim) shift text n characters.  

The two in conjunction create a certain uniformity. In some of the cases where there was previously shiftwidth without softtabstop, for instance, shifting moved 4 chars while entering new tabs would expand to 8 chars by default.

jstuder-gh:
It's also worth noting that these settings are per-buffer settings and will not affect other buffers in your vim session. So expandtab and softtabstop will not expand your tabs to spaces when editing the Makefile, for instance.

Also, expandtab does not convert existing tab characters, just newly inserted ones.

AlexDaniel:
👍 , but I'll let some vim user to merge this.

Altai-man:
Keep in mind, that all "atypical" files have the same extension and it should take just one or two lines in local users' vim config to do the same.

This PR is already merged, but I can also recommend you to improve your editor's config if you're working with an editor that needs it(vim/emacs/whatever). Thanks.

PR #1667, opened 2017-11-09
Add edit ico

### LINKED ISSUE #1622: Edit this page near the top.
From splitting #1212 

@AlexDaniel thinks it would improve things a lot to also have [an edit this page link] near the top of the page.

W4anD0eR96:
A try to fix #1622 .

### DIFF
--- a/html/images/pencil.svg
+++ b/html/images/pencil.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="14px" height="16px" viewBox="0 0 14 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 40.3 (33839) - http://www.bohemiancoding.com/sketch -->
+    <title>pencil</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Octicons" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="pencil" fill="#000000">
+            <path d="M0,12 L0,15 L3,15 L11,7 L8,4 L0,12 L0,12 Z M3,14 L1,14 L1,12 L2,12 L2,13 L3,13 L3,14 L3,14 Z M13.3,4.7 L12,6 L9,3 L10.3,1.7 C10.69,1.31 11.32,1.31 11.71,1.7 L13.3,3.29 C13.69,3.68 13.69,4.31 13.3,4.7 L13.3,4.7 Z" id="Shape"></path>
+        </g>
+    </g>
+</svg>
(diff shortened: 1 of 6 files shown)


W4anD0eR96:
Is this satisfies issue #1622 ?

AlexDaniel:
To your last comment: yes, if this is merged we can close #1622.

Building the docs currently to play with it. Right away I can tell that “Edit this page” text should be added somewhere. I probably don't mind it being alt text or something, but it has to be somewhere.

AlexDaniel:
Looks a little bit differently here:
2017-11-09-113221_1329x431_scrot

W4anD0eR96:
@AlexDaniel yes. do we have available js function to show message onmouseover?

I find GitHub pencil ico implements it by add `aria-label` attr inside `button`.
However, it's a Bootstrap feature.  I think we use jQuery and thus need to write a custom function.

AlexDaniel:
Well, you can use ``title="Edit this page"`` and it will work, but I don't know what would be the best practice.

W4anD0eR96:
@AlexDaniel  Is it?

I just check appearances of the ico locally in, all `/language/*`, all `/program/*`, all `/type/*` except most of `X::...`. Their looks are consistent. Could you check the html code relevant?

AlexDaniel:
@W4anD0eR96 what I meant by that is that the button looks differently on your screenshot and mine.

W4anD0eR96:
Yes, chrome and Safari look like mine, Firefox doesn't.

W4anD0eR96:
Firefox:




Chrome and Safari:




@AlexDaniel Now the ico is quite familiar, and "Edit this page" message shows correctly.

However, I notice that our sass style render a litter different between browsers. If we will re-design the doc site, this should be taken into consideration.

I'll merge this PR once build passed.

html/images/pencil.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="14px" height="16px" viewBox="0 0 14 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 40.3 (33839) - http://www.bohemiancoding.com/sketch -->
```
AlexDaniel: I don't think we should pollute the repo with ads.

html/images/pencil.svg
```diff
@@ -0,0 +1,12 @@
...
+    <title>pencil</title>
+    <desc>Created with Sketch.</desc>
```
AlexDaniel: Same here.

html/images/pencil.svg
AlexDaniel: (also, not only the repo, but every page we serve)

W4anD0eR96: agree

Issue #1679, opened 2017-11-18
Fix "Truncating slices" on Subscripts page


jstuder-gh:
Currently this section states that:

* All ranges (including non-lazy ones) will truncate the output/ omit undefined elements.

It might be a good idea to split the section into two related but logically separate topics.

1. If the element at index x does not exist then the result will be a Nil element (for Positionals).
    * This behavior can be modified using the :v adverb. 
2. Using lazy Iterables as subscripts yields a different behavior. Only returns defined elements (given that the elements of the Iterable are not reified past the index of the first undefined element)
    * Behind the scenes the elements of the lazy subscript are being reified up until the point that the element in the Positional is undefined.
    * Continued use of lazy Lists will begin to return Nil elements as new elements in the lazy List are reified.
    * Include justification for these behaviors. Sometimes returning undefined elements is desirable.
        * DESIRABLE: assignment to their containers. eg @a[^10] = 1..10;
        * NOT DESIRABLE: Running wild (if the following were to not truncate/end reification). eg. @a[1 .. *]

Related:
https://irclog.perlgeek.de/perl6/2017-11-18#i_15466865

jstuder-gh:
Hmm... "WRITE COMMIT MESSAGE LATER". Be assured that I am not going to commit upstream without amending that first. :ghost:

jstuder-gh:
This issue has been resolved and can now be closed.

See:
[PR #1681](https://github.com/perl6/doc/pull/1681)
Commit 5893559b9f7e7c159927f8251df41e0d3cf3474a

PR #1681, opened 2017-11-19
Rewrite, remove inaccurate Range truncation info

jstuder-gh:
Previously, this section stated that:

* All ranges (including non-lazy ones) will truncate the output/ omit
  undefined elements.

Rewrote the section to remove that inaccuracy and address the following:

* If the element at index [x] does not exist then the result will be an
  undefined element (when subscripting Positionals).
     * This behavior can be modified using the :v adverb.
* Using lazy Iterables as subscripts yields a different behavior. Only
  returns defined elements (most of the time; see bullets below).
    * Behind the scenes the elements of the lazy subscript are being
      reified up until the point that the element in the Positional
      is undefined.
    * Continued use of lazy Lists will begin to return undefined
      elements as new elements in the lazy List are reified.
* Include justification for these behaviors. Sometimes returning
  undefined elements is desirable.
    * DESIRABLE:     assignment. eg @a[^10] = 1..10;
    * NOT DESIRABLE: Out-of-memory due to uncontrolled indexing.

This removes some information about lack of protection against runaway
slices and reification of subscript elements in instances where the
undefined values are not being returned from the collection (eg,
instances where the same index is being used over and over again). This
seemed a little out of place in this section and probably fits better
elsewhere. Perhaps it would be appropriate in the Traps section?

SEE ALSO:
    [Issue #1679](https://github.com/perl6/doc/issues/1679)
    https://irclog.perlgeek.de/perl6/2017-11-18#i_15466865

### DIFF
--- a/doc/Language/subscripts.pod6
+++ b/doc/Language/subscripts.pod6
@@ -262,8 +262,8 @@ nonexistent elements):
     =end code
 
 This behavior, while at first glance may seem unintuitive, is desirable in
-instances when you want to assign a value at an index in which a defined
-value does not currently exist.
+instances when you want to assign a value at an index in which a value does
+not currently exist.
 
     =begin code
     my  @letters;
@@ -273,8 +273,8 @@ value does not currently exist.
     say @letters; # OUTPUT: «[a b c d e f g h i j]␤»
     =end code
 
-If you want the resulting slice to only include defined elements, you can
-silently skip the non-defined elements using the L<#:v> adverb.
+If you want the resulting slice to only include existing elements, you can
+silently skip the non-existent elements using the L<#:v> adverb.
 
     =begin code
     my  @letters = <a b c d e f>;
@@ -293,32 +293,6 @@ subscript, the resulting slice will be truncated.
 This behavior exists as a precaution to prevent runaway generation of massive,
 potentially infinite C<Lists> and the out-of-memory issues that occur as a
 result.
-
-Behind the scenes, the implementation is first eagerly accessing the collection
-via subscripts that are already L<reified|/language/glossary#index-entry-Reify>.
-After reaching an index of the lazy subscript-producer that has not yet been
-reified, the indices will be produced one-after-another up until the point that
-the subscript no longer yields a defined value from the collection. In other
-words, once an undefined value in the collection is encountered, the process of
-subscript reification is halted and the resulting slice is returned.
-
-If you are subscripting via a lazy C<List>, be aware that each subsequent use
-reifies and caches an additional element in the lazy C<List>, introducing
-undefined elements into the resulting slice. Some lazy C<Iterables> (such as
-C<List>) will cache each newly produced element while others (such as C<Seq>)
-will discard them. Being aware of how each C<Iterable> deals with it's lazily
-produced elements will help you to avoid unexpected results.
-
-    =begin code :preamble<my @letters = <a b c d e f>;>
-    my @i = 3..*;
-    .say for @letters[@i] xx 3;
-
-    #`( OUTPUT:
-        (d e f)
-        (d e f (Any))
-        (d e f (Any) (Any))
-    )
-    =end code
 
 =head2 Zen slices
 
--- a/doc/Language/subscripts.pod6
+++ b/doc/Language/subscripts.pod6
@@ -208,7 +208,7 @@ Be aware that slices are controlled by the I<type> of what is passed to
 (L<one dimension of|#Multiple dimensions>) the subscript, not its length.
 In particular the I<type> can be any of the following:
 
-=item a Range or a lazy Iterable, that L<truncates|#Truncating slices> in [ ]
+=item a lazy Iterable, that L<truncates|#Truncating slices> in [ ]
 
 =item '*' (whatever-star), that returns the full slice (as if all keys/indices were specified)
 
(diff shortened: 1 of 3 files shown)


jstuder-gh:
Looking for review from more senior members as this is a substantial rewrite of this section. Thanks.

AlexDaniel:
```perl6
my @letters = <a b c d e f>; @letters[4]:delete; say @letters[3..*]
```
```
(d (Any) f)
```
```perl6
my @letters = <a b c d e f>; @letters[4]:delete; say @letters[lazy 3..*]
```
```
(d)
```
I always thought that this was a bug… I guess this PR documents the current behavior, right?

doc/Language/subscripts.pod6
```diff
@@ -253,68 +251,74 @@ dimensions>) the subscript is preserved across the slice operation
...
-A similar thing is done for lazy sequences, but it is often impossible to
-determine whether the sequence is infinite.  Just as often, the first part
-of the sequence is already known, and it would be silly to pretend we
-did not know it.  As a stopgap measure to prevent runaway generation of huge
-lists, a lazy subscript will not truncate as long as it does not have to
-lazily generate values, but once it starts generating values lazily, it
-will stop if it generates a value that points to a nonexistent index.
+This behavior, while at first glance may seem unintuitive, is desirable in
+instances when you want to assign a value at an index in which a defined
+value does not currently exist.
```
zoffixznet: *[request for change]*

> in which a defined value

"defined" should be dropped here, since [undefined](https://docs.perl6.org/routine/defined) are perfectly acceptable elements as far as existence goes.

doc/Language/subscripts.pod6
```diff
@@ -253,68 +251,74 @@ dimensions>) the subscript is preserved across the slice operation
...
+    say @letters; # OUTPUT: «[a b c d e f g h i j]␤»
     =end code
 
-The runaway protection is not perfect.  The indices are eagerly evaluated,
-with the only stop condition being truncation.  This is to provide
-mostly consistent results when there is self-reference/mutation inside
-the indices.  As such, the following will most likely hang until all
-memory has been consumed:
+If you want the resulting slice to only include defined elements, you can
+silently skip the non-defined elements using the L<#:v> adverb.
```
zoffixznet: *[request for change]*

Same as above, change "defined" to "existent" ("extant"?), since it doesn't care about undefinedness:

```
$ perl6 -e '(Any, Int)[1..*]:v.say'
((Int))
```

doc/Language/subscripts.pod6
```diff
@@ -253,68 +251,74 @@ dimensions>) the subscript is preserved across the slice operation
...
     =end code
 
-If you I<don't> want to specify your slice as a range/sequence but still want
-to silently skip nonexistent elements, you can use the L<#:v> adverb.
+This behavior exists as a precaution to prevent runaway generation of massive,
+potentially infinite C<Lists> and the out-of-memory issues that occur as a
+result.
+
+Behind the scenes, the implementation is first eagerly accessing the collection
+via subscripts that are already L<reified|/language/glossary#index-entry-Reify>.
```
zoffixznet: *[feedback]*

IMO this paragraph goes into too much detail:

* Most users shouldn't care about these details and shouldn't need to read all this extra info, at least in this document
* This describes how *Rakudo* implemented this feature, while the docs site tries to be implementation-agnostic and only describe what [the spec](https://github.com/perl6/roast/) specifies the behaviour to be. I don't think the spec describes such minute details as when to reify things in this particular scenario.

doc/Language/subscripts.pod6
```diff
@@ -253,68 +251,74 @@ dimensions>) the subscript is preserved across the slice operation
...
-    @letters[0 xx *];
+    =begin code
+    my  @letters = <a b c d e f>;
+    say @letters[3..7]:v;  # OUTPUT: «(d e f)␤»
     =end code
 
-So, to safely use lazy indices, they should be one-shot things which
-are guaranteed to overrun the array.  The following alternate formulation
-will produce a fully lazy result (but will not truncate):
+The behavior when indexing a collection via lazy subscripts is different than
```
zoffixznet: *[request for change]*

Would probably be helpful to link `"lazy"` to https://docs.perl6.org/routine/is-lazy or something similar that explains what the term means.

doc/Language/subscripts.pod6
```diff
@@ -253,68 +251,74 @@ dimensions>) the subscript is preserved across the slice operation
...
+Behind the scenes, the implementation is first eagerly accessing the collection
+via subscripts that are already L<reified|/language/glossary#index-entry-Reify>.
+After reaching an index of the lazy subscript-producer that has not yet been
+reified, the indices will be produced one-after-another up until the point that
+the subscript no longer yields a defined value from the collection. In other
+words, once an undefined value in the collection is encountered, the process of
+subscript reification is halted and the resulting slice is returned.
+
+If you are subscripting via a lazy C<List>, be aware that each subsequent use
+reifies and caches an additional element in the lazy C<List>, introducing
```
zoffixznet: *[feedback]*

This problem too feels implementation-specific IMO and probably shouldn't be listed here. I opened a ticket for this problem in Rakudo: https://github.com/rakudo/rakudo/issues/1320 

It likely won't get fixed and in such a case a better place for this might be in [Reification](https://docs.perl6.org/language/glossary#index-entry-Reify) docs or in [Traps](https://docs.perl6.org/language/traps).

jstuder-gh:
@zoffixznet,
I've made the changes to the section that you requested in your review. I added the "drift" explanation to Traps and a link to a "Lazy List" section to explain the context of laziness a bit.

I made a few changes to the "Lazy List" section I linked to which is outside the scope of this PR, but I can remove that if you prefer the old language better. I added an index to the section as an explanation of the general concept of laziness. Let me know if you want that removed.

I had been holding off on updating this branch as I wasn't sure how to document lazy indexing and how it reacts to holes, but I suppose we can update the section in the future to address holes whichever way we decide to proceed on that issue. 

I rebased the branch so that its base is a more recent commit. I believe this is ready to be merged in now, assuming the recent revisions have addressed your concerns. Let me know what you think.

zoffixznet:
👍 Thanks!

PR #1686, opened 2017-11-24
be precise when to early is to early for constant

gfldex:


doc/Language/terms.pod6
```diff
@@ -190,7 +190,8 @@ Variables are discussed in the L<variable language docs |/language/variables>.
 X<|constant (Terms)>
 Constants are declared with C<constant>, followed by an
 L<identifier|/language/syntax#Identifiers> and do not require a sigil. The RHS
-is evaluated at compile time, what may be too early to make sense.
+is evaluated at compile time. For precompiled compilation units, that may be
```
fluca1978: While I agree that we need a rephrasing here, I'm not sure the proposed statement makes things clearer.
I would suggest a simpler statement as 'Be aware that compile time evaluation could be too early to make sense' and I would improvde the example with respect to the POW2 constant in order to emphasize this.
It could also be a good idea to separate the "too early" statement placing it at the bottom of the example.

zoffixznet:
*ping* any updates for the reviewer's comments?

fluca1978:
Sorry, but since the author did not provide any feedback on review, I'm going to close this PR and I've committed a possible rephrasing giving credits to the PR author.
I've reviewed this PR as soon as I found it, I would kindly ask the author to open a new one for a better wording.
Thanks for your effort.

PR #1699, opened 2017-12-06
Describe how to look up class members

lefth:
Lookup within packages and the global namespace is described. Classes should be covered as well.

### DIFF
--- a/doc/Language/packages.pod6
+++ b/doc/Language/packages.pod6
@@ -179,6 +179,14 @@ from a type name by use of the C<::> postfix:
 =for code :skip-test
 MyType::<$foo>
 
+=head2 Class member lookup
+
+Class attributes and methods are stored in the class meta object and can be looked up
+through by the L<lookup|/method/lookup> method.
+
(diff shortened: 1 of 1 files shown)


doc/Language/packages.pod6
```diff
@@ -179,6 +179,14 @@ from a type name by use of the C<::> postfix:
 =for code :skip-test
 MyType::<$foo>
 
+=head2 Class member lookup
+
+Class attributes and methods are stored in the class meta object and can be looked up
```
fluca1978: Isnt' 'lookup' searching only for methods (not members)?

zoffixznet:
👍 Thanks!

doc/Language/packages.pod6
zoffixznet: It'll show autogenerated accessor methods for public attributes. clarified that in https://github.com/perl6/doc/commit/4f53943fa08c1a1ef73587b23e8cbde5783f71e1

PR #1700, opened 2017-12-06
Fix readme

fluca1978:
Provides a little more information about nodejs and zef option while building.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ disk speed and Rakudo version, it might take a while.
 
 Install dependencies by running the following in the checkout directory:
 
-    zef --dep-sonly install .
+    zef --deps-only install .
 
 If you use [`rakudobrew`](https://github.com/tadzik/rakudobrew), also run the
 following, to update the shims for installed executables:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ disk speed and Rakudo version, it might take a while.
 
 Install dependencies by running the following in the checkout directory:
 
-    zef --depsonly install .
+    zef --dep-sonly install .
```
JJ: Wouldn't that be --deps-only?

README.md
```diff
@@ -60,6 +60,14 @@ To build the documentation web pages, simply run
...
     $ make html
 
+Please note that you will need to have [nodejs](https://nodejs.org)
+installed to produce HTML content with the above command.
+Moreover, depending on your system and installation, you could need
+to either *alias* or *link* the [nodejs](https://nodejs.org) executable
+so that you can run a command `node`. For instance:
+
+    $ ln `which nodejs` /usr/bin/node
+
```
JJ: I don't think such an explanation is needed. It's probably enough to say something along the lines of "You will need a "node" executable in your path. People can then figure out the rest.

fluca1978: Does it sounds better now?

README.md
```diff
@@ -60,6 +60,10 @@ To build the documentation web pages, simply run
 
     $ make html
 
+Please note that you will need to have [nodejs](https://nodejs.org)
+installed to produce HTML content with the above command, in particular
+a `node` executable must to be in your `PATH`.
```
JJ: "should be" sounds better. Thanks for the changes.

fluca1978: Fixed.

Please note I removed also a reference to panda (different commit, so cherry-picking is available).

PR #1702, opened 2017-12-06
Fix slice table.

fluca1978:
The L<> directive does not work within a table, while it seems to work
within an iterable list.

I'm not sure this is the best solution, in particular with regard to its verbosity in explaining the slicing.

### DIFF
--- a/doc/Language/subscripts.pod6
+++ b/doc/Language/subscripts.pod6
@@ -205,26 +205,21 @@ For associative slices, the angle-brackets form often comes in handy:
     say %color{*};                 # OUTPUT: «(red yellow green)␤»
 
 Be aware that slices are controlled by the I<type> of what is passed to
-(L<one dimension of|#Multiple dimensions>) the subscript, not its length:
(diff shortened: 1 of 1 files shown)


AlexDaniel:
@tbrowder by any chance can we have the POD bug fixed instead of merging this?

AlexDaniel:
Although, now I see that this PR gives more readable descriptions, so maybe we should merge it regardless.

fluca1978:
Allow me to explain better my intent: it is clear that there must be a more general fix of the L<> directive within the table one, but in the meantime we should at least deploy a readable content on both the web and the pod6.
That is why I converted the table to a list, as a workaround waiting for a better fix in the documentation directives handling (that I'm not able to face).
If people prefer to fix the L<> directive instead of merging this I'm fine, and that is of course the best solution. Anyway, if that requires too much time/effort, I would choose the workaround (we can also place a todo to revert the table back once the L<> works).

fluca1978:
Any more comments on this?

zoffixznet:
👍 Thanks!

PR #1704, opened 2017-12-07
Fix reference to Perl 5.

fluca1978:
It seems the comment working on regexp was referencing the similar feature of
Perl 5, not Perl 6.

Reading as it is now cannot make me  sure if this was referencing \x regexp in Perl 5 and comments or it was bad spelled out to explain that Perl 6 comments work _also_ in regexp.


AlexDaniel:
FWIW https://github.com/perl6/doc/issues/1705

JJ:
Right-on. But a focused effort should be needed...

PR #1708, opened 2017-12-11
Fix link to Promise.

fluca1978:
Very simple link from Proc::Async to Promise while documenting the start method.


moritz:
Thanks @fluca1978

PR #1735, opened 2018-01-20
first call to show_captures has wrong number of arguments in section Backtracking

MARTIMM:
In the first example where `show_captures` is defined, the call to it has also `$string` as its second argument(line 1937). this is not used in the sub signature.

Furthermore a small typo on line 1917 (matched instead of match).

Regards,
Marcel

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -1914,7 +1914,7 @@ my $string = 'PostgreSQL is an SQL database!';
 say $string ~~ /(.+)(SQL) (.+) $1/; # OUTPUT: ｢PostgreSQL is an SQL｣
 =end code
 
-What happens in the above example is that the string has to be match against the
+What happens in the above example is that the string has to be matched against the
 second occurrence of the word I<SQL>, eating all characters before and leaving out
 the rest.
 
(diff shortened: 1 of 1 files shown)


zoffixznet:
FWIW: You have the commit bit and can commit directly, without creating separate branches or PRs.

MARTIMM:
Hi Zoffix,
I am aware of this but I am still not confident enough to change things without forcing others to look into it before it is merged. When I will change document typos, I will commit directly though.

doc/Language/regexes.pod6
```diff
@@ -1934,7 +1934,7 @@ sub show_captures( Match $m ){
     say $result_split;
 }
 
-$string ~~ /(.+)(SQL) (.+) $1 (.+) { show_captures( $/, $string );  }/;
+$string ~~ /(.+)(SQL) (.+) $1 (.+) { show_captures( $/ );  }/;
```
W4anD0eR96: Two advice here.

1. use :preamble<my $string;> after `=begin code` to fit xtest.

2. rename `show_captures` to `show-captures` to fit naming convention.

Anyway, you can be brave to commit directly. We just trust our contributors and it's ok there is something not that good. Your caution is appreciated for me though.

zoffixznet:
👍 Thanks!

PR #1740, opened 2018-01-22
1=>1 vs. a=>1 as named arguments

gfldex:
this was asked about at https://irclog.perlgeek.de/perl6/2018-01-22#i_15723547

### DIFF
--- a/doc/Type/Signature.pod6
+++ b/doc/Type/Signature.pod6
@@ -568,7 +568,9 @@ as the parameters were declared.
     pos(4, 5);                          # OUTPUT: «x=4 y=5»
 
 In the case of named arguments and parameters, only the name is used for
-mapping arguments to parameters
+mapping arguments to parameters. IF a fat arrow us used to construct a 
+L<Pair|/type/Pair> only those with valid identifiers as keys are recognized
+as named arguments.
 
 =for code :allow<L>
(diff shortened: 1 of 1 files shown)


zoffixznet:
👍 Thanks!

doc/Type/Signature.pod6
```diff
@@ -568,7 +568,9 @@ as the parameters were declared.
     pos(4, 5);                          # OUTPUT: «x=4 y=5»
 
 In the case of named arguments and parameters, only the name is used for
-mapping arguments to parameters
+mapping arguments to parameters. IF a fat arrow us used to construct a
```
tommystanton: Ah, I noticed a small typo. :sweat_smile: This:

```text
IF a fat arrow us used
```

...should probably be:

```text
If a fat arrow is used
```

zoffixznet: Good catch! I invited you to **perl6** GitHub org. You can accept the invite [here](https://github.com/perl6/).

Once you do, you should be able to [edit the file](https://github.com/perl6/doc/blob/890bfa192679c33a318e80ffaa1d690adc8225c7/doc/Type/Signature.pod6#L569) and commit the typo fix directly.

Thanks.

PR #1762, opened 2018-02-08
add type smiley example

gfldex:


zoffixznet:
Can we make `type smiley` a link to what that is?

It's not easy to search for and users reading this document would know very little about Perl 6.

zoffixznet:
👍 Thanks!

PR #1811, opened 2018-03-04
c function tutorial added

finanalyst:
Curt Tilmes gave a great addition to a question on perl6 users. I thought it illustrated several parts of NativeCall, so I wrote it up as an example for the NativeCall documentation.


AlexDaniel:
I like it, it's awesome. But can we have it in a separate document? Similarly to how we have [Grammars](https://docs.perl6.org/language/grammars) and [Grammar Tutorial](https://docs.perl6.org/language/grammar_tutorial).

AlexDaniel:
Although, you know what… given that it's at the end of the nativecall document, I think we can just take it! Feel free to reorganize though.

PR #1841, opened 2018-03-10
Add a Q and A about Rakudo Star release cycle

### LINKED ISSUE #1830: The release cycle of both rakudo and rakudo-star aren't disclosed
AFAIK,
rakudo-star: released every three months.
rakudo: released every months.

titsuki:
Fix #1830 
I didn't describe the release cycle of the compiler-only version because this one is not a mainstream distribution for many users.
See also: https://github.com/perl6/perl6.org/issues/93

### DIFF
--- a/doc/Language/faq.pod6
+++ b/doc/Language/faq.pod6
@@ -693,6 +693,15 @@ bootstrapping files created by earlier runs of the build process.
 
     (not (not Nil))
 
+X<|Perl 6 Distribution (FAQ)>
+=head1 Perl 6 Distribution
+
+X<|Rakudo Star release cycle (FAQ)>
+=head2 When will the next version of Rakudo Star be released?
+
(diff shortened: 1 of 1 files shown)


nxadm:
I completely disagree with this statement. I fall into at least 3 categories of users where this does not apply and I think there are many people in the same boat:
- I don't want to install and maintain modules I don't use. Not on a server, not on my desktop.
- I want the smallest install size as possible for containers. In my opinion, Docker is pretty much the only sane deployment strategy for Perl 6 at the moment (until things like Fatpack for Perl 6 exist).
- I want precompiled packages. Compiling on target machines (say, production) is a **huge** NO-NO.

This was my trigger for providing pre-compiled Linux packages.

Regards,

C.

titsuki:
@nxadm I understand your opinion and I'm one of the such kind of users of the compiler-only packages.
I just want to make a chance to know the difference between these two in a way that doesn't confuse users.

PR #1842, opened 2018-03-10
Document module packages

cfa:
These commits add a new language documentation page covering _Module Packages_.

RfC: @niner, @ugexe.

CC: @moritz, @zoffixznet.

### DIFF
--- a/doc/Language/module-packages.pod6
+++ b/doc/Language/module-packages.pod6
@@ -12,17 +12,18 @@ focuses on use of the C<module> declarator.
 =head1 What are modules?
 
 Modules, like classes and grammars, are a kind of
-L<package|/language/packages>. Fundamentally, they are metaclasses
-that provide certain capabilities useful for namespacing, versioning,
-delegation and data encapsulation (see also L<class|/syntax/class>
-and L<role|/syntax/role>), and are created with the C<module>
-declarator:
+L<package|/language/packages>. Module objects are instances of the
+C<ModuleHOW> metaclass; this provides certain capabilities useful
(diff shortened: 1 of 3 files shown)


doc/Language/module-packages.pod6
```diff
@@ -0,0 +1,180 @@
...
+
+
+I<N.B.> "Module" is an overloaded term in Perl 6; this document
+focuses on use of the C<module> declarator.
+
+
+=head1 What are modules?
+
+Modules, like classes and grammars, are a kind of
+L<package|/language/packages>. Fundamentally, they are metaclasses
```
niner: Modules are not metaclasses. They are named containers for code and data. The module object itself is just an instance of the ModuleHOW metaclass.

niner:
All in all it reads like a great explanation of what it is and how it works. To become perfect, it would have to answer the question of why. When would one declare a module instead of e.g. a class or instead of just putting code in a file? But that's just a hint on how to improve it further. Certainly nothing that should prevent these comments from getting merged.

doc/Language/module-packages.pod6
cfa: Thanks; I've pushed 71cf557 to clarify this point. Better?

cfa:
@niner please take a look at 71cf557 re: metaclasses.

niner:
Sounds better, yes. Thanks!

cfa:
@niner Grand, thanks for reviewing. Okay for me to merge?

niner:
Sure :)

PR #1848, opened 2018-03-16
Remove leading indentation from (some) pod code blocks.

cfa:
There are numerous instances of `=for` and `=begin` code blocks marking already indented code peppered through the documentation. This indentation is preserved in the rendered output, which is typically unintentional. So, for example:

    =for code
        say 'hello';

is rendered as

        say 'hello';

rather than

    say 'hello';

Although this is a fairly boring commit, it touches a lot of files—hence the PR rather than direct push. Note that If you ignore whitespace (`git diff master -w`), no content has changed. There're also no `test` or `xtest` regresssions.


AlexDaniel:
Thanks!

Issue #551, opened 2016-05-26
description of the META6.json is not complete


titsuki:
See http://docs.perl6.org/language/modules#Distributing_Modules
It lacks "build-depends", "test-depends", and so on.

jonathanstowe:
Yeah, it probably needs a whole section of it's own to properly document the elements.

PR #1859, opened 2018-03-21
Update META6.json documentation to include "api" field

Tyil:
I'm considering to split up this section into a mandatory and optional section, to make it clearer what a user *needs*, and what they can additionally use to achieve certain goals (such as semantic versioning, in this case).

### DIFF
--- a/doc/Language/modules.pod6
+++ b/doc/Language/modules.pod6
@@ -537,6 +537,7 @@ lib
     {
         "perl" : "6.c",
         "name" : "Vortex::TotalPerspective",
+        "api" : "1",
         "version" : "0.0.1",
         "description" : "Wonderful simulation to get some perspective.",
         "authors" : [ "R<Your Name>" ],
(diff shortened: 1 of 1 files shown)


MARTIMM:
I was wondering how the user of a dependency has to select the proper version when there is a version and an api? How should zef process it ?

Tyil:
You can use `Module:ver<1.2.3>:api<1>` to select both. You can also leave off the `:api` if you don't care about that one, just like how you can leave out the `:ver` part.

In the `depends` field, the following are all correct:
- `Module` - Just give me Module, I don't care which version or API
- `Module:ver<1.0.0+>` - Give me Module, at least version 1.0.0
- `Module:api<1>` - Give me Module with API 1, no matter which version implements that API
- `Module:ver<1.0.0+>:api<1>` - Give me Module with at least version 1.0.0, but only if it also implements API 1

MARTIMM:
Thank you for the explanation.
The api field is optional so when I use Module:api<1>, zef has to devise that from its version. Is that correct? Then, when I read the docs about it, it would not give the newest version if The dependency would get version 2.0.0. That is nice! 
Regards,
Marcel

Tyil:
I'm not sure what `zef` will do if you specify a `Module:api<1>` as a dependency, but the dependency has no `api` field in its `META6.json`. Maybe @ugexe can expand on that.

ugexe:
`Foo:api<1>` is the same as `Foo:api<1>:ver<*>`. `Foo` is the same as `Foo:api<*>`

`@meta-data.sort({ Version.new($_<api>) }).sort({ Version.new($_<ver>) })`

If no `api` field is found in the META6 then it will be treated as 0 / * ( just like a `ver` that is declared as `*` ). It will not be guessed from the `ver` field.

JJ:
Related: #551. Maybe close it. I'll have to check.

PR #1881, opened 2018-03-29
return type constraints also work for pointy blocks

gfldex:
also, links are nice


gfldex:
You can use L<Typethatstartswithacapital> but that's just a heuristic in htmlify.p6 and not standard POD6. That's why I avoid that shortcut.

PR #1892, opened 2018-04-02
Fix typo

sfischer13:


doc/Type/Signature.pod6
```diff
@@ -615,7 +615,7 @@ as the parameters were declared.
     pos(4, 5);                          # OUTPUT: «x=4 y=5»
 
 In the case of named arguments and parameters, only the name is used for
-mapping arguments to parameters. IF a fat arrow us used to construct a
+mapping arguments to parameters. IF a fat arrow is used to construct a
```
JJ: Is there some reason why the F in IF is capitalized?

sfischer13: Maybe for emphasis, but it seems to be the only occurrence of ´IF´ in the documentation.

JJ:
Thanks!

PR #1941, opened 2018-04-19
Document NativeCall explicitly-manage subroutine

Kaiepi:
## The problem

NativeCall's`explicitly-manage` subroutine is exported by default but undocumented.

## Solution provided

I added information on it to NativeCall's documentation, as well as a strings section that goes into more detail on how it should be used, where documentation on buffers and blobs could also be placed later on. Please review before merging

### DIFF
--- a/doc/Language/nativecall.pod6
+++ b/doc/Language/nativecall.pod6
@@ -545,7 +545,61 @@ for 1 ..^ $n {
 }
 =end code
 
-=head1 Buffers and Blobs
+=head1 Strings
+
+=head2 Explicit memory management
+
+Let's say there is some C code that caches strings passed, like so:
+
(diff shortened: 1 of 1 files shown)


doc/Language/nativecall.pod6
```diff
@@ -545,6 +545,58 @@ for 1 ..^ $n {
 }
 =end code
 
+=head1 Strings
+
+Let's say there is some C code that caches strings passed, like so:
+
+=begin code :skip-test
```
MasterDuke17: I think you can add `:lang<C>` here and it will syntax highlight it correctly.

Kaiepi: Updated

PR #1977, opened 2018-04-29
m/ / and / / examples added

COMBORICO:
Added examples to help illustrate difference between m/ / and / /.  Zoffix provided the examples.  I provided the learning.  An example for rx/ / is needed.  As a newbie, this info is very helpful and rare.  It is not in Think Perl 6, ttbomk.  And definitely not in Perl6Intro.com.  Needed info or something better.


zoffixznet:
👍 Thanks!

JJ:
Thanks. Do you mean that, additionally, an example for `rx//`is needed?

COMBORICO:
JJ, yes.

JJ:
@COMBORICO please open an issue for that...

PR #1978, opened 2018-04-30
Clarify 'Bar.pm6' vs 'how to use Bar.pm6'

mryan:
## The problem

Reference to Bar.pm6 is arguably ambiguous

## Solution provided

Added three words


W4anD0eR96:
Thanks.

Issue #1984, opened 2018-05-01
Translate README.md to languages


JJ:
## The problem

We have right now a [Chinese version](https://github.com/perl6/doc/blob/master/README.zh.md), we could have more at least to populate search results in a number of languages.

## Suggestions

Translate to

- [x] Spanish
- [x] German
- [x] Dutch
- [x] Portuguese
- [x] French
- [x] Japanese

fluca1978:
I placed the first italian translation of the readme, and backlinked in the official readme.

JJ:
Thanks!

JJ:
@terceranexus6 is working on the Japanese translation.

JJ:
@antoniogamiz and @elenamerelo will be working on the Spanish translation.

zaucker:
I can do a German translation.

JJ:
Please do that!

zaucker:
Ok, I will do the translation into German.

Reko98:
Hey! I can do the french Readme, i'm not French but i know a little bit of that language and I have a few friends who could help me with the translation, since they study this language.
Do you want me to do it?

JJ:
Please do it! Thanks a lot.

Or better

S'il vous plait, vous les faites. Merci beaucoup.

And that's the full extent of my knowledge of French...

JJ:
nohighlight runs htmlify.p6, which needs node and graphviz. Mojolicious run
the local website, which is run from make run. And yes, make html has all
those dependencies.

PR #2011, opened 2018-05-08
README.de.md

zaucker:
## Problem

Missing translation to German [issue 1984](https://github.com/perl6/doc/issues/1984)

## Solution provided

Translated README.md to German and provided README.de.md

I left some words in English which I think are clearer than a translation into German (e.g. GitHub repository). If others who know the German-speaking Perl6 community better, please feel free to translate.

## Open questions

- Is  it still appropriate to reference `rakudobrew` here? I thought this was not recommended anymore.
- The whole section "Building the HTML documentation" seems a bit ackward in regards to dependencies. I think this should be cleaned up. May be worth a separate issue?
- This leads to the question on how these translations are kept in sync with README.md


zaucker:
I am sure it is "ok" :-) ... I am a native German speaker (living in Switzerland).
I am just not sure about leaving some words in English. May be Moritz or other German people from the German Perl6 community have a better feeling for that.

zaucker:
@JJ I think it would be good to add references to new translations to all the translated versions. I could make PRs  for the (western) European versions (or include it in this PR), but certainly not for chinese and japanese.
I just wasn't sure if I should include that in this PR or make a separate one once the German version is merged.

zaucker:
Oh well, I went ahead and added the links to the various languages (hoping Google Translate knows enough Chinese and if not those who do can correct).

PR #2012, opened 2018-05-08
Document CArray.allocate

Kaiepi:
## The problem

CArray.allocate from https://github.com/rakudo/rakudo/pull/1779/commits/a02131d53d52bf96ba8f7bd56d04c45771dbb08c is undocumented

## Solution provided

The old example showing how to allocate a CArray was updated to use the new method


zoffixznet:
You should already have a commit bit to this repo and can commit directly.

Kaiepi:
Ah, right

PR #2020, opened 2018-05-09
hint and and link to set operators from Range

gfldex:

### DIFF
--- a/doc/Type/Range.pod6
+++ b/doc/Type/Range.pod6
@@ -69,6 +69,15 @@ It is possible to shift or scale the interval of a range:
     say (1..10) - 1;       # OUTPUT: «0..9␤»
     say (1..10) * 2;       # OUTPUT: «2..20␤»
     say (1..10) / 2;       # OUTPUT: «0.5..5.0␤»
+    
+=head2 Matching against Ranges
+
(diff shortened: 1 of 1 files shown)


JJ:
It would have helped if you had said something about the problem (issue, for instance) and what the solution is about. I'll check it anyway.

doc/Type/Range.pod6
```diff
@@ -69,6 +69,15 @@ It is possible to shift or scale the interval of a range:
     say (1..10) - 1;       # OUTPUT: «0..9␤»
     say (1..10) * 2;       # OUTPUT: «2..20␤»
     say (1..10) / 2;       # OUTPUT: «0.5..5.0␤»
+    
+=head2 Matching against Ranges
+
+Use L<set operators|/language/setbagmix> to match against Ranges. Please note
+that set operators may cast to L<Set|/type/Set> and can create heavy memory load.
```
JJ: Maybe clarify when that would happen? Does it happen always? Why would that create heavy load?

gfldex:
forget it, the docs are all yours

JJ:
Hey, no, they are not... It's just a small change...

Issue #2015, opened 2018-05-09
Terminology normalization


JJ:
## The problem

Looking for `handles` for some reason I have found that file handle appear at least in that form or in the `filehandle` form. We should try and normalize terminology across all the documents, and this is only an example. When normalization is done, we should turn it into a test later on. 

## Suggestions

Agree on one single form for terms that are frequently appearing in the documentation. Please add terms below, and let's discuss in the comments which form is preferred.

* [x] file handle, filehandle, file-handle → filehandle
* [x] runtime, run-time → runtime
* [x] short-hand, shorthand →shorthand
* [x] look-ahead, lookahead →lookahead
* [x] look-behind, lookbehind →lookbehind
* [x] smart-match, smartmatch, smart-matching, smart-matches → smartmatch

* [x] Add the words to the xt/word-variants.t test
* [x] Add them to the writing-docs/STYLEGUIDE.md

JJ:
@ufobat I see you have edited this, but could you please indicate the commit where you made that change?

AlexDaniel:
“filehandle” is done now, but we need tests.

rafaelschipiura:
@chsanch You can also add the words you changed to the doc/Type/X/IO/DoesNotExist.pod6 test.

JJ:
El mar., 29 may. 2018 a las 20:12, Rafael Schipiura (<
notifications@github.com>) escribió:

> @chsanch <https://github.com/chsanch> You can also add the words you
> changed to the doc/Type/X/IO/DoesNotExist.pod6 test.
>

You probably mean   xt/word-variants.t

chsanch:
Ok, I'm on it.

PR #2040, opened 2018-05-21
Change "file handle" to "filehandle"

chsanch:
This is part of the fixes suggested in #2015

## The problem
Looking for handles to solve #761, I have found that file handle appear at least in that form or in the filehandle form. We should try and normalize terminology across all the documents, and this is only an example. When normalization is done, we should turn it into a test later on.

## Solution provided

 Change: file handle, filehandle, file-handle → filehandle


nxadm:
THx!

PR #2075, opened 2018-05-29
Change terms as suggested in #2015

chsanch:
Terms fixed: runtime, shorthand, lookahead, lookbehind

## The problem
Looking for handles to solve #761, I have found that file handle appear at least in that form or in the filehandle form. We should try and normalize terminology across all the documents, and this is only an example. When normalization is done, we should turn it into a test later on.

## Solution provided

Changes applied: 
runtime, run-time → runtime
short-hand, shorthand →shorthand
look-ahead, lookahead →lookahead
look-behind, lookbehind →lookbehind

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -1779,7 +1779,7 @@ also modify the other.
 X<|smartmatch operator>
 =head2 infix C«~~»
 
-The X<smart-match operator> aliases the left-hand side to C<$_>, then evaluates
+The smartmatch operator aliases the left-hand side to C<$_>, then evaluates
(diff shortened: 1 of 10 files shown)


doc/Language/operators.pod6
```diff
@@ -1779,7 +1779,7 @@ also modify the other.
 X<|smartmatch operator>
 =head2 infix C«~~»
 
-The X<smart-match operator> aliases the left-hand side to C<$_>, then evaluates
+The X<smartmatch operator> aliases the left-hand side to C<$_>, then evaluates
```
JJ: You might want to leave this one as it is, since it's used for indexing. If you don't, just eliminate the `X<>` since it's already indexed a few lines above.

PR #2078, opened 2018-05-29
Change use of WHAT to ^name

rafaelschipiura:
It's the recommended way. WHAT is used to actually get a type object, which isn't what we want. We want the name of the class the object was instantiated from.

### DIFF
--- a/doc/Language/js-nutshell.pod6
+++ b/doc/Language/js-nutshell.pod6
@@ -257,15 +257,15 @@ say $str;           # def
 =end code
 
 While we are talking about C<instanceof>, the equivalent to C<typeof> or the
-C<constructor> property on Node.js objects in Perl 6 is the C<WHAT> attribute:
+C<constructor> property on Node.js objects in Perl 6 is the C<^name> meta-attribute:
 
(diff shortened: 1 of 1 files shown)


b2gills:
The equivalent of `typeof` is actually `.WHAT`.

This is why it has `:ok-test<WHAT>`, which opts it out of the test to make sure `.WHAT` isn't used when it isn't needed.

JJ:
El mar., 29 may. 2018 a las 21:55, Brad Gilbert (<notifications@github.com>)
escribió:

> The equivalent of typeof is actually .WHAT.
>
> This is why it has :ok-test<WHAT>, which opts it out of the test to make
> sure .WHAT isn't used when it isn't needed.
>

What's happened to this comment?

rafaelschipiura:
Thanks @JJ For the improvement.
Teaching people to use .NAME is not good, even if that's what JS actually does, because in many places it won't work. Since this is an introductory tutorial, I think it's better to give them the tool they should be using anyway:
https://irclog.perlgeek.de/perl6/2018-05-29#i_16222140

PR #2085, opened 2018-05-31
Fix a typo

ismailarilik:
memoized -> memorized


zoffixznet:
I sent you an invite to **perl6* GitHub org. You can accept it here: https://github.com/perl6/

That gives you a commit bit to most of our repos, including this one, so once you accept the invite you can merge this PR yourself.

zoffixznet:
or rather reject this PR and instead commit a change to make `memoized` (which is a proper word) to a link that explains what it is: https://en.wikipedia.org/wiki/Memoization :smile_cat:

ismailarilik:
> or rather reject this PR and instead commit a change to make memoized (which is a proper word) to a link that explains what it is: https://en.wikipedia.org/wiki/Memoization 😸

It is not necessary since this document doesn't seem to target newbies and all experts should know this term. =) I am closing the PR.

PR #2090, opened 2018-06-02
Fix some typos and wording

uzluisf:
## Solution provided
Fix some typos and wording.

### DIFF
--- a/doc/Language/numerics.pod6
+++ b/doc/Language/numerics.pod6
@@ -60,8 +60,8 @@ to have the extra precision.
 The L<Num> type offers
 L<double-precision floating-point|https://en.wikipedia.org/wiki/Double-precision_floating-point_format> decimal numbers, sometimes called "doubles" in other languages.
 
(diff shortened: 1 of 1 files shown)


doc/Language/numerics.pod6
```diff
@@ -138,7 +138,7 @@ how-is-it 3+2i;  # OUTPUT: «meh␤»
 The types that do the L<Rational> role offer high-precision and
 arbitrary-precision decimal numbers. Since the higher the precision the
 larger the performance penalty, the L<Rational> types come in two flavours:
-L<Rat> and L<FatRat>. The L<Rat> is the most oft-used variant
+L<Rat> and L<FatRat>. The L<Rat> is the most often-used variant
```
JJ: Actually, "oft" is OK. If you use often, you shouldn't use the dash.

uzluisf: Oh, I wasn't aware of that. Thanks for pointing it out!

PR #2094, opened 2018-06-07
Fix missing nodejs dependency in the Dockerfile.

kjkuan:
- It appears that the nodejs package no longer exists for the base image
    used by the Dockerfile, and since it's already using the `n` script to
    install the latest stable version of nodejs, I simply use `n` to
    install nodejs in the first place instead.

- Also fixed a small code formatting issue in packages.pod6.

- Corrected the instructions on running the built docker image in CONTRIBUTING.md.

Initially, I was only trying to fix the formatting issue in packages.pod6, but couldn't get
Mojolicious working to be able to preview my changes, so I ended up building a docker
image from the Dockerfile.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -273,11 +273,11 @@ documentation.
 You can skip all the above and just build and view documentation with these simple commands (if you have docker already installed):
 
     $ docker build -t perl6-doc .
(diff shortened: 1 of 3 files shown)


JJ:
Thanks!

Issue #1084, opened 2016-12-27
Need higher contrast for highlighted code


AlexDaniel:
I am grateful for the latest switch to atom highlighter, however, with current CSS the code is significantly harder to read.

For example, comments are ``#888`` on ``#CFCFCF``, this is not very readable. ``#489047`` is not very nice as well.

So I suggest to change both the background color for snippets and code colors to add more contrast. You can use one of the contrast checkers ([1](http://contrastchecker.com/), [2](http://leaverou.github.io/contrast-ratio/), [3](http://webaim.org/resources/contrastchecker/), [4](https://snook.ca/technical/colour_contrast/colour.html#fg=33FF33,bg=333333), …)  to verify the results.

PR #2100, opened 2018-06-15
add a "colors off" button to code sections

timo:
for people who can't stand syntax highlighting
for any reason.

It will remember the choice in sessionStorage, just like
debug mode, and will remove colors on page load if the
setting was turned on. Restoring colors is also possible
by just clicking the button again.

It'd be nice if someone with battle-tested JS knowledge could check if there's something wrong with the code, for example for older browsers.

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -3,6 +3,7 @@ $(function(){
     setup_auto_title_anchors();
     setup_debug_mode();
     setup_tables();
+    add_syntax_highlighting_toggle_buttons();
     $(window).resize(setup_search_box);
 });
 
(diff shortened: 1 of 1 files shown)


timo:
Also, the button could be styled to fit in better, both in terms of looks and positioning.

zoffixznet:
FWIW, we have jQuery on the site, so stuff like 

```js
        const buttonDiv = document.createElement("DIV");
        const button    = document.createElement("BUTTON");
        const text      = document.createTextNode("colors " + colorsAre);
        buttonDiv.style.float = "right";
        button.appendChild(text);
        buttonDiv.appendChild(button);
        editor.insertAdjacentElement("afterBegin", buttonDiv);
```  

Can be written simply as
    
```js
$(editor).prepend( 
  $('<div><button>colors ' + colorsAre + '</button></div>')
    .css({'float': 'right'})
);
```

AlexDaniel:
I'm :-1: on this. I don't think syntax highlighting can hurt, and those who religiously (or maybe even if for a good reason) can't stand it are much better off with a userscript that will unhighlight code blocks on *all* pages.

zoffixznet:
> people who can't stand syntax highlighting

Is there really a large number of users who'd find this feature useful? Feels like featuritis to me and those who are strongly bothered by the colors can add a single line of CSS to their user sheet.

rafaelschipiura:
I think it's useful because it increases contrast by removing the colors and that's an important accessibility feature.

AlexDaniel:
There were problems with contrast in the past (https://github.com/perl6/doc/issues/1084), please file a ticket if that is still an issue.

But accessibility is not an argument for a toggle like this. Firefox has a built-in “reader mode” which does that already:

screenshot of a reader mode

It's fully a thing for user agents and we shouldn't be inventing our own.

JJ:
If this was an issue, I would probably be against. But it's a pull request, so he's bothered enough to put some work into it. So I'm not in principle against, but I have a couple of issues
* First, it should really be tested for different browsers. We don't do that in Travis, so it's better if that's done in advance.
* I would also like to see a sample of how it displays, i. e., it's not confusing, it's not hiding part of the code, things like that. I mean, how usable is it. Can you please add that?

zoffixznet:
So... what's the verdict? ping @timo

AlexDaniel:
> So... what's the verdict? ping @timo

I think we can close this. Maybe it'd make sense to have a short section somewhere titled “[how to read these docs](https://xkcd.com/1343/)” and there we can explain how this can be done locally, but I'm not sure if it's worth it. Also, if current syntax highlighting is too colorful or has too low contrast, then we can tweak it (please open an issue or a PR).

PR #2103, opened 2018-06-17
Fix p6doc build errors

Kaiepi:
## The problem
`zef install p6doc` fails after building.

## Solution provided
After looking through zef's code, I found that it considers the build to have passed only if a truthy value is returned, so I made the build method return `True`.

### DIFF
--- a/Build.pm
+++ b/Build.pm
@@ -2,6 +2,7 @@
 # need this to install pod under .../share/perl6/doc
 # This fixes p6doc command line use.
 
+use v6;
 use File::Find;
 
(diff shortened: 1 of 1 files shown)


Kaiepi:
Would checking if all the files were copied over and returning `True`/`False` depending on whether or not it succeeded be a better idea?

Kaiepi:
Something like this

zoffixznet:
Can this be merged now?

coke:
Looks good to me, addresses JJ's concern.

PR #2106, opened 2018-06-18
Document using dynamic variables for state in grammars

Kaiepi:
## The problem

How to embed code in grammars is largely undocumented.

## Solution provided

Add documentation on how to do that, starting with using dynamic variables for state.

### DIFF
--- a/doc/Language/grammars.pod6
+++ b/doc/Language/grammars.pod6
@@ -312,6 +312,53 @@ provided by parse methods:
     # OUTPUT: «12␤»
     =end code
 
+=head2 Dynamic Variables in Grammars
+
+Variables can be defined in tokens by prefixing the lines of code defining them
+with C<:>. Arbitrary code can be embedded anywhere in a token by surrounding it
+with curly braces. This is useful for keeping state between tokens, which can be
+used to alter how the grammar will parse text. Using dynamic variables
+(variables with C<$*>, C<@*>, C<&*>, C<%*> twigils) in tokens cascades down
+through I<all> tokens defined thereafter within the one where it's defined,
+avoiding having to pass them from token to token as arguments.
+
+One use for dynamic variables is guards for matches. This example uses guards to
+explain which regex classes parse whitespace literally:
+
+    grammar GrammarAdvice {
+        rule TOP {
+            :my Int $*USE-WS;
+            "use" <type> "for" <significance> "whitespace by default"
+        }
+        token type {
+            | "rules"   { $*USE-WS = 1 }
+            | "tokens"  { $*USE-WS = 0 }
+            | "regexes" { $*USE-WS = 0 }
+        }
+        token significance {
+            | <?{ $*USE-WS == 1 }> "significant"
+            | <?{ $*USE-WS == 0 }> "insignificant"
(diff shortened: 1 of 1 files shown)


JJ:
I have created an example that uses hidden class variables for state, instead of dynamic variables. I think it's a better alternative, since dynamic variables can be accessed from outside. Can you please check if this covers your intended behavior?

jnthn:
@JJ Did you test that example alternative? I'm pretty sure it won't work, since every rule that is called gets a fresh instance of the grammar class (which serves as its cursor object).

By contrast, dynamic variables are visible to subrule calls. When the parse is recursive, then the most recent declaration wins, which is also desirable. Furthermore, calling them "global variables" is wrong; they are dynamically scoped variables. Were they globals, parsing the same grammar in two threads at once would break horribly. But it doesn't, because each has its own call stack, and thus its own dynamic variables.

JJ:
@jnthn It's tested as written. It does not cover all the three examples in the original code, and it produces the self same output. I mean, just download it. Maybe using a class variable is not the best, but are you sure an instance variable wouldn't be much better in this case? Point is, dynamic variables can also be changed from outside, and it would be desirable to avoid side effects as a good practice.

jnthn:
@JJ D'oh, I somehow read `has` instead of `my`, because using `my` for this is so totally wrong that I didn't even imagine somebody doing it. :-) Yes, it will "work", until you use the grammar from multiple threads and then they will trample on the same state and do the totally wrong thing. So in trying to avoid something that isn't global state, your proposed alternative introduces real global state, and so isn't thread safe. Please don't advise this.

I'm not sure what "outside" you're worrying about with dynamic variables. It's visible in the dynamic scope of the parse, which means it can be seen by the grammar's own rules (which we want) and action methods (which is sometimes useful too). After the parse is done, it's no longer in scope.

JJ:
El lun., 18 jun. 2018 a las 11:40, Jonathan Worthington (<

> @JJ <https://github.com/JJ> D'oh, I somehow read has instead of my,
> because using my for this is so totally wrong that I didn't even imagine
> somebody doing it. :-) Yes, it will "work", until you use the grammar from
> multiple threads and then they will trample on the same state and do the
> totally wrong thing. So in trying to avoid something that isn't global
> state, your proposed alternative introduces real global state, and so isn't
> thread safe. Please don't advise this.
>

:-) That was me, doing the unimaginable. My rationale for doing that I
wasn't sure "parse", which is the usual way of doing this, actually
instantiates a grammar. You are, of course, right, but I didn't think of
that. And then the problem with class variables is quite general, it's not
reduced to grammars, right?

> I'm not sure what "outside" you're worrying about with dynamic variables.
> It's visible in the dynamic scope of the parse, which means it can be seen
> by the grammar's own rules (which we want) and action methods (which is
> sometimes useful too). After the parse is done, it's no longer in scope.
>

It can be seen by sub-rules, which you might or might not want; you could
imagine them doing someting bad along the class hierarchy. I don't know.
There should be a thread-safe way of encapsulating state, other than
dynamic variables. At any rate, if you think this PR is OK such as it is
now, we can merge right away.

PR #2110, opened 2018-06-20
Fix links to NativeCall

dmaestro:
There is currently no /type/NativeCall page, everything seems to be in /language/nativecall

## The problem
Broken links

## Solution provided
Link to /language/nativecall

### DIFF
--- a/doc/Language/nativetypes.pod6
+++ b/doc/Language/nativetypes.pod6
@@ -22,7 +22,7 @@ Some simple types in Perl 6 have a I<native> representation, indicating that th
 =end table
 
 However, these types do not necessarily have the size that is required by the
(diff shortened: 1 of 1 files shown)


zoffixznet:
FWIW: you're a member of **perl6** github org and you have sufficient permissions to just commit these changes directly.

dmaestro:
@zoffixznet Thanks! As I'm getting more familiar with the contribution process, I'll start taking those actions. Just waiting now for the checks to finish ...

zoffixznet:
FWIW, currently in release announcements you're listed as `dmaestro`. if you want to be listed under a different name, add an entry into https://github.com/perl6/doc/blob/master/CREDITS file

Issue #1513, opened 2017-09-03
TBDs on nativecall page


AlexDaniel:
See https://docs.perl6.org/language/nativecall and search for TBD.

JJ:
* [ ] Typed pointers
* [ ] Buffers and blows
* [ ] Library path and names.

Kaiepi:
Library paths and names I've had to work out a lot of problems with because of OpenBSD's... rather interesting way of naming, versioning, and storing them, so I could probably write that up.

JJ:
Please do it :-) Thanks!

JJ:
@Kaiepi did you have the chance to work on that part?

JJ:
:loudspeaker: 
Just library path and names to go!

JJ:
@Kaiepi can work on that if he wants, too.

Kaiepi:
I'll get started on this tomorrow

PR #2140, opened 2018-07-03
More documentation on library paths and names

Kaiepi:
## The problem
Library paths and names needed more documentation. See https://github.com/perl6/doc/issues/1513

## Solution provided
More was written

### DIFF
--- a/doc/Language/nativecall.pod6
+++ b/doc/Language/nativecall.pod6
@@ -646,18 +646,27 @@ to callbacks.
 
 =head1 Library Paths and Names
 
-=comment TODO
+The C<native> trait accepts the library name, the full path, or a subroutine
+returning either of the two. When using the library name, the name is assumed
+to be prepended with "lib" and appended with ".so" (or just appended with
+".dll" on Windows), and will be searched for in the paths in the
+LD_LIBRARY_PATH (PATH on Windows) environment variable.
 
-TBD more
-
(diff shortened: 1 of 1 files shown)


Skarsnik:
This example is misleading. the function does what specify the ABI version to the native trait is supposed to do, and people SHOULD always specify an ABI version (very few execption like libmysqlclient ><)

Kaiepi:
Updated

Skarsnik:
No, I mean the content of the function itself. The example was correct previously, it's just we should not encourage users to do what the ABI version does already :)

Kaiepi:
You mean like returning `/usr/local/lib/ebar/bar`?

Skarsnik:
Just read the https://docs.perl6.org/language/nativecall#ABI/API_Version
If I write
`is native('bar', 'v1)` it should give you /usr/lib/libbar.so.1 on openbsd (/usr/lib/ will be the default system library path, (I think NC/moar query ld config).) If it's not the case it should be patched in NC or Moar to follow the proper OpenBSD intended behavior.

But what I mean is users should be encouraged to use the ABI version argument. Giving the user an example of the use of the Routine argument type witch does what the API feature does is probably a bad idea/misleading. Maybe a better example could be just a function that call pkg-config to get the path (or something similar)

Skarsnik:
Thank :)

Issue #302, opened 2015-12-30
Need a way to mark which version something is part of.


coke:
Don't need this right away, but eventually, need to be able to 1) handle multiple implementations, 2) handle multiple specifications.

(e.g. ".sort was introduced in Perl 6.d; it is available in rakudo 2017.01, niecza 2017.05")

Ideally, we'd like this to be a simple notation in the original POD (or in a data file) rather than a formatted block we have to copy around.

AlexDaniel:
Does it need any special markup? Why can't it be mentioned in text (free-form)?

coke:
Because in a perfect world, we want to be able to write it (or perhaps read it) programmatically, which free-form text will make difficult.

AlexDaniel:
@coke perhaps, but what's a real use case there?
For example, differences between implementations can be read programmatically from roast.

coke:
So we can, for example:

- write a test to make sure we're tracking this information.
- automate generating an index based on spec version
- automate generating docs for a specific rakudo version

We can split this into 2 tickets if you like, one for specs, one for implementations. The spec one is much more important in my opinion.

coke:
if you search for "201\d.\d\d" you'll find many instances talking about the version of rakudo something is in. Some of these have some brackets, e.g.:

```
[Starting in 2017.02]
```

I would recommend picking a standard showing which versions maybe as a range, and including the compiler

```
[Available: Rakudo 2017.02..*]
```

AlexDaniel:
@coke yup! This totally makes sense. At least for now, this will work just fine.

AlexDaniel:
Relevant: https://github.com/perl6/doc/pull/1484 https://github.com/perl6/doc/pull/1485

AlexDaniel:
Some discussion: http://colabti.org/irclogger/irclogger_log/perl6-dev?date=2018-07-07#l193

I think that's not just a wishlist, this is something we *must* figure out as soon as possible.

Altai-man:
Just a thought: the issue is current docs structure assumes that docs are a bunch of files, they can evolve, but not diverge into branches(which is essentially what is needed to decide what docs parts are shipped for exact version). if we could map docs into more flexible format, we could then assemble it from pieces(e.g. `generic`, `c` = generic + c specific, `d` = generic + d specific etc), so we can update parts separately(e.g. wording) or add specific parts/notes into some version. Or a possibility to "overload" some documentation chunk in a OOP way.

[...]

PR #2155, opened 2018-07-07
Update Str.pod6

interlab:
starts-with, ends-with returns Bool

### DIFF
--- a/doc/Type/Str.pod6
+++ b/doc/Type/Str.pod6
@@ -837,7 +837,7 @@ one of the following will:
 
 =head2 method starts-with
 
-    multi method starts-with(Str:D: Str(Cool) $needle --> True:D)
+    multi method starts-with(Str:D: Str(Cool) $needle --> Bool:D)
 
 Returns C<True> if the invocant is identical to or starts with C<$needle>.
 
@@ -846,7 +846,7 @@ Returns C<True> if the invocant is identical to or starts with C<$needle>.
 
 =head2 method ends-with
 
-    multi method ends-with(Str:D: Str(Cool) $needle --> True:D)
+    multi method ends-with(Str:D: Str(Cool) $needle --> Bool:D)
 
 Returns C<True> if the invocant is identical to or ends with C<$needle>.
 


tbrowder:
Thanks!

interlab:
```perl6
perl6 -e "'qwerty'.starts-with('q').WHAT.^name.say;"

# Bool
```

JJ:
@interlab that's exactly what it returns, but please read what I have pointed to. The function is not defined to return Bool (so it could return something else), and it's not tested to return Bool, so it could return something else. We should be careful to document not actual behavior, but specified behavior, and in this case the specification is vague.
In practice, the definition of the functions you are documenting is not 

     multi method starts-with(Str:D: Str(Cool) $needle --> Bool:D) 

but [actually](https://github.com/rakudo/rakudo/blob/72ccd432a7155e408791da6cafd3b4c408661c04/src/core/Str.pm6#L137-L142)

    proto method starts-with(|) {*}
    multi method starts-with(Str:D: Cool:D $needle) {self.starts-with: $needle.Str}
    multi method starts-with(Str:D: Str:D $needle) {
        nqp::p6bool(nqp::eqat(self, $needle, 0))
    }

Not even the the definition of a method of Cool, which is the closest to the one you're defining, is like that. Please [check it out](https://github.com/rakudo/rakudo/blob/e4e753b0232692322560d6def6f6f0ec29fa01e5/src/core/Cool.pm6#L145-L147)

    method starts-with(Cool:D: |c) {
        self.Str.starts-with(|c)
    }

Please note that it does not specify a `return Bool` either.

tbrowder:
Oops, nice catch, @JJ, I’ll see if I can revert this.

AlexDaniel:
Don't revert, it's better than it was.

AlexDaniel:
See https://github.com/perl6/doc/issues/302 for further discussion on how to document versioned stuff.

JJ:
Agree, don't revert, just change a little bit...

AlexDaniel:
@JJ Change how? See [this discussion](http://colabti.org/irclogger/irclogger_log/perl6-dev?date=2018-07-07#l193). TL;DR it is fairly obvious that starts-with should return a Bool, so in this particular case it is fine to document that. The whole situation is LTA but until something is done with https://github.com/perl6/doc/issues/302 I don't know if there's a way to do better.

JJ:
Well, the declaration of the methods should reflect what's actually there.
I'm OK with the text saying that it's returning a Bool, since that's what's
doing _now_. But maybe add a non-visible note that the behavior is not
specced. Of course, we can also say what is actually specced, which is it's
returning a truish value. It including what's happening now, and if it's
eventually specced that way, it can change.

JJ:
And... @zoffixznet just specced that in Roast, so it's actually tested that it's returning a Bool, but still the method definition does not match what's there. So maybe change it just a tiny bit to match that?

PR #2157, opened 2018-07-07
Modify to allow pod6 page sorting by :page-order value

tbrowder:
+ added: sub get-pod6-page-order

+ added: process 'Language' with page-order

+ added: various comments


tbrowder:
Thanks, @JJ, and, yes, I have tested it.

Issue #1391, opened 2017-06-21
Always display google search option


zoffixznet:
If you type a search term and no results show up, you're given an option to do a google search for that term. However, that option isn't available when some results turned up.

I think the option to do a google search should always be present because the found results might not be what the user is looking for (e.g. I searched for "binder"), yet the site might still have the word used somewhere the search index didn't grab it, so a google search would still be useful.

JJ:
I have been testing with this. It looks easy in principle, just a matter of commenting out the part of the JS in search.js that adds this only if there are no results. However, there are some minor and some major problems. Minor one is the message, which is "Not found". That can be changed. Major one is that the message is not "added", but rather pops up in a separate div when the drop-down menu is closed. That's not noticed with no result because there's no result. 
So, for a quick-and-dirty solution, if not ideal in UX terms, it can be done fast addressing this issue (it displays google search option alright). However, I wonder if you were thinking more along the lines of adding it as the last option in the dropdown menu or something like that...

AlexDaniel:
> However, I wonder if you were thinking more along the lines of adding it as the last option in the dropdown menu or something like that...

Yes.

JJ:
I was afraid you would say so :-)

randyl:
Closed by #2171

PR #2171, opened 2018-07-13
Google search option

randyl:
## The problem

Issue #1391 describes a desire to always have a google search option in the search autocomplete menu, not just when there are no results.

## Solution provided

In the _renderMenu function of search_template.js, a new menu item is added after all other items are sorted. It uses a new category of "Site Search", and contains a link to a google search for the current keywords. See screenshot.

menu_screenshot

If the wording of the category and menu item is deemed appropriate, we might consider changing the no-results text from "try web search" to "try site search" to be consistent.

I did not see any existing tests for the search autocomplete menu, so there are no tests to accompany this change.


zoffixznet:
👍 Thanks!

zoffixznet:
> If the wording of the category and menu item is deemed appropriate, we might consider changing the no-results text from "try web search" to "try site search" to be consistent.

I sent you an invite to **perl6** GitHub org. You can accept it here https://github.com/perl6/ and commit any further changes you wanted to make directly.

randyl:
Thanks. I did change the no-results text to "try site search". I'll close #1391 as well.

PR #2172, opened 2018-07-13
Do the correct Fahrenheit to Kelvin conversion.

dmaestro:
Fix the Fahrenheit to Kelvin / Celsius factor, and update expected result.

## The problem
Example for Bridge uses the incorrect (inverted) factor for Temperature unit conversion.
The sum (expressed in Kelvin) was obviously wrong.

## Solution provided
Correct the conversion factor, and the example result


zoffixznet:
👍 Thanks!

PR #2177, opened 2018-07-15
Phase2

tbrowder:
This PR implements the Phase 2 Language doc list order control described on the doc wiki.


manage-page-order.p6
```diff
@@ -0,0 +1,414 @@
...
+sub write-group-page($group-char, $title, $page-order) {
+    my $fname = $todir ~ "/00-Group{$group-char}.pod6";
+    my $fh = open $fname, :w;
+
+    # x2588 big solid block <== using this char to make the separator
+    # x2501 solid thick horizontal line
+    $fh.print: qq:to/HERE/;
+        =begin pod :page-order<{$page-order}>
+        =TITLE $title
+        =SUBTITLE █████████████████████████████████████████
```
AlexDaniel: Why are we doing unicode art when we can output html?

AlexDaniel: Or what's the point of this?

tbrowder: Until the html can be affected, the bar gives a good visual separator between groups. This step is intended to get the idea of specific grouping and sorting onto the screen with existing tools before the rendering tools are modified.

There is a thinner bar available, or other choices.

I also am investigating the travis pod failure...

tbrowder: Fixed typo

Issue #2178, opened 2018-07-16
Circular and Incorrect link on NativeCast HTML page.


Xliff:
## The problem
On the following page:

https://docs.perl6.org/routine/nativecast

The link to "Function Pointers" points to https://docs.perl6.org/routine/nativecast#Function_Pointers, rather than https://docs.perl6.org/routine/nativecall#Function_Pointers

Altai-man:
Closed with @Xliff++ PR.

PR #2180, opened 2018-07-16
- Fixes bad link in route page for nativecast.

Xliff:
## The problem
A potential fix for #2178.
The link to "Function Pointers" in the nativecast page didn't point to the right location.

## Solution provided

This fixes the link, however I was not able to view the change properly due to the app interpreting forward slashes as "$SOLIDUS".


tison1:
LGTM, you might edit `util/trigger-rebuild.txt` to trigger rebuild to make CI happy. I see the error log, which complains about SEGV. I am sure it is irrelevant about this PR

Altai-man:
Just checked it: unfortunately, it seems that the proposed solution is incomplete. It starts link forming from `/routine` root, which is why `SOLIDUS` url transformation is applied. But url must be absolute to point to correct page.

Issue #2185, opened 2018-07-17
new manage-page-order.p6 script


coke:
These are think should be done:

* Move the p6 script be moved to the util/ directory
* Add the new makefile target should be added to PHONY.
* Fix .gitignore to ignore these newly generated files

Should be considered:

* do this as part of the build process (on master, htmlify.p6) - (in coke/build branch, I'm working on moving things out of htmlify and using sake, happy to do that there without moving things around on master first)
* Don't generate files under the docs/ folder. we already have a confusion under html of generated vs. checked in files.  Generated artifacts might want to go elsewhere.

tbrowder:
Is htmlify.p6 to be moved to /util also?

tbrowder:
My proposal (and I’m working on a PR for it) is to:

1. rename “***doc***/“ to “***doc-src***/“
2. copy or generate a transformed copy of ***ALL FILES*** in “***doc-src***/“ to matching dirs in “***build***/“
3. point “***pod2bigpage***” and “***htmlify.p6***” to “***build***/“ for the build

We can leave the “***html***/“ dir alone for now.

coke:
I had two very specific items to address here (just added a third).

There's a lot of scope creep in the comments here, but since we've opened the can of worms: my thoughts (again, this is bigger than I am proposing for just this ticket, which was just in regards to the things that were just added to the repo):

* leave doc/ alone - I'm not sure what the rename here was intended to address 
* I'm working to move most of htmlify.p6 into a separate (hopefully more parallelizable build using sake). We could move it, but it's not part of this ask.
* what's to transform? it's not a one to one for file in doc to built file.
* have html/ just contain the built site, not a combination of checked in content and generated files.
* ok to have a build/ step that contains intermediate generated files. I'm wondering if we even need these generated files, but that's beyond the scope of my ask here.
* Do we even need to have these files generated? why can't we do the sort order in the htmlify logic itself?

But anything aside from those 3 comments should be done in another ticket.

tbrowder:
Keeping doc name is not a problem, but using ***htmlify.p6*** to provide the sort is IMHO a problem (by file name?, title?, change the order easily?). That’s why I came up with the control file: make it a bit easier for a human to choose or modify the sort order. 

The ***pod2bigpage*** exec looks at only one source dir so all files to be sourced need to be in one dir, hence my choice of copying direct or transforming the pod6 source files to the intermediate dir ***build***.

JJ:
[...]

PR #2189, opened 2018-07-19
Doc reorg

tbrowder:
Should address doc issue #2185.


build/Programs/000-README
```diff
@@ -0,0 +1 @@
+This is a place holder to preserve this dir in the doc repo.
```
AlexDaniel: I don't like this. There has to be a better way.

util/manage-page-order.p6
```diff
@@ -48,20 +53,89 @@ given $mode {
...
+    write-Language-files();
+
+    # THEN
+    # copy all under the doc dir to the build dir
+    # NOTE this assumes the sort order is alpha by file name
+    # NOTE we exclude the Language dir
+    copy-dir-tree(:fromdir('doc'), :todir('build'), :exclude('Language'));
+}
+
+=begin comment
```
AlexDaniel: Commented out code?

util/manage-page-order.p6
```diff
@@ -393,13 +467,13 @@ sub is-done(*@args) {
 }
 
 sub write-group-page($group-char, $title, $page-order) {
-    my $fname = $todir ~ "/00-Group{$group-char}.pod6";
+    my $fname = $todir ~ "/{$page-order}-Group{$group-char}.pod6";
     my $fh = open $fname, :w;
 
-    # x2588 big solid block <== using this char to make the separator
-    # x2501 solid thick horizontal line
+    # note no SUBTITLE
```
AlexDaniel: You sure that's OK? I think Pod::To::BigPage is unhappy about stuff like this, is it actually alright in this case?

tbrowder: it works on my local debian box with firefox looking at the pages

util/manage-page-order.p6
tbrowder: I was savng it until all checked out—it can be removed.

PR #2197, opened 2018-07-20
revert file names to no number prefix

tbrowder:
Keep reorg, but remove number prefixes from transformed files for now.


AlexDaniel:
Now we're back to this stuff: https://docs.perl6.org/build-log/build-2018-07-20T17:05+00:00.log

JJ:
Will have to check this out

Issue #1531, opened 2017-09-03
How to use wincompose (unicode_entry)


AlexDaniel:
We should mention https://github.com/samhocevar/wincompose on https://docs.perl6.org/language/unicode_entry

JJ:
Can anyone who actually uses Windows do this? Or I can just point to that resource and tell people to use it.

JJ:
Do you mind if I assign this issue to you and you do that part? I don't have Windows even in a VM...

JJ:
So, does anyone want this issue?

mienaikage:
I use WinCompose on my gaming machine with [kragen/xcompose](https://github.com/kragen/xcompose/blob/master/dotXCompose). I could look into putting some instructions together :+1:

JJ:
Thanks!

PR #2206, opened 2018-07-22
Add WinCompose to unicode_entry

mienaikage:
Closes https://github.com/perl6/doc/issues/1531.

### DIFF
--- a/doc/Language/unicode_entry.pod6
+++ b/doc/Language/unicode_entry.pod6
@@ -90,9 +90,6 @@ These sequences can be viewed by right clicking the tray icon and selecting C<Sh
 If you wish to add your own sequences, you can do so by either adding/modifying
 C<.XCompose> in C<%USERPROFILE%>, or editing user-defined sequences in the options menu.
 
-By default, WinCompose allows you to enter any unicode character by pressing
-your compose key, followed by C<u>, followed by the unicode code point.
-
 =head1 Editors and shells
 
 =head2 Vim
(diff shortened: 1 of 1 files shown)


AlexDaniel:
> By default, WinCompose allows you to enter any unicode character by pressing
> your compose key, followed by C<u>, followed by the unicode code point. 

I don't like this remark. The way the sections ends with this, it makes me feel like the suggested way is to enter characters by codepoint, which is generally what you can avoid by using the compose key.

mienaikage:
Shall I throw in an 'also' after 'WinCompose' or change it further?

AlexDaniel:
I'd say remove that paragraph completely. I've seen people assume that alt codes is the only way to write non-ascii chars, I don't think there's a need to reinforce the belief that this way of typing non-ascii characters is even remotely sensible. I know that some people do it, and that's fine, but I don't think it is something we should suggest or even mention.

For the [survey](https://docs.google.com/forms/d/19qSBpGWWc9RNruxPWySHkmy3tppaqeqUNGfZpHhdsMs/viewanalytics) I listed these ways to type non-ascii chars:
* Compose key
* Using features of a text editor
* Custom keyboard layout
* Perl6::Tidy
* Input-method frameworks (e.g. dbus)
* Helper tools provided by the OS
* `digraphs` utility
* Software for keyboard macros
* Alt codes
* Copy-pasting

Given all the great options (excluding things like copy-pasting), I don't think entering by codepoint is worth a mention at all.

AlexDaniel:
Looks good to me, thanks!

FWIW if someone disagrees regarding the entry by codepoint, feel free to comment or even put the paragraph back.

PR #2209, opened 2018-07-23
move i18n files to new subdir

tbrowder:
## The problem

Translations clutter main dir

## Solution provided

Move to new subdir and link to main README.md in main dir


tbrowder:
Oops, I didn't check inside the files. Unless there is an objection, I'll try to make them all link to the "correct" place.

I recognize that Perl 5 may have all READMEs at the top, and we are trying to cater to non-English speakers in a fair and somewhat accepted manner. I'll ask AlexDaniel to give directions.

Issue #2187, opened 2018-07-19
substr-rw is missing an explanation into the default value of $length


Xliff:
## The problem
In the description of Str.substr-rw (https://docs.perl6.org/type/Str#method_substr-rw), the default value for $length is not specified, even though $length is defined as optional.

## Suggestions
According to IRC and experiments, a simple line saying that "$length, if not specified, defaults to the end of the string"

PR #2217, opened 2018-07-25
- Addresses issue #2187.

Xliff:
## The problem
The method section for Str#substr-rw did not mention the default value of $length.

## Solution provided
Patch supplies the proper wording, as discussed on IRC, and on the pages for #2187


zoffixznet:
IMO it'd be more useful to specify the actual default in the signature `method substr-rw($from, $length = Inf)` or `method substr-rw($from, $length = *)`

That way the reader can see the default value just by glancing at the signature, instead of trying to skim the prose, hoping to find it.

Xliff:
I will add this as well as the explanation.

PR #2219, opened 2018-07-26
Document grammar attributes

Kaiepi:
## The problem
Using attributes in grammars is undocumented.

## Solution provided
Document and provide an example using them.

### DIFF
--- a/doc/Language/grammars.pod6
+++ b/doc/Language/grammars.pod6
@@ -288,7 +288,7 @@ the number for the currently parsed C<digit> digits.
     say Digifier.parse('255 435 777', actions => Devanagari.new).made;
     # OUTPUT: «(२५५ ४३५ ७७७)␤»
 
-=head2 Methods in Grammar
+=head2 Methods in grammars
 
 It's fine to use methods instead of rules or tokens in a grammar, as long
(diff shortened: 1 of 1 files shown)


doc/Language/grammars.pod6
```diff
@@ -288,7 +288,7 @@ the number for the currently parsed C<digit> digits.
     say Digifier.parse('255 435 777', actions => Devanagari.new).made;
     # OUTPUT: «(२५५ ४३५ ७७७)␤»
 
-=head2 Methods in Grammar
+=head2 Methods in Grammars
```
JJ: Can you please not use capitals here? Either you use C<Grammar>s or simply grammars.

Kaiepi:
Updated

Kaiepi:
I'll see if I can come up with a better example than this. I don't think this is very legible

JJ:
@Kaiepi the example is correct, but what you are explaining, attributes and methods, take a very small part of the total real estate. Maybe a simpler grammar would be enough...

Kaiepi:
Updated
I added more details since there was some stuff I was forgetting to mention, along with why attributes don't work in tokens. The example isn't much shorter, but should better illustrate how they behave

PR #2231, opened 2018-07-30
add missing 'class' keyword for consistency with other types

hankache:
## The problem
`X::Method::NotFound`, `X::Parameter::MultipleTypeConstraints`, `X::Parameter::WrongOrder` and `X::Placeholder::Mainline` have no `class` keyword in their `=TITLE`. All other classes do.

## Solution provided
Add `class` keyword in their `=TITLE`

### DIFF
--- a/doc/Type/X/Method/NotFound.pod6
+++ b/doc/Type/X/Method/NotFound.pod6
@@ -1,6 +1,6 @@
 =begin pod
 
-=TITLE  X::Method::NotFound
+=TITLE class X::Method::NotFound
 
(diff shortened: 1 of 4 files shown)


zoffixznet:
FWIW: You should have a commit bit and can commit directly.

hankache:
@zoffixznet I do. Just waiting for Travis to finish before I merge.

hankache:
@zoffixznet on a side note, you're right I should commit directly without submitting a PR

PR #2242, opened 2018-08-07
Fix link to Inline::Ruby

mienaikage:
Existing URL 404s


mienaikage:
I figured the ecosystem link may be the better option in case the repo name were to ever change again 🙂

Issue #2246, opened 2018-08-09
Use HTTPS URLs, where available


zoffixznet:
We have a bunch of HTTP URLs in the docs:
```
2018.06 zoffix@VirtualBox~/R/doc (master)$ grep -FIRn 'http://' . | wc -l
1705
```

It's possible many of them actually have HTTPS versions (like [here](https://github.com/perl6/doc/commit/426579ee538737983665c822b854caeb97636213)) and perhaps we should link to HTTPS directly.

JJ:
And we should also include a test so that it does not happen, if possible,
whitelisting those that have no alternative.
But, wow, 1705 is a lot.

zoffixznet:
> But, wow, 1705 is a lot.

Tho in `doc/` dir it's just:
```
2018.06 zoffix@VirtualBox~/R/doc (master)$ grep -FIRn 'http://' doc | wc -l
62
```

And after running `make clean`, the whole repo:

```
$ grep -FIRn 'http://' . | wc -l                                                                                                                                                                                      
197
```

So it looks like something gens a ton of HTTP URLs or something.

JJ:
El jue., 9 ago. 2018 a las 12:46, Zoffix Znet (<notifications@github.com>)
escribió:

> But, wow, 1705 is a lot.
>
> Tho in doc/ dir it's just:
>
> 2018.06 zoffix@VirtualBox~/R/doc (master)$ grep -FIRn 'http://' doc | wc -l
> 62
>
> And after running make clean, the whole repo:
>
> $ grep -FIRn 'http://' . | wc -l
> 197
>
> So it looks like something gens a ton of HTTP URLs or something.
>

Chalk another one for htmlify.p6...

Also there are lots in README.md and translations... Multiplied by 7, no
less.

MorayJ:
That should be it for all of the ones that can currently change to https in doc/Types and doc/Language. Nothing in doc/Programs. I get the idea that the others are generated so need a different approach...

PR #2247, opened 2018-08-09
Makes some http links across 'Language' documents https

MorayJ:
For format L<name|url> in 'Language' docs, changes http to https where appropriate.

## The problem
This is a start on the following:
Use HTTPS URLs, where available #2246

## Solution provided
I've switched http to https in the 'Language' docs where the format is L<name|url>

https://vimdoc.sourceforge.net/htmldoc/digraph.html is not correctly configured for https so left as http

https://vim.wikia.com/wiki/Entering_special_characters redirects to a non-secure version so have left as http

### DIFF
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -486,7 +486,7 @@ apply.
     say $subref->($foo, $bar);
 
 In relatively recent versions of Perl 5 (5.20 and later), a new feature allows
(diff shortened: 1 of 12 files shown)


zoffixznet:
I sent you an invite to **perl6** GitHub org. You can accept it here: https://github.com/perl6/

And once you do, you should be able to merge this PR yourself and commit any further changes directly.

Issue #2250, opened 2018-08-09
404 on link on https://docs.perl6.org/language/5to6-nutshell


MorayJ:
## The problem
Page:
https://docs.perl6.org/language/5to6-nutshell
Has a broken link:
http://www.perlfoundation.org/perl6/index.cgi?perl_6_delta_tablet

## Suggestions
www.perlfoundation.org perl6 section forwards to https://perl6.org/ now so I think the link target is long gone.

Suggest linking to the overview document:
https://docs.perl6.org/language/5to6-overview

This page (5to6-overview) also contains a link back to:

 https://docs.perl6.org/language/5to6-nutshell

which might make it a bit self-referential (amongst other useful links). Alternative is to simply remove the link, but I think this way adds more value than irritation.

PR #2251, opened 2018-08-09
Replace broken link with a new link

MorayJ:
Replace http://www.perlfoundation.org/perl6/index.cgi?perl_6_delta_tablet with https://docs.perl6.org/language/5to6-overview for tips on translating Perl 5 to Perl 6

## The problem
404 on link on https://docs.perl6.org/language/5to6-nutshell #2250

## Solution provided
I've replaced the link with the one to the 5to6-overview, which seems best to me.


rafaelschipiura:
Looks good to me!

Issue #2176, opened 2018-07-14
The "1-item list" in Any docs is inaccurate


zoffixznet:
(I think I filed some general "Any improvement" Issue before, but I think it got fixed)

A lot of routines on [`Any` docs](https://docs.perl6.org/type/Any) (eg. `.classify`) say "Treats the Any as a 1-item list". However that's inaccurate.

I believe the actual behaviour is *"coerces the object to a list by calling `.list` method"*. While by default, it does indeed create a 1-item list with the object in it, nothing stops users declaring their own `.list` returning multiple items.

The "Treats the Any as a 1-item list" phrase suggests that even if you define your own `.list` the methods would not call it and still treat your object as 1-item list.

```perl6
<Zoffix_> m: class Foo { method list { <a b c> } }; dd Foo.classify: * eq "a"
<camelia> rakudo-moar 9dce63318: OUTPUT: «(my Any %{Any} = Bool::False => $["b", "c"], Bool::True => $["a"])␤»
```

MorayJ:
So, could they say "Coerces the Any to a list by calling its .list method. By default, this will be a 1-item list"

Or is the 1-item list not necessary information? Or could it be more explicit in saying what that default list item is? (though I don't know what it is...asking)?

zoffixznet:
> Or is the 1-item list not necessary information

I'd say that's unnecessary in all the methods but the actual docs for `.list`

zoffixznet:
Fixed by the merged PR

PR #2252, opened 2018-08-10
Any list clarity

MorayJ:
## The problem
The "1-item list" in Any docs is inaccurate #2176

## Solution provided

I have standardised the language based on other language in the document, though will need checking by someone with a greater understanding of what's actually being said.

Looking at the Cool type document, I have changed all descriptions which say 'coerce' to 'coerces' as there were both in the document.

I have added 'the invocant' in many places and hope this is correct to fit in with the language I was using in this update. Again, this was something used sporadically and I had to choose what to follow. It is now more consistent.

I have referred to C\<List\> meaning a class, rather than just 'a list'. Again, this seems more consistent to me with the rest of the document.

### DIFF
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -472,7 +472,7 @@ Defined as:
 
     method Supply(--> Supply:D) is nodal
 
-Coerces the invocant first to L<List> by applying its
+Coerces the invocant first to a C<list> by applying its
 L«C<.list>|/routine/list» method, and then to L<Supply>.
 
 =head2 method min
@@ -567,7 +567,7 @@ Defined as:
 
     method flatmap(Any:U: &code --> Seq)
 
-Coerces the invocant to L<List> by applying its
+Coerces the invocant to a C<list> by applying its
 L«C<.list>|/routine/list» method and uses
 L«C<List.flatmap>|/type/List#method_flatmap» on it.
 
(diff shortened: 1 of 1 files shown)


zoffixznet:
Looks good except for the:

> I have referred to C\<List> meaning a class, rather than just 'a list'. 

 There's a deliberate difference there. Note there are both `.List` and `.list` methods. Only `.List` specifically coerces to a `List` type; `.list` is allowed to coerce to any `Positional` type:

```
[].List.^name.say; # List
[].list.^name.say  # Array
```

------------

(the same thing exists for `.Hash` vs `.hash`; the latter is allowed to return any `Associative`)

MorayJ:
Ok, will have another go...see it's failed a test as well.

MorayJ:
Having doubts about methods pairs, anti-pairs and kv. Based the re-wording on method pairs and not sure that it's correct in List v list..? Bit out of my depth there. Should I revert that part?

But otherwise, I think this is ok as far as fixing the related bug?

zoffixznet:
> Having doubts about methods

The only thing I see is the use of word `undefined`. Should be changed to something like `…if the invocant is a type object`. The [non-definite](https://docs.perl6.org/syntax/DEFINITE) objects (i.e. type objects rather than instances) are slightly different from ["undefined" objects](https://docs.perl6.org/routine/defined). An object can override the `.defined` value (and some things like `Failure`s do) and can be false even for instances:

```
say Failure.new.DEFINITE; # True
say Failure.new.defined;      # False
```

zoffixznet:
👍 Thanks!

(gonna fix up the `defined` thing myself; need to do some other edits in the same area and don't want to introduce merge conflicts).

PR #2253, opened 2018-08-10
Add some small advise about module's update

uzluisf:
## The problem
After struggling for half an hour with a change to a local module that wasn't being reflected, 
I realized that I needed to uninstall and install the local module again for the change to take place.

If this is mentioned somewhere else, this can be disregarded.

## Solution provided
Point out that one must uninstall and install a given local module for the latest changes to take place.

### DIFF
--- a/doc/Language/modules.pod6
+++ b/doc/Language/modules.pod6
@@ -701,6 +701,9 @@ module directly from the module folder you just created.
 zef install ./your-module-folder
 =end code
 
+Be advised that you must uninstall and install the given module for the latest
+changes to take place.
+
 =end item
 
 =head2 Upload your Module to CPAN


zoffixznet:
:+1: Thanks!

ugexe:
This isn't entirely true. You can force install ( part of the CompUnit::Repository api ), change the auth, or bump the version or api.

zoffixznet:
> After struggling for half an hour with a change to a local module that wasn't being reflected,

What was the problem, actually? Doesn't zef say "already installed" and refuses to install again?

uzluisf:
@ugexe Thanks for the advice. I'll look into the CompUnit::Repository api.

@zoffixznet 
It was more of a problem from my lack of knowledge. The thing is that I installed the local module, made some changes to it afterward but when I proceeded to use it, none of those changes were reflected. I decided to uninstall and then install the module again with the latest changes. However, that doesn't seem to be the most appropriate course of actions as per @ugexe's suggestion.

zoffixznet:
I [reworded the text](https://github.com/perl6/doc/commit/2dbf66ef11f52ff6871853655ebe2ce00a9b61ec) to address both ugexes comments and to reduce the chances of the problem that happened to you happening to users.

uzluisf:
That addresses my concern. Thanks!

Issue #650, opened 2016-06-29
Squished table cells on mobile


zoffixznet:
On page https://docs.perl6.org/type-exceptions.html

Last column is squished and can't scroll sideways:
![screenshot_2016-06-29-17-10-57](https://cloud.githubusercontent.com/assets/5747918/16468993/aa502552-3e1c-11e6-9978-64e0855b0684.png)

JJ:
Still bad...

MorayJ:
Fixed with #2260

PR #2258, opened 2018-08-13
Make pod-tables scroll horizontally on smaller views

MorayJ:
If a table is too wide for a page a horizontal scroll bar will be added to the bottom.

If the page is focused, the right key will scroll right and left to come back.

I'm presuming this will work on mobile with finger dragging, but no testing set up for that at the moment.

## The problem
Squished table cells on mobile #650 

Table columns that are outside the viewport are not viewable. Problem on mobile. This occurs in the bug report and is also in the 'Routines' page, directly linkable from the front page of the site making the bug quite prominent.

## Solution provided

Add css via sass generator following advice from someone's question on stackoverflow (https://stackoverflow.com/questions/5533636/add-horizontal-scrollbar-to-html-table) on best way for creating scrollable table.

Requested reviews as not sure if other solutions in the offing and as a site wide change with potential repercussions, would like it viewed.

### DIFF
--- a/assets/sass/style.scss
+++ b/assets/sass/style.scss
@@ -74,6 +74,12 @@ svg {
     height: auto;
 }
 
+table.pod-table {
+    display: block;
+    overflow-x: auto;
+    white-space: nowrap;
+}
+
 table.pod-table tr:nth-child(odd) {
     background-color: rgba(0, 0, 0, 0.031373);
 }


AlexDaniel:
FWIW currently github does not link your commits to your github account. You can simply add the email address that you have in git user.email to the list on https://github.com/settings/emails and it will pick it up. Also, feel free to add yourself to https://github.com/perl6/doc/blob/master/CREDITS so that your preferred name is used in rakudo announcements.

Issue #2146, opened 2018-07-04
Inconsistent capitalization and use of period in the "Language" page


uzluisf:
Just like issue #1945, some section descriptions in the [Language](https://docs.perl6.org/language.html) page seem to have inconsistent casing and use of periods.

Examples:

* Perl 6 Terms
* Unicode Support in Perl 6
* What are contexts and how to get into them.

Unlike 'Functional Programming' or 'Object Orientation'  in other descriptions, I'm not sure if 
'Unicode Support' and 'Perl 6 Terms' are being used as proper noun here though.

Regarding the period, most of the descriptions don't have it.

JJ:
Inconsistent capitalization is everywhere: #1945. I'm all for using capital only in the first letter.

AlexDaniel:
> Inconsistent capitalization is everywhere: #1945. I'm all for using capital only in the first letter.

Yes, use sentence case. When in doubt, just do whatever [wikipedia does](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Capital_letters#Headings,_headers,_and_captions). What's good for them can't be harmful for us. Maybe even add that idea to the styleguide.

tbrowder:
Closed by several individual commits.

JJ:
I'm sorry, I'm opening this back. Although it's been done as requested, I have noticed that now many links to anchors are no longer correct, since they kept the former capitalization. I would you ask to please go back to those changed, and check them extensively. In general, any link-to-anchor with capitals is now suspect.

MorayJ:
I'm going to close this as think it's done with the merge. The hrefs are quite odd generally and the links too (in terms of consistency of formation, I think) so will maybe have a look into it (i.e. not consistent whether the anchor links to the id/name or a special class).

Altai-man:
Please also do `make test` as well as `make xtest` to make sure tests were not broken by the merge. I see some issues with travis.

JJ:
Yep. If I know my sh*t, I would say it's related to the placement of an X indexing mark. I'll try to check it out.

JJ:
And while I'm at it, @MorayJ , great job, thanks! Thanks also @Altai-man for checking everything here.

PR #2273, opened 2018-08-21
Anchor case

MorayJ:
## The problem
Broken anchors with headings having changed case #2146

## Solution provided

Changed links to anchors so that they fit the new headings.


Altai-man:
Are those the only fallout after heading changes?

MorayJ:
They're the only ones that I've found...I've tried to be thorough

Altai-man:
No issues with your PR, just thought is there more broken links coming. Thanks.

PR #2274, opened 2018-08-22
Update control.pod6

kjpye:
## The problem

The two variables in the final example for "for" don't match.

## Solution provided

Change first variable name to match the second - -and the comment.


ZzZombo:
I don't really see the point of the change as given, because it promotes a bogus practice of declaring the same variable twice in a single scope. I agree that the code should be rather changed to print the `@b` variable.

ZzZombo:
Whoops, I'm sorry about this mess. BTW, how is it even possible somebody else but the OP managed to edit the PR directly?

JJ:
I guess you can if you have the privs...

Issue #2232, opened 2018-07-31
TODOs in typesystem


JJ:
## The problem

Several `TODO`s in [`typesystem.pod6`](https://github.com/perl6/doc/blob/master/doc/Language/typesystem.pod6)

* [ ] `module`
* [ ] `package`
* [ ] `grammar`

## Suggestions

They are explained elsewhere in the documentation. A small intro and link to corresponding language or type doc might be more than enough.

PR #2285, opened 2018-08-29
Types

MorayJ:
## The problem

On the type system page there are TODOs to be replaced by text explaining Modules, Packages, and Grammars. #2232

## Solution provided

Have put in a short explanation and a link to the page specifically about that subject.


lizmat:
Thank you!

PR #2299, opened 2018-09-07
Write networking API section for JS nutshell

Kaiepi:
## The problem
This section of the page was a TODO.

## Solution provided
I wrote what needed to be documented.

### DIFF
--- a/doc/Language/js-nutshell.pod6
+++ b/doc/Language/js-nutshell.pod6
@@ -393,9 +393,110 @@ say ++$foo; # 0
 
 =head1 The networking API
 
-=comment TODO
+=head2 Net
 
-# TBD
+In Perl 6, there are two APIs for dealing with networking: C<IO::Socket::INET>
+(for synchronous networking), and  C<IO::Socket::Async> (for asynchronous
+networking).
+
+C<IO::Socket::INET> currently only supports TCP connections. Its API resembles
+that of C's socket API. If you're familiar with that, then it won't take long
+to understand how to use it. For example, here's an echo server that closes the
+connection after receiving its first message:
+
+=begin code
+my IO::Socket::INET $server .= new:
+    :localhost<localhost>,
+    :localport<8000>,
+    :listen;
+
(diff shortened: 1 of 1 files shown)


doc/Language/js-nutshell.pod6
```diff
@@ -393,9 +393,100 @@ say ++$foo; # 0
...
+=head2 Net
 
-# TBD
+In Perl 6, there are two APIs for dealing with networking: C<IO::Socket::INET>
+(for synchronous networking), and  C<IO::Socket::Async> (for asynchronous
+networking).
+
+C<IO::Socket::INET> currently only supports TCP connections. Its API resembles
+that of C's socket API. If you're familiar with that, then it won't take long
+to understand how to use it. For example, here's an example of an echo server
```
Tyil: "example" being used twice in one sentence reads a little odd, you might want to rephrase it to "For example, here's how to create an echo server that closes after receiving one message".

doc/Language/js-nutshell.pod6
```diff
@@ -393,9 +393,100 @@ say ++$foo; # 0
...
+=end code
+
+=head2 HTTP/HTTPS
+
+Perl 6 doesn't natively support HTTP/HTTPS. However, CPAN packages such as
+L<Cro|https://cro.services/> help fill the gap.
+
+=head2 DNS
+
+Perl 6 does not support DNS currently, though there are some CPAN modules in
```
Tyil: This makes it read as if I can't use a domain name as argument to `connect`. I don't use the sockets myself, so I'm not sure if this is actually true, but generally DNS lookups are handled by the OS and should Just Work. Is it indeed not possible to use something like `IO::Socket::Async.connect('tyil.nl', 8000);`?

Kaiepi: It is possible. What I was trying to say is that most of the features of Node's DNS module, like record lookups and reverse IP lookups, aren't yet implemented. I'll reword it

Kaiepi:
Updated

PR #2311, opened 2018-09-15
Fix broken link

dmc00:


lizmat:
The link appears to be correct on docs.perl6.org ?

dmc00:
Hmm.  For me, the link leads to "https://docs.perl6.org/routine/type$SOLIDUSValueObjAt" which is a 404.

zoffixznet:
Broken for me too and the name of the type is wrong. The PR will indeed fix the issue and can even be shortened to just `L<ValueObjAt>` without any URL.

@dmc00 I sent you an invite to **perl6** GitHub org. You can accept that invite here https://github.com/perl6/

That gives you access to most of our repos and once you accept the invite you should be able to merge this PR yourself and commit future changes directly.

dmc00:
Cool, thanks for the commit bit!

Issue #2308, opened 2018-09-14
Methods, routines, subs misnamed on Cool page


MorayJ:
## The problem
In https://docs.perl6.org/type/Cool some routines are titled method when they have a sub as well (eg method exp). This should be a routine as it is not specifically a method. There are others.

Moritz confirmed on #perl6 "Routine is the parent class of Sub and Method" and that if only a method, it should be **method _name_**, if only a sub it should be **sub _name_** and if both **routine _name_**

## Suggestions
Edit to reflect the above convention.

MorayJ:
Potentially a bigger issue here... when I look at https://docs.perl6.org/type/Complex.html, it lists all the routines under 'Methods' even though that list contains a sub.

Some classes have a header 'Methods' because they're all methods. But what should the layout be for consistency? Everything under 'Routines' with just 'method name' or 'sub name' or 'routine name'. Or separate headers for each -> Methods, Subs, Routines?

JJ:
Um, tricky one. Since we are talking about a class, in general "methods" is used. I don't know if there's some generation magic involved, there might be, but in any case, changing it to "routines" if there are some subs would make everything a bit more confusing. While in general I agree with you that we should take care a little bit more about this, I'm worried about the implications.

MorayJ:
I'm not sure it's consistent as it is, or I'm not clear on why there are routines and subs when there are...

jnthn:
> if only a method, it should be method name, if only a sub it should be sub name and if both routine name

Tread carefully here: there are numerous cases where things exist as both a `method` and a `sub`, but have entirely different signatures. Take `map`, for instance, which:

* As a `method` takes the first (invocant) argument as the data and the second argument as the mapper
* As a `sub` takes a first argument which is the mapper and then slurps the rest of the arguments as the data

In many other cases, the `sub` is simply dispatching to the method, rather than doing the operation itself.

MorayJ:
So, in the case of a) Method and sub, entirely different signatures:
Is that a sub and a method documented separately? Or does it have an effect on if that's called a method or a routine?

In b) The sub is dispatching to a method:
Is that a method or a routine? They seem to be routines for the Cool class page (link below)

An example I am struggling with on Cool https://docs.perl6.org/type/Cool:
Method uniprops is only defined as:
`sub uniprops(Str:D $str, Stringy:D $propname = "General_Category")`
I would want to call that sub uniprops.

[...]

PR #2318, opened 2018-09-18
Changes routine headlines

MorayJ:
If a routine is only defined as sub, change to 'sub name', if only by
method to 'method name', and if by both to 'routine name' - where this was
not already the case.

## The problem
Methods, routines, subs misnamed on Cool page #2308

## Solution provided
Made changes


zoffixznet:
:+1: Thanks!

PR #2325, opened 2018-09-23
Closes #2321

### LINKED ISSUE #2321: Incorrect description of smiley constraints
This heading https://docs.perl6.org/type/Signature#Constraining_defined_and_undefined_values reads `Constraining defined and undefined values`. However, that's not what `:D`/`:U` type constraints constrain for.

They constrain for [definiteness](https://docs.perl6.org/language/classtut#index-entry-DEFINITE) instead of definedness. That is type objects vs. instances. An instance may be undefined (e.g. `Failure.new`), but it fit into a `:U` constraint.

I haven't read the section's body, but the title along with any similar errors in the body, if any present, need to be corrected.

threadless-screw:
## The problem
Incorrect heading for section describing smiley constraints :U/:D.

## Solution provided
Heading corrected, and added a brief clarification of the difference between "definiteness" and "definedness". Clarification based on: https://stackoverflow.com/questions/48986414/whats-the-difference-between-perl-6s-definite-and-defined-methods

### DIFF
--- a/doc/Type/Signature.pod6
+++ b/doc/Type/Signature.pod6
@@ -396,7 +396,7 @@ be confusing, since the terms have subtly different meanings.
 As explained above, I<definiteness> is concerned with the distinction between 
 type objects and object instances. A type object is always indefinite, while an
 object instance is always definite. Whether or not an object is a type 
-object/definite or an object instance/indefinite can be verified using the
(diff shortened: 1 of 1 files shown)


zoffixznet:

Issue #1753, opened 2018-02-02
You can have colonpairs in identifiers. What's the justification?


AlexDaniel:
From the [docs](https://docs.perl6.org/language/syntax#Identifiers):

> Identifiers can contain colon pairs. The entire colon pair becomes part of the name of the identifier.

Here is a nicer example of that:
```perl6
my $foo:bar<2> = 5;
say $foo:bar(1+1) # OUTPUT: «5␤»
```

… but why?

This is probably some material for [faq-why](https://github.com/perl6/doc/issues/114) (“Why things are the way they are in Perl 6”).

JJ:
Why they must be pairs? Or why do the use colons at all?

zoffixznet:
> … but why?

We already have two reasons in core:

1) `sub infix:<+>` uses the main name as a category of the op and the colonpair as the symbol for the op.
2) In grammars, we have proto regexes. `token foo:sym<bar>` is one of the variations of `foo` token and the value of `:sym` colonpair can be used inside of it for matching that string.

Since colonpairs are allowed in identifiers for the above cases, it naturally extends to `$foo` also having them allowed.

The Slang-writers can get all the QASTs for colonpairs in `<colonpair>` token part of the `<longname>`

AlexDaniel:
> Why they must be pairs? Or why do the use colons at all?

Why is this allowed at all. FWIW, there are many things I wish we had written justification for, but I was unsure if this should be included in the docs. Now I learned that there were plans for this, so perhaps more tickets to come :)

Anyway, Zoffix++, this explanation is short and very clear, I like it.

TimToady:
[...]

PR #2335, opened 2018-09-28
Rewrite of Identifier-section

threadless-screw:
## The problem
The identifier-section was perceived as a bit messy.

## Solution provided
Clearer distinction between different identifiers types: ordinary identifiers, extended identifiers, and compound identifiers. Particulars for each type can now be discussed in their own section.

### DIFF
--- a/doc/Language/syntax.pod6
+++ b/doc/Language/syntax.pod6
@@ -223,109 +223,193 @@ say 'code again';
 
 =head2 X<Identifiers|identifier,identifiers>
 
-Identifiers are grammatical building blocks that occur in several kind
-of statements, giving a name to objects, data structures and blocks of
-code. These names must start with an alphabetic character (or an
-underscore), followed by zero or more word characters (alphabetic,
-underscore or number). You can also embed dashes C<-> or single quotes
-C<'> in the middle, but not two in a row, and only if followed
-immediately by an alphabetic character.
+Identifiers are grammatical building blocks that may be used to give a name
+to entities/objects such as constants, variables (e.g. Scalars) and routines
+(e.g. Subs and Methods). In a variable name, an identifier is normally preceded
+by a sigil; i.e. the sigil is not part of the identifier, only of the variable
+name.
+
+    constant c = 299792458;         # Identifier "c" names an Int
+    my $a = 123;                    # Identifier "a" in the name "$a" of a Scalar
+    sub hello { say "Hello!" };     # Identifier "hello" names a Sub
+
+Identifiers come in different forms: ordinary identifiers, extended
+identifiers, and compound identifiers.
+
+=head3 Ordinary identifiers
+
+An ordinary identifier is composed of a leading alphabetic character
+that may be followed by one or more alphanumeric characters. It may also
+contain isolated, embedded apostrophes C<'> and/or hyphens C<->, provided
+that the next character is each time alphabetic.
+
+The definitions of "alphabetic" and "alphanumeric" include appropriate Unicode
+characters. Which characters are "appropriate" depends on the implementation.
+In the Rakudo/MoarVM Perl 6 implementation alphabetic characters include
+Unicode characters in the character category I<Letter>, and the underscore
+C<_>. Alphanumeric characters additionally include Unicode characters in the
+general category I<Number, Decimal Digit> (Nd).
 
 =begin code :skip-test
-# valid identifiers:
+# valid ordinary identifiers:
 x
 something-longer
 with-numbers1234
-don't
-fish-food
-π
+don't-do-that
+piece_of_π
+駱駝道                 # "Rakuda-dō", Japanese for "Way of the camel"
 =end code
 
 =begin code :skip-test
-# not valid identifiers:
-with-numbers1234-5
-42
-is-prime?
-fish--food
+# invalid ordinary identifiers:
+42                    # Identifier does not start with alphabetic character
+with-numbers1234-5    # Embedded hyphen not followed by alphabetic character
+is-prime?             # Question mark is not alphanumeric
+x²                    # Superscript 2 has Unicode general category No
 =end code
 
-You use identifiers to name of constants, types (including classes and
-modules) and routines (subs and methods); they also appear in variable
-names usually proceeded by a sigil; see L<variables|/language/variables>
-for more details.
+In the C<x²> example above, the Unicode superscript numeral C<²> is not, and
+cannot be, a part of an ordinary identifier. Still, it may form part of a valid
+expression as an operator. The expression C<$x²>, for instance, produces the
+square of variable C<$x>. This should be somewhat surprising at this point,
+since operators are L<Subs|/type/Sub>, and, as mentioned above, Subs are named using
+identifiers. So how does the superscript numeral get accepted as (part of) an
+identifier? The key to this riddle lies in the concept of I<extended
+identifiers>, which is elaborated upon in the next section.
 
-Namespaces are provided by L<packages|/language/packages>. By separating
-identifiers with double colons, the right most name is inserted into
-existing or automatically created packages.
+=head3 Extended identifiers
 
-    my Int $Foo::Bar::buzz = 42;
-    say $Foo::Bar::buzz; # OUTPUT: «42␤»
+It is often convenient to have names that contain arbitrary characters.
+Use cases include situations where a set of entities shares a common "short"
+name, but still needs for each of its elements to be identifiable individually.
+For example, you might use a module whose short name is C<ThatModule>, but the
+complete long name of a module includes its version, naming authority, and
+perhaps even its source language. Similarly, sets of operators work together in
+various syntactic categories with names like C<prefix>, C<infix>, C<postfix>,
+etc. The (long) names of these operators, however, often contain characters that
+are excluded from ordinary identifiers.
 
-Identifiers can contain colon pairs. The entire colon pair becomes part
-of the name of the identifier.
+For all such uses, you can append one or more
+L<adverbial pairs|/language/syntax#Adverbial_pairs_(colon_pairs)> to an ordinary
+identifier to create a so-called I<extended identifier>. When appended
+to an identifier (that is, in postfix position), the adverbial pair syntax
+generates unique variants of that identifier.
 
-    my $foo::bar = 1;
-    say MY::.keys;
-    # OUTPUT: «($=pod !UNIT_MARKER EXPORT $_ $! ::?PACKAGE
-    #          GLOBALish $¢ $=finish $/ $foo:bar $foo:bar<2>
-    #          $?PACKAGE)␤»
-    say OUR::.key; # OUTPUT: «foo␤»
+The generic form of an adverbial pair or "colon pair" is C<:key<value>> :
+it starts with a single colon C<:>, followed by an ordinary identifier C<key>,
+in turn followed by some quoting bracketing construct, such as C«< >», C<« »> or
+C<[' ']>, which quotes one or more arbitrary characters C<value>:
 
-The last sentence shows how the C<foo> package has been created
-automatically, as a deposit for variables in that namespace.
+=begin code :skip-test
+# exemplary extended identifiers:
+infix:<+>                           # The official long name of the operator in $a + $b
+postfix:<²>                         # The official long name of the operator in $x²
+WOW:That's<<🆒>>
+ThatModule:auth<Somebody>:ver<2.7.18>
+=end code
 
-Identifiers can also contain single colons, and in that case they can
-add I<symbols> within angle brackets. This is a mechanism that supports
-interpolation; which is also supported by colonpairs if the interpolated
-variable is inserted in parentheses. Please note that resolution of
-names often happens at compile time, so interpolation values must be
-known at compile time.
+The C<key> I<or> the quoted C<value> is optional. Colon pairs with a null key
+C<key> are normally used for naming operators, as in the first two examples
+above. Starting with Perl 6 language version 6.d, colon pairs with C<sym> as
+the C<key> (e.g. C«:sym<foo>») are reserved for possible future use.
 
-    my $foo:bar<2> = 2;
-    constant $c = 42;
-    my $a:foo<42> = "answer";
-    say $a:foo«$c»;    # OUTPUT: «answer␤»
-    my $buz = "qux";
-    my $bur::quux = 7;
-    say $bur::($buz);  # OUTPUT: «7␤»
+In an extended identifier, the adverb is considered an integral part of the name,
+so C<infix:<+>> and C<infix:<->> are two different operators. The bracketing
+characters used, however, do not count as part of the name; only the quoted data
+matters. So these are all the same name:
 
-The first statement is an example of I<X<extended identifiers>>. Extended
-identifiers are used extensively in the definition of the language, for
-instance, to define L<terms|language/syntax#Term_term:%3C%3E>. They include a
-colon C<:> in their name, and might include a bracketing-quoting construct such
-as C<«»>, C«<>» or C<['']>. These bracketing constructs are interchangeable. For
-instance
+    infix:<+>
+    infix:<<+>>
+    infix:«+»
+    infix:['+']
+    infix:('+')
+
+Similarly, all of this works:
 
     my $foo:bar<baz> = 'quux';
-    say $foo:bar«baz»; # OUTPUT: «quux␤»
+    say $foo:bar«baz»;                                  # OUTPUT: «quux␤»
     my $take-me:<home> = 'When the glory has no end';
-    say $take-me:['home'];# OUTPUT: «When the glory has no end␤»
+    say $take-me:['home'];                              # OUTPUT: «When the glory has no end␤»
+    my $foo:bar<2> = 5;
+    say $foo:bar(1+1);                                  # OUTPUT: «5␤»
 
-This extended syntax is used, for intance, for module versions and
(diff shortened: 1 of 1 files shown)


doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
-code. These names must start with an alphabetic character (or an
-underscore), followed by zero or more word characters (alphabetic,
-underscore or number). You can also embed dashes C<-> or single quotes
-C<'> in the middle, but not two in a row, and only if followed
-immediately by an alphabetic character.
+Identifiers are grammatical building blocks that may be used to give a name
+to entities/objects such as constants, variables (e.g. Scalars) and routines
+(e.g. Subs and Methods). In a variable name, an identifier is normally preceded
+by a sigil; i.e. the sigil is not part of the identifier, only of the variable
+name.
```
JJ: As long as you're making a precision here, it might be interesting to include sigilless containers and how they lose the \ when they are used.

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+that may be followed by one or more alphanumeric characters. It may also
+contain isolated, embedded apostrophes C<'> and/or hyphens C<->, provided
+that the next character is each time alphabetic.
+
+The definitions of "alphabetic" and "alphanumeric" include appropriate Unicode
+characters. Which characters are "appropriate" depends on the implementation.
+In the Rakudo/MoarVM Perl 6 implementation alphabetic characters include
+Unicode characters in the character category I<Letter>, and the underscore
+C<_>. Alphanumeric characters additionally include Unicode characters in the
+general category I<Number, Decimal Digit> (Nd).
```
JJ: Maybe just saying above that _ is considered alphanumeric is enough. That is not a general consideration, however. It's generally considered punctuation https://stackoverflow.com/questions/8509262/will-alphanumeric-contain-and-space

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+is-prime?             # Question mark is not alphanumeric
+x²                    # Superscript 2 has Unicode general category No
 =end code
 
-You use identifiers to name of constants, types (including classes and
-modules) and routines (subs and methods); they also appear in variable
-names usually proceeded by a sigil; see L<variables|/language/variables>
-for more details.
+In the C<x²> example above, the Unicode superscript numeral C<²> is not, and
+cannot be, a part of an ordinary identifier. Still, it may form part of a valid
```
JJ:
```perl6
say '²' ~~ /<alnum>/ # Nil
```

So it would be simpler to just say it's not alphanumeric in the Unicode sense as explained above.

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+square of variable C<$x>. This should be somewhat surprising at this point,
+since operators are L<Subs|/type/Sub>, and, as mentioned above, Subs are named using
+identifiers. So how does the superscript numeral get accepted as (part of) an
+identifier? The key to this riddle lies in the concept of I<extended
+identifiers>, which is elaborated upon in the next section.
+
+=head3 Extended identifiers
+
+It is often convenient to have names that contain arbitrary characters.
+Use cases include situations where a set of entities shares a common "short"
```
JJ: Clarify if those characters are really arbitrary or must follow some rules.

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+name, but still needs for each of its elements to be identifiable individually.
+For example, you might use a module whose short name is C<ThatModule>, but the
+complete long name of a module includes its version, naming authority, and
+perhaps even its source language. Similarly, sets of operators work together in
+various syntactic categories with names like C<prefix>, C<infix>, C<postfix>,
+etc. The (long) names of these operators, however, often contain characters that
+are excluded from ordinary identifiers.
```
JJ: Instead  of long, which is imprecise, and maybe even wrong (² is not long), express what you mean. As a matter of fact, it might be the other way round: the identifier you declare those operators with include ordinary characters...

JJ: OK, I see this is taken from the synopse. It's hard to say it's wrong then, but, well, maybe it can be expressed in some other way (also: if you take paragraphs from the synopse it would be better if you quote it explicitly)

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+
+For all such uses, you can append one or more
+L<adverbial pairs|/language/syntax#Adverbial_pairs_(colon_pairs)> to an ordinary
+identifier to create a so-called I<extended identifier>. When appended
+to an identifier (that is, in postfix position), the adverbial pair syntax
+generates unique variants of that identifier.
+
```
JJ: Not sure about this. Advervial pairs have this type of syntax: :foo<bar>. The syntax you are describing here is rather foo:<bar>. Everything with a colon needn't be an adverbial pair, and I don't know if you're gaining anything by describing it as such. The synopsis talks about an adverbial "form", which I guess describes anything that follows a colon. It's not a pair, since, well, there's not a pair of things. Those [adverbial forms are also in the glossary under "Adverb"](https://docs.perl6.org/language/glossary#Adverb)

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+The generic form of an adverbial pair or "colon pair" is C<:key<value>> :
+it starts with a single colon C<:>, followed by an ordinary identifier C<key>,
+in turn followed by some quoting bracketing construct, such as C«< >», C<« »> or
+C<[' ']>, which quotes one or more arbitrary characters C<value>:
```
JJ: This is really unnecesary, since, as said above, it's not an adverbial pair, but an adverbial form.

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+postfix:<²>                         # The official long name of the operator in $x²
+WOW:That's<<🆒>>
+ThatModule:auth<Somebody>:ver<2.7.18>
+=end code
 
-    my Int $Foo::Bar::buzz = 42;
-    say $Foo::Bar::buzz; # OUTPUT: «42␤»
+The C<key> I<or> the quoted C<value> is optional. Colon pairs with a null key
+C<key> are normally used for naming operators, as in the first two examples
+above. Starting with Perl 6 language version 6.d, colon pairs with C<sym> as
```
JJ: This should be a note.

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
 
-    constant $what= 'are';
+Furthermore, the adverbial pair syntax in extended identifiers supports
+compile-time interpolation, which mandates the use of
+L<constants|/language/terms#Constants> for the interpolation values:
+
+    constant $c = 42;               # Constant binds to Int; $-sigil enables interpolation
+    my $a:foo<42> = "answer";
+    say $a:foo«$c»;                 # OUTPUT: «answer␤»
+
```
JJ: I know this is exactly how it was in the original, but maybe, and as long as we're trying to establish some best practice for identifiers here, it would be better to define the constant as \c to differentiate it from run-time variables.

doc/Language/syntax.pod6
```diff
@@ -223,109 +223,193 @@ say 'code again';
...
+    say @we:<$what>;                # Compilation error: Variable '@we:<$what>' is not declared
+
+=head3 Compound identifiers
+
+A compound identifier is an identifier that is composed of two or more
+ordinary and/or extended identifiers that are separated from one another by a
+double colon C<::>. Note that the C<::> is referred to as a I<double colon>,
+and not as a I<colon pair>. The latter term refers to the adverbial pair syntax
+C<:key<value>> that may be used to create a Pair, or, as described above, to
+create an extended identifier.
```
JJ: That's an adverbial pair, not a colon pair. Colon pairs include other constructs such as :a.

doc/Language/syntax.pod6
threadless-screw: It is my understanding that in Perl6 the underscore is considered "alphabetic", and thus also "alphanumeric", despite its Unicode classification al punctuation. Hence its special mention in the text above.

 say '_' ~~ /<alpha>/
｢_｣
 alpha => ｢_｣

The rest of the text clarifies precisely what Unicode code points are acceptable, which I'd say is useful information:
https://stackoverflow.com/questions/34689850/whats-allowed-in-a-perl-6-identifier

doc/Language/syntax.pod6
threadless-screw: The clarifying remark was prompted by this post: https://stackoverflow.com/questions/34689850/whats-allowed-in-a-perl-6-identifier, and the current text of the section which merely says "Unicode superscript numerals are exponents and not part of an identifier; e.g. $x² does the square of variable $x. Subscript numerals are TBD."

The proposed text clarifies that the superscript numeral is actually part of an identifier (it has to be: it is the name of a Sub), but not of an ordinary identifier, but rather of an extended identifier.

doc/Language/syntax.pod6
threadless-screw: I would love to clarify this in all detail, but I don't have the info, and I'm not sure how to find
out other than by asking people in the know (I'm not exactly at home in the Perl6 source code repo). Note that the term "arbitrary characters" was taken from the synopses. I have now amended the proposed text to avoid the term. The text sounds cleaner now, but contains no more information.

doc/Language/syntax.pod6
threadless-screw: I tried to elucidate the text by including examples.

doc/Language/syntax.pod6
JJ: It's not a matter of documenting "dubious" practices. The problem is the use of the word normally. If you are focusing on identifiers, you can simply say that when there's a sigil, it's not a part of the identifier, without saying if that's the "normal" practice or not.

doc/Language/syntax.pod6
threadless-screw: I adapted the proposed text to stay a bit closer to the synopses, and to clarify some differences between colon pairs allowed in identifiers (and, implicitly, colon pairs used for Pair construction).  I'm not entirely sure what you are saying though. The text does not speak of `foo:` ; it talks about "appending" rather than "prepending".

doc/Language/syntax.pod6
JJ: Yes it is, and it's not general, that's why just saying that _ is considered alphanumeric in Perl 6 is enough.

doc/Language/syntax.pod6
JJ: It's an identifier, but it's not part of _that_ identifier. The previous text was probably LTA also.
It's probably better to clarify that `x²`, together, do not form a single indentifier, but that they are identifiers separately, with a forward call to the extended syntax.

doc/Language/syntax.pod6
JJ: Just say characters "others than the ones in the regular identifiers". That's a precise definition, but it's not exhaustive, as "arbitrary", which implies you can use any character (smileys, anyone?)

doc/Language/syntax.pod6
JJ: The synopse is not the final say in the *description* of the function; that goes to the roast test; some vocabulary may have drifted from the original. In general, why give a name to that construct? It's clear enough if you describe it.

doc/Language/syntax.pod6
threadless-screw: I don't understand what you are trying to say. In all honesty, I have that problem with some other comments of yours as well. I appreciate that you take the time to review the PR, but I think we could move faster if you'd be a bit more explicit about what you have in mind.

I'm currently guessing that the only thing you want the text to say it that the underscore is alphanumeric (which I think should be: alphabetic), and hence that the other information should be deleted because it is not "general" in the sense that in does not relate to the Perl6 language, but to an implementation thereof. Is that correct?

If so: I would disagree. This information is pure gold. I am a newbie to Perl6 but have experience with other programming languages (other than Perl). Never have I experienced such a high barrier to entry as with Perl6. This is largely because clear, practical information is oftentimes difficult to find. A question that I think should be answered in every Perl6 101 is which characters are acceptable as part of an identifier. Yet, I have found no documentation that documents this in complete detail. Not in the current docs, not in the synopses, not in Think Perl 6 (Rosenfeld), not in Perl6 Fundamentals (Lenz) and not in Learning Perl6 (Foy). I'd say that's ridiculous. If I look up the Python documentation, I find this: https://docs.python.org/3/reference/lexical_analysis.html#identifiers. Holy Moly - what do you know: a clear answer! 

Why withhold this information, especially having regard to the fact that one of Perl6' flagship features is all-round Unicode support?

doc/Language/syntax.pod6
threadless-screw: Can you provide a working example? I can't get the interpolation to work with a sigilless "variable"; i.e.

 constant \c = 42;
 my $a:foo<42> = "answer";
 say $a:foo«c»;

does not compile. I assume the sigil is necessary to perform the interpolation, which is one of the reasons for having sigils in the first place, so that makes sense. It is also why the proposed text (in the example) states that the $-sigil enables interpolation.

doc/Language/syntax.pod6
threadless-screw: I don't understand what you are saying.

doc/Language/syntax.pod6
threadless-screw: Already done in the latest commit.

doc/Language/syntax.pod6
JJ: I'm sorry. I try to, but sometimes it's difficult to say precisely what I mean. Please consider them suggestions, and at the end of the day, none of them will break anything; if something is LTA, we can eventually fix it when merged. But let's get to the point. I kind of try and make the point that anything that is considered alphanumeric *in Perl 6* is OK as an indentifier. Plus `_` is also considered alphanumeric in Perl 6 (not elsewhere). The best way to express that is up to you; maybe the first sentence is already OK in the text, and then a note about `_`.

threadless-screw: Well, everything that is considered alphanumeric in Perl6 may be OK as an identifier, but not for every position in the identifier. For instance, as a(n unquoted) part of the proposed text states, an ordinary identifier should start with an 'alphabetic' character (which may be an underscore, because the underscore is considered alphabetic, not just alphanumeric). This is why it is important to make a distinction between alphabetic and alphanumeric characters, and why I think that the text as currently proposed is concise and precise (to the best of my knowledge anyway).

JJ: That's correct, but it does not belong here, where you're only talking about _ as alphanumeric.

JJ: Anyway, if you think it's OK I'll just resolve this.

doc/Language/syntax.pod6
threadless-screw: OK, I see what you mean now. I suppose `In a L<variable name|/language/variables>, any sigil
(and twigil) precedes the identifier and does not form a part thereof.` would work?

doc/Language/syntax.pod6
threadless-screw: I'm sorry; I'm trying hard to see where the flaw is in the text, or how you are reading it so the flaw comes out, but I don't see it. In particular the text never explicitly states that the underscore _ is alphanumeric. It first says it is *alphabetic*, and then implies - through the phrase "Alphanumeric characters *additionally* include..." that is is also alphanumeric. But I need to know it is alphabetic too, otherwise I don't know I can use it in certain positions in identifiers.

Perhaps it's better to resolve this conversation.

doc/Language/syntax.pod6
JJ: :+1:

doc/Language/syntax.pod6
JJ: :+1:

doc/Language/syntax.pod6
threadless-screw: I just deleted the paragraph.

doc/Language/syntax.pod6
threadless-screw: I'm also not super sure about the terminology. In this [thread](https://github.com/perl6/doc/issues/1753), at least Zoffixznet and TimToady seem comfortable with the term "colon pair" which, to my mind, makes sense as a name for the syntax, even though no actual Pair is constructed.

doc/Language/syntax.pod6
threadless-screw: Done.

doc/Language/syntax.pod6
JJ: Simply say that :: is a double colon. The rest of the paragraph do not help clarify that fact.

doc/Language/syntax.pod6
threadless-screw: "Adverbial pair" terminology removed.

doc/Language/syntax.pod6
threadless-screw: Done.

doc/Language/syntax.pod6
JJ: I guess just describing the syntax without making any reference to how it be called might be the best bet.

PR #2352, opened 2018-10-04
Mention phasers a bit more explicitly in the Control page

amire80:
## The problem
The Phasers were mentioned only in passing, and NEXT and LAST weren't mentioned at all.

## Solution provided
This patch mentioned the concept of phasers explicitly immediately after discussing blocks, and mentions the two common phasers NEXT and LAST when discussing the relevant tokens.

I think that it's my first pull request here, please be gentle :)

### DIFF
--- a/doc/Language/control.pod6
+++ b/doc/Language/control.pod6
@@ -85,6 +85,12 @@ You have to watch out for this in most languages anyway to prevent things
 from getting accidentally commented out.  Many of the examples below may
 have unnecessary semicolons for clarity.
 
+=head1 X<Phasers|control flow,Phasers>
+
+Blocks may have phasers: special labeled blocks that break their execution
+into phases that run in particular phases. See the page
+L<phasers|/language/phasers> for the details.
+
 =head1 X<do|control flow,do>
 
 The simplest way to run a block where it cannot be a stand-alone statement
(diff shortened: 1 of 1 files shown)


doc/Language/control.pod6
```diff
@@ -927,6 +951,25 @@ for @x -> $x {
 
 prints "12".
 
+If the C<LAST> L<phaser|/language/phasers> is present, it runs:
```
JJ: It runs when? No problem if you repeat part of the statement before the example.
Also, I'm not sure it belongs here. As long as you have introduced a new section, wouldn't it be better if you put it there?

doc/Language/control.pod6
```diff
@@ -912,6 +918,24 @@ for @x -> $x {
 
 prints "1245".
 
+If the C<NEXT> L<phaser|/language/phasers> is present, it runs before the next iteration:
+
```
JJ: NEXT is also here https://docs.perl6.org/syntax/%20NEXT Maybe link for further explanation. Same goes to LAST (although there's not much to see there)

amire80: Thanks for the pointer. If I understand correctly, it's the same as the big page on phasers, so I just linked to the section there.

doc/Language/control.pod6
```diff
@@ -927,6 +951,26 @@ for @x -> $x {
 
 prints "12".
 
+If the L<C<LAST> phaser|/language/phasers#LAST> is present, it runs
+before exiting the loop:
```
JJ: Still not sure about the placement of these examples, within the section that talks about the command with the same (lowercase) name. Is there some reason you haven't placed them within the new `Phasers` section you have created?

amire80: Yes, there is a reason: next/last are closely related to NEXT/LAST, and should be mentioned together. Without mentioning NEXT/LAST in the same place the reader may think that they always go to the next iteration or exit the loop immediately, and this is not the case. It's also similar to how "return" is already documented in this same document: "The sub return will stop execution of a subroutine or method, run all relevant phasers and provide the given return value to the caller."

JJ:
Thanks for the great work!

PR #2358, opened 2018-10-05
Document the Setty ops for MixHashes per TODO

Coleoid:
Also added code examples to a note in BagHash.pod6 about .reverse
and included that note in MixHash.pod6.
Also a few grammar/accuracy touches.

## The problem

The documentation was incomplete, waiting on the Setty operator rework.

## Solution provided

Per the todo, I used BagHash.pod6 as the template.
Along the way, running the code samples, I found that the output was now 'Mix' and 'Bag' instead of 'mix' and 'bag', so fixed.
I thought the note on C<reverse> was excellent, so I added samples to it, and moved that also into MixHash.pod6.


Coleoid:
Er, the "note on C" was me trying to use Pod in my comment--intended to be "note on reverse" with 'reverse' tagged as code.

PR #2361, opened 2018-10-06
Add a missing comma

shmup:
## The problem
A missing comma in a META6.json example

## Solution provided
The comma was added, validating the JSON


AlexDaniel:
Nice! Thanks!

AlexDaniel:
@shmup by the way, I've sent you an invitation to perl6 org, you should be able to accept it here: https://github.com/perl6
This way you can commit your further changes directly if you want.

PR #2362, opened 2018-10-06
Corrects some imprecisions in the "Doing math with Perl 6" article

cronos2:
## The problem

The cited article contains a few mathematical imprecisions which ought to be addressed. One of them, about sequences and the real numbers, is particularly troublesome since there seemingly is no justification for this claim and doesn't really add much value to the rest of the article. 

## Solution provided

* Disclaimer about the purported mathematical field nature of Rat, Num
  and Complex
* Rephrases the Int section to be more consistent
* Matrices are not, in general, a field but a ring
* Hamilton's quaternions are not a field, but an algebra over the real
  numbers
* Claiming that ℝ is a sequence makes very little sense from a
  mathematical standpoint. Lacking further explanation/justification,
  the reference should be dropped until it can be given some meaning.

### DIFF
--- a/doc/Language/math.pod6
+++ b/doc/Language/math.pod6
@@ -84,7 +84,7 @@ mathematical notation.
 Perl 6 can do arithmetic using different data types. L<Num>, L<Rat> and
 L<Complex> can all operate as a L<field under the operations of addition, subtraction, multiplication and division|https://en.wikipedia.org/wiki/Field_(mathematics)>
 (technically, it should be noted that data types dealing with floating point number representations are not a field in the mathematical sense due to the inherent
-imprecissions of their arithmetic. However, they constitute an approximate enough, computer friendly version of such mathematical objects for most of the cases).
+imprecisions of their arithmetic. However, they constitute an approximate enough, computer friendly version of such mathematical objects for most of the cases).
 The equivalent mathematical fields are:
 
 =begin table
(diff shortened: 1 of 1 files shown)


doc/Language/math.pod6
```diff
@@ -82,7 +82,9 @@ mathematical notation.
 =head1 Arithmetic
 
 Perl 6 can do arithmetic using different data types. L<Num>, L<Rat> and
-L<Complex> can all operate as a L<field under the operations of addition, subtraction, multiplication and division|https://en.wikipedia.org/wiki/Field_(mathematics)>.
+L<Complex> can all operate as a L<field under the operations of addition, subtraction, multiplication and division|https://en.wikipedia.org/wiki/Field_(mathematics)>
+(technically, it should be noted that data types dealing with floating point number representations are not a field in the mathematical sense due to the inherent
+imprecissions of their arithmetic. However, they constitute an approximate enough, computer friendly version of such mathematical objects for most of the cases).
```
AlexDaniel: `imprecissions` → `imprecisions`?

cronos2: Oops, typo. Let me fix that.

AlexDaniel: By the way, if you run `make xtest` it will run some extra tests (one of the test files will do spell checking).

PR #2363, opened 2018-10-06
Take a small paragraph out of verbatim mode

kensanata:
The paragraph between two code blocks is not itself a code block and should not be indented by four spaces.


AlexDaniel:
Thanks! Invitation to join perl6 org sent. Next time you can commit this kind of stuff directly :)

Issue #2266, opened 2018-08-16
Create a HOWTO for newcomers to the repo


JJ:
## The problem

During the BoF meeting in TPCiG, it was suggested that a how to guide to get into the documentation would be good.

## Suggestions

Create it as a markdown or POD6 page so that everyone can follow it. Link all relevant pieces of knowledge that are needed, from the POD6 specification to how to work with PRs and issues.

finanalyst:
I just wrote a post on perl6 users and suggested this. I have begun to review documentation with the intent of drafting something.

JJ:
Great. It will really help this hacktoberfest, and beyond.

PR #2369, opened 2018-10-08
Language page categories

finanalyst:
## The problems

1. The Language page (`https://docs.perl6.org/language.html`) contains a table of pages in an apparently random order because it depends on the name of the underlying pod6 file.
  - Tom Browder did most of the work to change this, but there was no distinction between the headings and the pages, and there was a link problem because links defined in the original pod files do not correspond to the re-ordered pod file
2.  A brief introduction to the Documentation is added to resolve Issue #2266

## Solution provided
- Tom's work is modified to produce a language page that is based on his Control file.
- The Headings of each section become  level 2 headings on the language page, and the sub-pages are combined into tables.
- The Makefile has been modified to include a new variable USE_CATEGORIES, which is set to False by default
- Only when USE_CATEGORIES is set to True in the Makefile will the Categories be introduced. Thus merging these changes will not affect the current behaviour.
- The mechanism for keeping the section headings and distinguishing between section pods and content pods is by creating a :class<section-heading> in the section pods.
- The mechanism for keeping the old pod filenames is to introduce a :link<$key-fname> into the pod, and to ensure that the url attribute takes the old name.
- The Language page now (if USE_CATEGORIES is set to True) looks much better and is more easily understandable. It also has a TOC.

### DIFF
--- a/util/manage-page-order.p6
+++ b/util/manage-page-order.p6
@@ -188,7 +188,7 @@ sub write-Language-files() {
                     else {
                         #$fh.say: "$line :page-order<{$page-order}>";
                         $fh.say: "# THIS FILE IS AUTO-GENERATED--ALL EDITS WILL BE LOST";
-                        $fh.say: $line;
+                        $fh.say: $line ~ " :link<$key-fname>";
                         #$fh.say: "=comment THIS FILE IS AUTO-GENERATED--ALL EDITS WILL BE LOST";
                     }
                     next;
@@ -450,6 +450,7 @@ sub help {
         -n      - create  (for mode 'control')
         -h      - extended help
         -d      - debug
+        -m    - manage (include the categories)
 
     Note: The modes are selected by entering either the first letter
           of the mode name or its complete name.
--- a/.gitignore
+++ b/.gitignore
@@ -29,4 +29,3 @@ highlights/package-lock.json
 .pod-cache
 retest
 build/
-notesOnInstalling.md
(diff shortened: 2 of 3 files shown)


util/manage-page-order.p6
```diff
@@ -450,6 +450,7 @@ sub help {
         -n      - create  (for mode 'control')
         -h      - extended help
         -d      - debug
+        -m    - manage (include the categories)
```
AlexDaniel: This is not aligned vertically with the rest of the lines

util/manage-page-order.p6
```diff
@@ -51,11 +52,13 @@ given $mode {
...
-        =begin comment
         EXIT("Options 'refresh' and 'new' are not used with 'update'.")
             if ($new || $refresh);
-        =end comment
-        do-update();
+        if $manage {
+            do-update()
+        } else {
+            copy-dir-tree :fromdir('doc'), :todir('build')
+        };
```
AlexDaniel: No need for the semicilon here.

doc/Language/intro.pod6
```diff
@@ -0,0 +1,46 @@
+=begin pod
+
+=TITLE Brief Introduction
+
+=SUBTITLE Using the Perl6 Documentation
```
AlexDaniel: This whole document says “Perl6” when I think we settled on “Perl 6” with no-breaking space. Maybe it is worth changing it, although with recent discussions regarding an alias it may be worth to reword some sentences so that the language name is not spelled out. @JJ what do you think?

AlexDaniel:
@tbrowder [pinged me](http://colabti.org/irclogger/irclogger_log/perl6?date=2018-10-09#l95) on IRC. I looked at the code, seems to be OK. I left some nitpicking comments. This seems not to have blocking downsides that we previously had with tbrowder's work, so I recommend setting it to True by default.

PR #2371, opened 2018-10-09
Add whitespace and reflow

uzluisf:


uzluisf:
A newest commit seems to have taken care of what this commit was trying to fix except for the text reflow. I guess it depends on the keyboard but is there an easy way to type in that character?

JJ:
@uzluisf there's a plugin for Atom that makes it easy, or just the average keyboard input in emacs. There's probably some way in other editors...

coke:
in vim, when in insert mode: control-k space space

uzluisf:
@JJ @coke Thanks for the suggestion! The vim command will do.

PR #2375, opened 2018-10-10
search urls corrected

finanalyst:
## The problem
during the refactor to use categories, a variable $link was used to hold the source url. It was not applied to the code that created definitions and references 

## Solution provided
use the $link for the urls


finanalyst:
Yes. Checked locally. Everything working.

PR #2382, opened 2018-10-13
Add minor fixes, rephrase sentences and reflow text

uzluisf:


zoffixznet:
FWIW: you have a commit bit and can just commit all this stuff directly.

uzluisf:
I know but I'd rather have someone else review it so as to avoid reckless merges.

zoffixznet:
There's always `git revert` 😉

uzluisf:
Thanks, @zoffixznet !

Issue #2292, opened 2018-09-03
whenever can be used in a supply block


frithnanth:
## The problem

In [https://docs.perl6.org/language/concurrency](Concurrency), the `whenever` keyword seems limited to appear only in a `react` block, while it can appear also in a `supply` block.

## Suggestions

Example:

```perl6
my $barman = supply {
  whenever Supply.interval(60) {
    done when * > ∞;
    emit "🍺";
  }
}
$barman.tap: { say "Got: $_" };
await $barman;
```

MorayJ:
Can react be used without 'whenever'? I would probably be looking to have a headline for 'whenever' rather than 'react and whenever', and putting the supply example under that by saying you can also do this using the 'whenever' keyword. And then having just 'react' as a headline and saying that react has to use the 'whenever' keyword, as per the example if that sounds about right...

jnthn:
> Can react be used without 'whenever'?

There's no obvious use I can think of for that, no. For `supply`, however, it *is* useful to use it without a `whenever`, if writing an async data source.

MorayJ:
Hopefully this covers it, though possibly the whole page could do with a general revision, but I think whenever is in there now for supplies, which is important to have done.

Issue #2135, opened 2018-07-02
So what the hell is react/whenever


AlexDaniel:
*Note: Please feel free to reword the title. I'm just frustrated with how little our docs say about such an important feature of perl 6.*

Imagine you stumbled upon react block somewhere in the wild, so you go to the documentation to learn about it.

The search functions gives you just one link: https://docs.perl6.org/language/concurrency#index-entry-react-react

The section starts with:

> This could also be written using the react keyword:

Sorry, what can be written? The previous section takes a full page and talks about supplies and suppliers, and moreover that's not needed at all to understand react (IMO). The basics can be shown with simple promises.

Then it goes with the only paragraph that explains react block:

> Here the whenever keyword uses .act to create a tap on the Supply from the provided block. The react block is exited when done() is called in one of the taps. Using last to exit the block would produce an error indicating that it's not really a loop construct.

Uh… and? So if it's using taps, why use the react block instead of just using .tap? Nothing is explained.

----

If I had to summarize react block in one sentence I would say that it does *sequential processing of asynchronous events*. That, in my opinion, is its main feature. I don't know if this description is fully correct and if it is understandable, but perhaps that's a good start.

Moreover, it should probably discuss how exceptions happening in whenevers/react can be noticed easily, compared to threads dying in regular .taps.

FWIW there's a huge example using react/whenever in https://docs.perl6.org/type/Proc::Async (maybe it can be used for inspiration).

bduggan:
>If I had to summarize react block in one sentence I would say that it does sequential processing of asynchronous events. That, in my opinion, is its main feature. I don't know if this description is fully correct and if it is understandable, but perhaps that's a good start.

I usually say "`react` declares an event loop".

jnthn:
I'm quite fond of this [minimal web crawler example](https://gist.github.com/jnthn/b358ebfb6ea6c366834896b9f62ba92b) for introducing `react`/`whenever`, in that it allows for discussion of various key things that the construct does:

* The responses to the web requests may arrive on multiple threads, but we will only be inside the code of the `react` block processing one of those at a time. Thus the state declared in the `react` block is protected.
* The `react` block lives so long as there are outstanding `whenever`s to be run
* The `whenever` construct is dynamic in nature - that is, you can run it many times within a single `react` block's lifetime
* Any unhandled exception is propagated to the code that started the `react` block

Besides that, the problem is:

* Practical
* Something many developers would be able to relate to, or at least understand the problem being solved
[...]

PR #2386, opened 2018-10-16
Concurrency

MorayJ:
## The problem
So what the hell is react/whenever #2135

whenever can be used in a supply block #2292

## Solution provided

Explains react.

Gives an whenever its own explanation saying it's for supply blocks and react blocks


timo:
looks very good! i'm not sure if it's already explained somewhere else, but perhaps it would be good to point out (even if it's redundant) that a supply block's mainline will be run every time it is tapped.

doc/Language/concurrency.pod6
```diff
@@ -306,9 +306,50 @@ following code outputs 0 .. 5 :
...
+     
+     my $supply = supply {
+            whenever $bread-supplier.Supply {
+                emit("We've got bread: " ~ $_);
+            };
+            whenever $vegetable-supplier.Supply {
+                emit("We've got a vegetable: " ~ $_);
+            };
+        }
+
```
timo: shippable complains that there is a tab on this line
```suggestion

```

MorayJ:
Thanks for looking @timo I think there is an example of what you're mentioning there...that said there's probably room for an issue of a general tidy up of this page as there's a lot of information but it's not necessarily well structured. I will look again...

PR #2427, opened 2018-10-29
Fix example for Pod::Block::Named. Fixes #2425

### LINKED ISSUE #2425: Example for Pod::Block::Named is not rendered correctly
## The problem

Looking at https://docs.perl6.org/type/Pod::Block::Named, you can see that example does not demonstrate its real code, rather it is rendered as pod heading, not a code piece.

## Suggestions

Simple wrapping of this example(with removing its indentation) with `=begin code :skip-test`, `=end code` pair should be enough(exact code is around https://github.com/perl6/doc/blob/master/doc/Type/Pod/Block/Named.pod6#L12). Don't forget to properly build docs locally and check if that _really_ solves the issue or ping here if not.

chsanch:
## The problem
Looking at [https://docs.perl6.org/type/Pod::Block::Named](https://docs.perl6.org/type/Pod::Block::Named), you can see that example does not demonstrate its real code, rather it is rendered as pod heading, not a code piece.

## Solution provided
Wrap the example in:
```
=begin code :skip-test
=end code
```

### DIFF
--- a/doc/Type/Pod/Block/Named.pod6
+++ b/doc/Type/Pod/Block/Named.pod6
@@ -7,11 +7,11 @@
     class Pod::Block::Named is Pod::Block { }
 
 Class for a named Pod block. For example
-
-    =begin MySection
-    ...
-    =end MySection
-
+=begin code :skip-test
+=begin MySection
+...
+=end MySection
+=end code
 creates a C<Pod::Block::Named> with name C<MySection>.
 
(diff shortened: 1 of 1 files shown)


coke:
There's no need to skip the test - that's valid Perl 6

chsanch:
> There's no need to skip the test - that's valid Perl 6

Fixed

Issue #2353, opened 2018-10-04
Improve the math page


JJ:
## Introduction

You might need to have a quick introduction to the language. [This is as short as it gets](https://docs.perl6.org/language/101-basics). If you have any question about Perl 6, it's probably better to either [take it to StackOverflow](https://stackoverflow.com/questions/tagged/perl6) or ask [in the IRC channel](https://webchat.freenode.net/?channels=#perl6), which you can access from the web.

The language used for the documentation is called [Perl6 pod](https://docs.perl6.org/language/pod). You can pretty much follow the conventions that you already see in the document, but that guide above might help you find whatever is not there.

We are talking about [this page, "Doing math with perl6"](https://docs.perl6.org/language/math), which you can edit directly on the GitHub page by clicking on the pencil right above the butterfly (a.k.a Camelia); saving it will generate a pull request, or via forking, editing locally and doing a pull request in the usual way. 

## The problem

Perl 6 includes extensive facilities for using math notation, and doing math with different types. The current page is short and includes only a few examples of sets, or arithmetic.

## Suggestions

The page can be improved in many different ways, obviously not limited to the ones below.

- [ ] Revise page for possible math terminology, or other, errors.
- [ ] The `Set` section includes, effectively, only Sets, while Perl 6 includes [several other kind of set-ty classes](https://docs.perl6.org/language/setbagmix). Math on them could be added to that section
- [ ] Most Set operations are not mentioned. Some examples of theorems proved using them, or simply solving some mathematical problems, would be interesting.
- [ ] Perl 6 is a functional language. A section on type algebra would enhance this page.

ab5tract:
Not directly math related, but I'm currently working on updating the documentation for sets, bags, and mixes page to be actually truthful to the current implementation.

So if you are adding anything set related, make sure you ignore most of what is currently written on that page, at least with regards to how the operators work with bags and mixes (and sets when they are involved in operations with bags and mixes).

PR #2428, opened 2018-10-31
Completing maths in Perl6 #2353

luisbalru:
## The problem
Allee name was wrongly written. Besides, the Allee Effect section was not completed.

## Solution provided
I've changed the name and I've added a new section: Weak Allee Effect


taboege:
There are four other typos in that section. Sorry to bring them up here, as they are unrelated to your changes, but maybe you want to fix those, too, while you are in the file:

> Well, this is and interesting phenomenon

`and` → `an`

> due to the individuals' inhability

`individuals'` → `individual's`
`inhability` → `inability`

> to see the differents behaviours

`differents` → `different`

PR #2432, opened 2018-11-02
- Should be a fix for rakudo/issues/#2466

Xliff:
## The problem

See rakudo/rakudo/#2466.

This addresses the issue with the example for IO::Handle.Supply.

-->


Xliff:
Yes. The code is fine, it's the output that is incorrect. This fixes that problem.

jnthn:
However, there is a problem still, because it says `or C<65536> characters in non-binary mode`, which is not true: the size argument (I've grumbled about the name on the Rakudo issue) is actually the I/O buffer size, which is always measured in bytes.

PR #2439, opened 2018-11-03
Fix #2438. Change items to level-2 headers in pragmas page

### LINKED ISSUE #2438: Rearrange the pragmas page
## The problem

[It](https://docs.perl6.org/language/pragmas) currently uses `=item` to introduce the different pragmas, which makes for LTA layout (and no table of contents)

## Suggestions

Change items to level-2 headers.

chsanch:
## The problem

The pragmas page currently uses =item to introduce the different pragmas, which makes for LTA layout (and no table of contents)

## Solution provided

Change items to level-2 headers


zoffixznet:

doc/Language/pragmas.pod6
```diff
@@ -41,25 +43,35 @@ and turns on its features if they are optional.
 Since these pragmas turn on the compiler version, they should be the
 first statement in the file (preceeding comments and Pod are fine).
 
-=item X<B<MONKEY-GUTS>|MONKEY-GUTS>
+=head2 MONKEY-GUTS
```
JJ: Please carry the indexing (X<>) over to this new version too.

zoffixznet: Doesn't `=head` stuff get automatically indexed? Also, I recall there being a problem with `X<>` not supported in `=head` and we had to put them above `=head` instead

JJ: It does only in /Types, if I remember correctly. If it's in a `head` it does strange stuff (creating categories). Yep, it should be over the header.

chsanch:
@JJ just to be sure, about the syntax of the index element, it should be something like:

`X<|trace, pragma>`

I couldn't find this element in the pod documentation, maybe we need to add it

JJ:
@chsanch it's not exactly POD, that's rather documentation specific. To be sure, we need to do something about that. But yes, that would work.

chsanch:
@JJ Could you check if this is OK now? Thanks!

PR #2448, opened 2018-11-05
Example use of whenever should use a react block

jmaslak:
## The problem

whenever needs to be in a react block (looks like a 6.d feature).

## Solution provided

Wrapped the usage of whenever in a react block within the example for listen().

Also added myself to CREDITS.


zoffixznet:

jmaslak:
Thanks!

I'll merge this tomorrow assuming nobody comments that they have issues with it.

PR #2451, opened 2018-11-05
Fix typo and link in Glossary

taboege:
This fixes a typo and an incorrect link on the Glossary page.

From what I read, it seems like "an historical" is not up to taste. It's "a" unless the "h" is silent, which it isn't in "historical", unless you're French :-)


lizmat:
Thanks!

PR #2461, opened 2018-11-14
fix output value for ^^ operator

interlab:


AlexDaniel:
Thanks! I invited you to the perl6 organization, so now you can commit changes like this directly.

Issue #2476, opened 2018-11-26
HTML table headers for the type index


lucasbuchala:
In https://github.com/perl6/doc/commit/f67e0008a3134b790278a4eb2b98929fc70d7907, the headers were introduced for the type index table (https://docs.perl6.org/type.html). But why "Sources" for the header of the 3rd column? Can it be changed to either "Description", "Summary", "Abstract" or "Subtitle"?

Line: https://github.com/perl6/doc/blob/master/htmlify.p6#L974

(Also, but less important, the header for the 2nd column is "Declarator". Ok. Hmm... but what about "Type"?)

PR #2478, opened 2018-11-28
Fix for issue #2476

albertferrico:
## The problem

https://github.com/perl6/doc/issues/2476

## Solution provided

Changing the headers for the table as suggested in the issue mentioned above.


lucasbuchala:
@albertferrico Thank you!

albertferrico:
My pleasure!

PR #2485, opened 2018-11-30
Added test for consisten header capitalization

### LINKED ISSUE #1945: Inconsistent capitalization for doc headers
I've seen doc headers capitalized like this:

- Foo bar baz
- Foo Bar Baz

Which way of capitalizing them would be more appropriate?

kensanata:
This fixes #1945, by changing the capitalization of a few titles and subtitles, and it adds a test to prove that it is fixed. The test includes a bunch of whitelisted terms (programming languages, acronyms, classes, and so on).


kensanata:
It’s caps after colon in German – I thought English was different?

PR #2487, opened 2018-12-04
Mention value types in eqv documentation

taboege:
## The problem
The `eqv` documentation demonstrates that two arbitrary (from user-defined
classes) objects are equivalent when their types agree and, recursively, their
properties are equivalent. But if you put these two equivalent objects each into
a `Set`, the sets do not compare `True` when using `eqv`. This is a WAT to me.

## Solution provided
`eqv` on Sets works differently than I had imagined. It doesn't unpack the Sets,
but delegates the check to Set.ACCEPTS, which in turn uses `===` to perform
the element-wise comparisons. To get the expected behaviour, as timotimo [explained](https://colabti.org/irclogger/irclogger_log/perl6?date=2018-12-04#l49),
the class must act like a value type.

Since I had never once heard of value types before, I think it is a good idea
to show the problem and link to an explanation of "value type" to increase its
page rank, which this commit does.

I'm open to suggestions. For example, I don't know if the "**Trap:**" tag should
be removed from my additions. One could also consider rewriting the previous
paragraphs, so that the impression doesn't even arise that `eqv` is really recursively
applied to anything "container-y", as that clearly isn't the case for Sets.

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -1822,6 +1822,29 @@ implement an appropriate infix C<eqv> operator:
     multi infix:<eqv>(A $l, A $r) { $l.a eqv $r.a }
     say A.new(a => 5) eqv A.new(a => 5);            # OUTPUT: «True␤»
 
+
+Note that C<eqv> does not work recursively on every kind of container type,
+e.g. C<Set>:
+
+    my class A {
+        has $.a;
+    }
+    say Set(A.new(a => 5)) eqv Set(A.new(a => 5));  # OUTPUT: «False␤»
+
+Even though the contents of the two sets are C<eqv>, the sets are not.
+The reason is that C<eqv> delegates the equality check to the C<Set> object
+which relies on element-wise C«===» comparison. Turning the class C<A>
+into a L<value type|/type/ValueObjAt> by giving it a C<WHICH> method
+produces the expected behavior:
+
+    my class A {
+        has $.a;
+        method WHICH {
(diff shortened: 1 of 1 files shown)


doc/Language/operators.pod6
```diff
@@ -1822,6 +1822,28 @@ implement an appropriate infix C<eqv> operator:
     multi infix:<eqv>(A $l, A $r) { $l.a eqv $r.a }
     say A.new(a => 5) eqv A.new(a => 5);            # OUTPUT: «True␤»
 
+B<Trap:> If these equivalent C<A> objects are inside a container like
```
JJ: Not sure about this, or the way it's formulated. If it's a Trap, there's a specific page for that. If you want to keep it here, it's better if you formulate it in a different way...

taboege: Yes, let me try to remove that and make it sound more directly relevant to the documentation of eqv, instead of discussing a special case involving eqv:

------

Note that eqv does not work recursively on every kind of container type, e.g. Set:
```
my class A {
    has $.a;
}
say Set(A.new(a => 5)) eqv Set(A.new(a => 5));  # OUTPUT: «False␤»
```

Even though the contents of the two sets are `eqv`, the sets are not. The reason is that
`eqv` delegates the equality check to the `Set` object which relies on element-wise `===`
comparison. Turning the class `A` into a *value type* by giving it a `WHICH` method
produces the expected behavior: [...]

JJ: OK, thanks.

PR #2512, opened 2018-12-17
Fix typo in example

MorayJ:
## Problem
Missing comma in example

## Solution provided
Added comma


Altai-man:
Thanks! I'd say that such tiny issues should be committed directly into master. :+1:

Issue #2537, opened 2019-01-06
Hash.pick is intrinsically slow on big hashes, maybe it should be reflected in the docs


AlexDaniel:
See https://github.com/rakudo/rakudo/issues/2586.

> This is a simple matter of algorithm. Each time you call .roll / .pick without any parameters, it needs to walk half of the keys in the hash (on average).

So we should recommend `%h.pick($how-many-elements-you-want)`.

JJ:
Maybe as a trap?

cfa:
I'm inclined to agree with @JJ's trap proposal—closing the PR.

2colours:
Maybe it should be reflected in the docs - maybe not. I'm more leaning towards "not".

It's half an implementation detail of runtimes, not even a universal thing for the language, and anyway, I think knowing these things is a whole engineering discipline we can't (nor should) take up on.

PR #2538, opened 2019-01-06
Add a note about coercion overhead to Any.pick (addresses #2537).

cfa:
Draft wording for #2537.

(Please only merge this if `Any` is deemed an appropriate place for this warning.)


cfa:
Sure, happy to pop this in a footnote if that's better/clearer.

PR #2540, opened 2019-01-07
start of test suite for rendered files

finanalyst:
## The problem
More rigorous testing is needed at each stage of the documentation process, including testing of the pod renderers.

## Solution provided
This is more of a proof of concept.
A new subdirectory is added. 'render-test'

The intention is that the pod renderer creates the html file `html/generic-pod.html` from the generic pod file in `doc/generic-pod.pod6`. Then the tests in t/ are run against `html/generic-pod.html`

The proof of concept has three tests for consideration. The tests would be run from inside `render-test`,eg., 
```
cd render-test
prove -e perl6
```


finanalyst:
different plan suggested. So closing here

PR #2563, opened 2019-01-13
Document trap on set subroutines on Associative

Kaiepi:
Fixes https://github.com/rakudo/rakudo/issues/2604

## The problem
`Map/Bag/BagHash/Hash (cont) <string>` returns `False` if the value of the key is falsey.

## Solution provided
Document the trap.


lizmat:
Cool!

Perhaps also mention this applies to all related ops (cont), ∋, ∌ and (elem), ∈, ∉ ?

Kaiepi:
What is the name for those ops?

Kaiepi:
Updated

lizmat:
Thanks!

Issue #1433, opened 2017-08-03
Make link checker work locally; make it a test


coke:
We currently have a file called "links.txt" that is generated and then pushed out to the live site.

There is also util/test-links.sh which goes out to the site, grabs the links file, then verifies each of the links against the live file.

We should be able to verify all the internal links either against the generated HTML on disk, or against a locally running copy of the site; either way, this should probably be an xt/ test, rather than a one off script.

AlexDaniel:
Related issue: https://github.com/perl6/doc/issues/561

JJ:
If we make it a test, we will have to first fix all broken links, as said in #561. Would converting util/test-links.sh to a test just work?

coke:
No because that test is going against the live website we would  need to update it so that it goes against the local running copy

JJ:
Modify it or create a new and different version that starts up a local copy and checks links?

coke:
Yes.

coke:
In the past, when converting things to tests that were not 100%, we've builtin skips/todos in the test for known issues, then slowly chipped away at the list until it's clear; any new things are expected to pass, this lets us get the advantage of the test without having to fix everything immediately.

JJ:
We do have now a test for links. Maybe we could extend it to actually check those links...

coke:
Not the same kind of links. Your test is checking that mentions of things that look like URLs are in a pod link.

This ask is for every link tag on the site, regardess of what the source of that link looks like.

JJ:
Related #935 #1542 #1925 #2328 #1838  (some of them might be linked already)

JJ:
As a matter of fact, we can simply use the generated HTML source to check local links.

finanalyst:
I have done quite a lot of work on links recently. Several types of links:
1. In primary files (under Language|Type|Programs) of the type L<.+ |https://...> - external.
1. In primary files of the type L<...|somereference> - which points to a local file.
1. In primary files of the type L<...|/routine/somereference> - points to a generated file.
1. In the generated html files of the type <a href="..."> 

The first type are tested by Pod::To::HTML. I also have tests for them. In general they are all working.

For types 2 & 3, I can write a test, but it needs to be run after the html files have all been generated because we cannot determine in advance whether they will be generated (I have found errors).

For type 3. I can write a test that looks at each html file, extract every <a href> then use eg. curl to test whether the link works. 

Have I missed anything?

JJ:
My main concern is 3.

[...]

PR #2595, opened 2019-01-28
new test for generated html files

finanalyst:
## The problem
When all the html files have been generated, all the `<a href="somelink">...</a>` should point to valid targets.

Currently, this does not happen because of bugs in the current tool-chain, particularly htmlify. 

Although, it is also possible to test whether each link in the source Pod6 file, namely each `L< >` is valid, this does not guarantee that the final html file generated to render the Pod6 source has correct links. In fact, currently, the html files contain approximately 120 incorrect links.

## Solution provided
A test file `verify-generated-html-link.t` goes through each html file in the html directory, finds each `<a` link and then verifies if the link exists (a) against a live static file server defined in P6_DOC_TEST_SERVER or (b) against the html files in the html directory.

The test uses 16 threads, and TAP syntax errors may occur when tests are returned in a different order.

### DIFF
--- a/xt/verify-generated-html-link.t
+++ b/xt/verify-generated-html-link.t
@@ -2,7 +2,6 @@
 
 use v6.*;
 use Test;
-use YAMLish;
 
 my SetHash $tested .= new;
 
--- a/xt/verify-generated-html-link.t
+++ b/xt/verify-generated-html-link.t
@@ -1,0 +1,153 @@
+#!/usr/bin/env perl6
+
+use v6.*;
+use Test;
+use YAMLish;
+
+my SetHash $tested .= new;
+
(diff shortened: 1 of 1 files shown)


coke:
See #1433

coke:
When I run this locally with no args, it runs about a 100 tests (104, 105, 91, then 97), hangs for a bit, then ends reporting a few failures.

coke:
Also, fails https://raw.githubusercontent.com/perl6/mu/master/misc/camelia.txt, but that is responding with content.

JJ:
We are finally developing a minimal set of documents for testing in a separate repo as part of Google Summer of Code. Thanks a lot.

Issue #502, opened 2016-05-08
$*REPO and CompUnit::Repository::* classes not documented


AlexDaniel:
Code:

``` perl6
say WHAT $*REPO
```

Result:

```
(Installation)
```

jonathanstowe:
I think this wants to be done with all the other CUR stuff, note also that it not always a ```CompUnit::Repository::Installation```  : 

```
[jonathan@coriolanus doc.master]$ perl6 -Ilib -e 'say $*REPO.^name'
CompUnit::Repository::FileSystem
[jonathan@coriolanus doc.master]$ perl6  -e 'say $*REPO.^name'
CompUnit::Repository::Installation
```

JJ:
There's not much about [compunit in the documentation](https://docs.perl6.org/language/modules#index-entry-compunit-need). Maybe create a new page?

ugexe:
`$*REPO` is a linked list -- what it contains depends on where it is in the list. It has nothing to do with `.git` nor does it have anything to do with the directory a program is run it. At the risk of sounding harsh: nothing in the commit that closed this issue makes sense.

JJ:
And what would make sense?

ugexe:
Some things that would make sense:
1) revert the offending commit.
2) ask what “makes sense” before commiting work that isn’t understood.

And I hate to say it but this was part of grant related work so it looks even worse.

niner:
JJ: please have a look at https://github.com/rakudo/rakudo/blob/master/docs/module_management.md
and http://niner.name/talks/A%20look%20behind%20the%20curtains%20-%20module%20loading%20in%20Perl%206/
"Not documented" has not been entirely accurate.

ohmycloud:
There are two broken links in the https://docs.perl6.org/type/CompUnit::Repository::Installation page:
CompUnit::Repository and CompUnits, shows 404 when clicked.

lizmat:
With the addition of lemmas for `CompUnit::Repository` and `CompUnit`, I think the broken links are now fixed.

rafaelschipiura:
$*REPO and CompUnit::Repository::* have been documented, closing the bug report.

JJ:
Most classes are still not documented; CompUnit::Repository is not either for the time being. Sorry, I'm reopening.

JJ:
[...]

PR #2602, opened 2019-02-02
CURFS and CURI basics

ugexe:
This adds basic documentation on the parts of CURFS and CURI that can be considered stable.

This probably needs some cleanup. For instance I wasn't exactly sure how to express `need()`s signature since it includes a private method call in its default value:
`CompUnit::PrecompilationStore :@precomp-stores = self!precomp-stores()`

### DIFF
--- a/doc/Type/CompUnit/Repository/FileSystem.pod6
+++ b/doc/Type/CompUnit/Repository/FileSystem.pod6
@@ -1,0 +1,89 @@
+=begin pod
+
+=TITLE class CompUnit::Repository::FileSystem
+
+=SUBTITLE CompUnit::Repository::FileSystem
+
+    class CompUnit::Repository::FileSystem
+        does CompUnit::Repository::Locally
+        does CompUnit::Repository
+        { }
+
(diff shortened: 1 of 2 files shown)


cfa:
Are there plans to document `CompUnit` too?

Context:

- There're a few (broken) ` L<CompUnit>` on
  - `Type/CompUnit/Repository/Installation.pod6`,
  - `Type/CompUnit/Repository/FileSystem.pod6`.
- [This](https://docs.perl6.org/language/modules-core#CompUnit::*_modules_and_roles) core modules entry seems outdated now? (_"These modules are mostly used by distribution build tools, and are not intended to be used (at least until version 6.c) by the final user."_)

JJ:
There's #502 still to go...

jnthn:
> I wasn't exactly sure how to express `need()`s signature since it includes a private method call in its default value

Perhaps just document it as an optional parameter (`:@precomp-stores` already is one, given it's named), and consider the default value an implementation detail.

lizmat:
`CompUnit` now documented with 6be9b0f146

Issue #2632, opened 2019-02-23
Checklist for 6.d


JJ:
## The problem

Many things have changed in the 6.d version, of which we have at least one release. [Here is the list](https://github.com/perl6/roast/blob/master/docs/announce/6.d.md). However, coverage is patchy. Deprecation notices take different form, some new Types and methods are there, some are not... 

Refer to this issue with `refs #2632` and check the item when you work on it either via change in the documentation of after seeing that no changes need to be done to it. In that case, please explain via comment or however why is that the case. 

## Version-Controlled Changes

- [x]  **[6.d]** `&await` no longer blocks *a thread* while waiting
- [x]  **[6.d]** `whenever` not in lexical scope of `react` throws
- [x]  **[6.d]** `$*ARGFILES` inside `sub MAIN` is always fed by `$*IN`
- [x]  **[6.d]** Constructs (literally) `$()`, `@()`, and `%()` are no longer magical
- [x]  **[6.d]** variables with `:D`/`:U` type constraints default to type object of
    the constrained type (e.g. so you could use `.new` with them)
- [x]  **[6.d]** `start` blocks in sink context attach exception handler
~- [ ]  **[6.d]** Routines must use `return-rw` to return a `Proxy`, even if
    routine is marked as `is raw` or `is rw`~
- [x]  **[6.d]** Native `num` types default to `0e0` instead of `NaN`
- [x]  **[6.d]** On subroutine names, the colonpair with key `sym` (e.g. `:sym<foo>`)
    is reserved, in anticipation of possible future use.

## Deprecations

These methods are deprecated in 6.d language and will be removed in 6.e.
Implementations may choose to emit deprecation warnings or to offer these
methods for a longer period than 6.e release.

- [x]  The use of `'-'` (single hyphen) as a special path to `&open` to mean the
    special handles (use `IO::Special` objects instead)
- [x]  `IO::Handle.slurp-rest` (use `.slurp` instead)
- [x]  `Any.flatmap` (use combination of `.flat` and `.map` methods instead)
- [x]  `Cool.path` (use `.IO` instead)
- [x]  `Pair.freeze` (use `Pair.new` with decontainerized arguments instead)
- [x]  `Str.subst-mutate` (use `Str.subst` with `.=` method call assign metaop instead)
- [x]  `Rational.norm` (`Rational` types are required to be normalized on creation now)
- [x]  `IO::Path.child` (use `.add` instead)
[...]

PR #2641, opened 2019-03-02
check negative .expmod #2632

antoniogamiz:


doc/Type/Int.pod6
```diff
@@ -81,6 +81,12 @@ can accept non-C<Int> arguments, which will be coerced to C<Int>.
     say expmod(4, 2, 5);    # OUTPUT: «1␤»
     say 7.expmod(2, 5);     # OUTPUT: «4␤»
 
+C<$y> argument can also be negative, in wich case, the result is
+equivalent to C< ($x ** $y) mod $mod.
```
JJ: Closing > is missing. Please use Perl6-mode for Atom that will warn you about these things.

PR #2652, opened 2019-03-06
Complex exponentiation involving zeros #2632

antoniogamiz:

### DIFF
--- a/doc/Type/Complex.pod6
+++ b/doc/Type/Complex.pod6
@@ -215,8 +215,8 @@ Declared as
     multi sub infix:<**>(Complex:D \a, Num(Real) \b --> Complex:D)
 
 The X<exponentiation operator> coerces the second argument to L<Complex|/type/Complex>
-and calculates the left-hand-side raised to the power of the right-hand side (both can
-be equal to zero).
+and calculates the left-hand-side raised to the power of the right-hand side. Since 6.d, 
+either argument can be equal to zero.
 
     say i ** i # OUTPUT: «0.20787957635076193+0i␤»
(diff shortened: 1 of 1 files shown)


doc/Type/Complex.pod6
```diff
@@ -206,6 +206,21 @@ coercion L<fails|/routine/fail> with C<X::Numeric::Real>.
 The C<:D> variant returns the result of that coercion. The C<:U> variant issues
 a warning about using an uninitialized value in numeric context and then returns value C<0e0>.
 
+=head2 sub infix:<**>
+
```
JJ: usually "Declared as" is inserted here.

doc/Type/Complex.pod6
```diff
@@ -206,6 +206,21 @@ coercion L<fails|/routine/fail> with C<X::Numeric::Real>.
...
+The X<exponentiation operator> coerces the second argument to L<Complex|/type/Complex>
+and calculates the left-hand-side raised to the power of the right-hand side (both can
+be equal to zero).
+
+    say i ** i # OUTPUT: «0.20787957635076193+0i␤»
+    say 2 ** i # OUTPUT: «0.7692389013639721+0.6389612763136348i␤»
+    say i ** 2 # OUTPUT: «-1+1.2246467991473532e-16i␤»
+    say 0 ** 1 # OUTPUT: «0+0i␤»
+
```
JJ: Since you are doing this as a response to issue #2632 , you should maybe add some clarification, as in "In 6.d this and that worked that way"

PR #2655, opened 2019-03-10
Document the :&tap parameter of Supply.tap

Kaiepi:
## The problem

The `:&tap` parameter of `Supply.tap` was undocumented.

## Solution provided

It is now documented.

### DIFF
--- a/doc/Type/Supply.pod6
+++ b/doc/Type/Supply.pod6
@@ -53,6 +53,7 @@ Further examples can be found in the L<concurrency page|/language/concurrency#Su
 method tap(Supply:D: &emit = -> $ { },
         :&done,
         :&quit,
+        :&tap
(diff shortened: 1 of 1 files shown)


Altai-man:
Was the parameter always(or, well, for a long time) present? Or is it a new change?

Kaiepi:
It's been around since https://github.com/rakudo/rakudo/commit/9d903408e3727cb2dd26e78f92a8e0a12345907a

jnthn:
It might be worth explaining *why* one might want to use this callback; the commit message of the referenced commit does a reasonable job of explaining why it exists.

Kaiepi:
Good point. I'll update this once I finish writing for my other pullreq I want to make

Kaiepi:
Updated

PR #2665, opened 2019-03-12
replace 'sigilled' with 'sigiled'

### LINKED ISSUE #2660: sigiled vs. sigilled
## The problem

Both versions appear in the docs, and exceptions for each are in the .pws file

## Suggestions

I think because the emphasis is on the first syllable, the l should not be doubled. Convert the usage, remove the other entry from xt/words.pws, and maybe add this to xt/word-variants since it's a common confusion.

kawaii:
This PR is a very rough draft to fix #2660.
Feedback welcome, @coke. :)


coke:
Looks good; given that it's a PR branch, you could rebase and force push to keep the history clean, but this is fine. Merging.

coke:
Done, thanks!

PR #2676, opened 2019-03-17
Document Int method on Match

Altai-man:
```
Match provides .Int and .actions
```
from 6.d changelog.
`actions` method was documented already, this adds `Int`.

### DIFF
--- a/doc/Type/Match.pod6
+++ b/doc/Type/Match.pod6
@@ -118,9 +118,10 @@ Defined as:
 
 Tries to convert stringified result of the matched text into Int.
 
-    say ('12345' ~~ /234/).Int;   # OUTPUT: «234␤»
+    say ('12345' ~~ /234/).Int;       # OUTPUT: «234␤»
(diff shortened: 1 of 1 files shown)


doc/Type/Match.pod6
```diff
@@ -110,6 +110,18 @@ Returns the matched text.
...
 
+=head2 method Int
+
+Defined as:
+
+    method Int(Match:D: --> Int:D)
+
+Tries to convert stringified result of the matched text into Int.
+
+    say ('12345' ~~ /234/).Int;   # OUTPUT: «234␤»
```
cfa: Perhaps add:

```perl6
say ('12345' ~~ /234/).Int.^name    # OUTPUT: «Int␤»
```

to clarify?

Altai-man: Added.

cfa: Thanks.

PR #2677, opened 2019-03-17
Clarify gcd with zero results

Altai-man:
See https://github.com/perl6/roast/blob/master/S03-operators/gcd.t

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -1164,7 +1164,7 @@ Please note that this has not yet been implemented.
     multi sub infix:<gcd>($a, $b --> Int:D)
 
 Coerces both arguments to L<Int|/type/Int> and returns the greatest common divisor.X<|greatest common divisor operator>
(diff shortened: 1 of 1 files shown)


doc/Language/operators.pod6
```diff
@@ -1164,6 +1164,7 @@ Please note that this has not yet been implemented.
     multi sub infix:<gcd>($a, $b --> Int:D)
 
 Coerces both arguments to L<Int|/type/Int> and returns the greatest common divisor.X<|greatest common divisor operator>
+If one of its arguments is 0, then the second one is returned. If both arguments are 0, it returns 0.
```
cfa: The phrasing is a little confusing—"second one" to me implies that the right argument is returned if either operand is 0, which isn't the case.

How about: "If one of its arguments is 0, the other is returned"? The `0 gcd 0` behaviour then follows; this could perhaps be placed in parens?

Altai-man: Changed wording, thanks!

JJ:
Testing really takes a looooong time...

PR #2691, opened 2019-03-22
Revise examples in the intro to lists.

uzluisf:

### DIFF
--- a/doc/Type/List.pod6
+++ b/doc/Type/List.pod6
@@ -24,31 +24,49 @@ information. The difference is that iteration generally treats a
 list (or any other list-like object, like a L<Seq|/type/Seq> or an L<Array|/type/Array>)
 inside a scalar as a single element, as long as it's part of another .
 
+    my $s = (1, 2, 3);
+    for $s { }      # one iteration
+    for $s.list { } # three iterations
+
+    my $t = [1, 2, 3];
+    for $t { }      # one iteration
+    for $t.list { } # three iterations
+
     my @a = 1, 2, 3;
     for @a { }      # three iterations
-
-    my $s = @a;
-    for $s { }      # one iteration
     for @a.item { } # one iteration
-    for $s.list { } # three iterations
+
+C<.item> can often be written as C<$( ... )>, and on an array variable
+even as C<$@a>.
 
 Lists generally don't interpolate (flatten) into other lists, except
 when they are not itemized or they are the single argument to an
 operation such as C<append>:
 
-    my @a = 1, 2, 3;
-    my @nested = @a, @a;         # two elements
-    my @flat = flat @a, @a;      # six elements, with explicit flat
-    my @b = 'a', 'b';
-    @b.append: @a;               # @b now has 5 elements, because @a
-                                 # is the sole argument to append
-    my @c = 'a', 'b';
-    @c.append: $@a;              # @b now has 3 elements, because of the
-                                 # itemization with $
-    say @c.elems;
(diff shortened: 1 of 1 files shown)


doc/Type/List.pod6
```diff
@@ -24,31 +24,49 @@ information. The difference is that iteration generally treats a
...
     my @a = 1, 2, 3;
     for @a { }      # three iterations
-
-    my $s = @a;
-    for $s { }      # one iteration
     for @a.item { } # one iteration
-    for $s.list { } # three iterations
+
+C<.item> can often be written as C<$( ... )>, and on an array variable
+even as C<$@a>.
```
treyharris: This is a bit confusing to me. "Often"? When are `@a.item`, `$(@a)`, and `$@a` different and when are they the same?

jnthn: Both `$(@a)` and `$@a` are sugar for `@a.item`, so will behave the same. Maybe it was meaning it in the sense of "you'll often also see it written as"?

treyharris: @jnthn Got it, thank you.

@uzluisf May I suggest you either remove `often` or change `can often be written` to `is often written`?

Also, punctuation or lack thereof causes some ambiguity in the following clause. Please change:

`and on an array variable even as` →
- `and on array variables can even be written as` or 
- `and, on array variables, even as`?

uzluisf: @treyharris That sentence was already there :-). I just moved it up closer to where `.item` is first used in the code snippet. Is `C<.item> is often written as C<$( ... )>, and on array variables, even as C<$@a>.` fine enough?

BTW, the sentence at lines 23-25:

> The difference is that iteration generally treats a list (or any other list-like object, like a <Seq|/type/Seq> or an L<Array|/type/Array>) inside a scalar as a single element, **as long as it's part of another .**

seems incomplete. I have no idea to what "it's part of another" is referring to. Probably "it's part of another list."?

treyharris: > @treyharris That sentence was already there :-). I just moved it up closer to where `.item` is first used in the code snippet.

Yes, I noticed that, but unless you need to preserve the old wording for some reason (like you need Git to detect the movement), there's no reason _not_ to write it better while you're in there.

> Is `C<.item> is often written as C<$( ... )>, and on array variables, even as C<$@a>.` fine enough?

Almost, but `s:s/and on/and, on/`, please.

> BTW, the sentence at lines 23-25:
> 
> > The difference is that iteration generally treats a list (or any other list-like object, like a <Seq|/type/Seq> or an L<Array|/type/Array>) inside a scalar as a single element, **as long as it's part of another .**
> 
> seems incomplete. I have no idea to what "it's part of another" is referring to. Probably "it's part of another list."?

I don't understand this addition either... it looks like it was added (complete with the trailing space before the final full stop) by @moritz in 645466d:

```diff
@@ -18,7 +18,7 @@ Arrays to have every value of the list stored in a container.
 In Perl 6, assigning a C<List> to a scalar variable does not lose
 information. The difference is that iteration generally treats a
 list (or any other list-like object, like a L<Seq> or an L<Array>)
-inside a scalar as a single element.
+inside a scalar as a single element, as long as it's part of another .

     my @a = 1, 2, 3;                                                           
     for @a { }      # three iterations
```

Unless @moritz can illuminate, I'd say revert that line back to its `645466d~` version (in 
4801399f.

moritz: Sorry, I don't know anymore what I meant to write there. Feel free to discard it.

Issue #2693, opened 2019-03-23
so … what?


isomorphisms:
## The problem

I don’t like this:

https://docs.perl6.org/type/Mu#method_so

> method so()
> Returns a Bool value representing the logical non-negation of an expression. One can use this method similarly to the English sentence: "If that is so, then do this thing".

## Suggestions

What is "logical non-negation" and why is it the simplest way to explain what `so` means? Please re-explain that.

JJ:
OK

PR #2697, opened 2019-03-25
Align Mu.so & Mu.not docs with design

softmoth:
This aligns with the docs for `prefix:<so>` and `:<not>`, too. Ref #2693.

## The problem

The previous explanation of so() wasn't consistent with implementation or mentioned in design docs. Double-negation is a common JavaScript idiom but not relevant to Perl 6.

The extra description of the example isn't necessary, since the parenthetical mention of Junction resolution hints at why the example is relevant. Although a better example might be found, since `if` creates a Bool context and `.so()` is redundant. But it's not wrong, so it's fine for now.

## Solution provided

Copied the text from operators#prefix_? with appropriate tweak to text. The implementation in Radudo for `method so()` calls `self.Bool`, and the docs can be just as simple.

### DIFF
--- a/doc/Type/Mu.pod6
+++ b/doc/Type/Mu.pod6
@@ -609,7 +609,7 @@ Returns the given item to the enclosing C<gather> block, without introducing a n
 
     method so()
 
-Evaluates the item in boolean context (and thus collapses Junctions),
+Evaluates the item in boolean context (and thus, for instance, collapses Junctions),
 and returns the result.
 It is the opposite of C<not>, and equivalent to the
 L«C<?> operator|/language/operators#prefix_?».
@@ -623,13 +623,19 @@ then do this thing".  For instance,
         say "Verbose option detected in arguments";
     } # OUTPUT: «Verbose option detected in arguments␤»
 
+The C<$verbose-selected> variable in this case contains a
+L<C<Junction>|/type/Junction>, whose value is C<any(any(False, False),
+any(False, False), any(False, False), any(True, False))>. That is actually a
+I<truish> value; thus, negating it will yield C<False>. The negation of that
+result will be C<True>. C<so> is performing all those operations under the hood.
+
 =head2 method not
 
     method not()
 
-Evaluates the item in boolean context (and thus collapses Junctions),
+Evaluates the item in boolean context (leading to final evaluation of Junctions, for instance),
 and negates the result.
-It is the opposite of C<so>, and equivalent to the
+It is the opposite of C<so> and its behavior is equivalent to the
 L«C<!> operator|/language/operators#prefix_!».
 
     my @args = <-a -e -b>;
--- a/doc/Type/Mu.pod6
+++ b/doc/Type/Mu.pod6
@@ -609,9 +609,10 @@ Returns the given item to the enclosing C<gather> block, without introducing a n
 
     method so()
 
-Returns a C<Bool> value representing the result of negating (with implicit
-conversion to C<Bool>) the expression and then again computing the negation of
-the result, thus computing the C<Bool> equivalent to the original expression.
+Evaluates the item in boolean context (and thus collapses Junctions),
+and returns the result.
+It is the opposite of C<not>, and equivalent to the
+L«C<?> operator|/language/operators#prefix_?».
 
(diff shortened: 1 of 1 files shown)


doc/Type/Mu.pod6
```diff
@@ -609,9 +609,10 @@ Returns the given item to the enclosing C<gather> block, without introducing a n
 
     method so()
 
-Returns a C<Bool> value representing the result of negating (with implicit
-conversion to C<Bool>) the expression and then again computing the negation of
-the result, thus computing the C<Bool> equivalent to the original expression.
+Evaluates the item in boolean context (and thus collapses Junctions),
+and returns the result.
```
JJ: Maybe say *for instance* collapses functions. It's a particular instance of what could be done, no the whole extent of it. Also give some hint of what might be the result in boolean context. Double negation at least describes the process to arrive to that value. "Collapses Junctions" does not give any hint.

doc/Type/Mu.pod6
```diff
@@ -609,9 +609,10 @@ Returns the given item to the enclosing C<gather> block, without introducing a n
...
+It is the opposite of C<not>, and equivalent to the
+L«C<?> operator|/language/operators#prefix_?».
```
JJ: Careful when you say equivalent, because things like precedence might vary. Maybe add an example.

doc/Type/Mu.pod6
```diff
@@ -622,27 +623,24 @@ then do this thing".  For instance,
...
-result will be C<True>. C<so> is performing all those operations under the hood.
-
 =head2 method not
 
     method not()
 
-Returns a C<Bool> value representing the logical negation of an expression.
-Thus it is the opposite of C<so>.
+Evaluates the item in boolean context (and thus collapses Junctions),
+and negates the result.
```
JJ: Repeating the same sentence again does not help. Be either exhaustive or say for instance.

doc/Type/Mu.pod6
```diff
@@ -622,27 +623,24 @@ then do this thing".  For instance,
...
+It is the opposite of C<so>, and equivalent to the
```
JJ: comma probably unnecessary.

doc/Type/Mu.pod6
```diff
@@ -622,27 +623,24 @@ then do this thing".  For instance,
         say "Verbose option detected in arguments";
     } # OUTPUT: «Verbose option detected in arguments␤»
 
-The C<$verbose-selected> variable in this case contains a
```
JJ: I don't see why eliminating an explanation helps. It was added last week when an user requested some clarification. If it's  incorrect or confusing, correct it, but eliminating text that might helpful to someone is not really the way to go here.

doc/Type/Mu.pod6
JJ: Also, what was incorrect about the original one? You are referring to a "Boolean context" that is not defined anywhere (and should be capitalized)

doc/Type/Mu.pod6
softmoth: The explanation that was there wasn't correct except to say that the value is "truthy", which is obvious by the example. So IMO it's completely understandable without an explanation and is less content to read.

doc/Type/Mu.pod6
softmoth: Good point. That wording is copied from operator prefix:<so>, but could be better in both places. There is no such process of double negation being done, so I think it's misleading to bring up. "Collapses Junctions" is not describing how it's done, it's a significant guarantee provided by the language (i.e., after .so() you will NOT have a Junction left over waiting to be resolved elsewhere, it'll definitely be either True or False right here). I agree that it's less than clear.

doc/Type/Mu.pod6
softmoth: Good point.

doc/Type/Mu.pod6
JJ: If it's not correct, then the best thing is to change it. That explanation was written to address a particular issue, if it's deleted that issue will be left unaddressed. What's understandable for you might not be for someone else. Unless it's incorrect it's better to leave it there, and if it's incorrect it's better to fix or clarify instead of deleting it.

PR #2707, opened 2019-04-03
Change instances of 'timezone' with 'time zone'

chsanch:
## The problem
timezone & time zone appear in the docs

## Solution provided
Change the few instances of 'timezone' with 'time zone'

### DIFF
--- a/doc/Language/variables.pod6
+++ b/doc/Language/variables.pod6
@@ -1329,7 +1329,7 @@ in your program.
 X<|$*TZ>
 =head4 C<$*TZ>
 
-C<$*TZ> contains the system's local timezone offset, as the number of B<seconds>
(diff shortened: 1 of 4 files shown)


AlexDaniel:
https://en.wiktionary.org/wiki/timezone:

> In the US, timezone seems to be favored considerably (unless comparing different specific zones), while the UK seems more likely to use time zone.

I think we need to replace “time zone” with “timezone” instead. Not because we prefer US variants over UK ones, but because all perl 6 methods use “timezone” (e.g. `.timezone` not `.time-zone`).

JJ:
Agreed. In most composite terms such as this one, the equivalent to timezone is preferred.

JJ:
This PR is not in agreement with the current terminology we are using. Thanks and keep them coming!

PR #2731, opened 2019-04-29
Add example of dynamically created enum

Altai-man:
## The problem

Part of 6.d changelog, `Dynamically created lists can be used to define an enum`.

### DIFF
--- a/doc/Language/typesystem.pod6
+++ b/doc/Language/typesystem.pod6
@@ -780,6 +780,19 @@ the keys.
     say @keys.map: *.enums;
     # OUTPUT: «(Map.new((one => 0, two => 1)) Map.new((one => 0, two => 1)))␤»
 
+With the use of B<()> parentheses, an enum can be defined using any
+arbitrary dynamically defined list. The list should consist of contain Pair
+objects:
+
+    # For example, in file `config` we have:
+    # a 1
+    # b 2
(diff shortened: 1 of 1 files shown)


doc/Language/typesystem.pod6
```diff
@@ -780,6 +780,19 @@ the keys.
     say @keys.map: *.enums;
     # OUTPUT: «(Map.new((one => 0, two => 1)) Map.new((one => 0, two => 1)))␤»
 
+With the use of B<()> parentheses, an enum can be defined using any
+arbitrary dynamically defined list. The list should consist of contain Pair
```
JJ: consist of *or* contain

doc/Language/typesystem.pod6
```diff
@@ -780,6 +780,19 @@ the keys.
...
+objects:
+
+    # For example, in file `config` we have:
```
JJ: Use normal text, not comments, for this...

Altai-man:
Edited, but not sure at all regarding `C<>` for the example config file content describing.
Of course, we can go without a config file, for example, sticking `rand` for enum creation, but creating enums with random values creates more of "Those folks are odd, why would one create an enum dynamically?" aura than we really need, where a config use-case is not perfect, but at least somewhat useful.

PR #2737, opened 2019-05-01
Map and Range: Iterable is a role; use 'does' not 'is'

softmoth:
## The problem

Docs for Map and Range classes say `is Iterable`, but Iterable is a class.

## Solution provided

Change to `does`.


moritz:
Thanks @softmoth, good catch!

Issue #2603, opened 2019-02-03
Set/Bag operations should include Mixes too


JJ:
## The problem

Most of the operations [here](https://docs.perl6.org/language/setbagmix#Set/Bag_operators) do work on Mixes. A paragraph says that Mixes will be coerced to Sets where necessary, but that's not always the case. Some operations work on Mixes directly, or on Bags.

## Suggestions

Revise definitions of methods and title to reflect the fact that they would work on mixes. Maybe change the title to simply Set-like operators (or something like that)

Xliff:
I am fine referring to them as "Set-like" operators, but would like to know HOW a Mix reacts when they are used with such operators. My experience with Sets, Bags, and Mixes is relatively non-existent. 

I'm fine with taking this on, but would like to have more guidance on the wording. I am just abour finished with #2761, but this will have to remain open until I fully understand the issue.

PR #2765, opened 2019-05-05
Operators and SetBabMix page updates

Xliff:
Addresses #2761 and sets the ground work for #2603

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -3071,7 +3071,9 @@ Equivalent to C«!(>)».
 
 =head3 infix (|)
 
-X<Union operator>. It is of precedence "Junctive Or".
+X<Union operator>.
+
+It is of precedence "Junctive Or".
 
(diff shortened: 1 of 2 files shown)


lizmat:
Too bad I had to go away before this was merged.

In my opinion, set operators are no different from other operators, and should as such be listed in the operators page.  All of them take any type of parameters (if they don't, it's a bug).  Most of them return a `Bool`, only some of them return a `Set`, `Bag` or `Mix`.  But since we consider these first class citizens in Perl 6, that in itself shouldn't make a difference, just like we don't separately handle operators returning an `Int`, `Order`, `List`, `Hash` or `Map`.

@Xliff: thank you for your work, but I'm afraid I'm  going to remove some of that and make the set operators first class citizens in the operators page, with the setbagmix page in the future referring to them.  `Set`s, `Bag`s and `Mix`es can have a full life without ever seeing a set operator, so this makes sense (at least to me).

PR #2772, opened 2019-05-07
Remove spaces around operators and reflow

uzluisf:

### DIFF
--- a/doc/Language/statement-prefixes.pod6
+++ b/doc/Language/statement-prefixes.pod6
@@ -64,8 +64,8 @@ my @a = hyper for ^100_000 { .is-prime }
 This code is around 3x faster than the bare for. But there are a couple of
 caveats here:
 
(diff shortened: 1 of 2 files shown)


doc/Language/statement-prefixes.pod6
```diff
@@ -65,12 +65,12 @@ This code is around 3x faster than the bare for. But there are a couple of
 caveats here:
 
 =item The operation inside the loop should take enough time to make threading
-make sense.
+makes sense.
```
JJ: Not really. "make sense" makes more sense here.

Altai-man: +1, but I would say that the overall sentence structure is hmm. "to make threading make sense"... Maybe we can rewrite it somehow?

JJ: ... "for threading to make sense"

PR #2774, opened 2019-05-09
Clarify that assigning a Seq to an Array consumes it

kjkuan:
## The problem
I thought the original sentence, "The same is true when assigning a Seq to an array."  is not very clear on what's the same here. So I rewrote it and provided an example too.

### DIFF
--- a/doc/Type/Seq.pod6
+++ b/doc/Type/Seq.pod6
@@ -8,10 +8,20 @@
 
 A C<Seq> represents anything that can lazily produce a sequence of values. A
 C<Seq> is born in a state where iterating it will consume the values. However,
-calling .cache on a Seq will return a List that is still lazy, but stores the
-generated values for later access. The same is true when assigning a Seq to an
-array.
+calling C<.cache> on a C<Seq> will return a List that is still lazy, but stores the
+generated values for later access. Assigning a C<Seq> to an array consumes the C<Seq>;
+alternatively, you can use the C<lazy> statement prefix to avoid it from being
+iterated during the assignment:
 
+    =begin code
+    # The Seq created by gather ... take is consumed on the spot here.
+    my @a = gather do { say 'consuming...'; take 'one' };  # OUTPUT: «comsuming...␤»
+     
+    # The Seq here is only consumed as we iterate over @a later.
+    my @a = lazy gather do { say 'consuming...'; take 'one' };  # outputs nothing.
+    .say for @a;  # OUTPUT: «comsuming...␤one␤»
+    =end code
+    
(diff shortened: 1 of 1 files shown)


coke:
Looks good. Small typo: comsuming.

doc/Type/Seq.pod6
```diff
@@ -8,10 +8,20 @@
...
 C<Seq> is born in a state where iterating it will consume the values. However,
-calling .cache on a Seq will return a List that is still lazy, but stores the
-generated values for later access. The same is true when assigning a Seq to an
-array.
+calling C<.cache> on a C<Seq> will return a List that is still lazy, but stores the
+generated values for later access. Assigning a C<Seq> to an array consumes the C<Seq>;
+alternatively, you can use the C<lazy> statement prefix to avoid it from being
+iterated during the assignment:
 
+    =begin code
```
JJ: No need to indent this.

doc/Type/Seq.pod6
```diff
@@ -8,10 +8,20 @@
 
 A C<Seq> represents anything that can lazily produce a sequence of values. A
 C<Seq> is born in a state where iterating it will consume the values. However,
-calling .cache on a Seq will return a List that is still lazy, but stores the
-generated values for later access. The same is true when assigning a Seq to an
-array.
+calling C<.cache> on a C<Seq> will return a List that is still lazy, but stores the
+generated values for later access. Assigning a C<Seq> to an array consumes the C<Seq>;
```
JJ: Maybe a "however" before Assigning?

doc/Type/Seq.pod6
```diff
@@ -8,10 +8,20 @@
...
+alternatively, you can use the C<lazy> statement prefix to avoid it from being
+iterated during the assignment:
 
+    =begin code
+    # The Seq created by gather ... take is consumed on the spot here.
+    my @a = gather do { say 'consuming...'; take 'one' };  # OUTPUT: «comsuming...␤»
```
JJ: Check typo in output.

doc/Type/Seq.pod6
```diff
@@ -8,10 +8,20 @@
...
+     
+    # The Seq here is only consumed as we iterate over @a later.
+    my @a = lazy gather do { say 'consuming...'; take 'one' };  # outputs nothing.
+    .say for @a;  # OUTPUT: «comsuming...␤one␤»
```
JJ: Again, typo.

doc/Type/Seq.pod6
kjkuan: Thanks for reviewing, I just addressed your comments, please check it out. Thanks.

PR #2775, opened 2019-05-09
Fix infinite range subscript examples

taboege:
cf. rakudo/rakudo#2872 and perl6/roast#536

### DIFF
--- a/doc/Language/list.pod6
+++ b/doc/Language/list.pod6
@@ -477,17 +477,34 @@ which is selecting the 4 to 6th element from the three  first dimensions (C<^3>)
 
 =head2 Range as slice
 
-A L<C<Range>|/type/Range> is a container for a lower and an upper boundary.
-Generating a slice with a C<Range> will include any index between those bounds,
-including the bounds. For infinite upper boundaries we agree with
-mathematicians that C<Inf> equals C<Inf-1>.
+A L<C<Range>|/type/Range> is a container for a lower and an upper boundary,
+either of which may be excluded. Generating a slice with a C<Range> will
+include any index between the bounds, though an infinite Range will
+L<truncate|/language/subscripts#Truncating_slices> non-existent elements.
+An infinite range with excluded upper boundary (e.g. C<0..^Inf>) is still
+infinite and will reach all elements.
 
     my @a = 1..5;
     say @a[0..2];     # OUTPUT: «(1 2 3)␤»
     say @a[0..^2];    # OUTPUT: «(1 2)␤»
     say @a[0..*];     # OUTPUT: «(1 2 3 4 5)␤»
-    say @a[0..^*];    # OUTPUT: «(1 2 3 4)␤»
+    say @a[0..^*];    # OUTPUT: «(1 2 3 4 5)␤»
     say @a[0..Inf-1]; # OUTPUT: «(1 2 3 4 5)␤»
+
+Note that when the upper boundary is a WhateverCode instead of just a Whatever,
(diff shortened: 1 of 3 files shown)


doc/Language/list.pod6
```diff
@@ -480,14 +480,17 @@ which is selecting the 4 to 6th element from the three  first dimensions (C<^3>)
 A L<C<Range>|/type/Range> is a container for a lower and an upper boundary.
 Generating a slice with a C<Range> will include any index between those bounds,
 including the bounds. For infinite upper boundaries we agree with
-mathematicians that C<Inf> equals C<Inf-1>.
+mathematicians that C<Inf> equals C<Inf-1>. Slicing with an infinite Range
+will L<truncate|/language/subscripts#Truncating_slices>.
```
JJ: Not clear the relationship between the two sentences here. If it's "so,..." they the example contradicts them. it's only the combination of -1 and using the ^ which truncates the range. How does that follow from the two sentences?

doc/Language/list.pod6
```diff
@@ -480,14 +480,17 @@ which is selecting the 4 to 6th element from the three  first dimensions (C<^3>)
...
 
     my @a = 1..5;
     say @a[0..2];     # OUTPUT: «(1 2 3)␤»
     say @a[0..^2];    # OUTPUT: «(1 2)␤»
     say @a[0..*];     # OUTPUT: «(1 2 3 4 5)␤»
-    say @a[0..^*];    # OUTPUT: «(1 2 3 4)␤»
+    say @a[0..^*];    # OUTPUT: «(1 2 3 4 5)␤»
     say @a[0..Inf-1]; # OUTPUT: «(1 2 3 4 5)␤»
+    say @a[0..*-1];   # OUTPUT: «(1 2 3 4 5)␤»
+    say @a[0..^*-1];  # OUTPUT: «(1 2 3 4)␤»
```
JJ: Infinite minus one minus one is still infinite. How does this apply here?

doc/Language/list.pod6
taboege: There is no causal relationship between the two. I just found the information about truncation was missing from this discussion.

To fix this properly, without just removing it again, maybe this entire section about Range slicing should be taken off the List context section there and instead go into http://docs.perl6.org/language/subscripts#Slices. What do you think?

doc/Language/list.pod6
taboege: It doesn't apply. `1..*` is the same as `1..Inf` with or without the caret. But `1..*-1` suddenly is a WhateverCode that generates a Range whose upper boundary is filled in by the subscript operator with the list's .elems. So subtracting 1 twice indeed loses the last value. The interesting point is that the infinite ranges `1..*` and `1..^*` behave as if they were also WhateverCode, making the use of `*` strangely consistent. I take it that this (up to the strangely consistent part) should be explained on the page, too. Correct?

doc/Language/list.pod6
JJ: Maybe you could move it up in the paragraph, in the sentence starting with "Generating a slice"

doc/Language/list.pod6
JJ: Yep, probably. Clarifying that would really be helpful.

taboege:
I pushed another commit in front, in addition to (hopefully) addressing your remarks. That additional commit adds to the Range documentation a paragraph about how WhateverCode is pulled through the .. operator, to which I refer in the list subscripting docs.

Issue #2684, opened 2019-03-19
Nativecall function callback syntax


CurtTilmes:
https://docs.perl6.org/language/nativecall#Function_arguments

has an example like this:

    my sub SetCallback(&callback:(Str --> int32)) is native('mylib') { * }

I'm not sure, but that syntax seems wrong to me.  I _think_ it should be: 

    my sub SetCallback(&callback (Str --> int32)) is native('mylib') { * }

(No ':', space instead).

Can someone who actually knows verify that?  (The space seems to work when I use it, while I get errors with the ':', but I don't actually know why)

JJ:
Well, it might be that the first has too many parentheses... But you are right, that syntax is wrong. However, we will have to look and check what's actually right...

taboege:
Everywhere else the colon is used to [constrain the signature of a callable](http://docs.perl6.org/type/Signature#Constraining_signatures_of_Callables), isn't it? The no-colon-but-space syntax reminds me of [destructuring](http://docs.perl6.org/type/Signature#Destructuring_arguments) which is not at all what is intended.

But I can reproduce what Curt said: with the colon I get an exception, and without the colon it works as expected, e.g. https://gist.github.com/taboege/2e95ce969d26d00342669723c1a11dd1.

I am so surprised that I bugged rakudo/rakudo#2878 about it.

JJ:
Closed via 1abe655

treyharris:
This is insufficient, @JJ — it makes the example work, but it has no explanation anywhere. Like @taboege noted, the “fixed” example _should_ be, by [Signature’s doc](http://docs.perl6.org/type/Signature#Destructuring_arguments), a destructuring, and a nonsensical one.

I don’t think there’s harm in leaving the “fixed” example, but I think this needs to stay open at least until we have some clarity from the rakudo bug on _how_ this syntax is doing what it’s doing.

I’ll annotate rakudo/rakudo#2878 about this.

treyharris:
@JJ I’m suggesting something along the lines of the pullreq #2791 — please take a look.

JJ:
Closed after merging #2791

PR #2791, opened 2019-05-14
nativecall: Add language to address #2684

treyharris:
The syntax for NativeCall callable signature constraints under Rakudo
deviates from the usual syntax. There is an issue at
rakudo/rakudo#2878 addressing this, but until this is addressed, some
minimal explanation for the strange syntax is needed.

### DIFF
--- a/doc/Language/nativecall.pod6
+++ b/doc/Language/nativecall.pod6
@@ -653,8 +653,17 @@ say $esponja;
 
 NativeCall also supports native functions that take functions as arguments.  One
 example of this is using function pointers as callbacks in an event-driven
-system.  When binding these functions via NativeCall, one need only provide the
+system.  When binding these functions via NativeCall, one should need only provide the
 equivalent signature as L<a constraint on the code parameter|/type/Signature#Constraining_signatures_of_Callables>:
+
+    # See below--this syntax does not work on Rakudo currently
+    use NativeCall;
+    # void SetCallback(int (*callback)(const char *))
+    my sub SetCallback(&callback:(Str --> int32)) is native('mylib') { * }
+
+However, in the case of NativeCall, currently Rakudo requires a space between the
+function argument and the signature, and the colon of a normal Signature
+literal is ommitted, as in:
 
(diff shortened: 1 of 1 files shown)


treyharris:
Please don’t merge this yet; it’s for discussion at #2684.

doc/Language/nativecall.pod6
```diff
@@ -653,9 +653,18 @@ say $esponja;
 
 NativeCall also supports native functions that take functions as arguments.  One
 example of this is using function pointers as callbacks in an event-driven
-system.  When binding these functions via NativeCall, one need only provide the
+system.  When binding these functions via NativeCall, one should need only provide the
```
JJ: need only provide is correct, should need only is not. *should need to only provide* could be an alternative.

doc/Language/nativecall.pod6
```diff
@@ -653,9 +653,18 @@ say $esponja;
...
 equivalent signature as L<a constraint on the code parameter|/type/Signature#Constraining_signatures_of_Callables>:
 
+    # See below--this syntax does not work on Rakudo currently
```
JJ: I really see no point in including an example of something that does not work. We can mark an issue as "external" and wait for it to be solved instead.

doc/Language/nativecall.pod6
```diff
@@ -653,9 +653,18 @@ say $esponja;
...
+    use NativeCall;
+    # void SetCallback(int (*callback)(const char *))
+    my sub SetCallback(&callback:(Str --> int32)) is native('mylib') { * }
+
+However, in the case of NativeCall, currently Rakudo requires a space between the
```
JJ: Again, this kind of statement is impossible to maintain. Currently when? Up to when? We can leave the issue open if needed, but the documentation should reflect what's happening right now.

doc/Language/nativecall.pod6
JJ: needs only provide, actually...

PR #2792, opened 2019-05-15
Clarify Setty/Baggy and add examples

Altai-man:
Resolves `Items with negative weights are removed when coercing a Mixy to Setty/Baggy` item in 6.d checklist.


lizmat:
Thanks!

PR #2797, opened 2019-05-17
add .map, .flat, .concise, and .summary for Backtrace

lukasvalle:
add .map, .flat, .concise, and .summary for Backtrace as list missed in https://github.com/perl6/doc/issues/2632

### DIFF
--- a/doc/Type/Backtrace.pod6
+++ b/doc/Type/Backtrace.pod6
@@ -68,6 +68,69 @@ Defined as:
 
 Returns a list of L<Backtrace::Frame> objects for this backtrace.
 
+=head2 method summary
+
+Defined as:
+
+    method summary(Backtrace:D: --> Str:D)
+
+Returns a summary string representation of the backtrace, filtered
+by C<!.is-hidden && (.is-routine || !.is-setting)>.
+
+    sub inner { say Backtrace.new.summary }
+    sub outer { inner; }
+    outer;
+
+    #OUTPUT:
+    #perl6 test.p6
+    #  in method new at SETTING::src/core/Backtrace.pm6 line 85
+    #  in sub inner at test.p6 line 1
+    #  in sub outer at test.p6 line 2
+    #  in block <unit> at test.p6 line 3
+
+=head2 method concise
+
+Defined as:
+
+    method concise(Backtrace:D: --> Str:D)
+
+Returns a concise string representation of the backtrace, filtered
+by C<!.is-hidden && .is-routine && !.is-setting>.
+
+    sub inner { say Backtrace.new.concise }
+    sub outer { inner; }
+    outer;
+
+    #OUTPUT:
+    #perl6 test.p6
+    #  in sub inner at test.p6 line 1
+    #  in sub outer at test.p6 line 2
+
+=head2 method map
+
+It invokes &code for each element and gathers the return values in a sequence and returns it.
+
+    sub inner { Backtrace.new.map({ say "{$_.file}: {$_.line}" }); }
+    sub outer { inner; }
+    outer;
+
+    #OUTPUT:
+    #perl6 test.p6
+    #  SETTING::src/core/Backtrace.pm6: 85
+    #  SETTING::src/core/Backtrace.pm6: 85
+    #  test.p6: 1
+    #  test.p6: 2
+    #  test.p6: 3
+    #  test.p6: 1
+
+=head2 method flat
+
+Defined as:
+
+    multi method flat(Backtrace:D: --> List:D }
+
+Returns the backtrace same as L<list|#method_list>.
+
 =end pod
 
 # vim: expandtab softtabstop=4 shiftwidth=4 ft=perl6


doc/Type/Backtrace.pod6
```diff
@@ -68,6 +68,69 @@ Defined as:
...
+    method summary(Backtrace:D: --> Str:D)
+
+Returns a summary string representation of the backtrace, filtered
+by C<!.is-hidden && (.is-routine || !.is-setting)>.
+
+    sub inner { say Backtrace.new.summary }
+    sub outer { inner; }
+    outer;
+
+    #OUTPUT:
```
Altai-man: I would moved the section out of comments, you can prepend the listing with e.g. `This program:` and then `prints:`, and then you can just show the output without it being commented out or showing `perl6 test.p6`.
Something like:
```
This program:
$code
results in:
$output-as-just-monospace-text-without-indentation
```

doc/Type/Backtrace.pod6
```diff
@@ -68,6 +68,69 @@ Defined as:
...
+    method concise(Backtrace:D: --> Str:D)
+
+Returns a concise string representation of the backtrace, filtered
+by C<!.is-hidden && .is-routine && !.is-setting>.
+
+    sub inner { say Backtrace.new.concise }
+    sub outer { inner; }
+    outer;
+
+    #OUTPUT:
```
Altai-man: This output can be done in the same way as previous method definition is suggested.

doc/Type/Backtrace.pod6
```diff
@@ -68,6 +68,69 @@ Defined as:
...
+
+=head2 method map
+
+It invokes &code for each element and gathers the return values in a sequence and returns it.
+
+    sub inner { Backtrace.new.map({ say "{$_.file}: {$_.line}" }); }
+    sub outer { inner; }
+    outer;
+
+    #OUTPUT:
```
Altai-man: And this one.

doc/Type/Backtrace.pod6
```diff
@@ -68,6 +68,69 @@ Defined as:
...
+    #perl6 test.p6
+    #  in sub inner at test.p6 line 1
+    #  in sub outer at test.p6 line 2
+
+=head2 method map
+
+It invokes &code for each element and gathers the return values in a sequence and returns it.
```
Altai-man: I wonder if we can prepend it with signature?

doc/Type/Backtrace.pod6
```diff
@@ -77,16 +77,16 @@ Defined as:
 Returns a summary string representation of the backtrace, filtered
 by C<!.is-hidden && (.is-routine || !.is-setting)>.
 
+    this program:
```
JJ: de-indent or comment this?

doc/Type/Backtrace.pod6
```diff
@@ -77,16 +77,16 @@ Defined as:
...
     sub outer { inner; }
     outer;
 
-    #OUTPUT:
-    #perl6 test.p6
-    #  in method new at SETTING::src/core/Backtrace.pm6 line 85
-    #  in sub inner at test.p6 line 1
-    #  in sub outer at test.p6 line 2
-    #  in block <unit> at test.p6 line 3
+    results in:
```
JJ: And this? Also try and run changed files through examples-compilation.t to detect this

doc/Type/Backtrace.pod6
```diff
@@ -97,31 +97,35 @@ Defined as:
 Returns a concise string representation of the backtrace, filtered
 by C<!.is-hidden && .is-routine && !.is-setting>.
 
+    this program:
```
JJ: Same as above (and below too)

doc/Type/Backtrace.pod6
```diff
@@ -97,31 +97,35 @@ Defined as:
...
 
 =head2 method map
 
+Defined as:
+
+    multi method map(Backtrace:D: &block --> Seq:D)
+
 It invokes &code for each element and gathers the return values in a sequence and returns it.
 
+    this program:
```
JJ: Same problem here.

doc/Type/Backtrace.pod6
```diff
@@ -97,31 +97,35 @@ Defined as:
...
 
-    #OUTPUT:
-    #perl6 test.p6
-    #  SETTING::src/core/Backtrace.pm6: 85
-    #  SETTING::src/core/Backtrace.pm6: 85
-    #  test.p6: 1
-    #  test.p6: 2
-    #  test.p6: 3
-    #  test.p6: 1
+    results in:
```
JJ: And here

doc/Type/Backtrace.pod6
```diff
@@ -97,31 +97,35 @@ Defined as:
...
     sub inner { say Backtrace.new.concise }
     sub outer { inner; }
     outer;
 
-    #OUTPUT:
-    #perl6 test.p6
-    #  in sub inner at test.p6 line 1
-    #  in sub outer at test.p6 line 2
+    results in:
+    in sub inner at test.p6 line 1
```
JJ: You will also have to mark this as :language<text>, since it's not Perl6

doc/Type/Backtrace.pod6
lukasvalle: First version was based on example [debugging docu](https://docs.perl6.org/programs/01-debugging#___top) but I did change to the current version as requested. Is there a good example somewhere already, how to document this?

JJ: Indented text is going to be treated as a program. Just de-indent it.

doc/Type/Backtrace.pod6
JJ: Maybe this could have been a bit more explicit
```
This program
    $code
results in
=for code :lang<text>
$output
```

doc/Type/Backtrace.pod6
lukasvalle: But without indentation all text will be rendered on the same line in think.

JJ: Just that first line, as above.

JJ: (as a matter of fact, I don't think the initial version was incorrect, but if you do it this other way, please try and take that into account)

Altai-man:
Let's see. Please, wait a little, I will build docs now and will write some explanations...

Altai-man:
Ok, so firstly I built docs using instructions at https://github.com/perl6/doc#building-the-html-documentation
Secondly, I had to install cpan-minus and install Mojolicious dependency to run it locally.

Then `make run` and opening the page in browser at `localhost:3000` address. 
2019-05-17-223209_1920x1080_scrot
^ here you can see how your changes look like (without highlighting enabled): the explanation lines and output are placed in a big code block, which is not good.

To improve the situation, we can:
1)Remove indentation from lines that are not planned to be in the code block, but rather that should be displayed as normal text(such as "This program" and "results in"), as was suggested.
2)The program code indentation ls left as is.
3)The output is not a Perl 6 code, so should be marked as a code(to get monospace font), but text, with `=for code :lang<text>

For example, one of methods after an edit looks this way:

```
    multi method map(Backtrace:D: &block --> Seq:D)

It invokes &code for each element and gathers the return values in a sequence and returns it.

This program:

    sub inner { Backtrace.new.map({ say "{$_.file}: {$_.line}" }); }
    sub outer { inner; }
    outer;

results in:

=for code :lang<text>
SETTING::src/core/Backtrace.pm6: 85
SETTING::src/core/Backtrace.pm6: 85
test.p6: 1
test.p6: 2
test.p6: 3
test.p6: 1

=head2 method flat
```

Note that for the text block, indentation is absent, because `=for code` directive already treats the next lines as code and if there will be indentation, it will be present in the rendered docs, which is less than awesome.

After another `make html-nohighlight` and `make run`, I can check the changes:

2019-05-17-225715_1920x1080_scrot

^ results. With highlighting on the real site enabled, there will be a clear separation between the code and the result's listing.

I have fixed this pull request committing the changes, so you can check them.

Thanks a lot for your contribution! If there are any questions, don't hesitate to ask.

I have sent a pull request with my changes and this PR can be merged now.

PR #2804, opened 2019-05-17
Explain $! and $/ in start block

Altai-man:
Addresses `start blocks and thunks get fresh $/ and $!` from 6.d changelog.

On the other hand, I am not sure where we can indicate `thunks` part. Any ideas?

### DIFF
--- a/doc/Language/control.pod6
+++ b/doc/Language/control.pod6
@@ -178,7 +178,7 @@ If the code inside the block has not finished, the call to C<.result> will
 wait until it is done.
 
 A C<start> may also be used on a bare statement (without curly braces).
-This is mainly just useful when calling a subroutine / method on an object
+This is mainly useful when calling a subroutine / method on an object
 is the only thing to do asynchronously.
 
     sub get42 { 42 }
@@ -187,7 +187,8 @@ is the only thing to do asynchronously.
 
 Note that code executed this way does not have access to the special
 variables L«C<$!>|/syntax/$!» and L«C<$/>|/syntax/$/» of its outer
-block, but receives new ones.
+block, but receives new ones, so every asynchronous task has its
+per-task state.
 
 Thus, C<try> expressions and regex matches executed in the
 asynchronous task have their per-task state.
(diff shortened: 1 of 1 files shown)


doc/Language/control.pod6
```diff
@@ -181,6 +181,22 @@ A C<start> may also be used on a bare statement (without curly braces).
 This is mainly just useful when calling a subroutine / method on an object
 is the only thing to do asynchronously.
 
+    sub get42 { 42 }
+    my $promise = start get42;
+    say $promise.result; # OUTPUT: «42␤»
+
+Note that code executed this way does not have access to special variables L«C<$!>|/syntax/$!»
```
jnthn: *to the

doc/Language/control.pod6
```diff
@@ -181,6 +181,22 @@ A C<start> may also be used on a bare statement (without curly braces).
...
+and L«C<$/>|/syntax/$/» of its outer block, but receives new ones.
```
jnthn: Not sure if it's worth mentioning the motivation for this: so that regex matches and `try` expressions executed in the asynchronous task will have their per-task state.

Altai-man: Hmmm, we probably should. Also, can you please explain `thunks` part? I know the term in its general meaning, we even have a glossary item, but in fact it tells no useful info, and the changelog entry says "start blocks and __thunks__ get fresh $/ and $!", while it is totally unclear where we can or if we even should note the thunks part.

jnthn: When we have `start { blah }` it's obvious that there's a new scope there: the `start` has a block. But when we have `start blah` there's no visible block there, but in fact we are creating an object of type `Code` that contains the `blah`. That `Code` object - which doesn't map to any visible block in the code - is the thunk.

jnthn: Ah, and the reason we must do this is because we don't want `blah` to execute *now*, but over on the thread pool - which is the point of `start`! :)

JJ: This is not resolved yet, right? Can you please take a look and solve it?

Altai-man: It is not, thanks for reminding.

I will address it a bit later today.

PR #2805, opened 2019-05-17
Add cando example with Block

Altai-man:
Tries to address `WhateverCode and Block provide .cando` item from 6.d changelog.

Both `WhateverCode` and `Block` inherit from `Code`, so have this method. On the other hand, it says "thunks of code" and it is unclear how to create them explicitly.

### DIFF
--- a/doc/Type/Code.pod6
+++ b/doc/Type/Code.pod6
@@ -146,7 +146,7 @@ L<Capture|/type/Capture>.  Since C<Code> objects do not have any multiple
 dispatch, this either returns a list with the object, or an empty list.
 
     my $single = \'a';         # a single argument Capture
-    my $plural = \'a', 42;     # a two argument Capture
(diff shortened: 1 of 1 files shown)


doc/Type/Code.pod6
```diff
@@ -145,6 +145,12 @@ Returns a list of candidates that can be called with the given
 L<Capture|/type/Capture>.  Since C<Code> objects do not have any multiple
 dispatch, this either returns a list with the object, or an empty list.
 
+    my $single = \'a';         # a single argument Capture
+    my $plural = \'a', 42;     # a two argument Capture
```
JJ: Maybe add a () around the constants? This yields:
```
Useless use of constant integer 42 in sink context (lines 1, 1)
```

Altai-man: D'oh! I changed that when been testing my code, but not in actual source. Thanks, will fix now...

Issue #2838, opened 2019-06-05
Use prove6 instead of prove in the Testing page


JJ:
## The problem

App::Prove6 has been available for some time now.

## Suggestions

 Use it instead of encouraging people to install perl just to run prove.

PR #2841, opened 2019-06-08
Suggest prove6 over prove

Altai-man:
https://github.com/perl6/doc/issues/2838

### DIFF
--- a/doc/Language/testing.pod6
+++ b/doc/Language/testing.pod6
@@ -68,11 +68,27 @@ command line:
 =for code :lang<shell>
 $ perl6 t/test-filename.t
 
-Or via the L<prove|https://perldoc.perl.org/prove.html> command from Perl 5,
-where using C<--exec> to specify the executable that runs the tests:
+To run all tests in the directory recursively,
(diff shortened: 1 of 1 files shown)


doc/Language/testing.pod6
```diff
@@ -68,11 +68,27 @@ command line:
...
+
+=for code :lang<shell>
+$ zef install App::Prove6
+
+You can run C<prove6> in a distribution directory this way:
 
 =for code :lang<shell>
-$ prove --exec perl6 -r t
+$ prove6 --lib t/
+
```
JJ: I am not sure you need --lib t/ Have you tried to run it by itself?

Altai-man: Sure, of course I tried to run it before constructing the command and writing about it.
The thing is, it _does_ work without explicit `t/` argument, but I wanted to be a bit more explicit about the directory with tests path.
We refer to prove6 docs here, if the user wants to customize the command, they can refer to it and construct something that is more specific. `t` directory is an assumption.

On toe other hand, I don't know why I omitted `-e` argument that has Perl 6 executor as the default one, though the situation is similar.

PR #2844, opened 2019-06-09
Teach p6doc to handle --format

clarkema:
As a long-term improvement, I'd like to able to read nicely-formatted documentation in the terminal using `man(1)`.

This is an early step in that direction, introducing a `--format` option to `p6doc` to allow it to output documentation in any format supported by a `Pod::To::*` module.  Specifying a `--format` disables paging, since that doesn't really make any sense when generating anything that can't be fed directly into `less(1)`

### DIFF
--- a/bin/p6doc
+++ b/bin/p6doc
@@ -52,8 +52,8 @@ sub locate-module(Str $modulename) {
     return $m;
 }
 
-sub show-docs(Str $path, :$section, :$no-pager, :$package is copy) {
-
+sub show-docs(Str $path, :$section, :$no-pager, :$format is copy, :$package is copy) {
+    $format ||= 'Text';
     my $pager;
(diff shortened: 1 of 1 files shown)


clarkema:
@JJ I'm glad to hear p6doc is being actively worked on!  This patch is certainly not a final, complete feature -- it doesn't support the SectionFilter path, for one thing (requests that go that route still use the default Text formatter).

It's really more an expression of a long-term "I wish...", partnered with my new `Pod::To::Man` module.  If it ends up combined with other work, or simply becomes inspiration for a feature in a re-written p6doc, it's done its job!

JJ:
As I said, there's some work being done on this script. Please check [it out](https://github.com/noisegul/perl6-p6doc) and add this feature there, if possible. Thanks anyway.

PR #2845, opened 2019-06-09
Adapt documentation for pairup 

lukasvalle:
Return value Seq consistent defined with commit https://github.com/rakudo/rakudo/commit/d73272e4c0f3d92ed82008eb366cfe756ae0c39b#diff-0bf1cf4b83dbeb7a105cd2e64f48fc44

### DIFF
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -339,7 +339,7 @@ Interprets the invocant as a list, and returns the last index of that list.
 
 Defined as:
 
-    multi method pairup(Any:U:)
+    multi method pairup(Any:U: --> Seq)
     multi method pairup(Any:D:)
 
 Interprets the invocant as a list, and constructs a list of


doc/Type/Any.pod6
```diff
@@ -339,7 +339,7 @@ Interprets the invocant as a list, and returns the last index of that list.
 
 Defined as:
 
-    multi method pairup(Any:U:)
+    multi method pairup(Any:U: --> Seq)
```
JJ: I really want to keep the description consistent with the source, and this is not the case here:
https://github.com/rakudo/rakudo/blob/2ff4a88b5b0018c490777aa7eceebed4b000a62a/src/core/Any-iterable-methods.pm6#L1891-L1892
When that is not the case, it's impossible to know if it's code drift or simply an error. So keeping it consistent is the only way to keep it maintenable. You can always add what it returns in the description. So I would really prefer this change to be reverted.

lizmat: Understood.

Issue #2892, opened 2019-07-08
Add non-destructive substitution to regex documentation.


tmtvl:
## The problem
The `S///` operator is documented in Operators, but not in Regexes.

## Suggestions
Add a small section on it to the Substitution chapter of the Regexes documentation.

tmtvl:
Fixed by #2894.

PR #2894, opened 2019-07-08
Add the `S///` section from operators to regexes

tmtvl:
## The problem
The S/// operator is documented in Operators, but not in Regexes.

See #2892 

## Solution provided
Copied the docs from Operators over.


tmtvl:
Thanks!

Issue #2668, opened 2019-03-13
Immediately view small changes in browser without rebuilding all pages


hakonhagland:
Say I make a small change to one of the `*.pod6` files in the `doc` folder, do I have to rebuild the
whole documentation with `make html` to view the change in my browser at http://127.0.0.1:3000/ ?

It seems it should be possible somehow to make small changes and see the result immediately without having to wait for the build (takes several minutes), but I could not find out how.

JJ:
It should, shouldn't it? Yet it's not possible. We are into a (slow) process of refactoring all the documentation tooling, and that would be one of our requirements. See #1823 and many other related issues. It's also an project in the Google Summer of Code, so we should expect to have something by Fall this year.

AlexDaniel:
Reopening because the issue is still there.

treyharris:
This should be left open until it is possible to do an incremental build.

antoniogamiz:
You can check the `update` option :D.

treyharris:
@antoniogamiz — Can you document how to use that option in
https://github.com/perl6/doc/blob/616e6b0e0afa4f0b95afb295bac48ced90fa6d50/README.md#L83

and/or in
https://github.com/perl6/doc/blob/616e6b0e0afa4f0b95afb295bac48ced90fa6d50/CONTRIBUTING.md#L250

and then we can close this out?

JJ:
@treyharris it's not deployed yet here.

coke:
This is now a problem for doc-website: see https://github.com/Raku/doc-website/blob/main/README.md#building-the-documentation-locally

PR #2912, opened 2019-07-21
Update issue templates for #2904

### LINKED ISSUE #2904: New GitHub issue and PR template
## The problem

If you go `.github` directory you will see this warning:

## Suggestions

Update the template to match the new format.

treyharris:
This is WIP for discussion. Closes #2904 when accepted.

Two things (at least) need to be added:

1. A pull request template
2. Where we should point people who have issues with the language that they are having the impulse to express via a doc fix. (There isn’t a one-size-fits-all answer here, since some things will go to rakudo and some to perl6 for instance, but a general rule of thumb should be included in the templates, I think?)


.github/ISSUE_TEMPLATE/author-bug-report.md
```diff
@@ -0,0 +1,10 @@
+---
```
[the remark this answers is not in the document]
AlexDaniel: So my suggestions in this review will be about reducing the templates to bare essentials.

.github/ISSUE_TEMPLATE/clarification-request.md
```diff
@@ -0,0 +1,74 @@
...
+  should be in a "New material" request instead.
+
+  -->
+  **Found in URL:** URL-HERE
+<!--
+  Also include the commit hash shown at the bottom of your browser
+  page, so we know what version you’re looking at. Better yet, include
+  the entire “From FILE.pod6 at perl6/doc on GitHub, commit a0b1c2d.”
+  line you’ll find as the first line of every doc page’s footer. -->
+**Found at commit:** COMMIT-OR-FROM-LINE-HERE
```
AlexDaniel: Just why? The vast majority of issues are reported for the latest version of the docs (which is visible in the web version). Specifying the commit is an extra step that is not very useful.

.github/ISSUE_TEMPLATE/clarification-request.md
```diff
@@ -0,0 +1,74 @@
...
+## Suggestions
+
+
+<!--
+  If applicable, tell us in "Suggestions" what could be done to solve
+  the problem, such as "Rephrase the description" or "Change the
+  example program so it doesn't introduce two unrelated concepts at
+  once."
+
+  Please note we do not have the ability to add diagrams at present
```
AlexDaniel: Remove this. Just because we're unable doesn't mean that people shouldn't suggest to do that.

.github/ISSUE_TEMPLATE/clarification-request.md
```diff
@@ -0,0 +1,74 @@
...
+  aside from the type graphs.
+
+  If you don't know how to fix the issue (this _is_ a documentation
+  site and we expect it to be used by people who are not expert!),
+  please don't worry about exact fixes. However, even a vague
+  description of what would help will help us help you to help
```
AlexDaniel: help help help help?

.github/ISSUE_TEMPLATE/clarification-request.md
```diff
@@ -0,0 +1,74 @@
...
+  everyone!
+
+  On the other hand, if you feel you know _exactly_ how to fix the
+  issue, feel free to submit a pull request if you're comfortable
+  doing so--see the CONTRIBUTING.md doc for details. We welcome
```
AlexDaniel: Replace `--` with `–` (we ♥ unicode).

.github/ISSUE_TEMPLATE/new-material-request.md
```diff
@@ -0,0 +1,62 @@
...
+  additional multi variant or a tweak to a method parameter. We
+  only document the language as it is today.
+
+  If you're not sure where your request could be best dealt with,
+  either provide the URL where you think it should be, provide
+  multiple possible URLs, or delete the line. If you just have an idea
+  of the section ("Language," "Types", etc.) include that in "The
+  problem" below.
+
+  Please don't provide a URL you wish existed that does not -- if you
```
AlexDaniel: Remove this. I don't think it's common enough to warrant a warning, and even if a URL is suggested it doesn't hurt anything.

.github/ISSUE_TEMPLATE/new-material-request.md
```diff
@@ -0,0 +1,62 @@
...
+  Please describe the content you feel needs to be added. As much
+  specificity as possible is appreciated. For example, if you thing a
+  new code example is what is called for, or new text, or both, or a
+  new tutorial, say so.
+
+  Please note we do not have the ability to add diagrams at present
+  aside from the type graphs.
+
+  If you don't know how to fix the issue (this _is_ a documentation
+  site and we expect it to be used by people who are not expert!),
```
AlexDaniel: > (this _is_ a documentation site and we expect it to be used by people who are not expert!)

Remove this.

.github/ISSUE_TEMPLATE/operational-issues.md
```diff
@@ -0,0 +1,14 @@
+---
+name: Operational issues
+about: Website and hosting-related issues
+title: ''
+labels: site
+assignees: ''
+
+---
+
+<!-- This issue type is mostly for contributors and developers. If you are seeing an issue with the website and you’re not involved in this project, you probably want a different issue type.
```
AlexDaniel: Meh. Remove this.

.github/ISSUE_TEMPLATE/operational-issues.md
```diff
@@ -0,0 +1,14 @@
...
+
+Also, note that the site is available in at least two places: https://docs.perl6.org/ and https://perl6docs.github.io/ (and the latter will usually be more up to date).
```
AlexDaniel: What is https://perl6docs.github.io it's the first time I see it.

.github/ISSUE_TEMPLATE/something-is-incorrect.md
```diff
@@ -0,0 +1,67 @@
...
+labels: docs
+assignees: ''
+
+---
+
+<!--
+  Feel free to delete anything that definitely does not apply to your
+  issue, as well as these comments. (Or don’t, it’s up to you.)
+
+  FIRST!: Check the site at https://perl6docs.github.io/ rather than
```
AlexDaniel: OK, I guess that's something new?

.github/ISSUE_TEMPLATE/clarification-request.md
treyharris: I’ve been reading all the docs issues for years—its the only perl6 repo whose issues I have diverted into my main inbox—and I tried to address issues that I felt came up with some frequency.

In this case, due to the lag from manual intervention between committing and presentation, I’ve seen the following happen all too often (and it’s very time wasting):
1. Person commits a change
2. Website gets rebuilt
3. Author notices an issue, often indexing or formatting (whether because they missed it on their private build or didn’t do one and saw it only once it went up on the site) and commits a fix
4. Time passes without a new build getting deployed
5. Somebody else posts an issue about the indexing or formatting issue
6. A third contributor, trying to be helpful, looks at the current source—not realizing it’s been corrected in master—and makes an unnecessary change to try to force the fix
7. Confusion ensues

So that’s the “just why”... if you think the cost-benefit analysis is in favor of the time saved by most issue submitters not copying and pasting a line from the footer over the well-meaning hyper-correctors’ time doing deckchair rearrangement, you’re probably less biased than me in making that call, since I’ve been on the pointy end of this particular stick a number of times and wished I’d known the submitter was looking at an old version of the site.

But that’s the why—I want to be sure that you’re aware of this and are making an informed cost-benefit call here rather than just looking for places to simplify.

.github/ISSUE_TEMPLATE/author-bug-report.md
```diff
@@ -0,0 +1,10 @@
...
+name: Author bug report
+about: For contributors and developers
```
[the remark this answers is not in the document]
treyharris: By default, there will be an extra option (that, you should be warned from experience with other repos, is very easy to miss because it looks like a tip-and-trick text rather than a big green button) that presents no template at all.

My only reason for including this besides visibility was because then it’ll get the `bug` label pre-assigned. If you don’t think that’s worth it, I can drop it. But when you write

>I'm pretty sure contributors and developers know how to submit bug reports?

when I see a file only containing
```
<!-- This is for bugs involving the software used to build the docs.
If you are experiencing an issue viewing the docs, please use
another issue type. -->
```
which isn’t giving instructions or questioning how contributors and developers submit bug reports, so maybe we’re not looking at the same thing?

.github/ISSUE_TEMPLATE/new-material-request.md
treyharris: > Remove this. I don't think it's common enough to warrant a warning, and even if a URL is suggested it doesn't hurt anything.

It isn’t common, but the point was to reinforce that it’s okay to delete the line entirely when it isn’t applicable and that we’d rather that be done than presenting a URL filled in because it’s a field and people fill fields in. OTOH, issues where people don’t really give us a clue where they’re referring to because they think their English description of the location will be specific enough when it isn’t _is_ common, so the field’s existence is useful.

.github/ISSUE_TEMPLATE/operational-issues.md
treyharris: Remove the entire “operational” issue type, or the HTML comment? Your rationale?

.github/ISSUE_TEMPLATE/operational-issues.md
treyharris: It’s referenced in several places, such as:

https://github.com/perl6/doc/blob/58b9ef05015db6b58d7300e967eb31fb9020c0f6/README.md#L5-L9

.github/ISSUE_TEMPLATE/something-is-incorrect.md
treyharris: It’s referenced in several places, such as:

https://github.com/perl6/doc/blob/58b9ef05015db6b58d7300e967eb31fb9020c0f6/README.md#L5-L9

.github/ISSUE_TEMPLATE/new-material-request.md
treyharris: Done.

.github/ISSUE_TEMPLATE/clarification-request.md
treyharris: Even in a setting that will usually be viewed in monospace text? (I ask because I actually manually changed those _back_ from `—` em dashes to `--` because I have that transformation set up in my editor automatically.

.github/ISSUE_TEMPLATE/clarification-request.md
treyharris: I’m not sure how to respond to a criticism of a joke.... If you think there shouldn’t be humor even in places that aren’t very visible then I’ll rewrite it.

.github/ISSUE_TEMPLATE/clarification-request.md
treyharris: Done.

.github/ISSUE_TEMPLATE/clarification-request.md
JJ: That does not happen so often, actually...

.github/ISSUE_TEMPLATE/clarification-request.md
AlexDaniel: The sentence is just hard to read.

JJ:
I'm commenting on the #2904 issue. Can we please discuss what's the issue there while we leave this for discussing the solution? TL;DR: I think what we need only is to update issue and PRs templates. New issue templates should be proposed, and discussed, separately.

.github/ISSUE_TEMPLATE/clarification-request.md
treyharris: > That does not happen so often, actually...

As you’re the one who has been doing the pushes, you may not have noticed how often it does happen. Looking at a git-blame of the omnibus docs (like control or operators) walking through a few levels deep, you’ll see how regressions and re-fixes (especially cargo-cultish ones involving indexing or unusual formatting corner cases) have happened in quick succession a few times a year.

Again, that may not be often enough to merit asking someone to mention the revision they noticed a problem in, but I, for one, would have noticed formatting or indexing issues (which are frequently hard to verify from inspection) had already been fixed in master when the issue was filed at least twice in recent memory if I saw it was a week-old revision and checked if the current one still had the issue. (In one case, I even went so far as committing the effective null-op, thinking I had fixed something, until I went to try to fix the “root cause” and discovered the “underlying issue” didn’t exist.)

If we always built the docs on the current rev whenever an issue came in, so we could see it for ourselves, that would be different. But sticking a cycle of building the docs at one revision, then making changes and rebuilding them again, is a lot of friction until incremental updating is easy.

(In another ticket #2668 I see that it may soon be easy; if that’s the case, then once it’s deployed and documented, I don’t think asking for a revision where the issue was noticed would be necessary anymore. If that’s soon, then I’ll just remove this field and resolve.)

.github/ISSUE_TEMPLATE/clarification-request.md
treyharris: > The sentence is just hard to read.

If you’re referring to the sentence with the parenthetical—PTAL¹, as I’ve removed it. If you’re referring to the (edited) jokey sentence, _“However, even a vague description of what would help
  will help us help you to help everyone!”_ then I’d just point out that _was_ the joke, and I don’t think it would honestly confuse anyone. But I’ll cut it to, _“...even a vague description of what would help is useful”_ if you truly think it’s confusing in a meaningful way (perhaps to less fluent English speakers?).
—
¹ PTAL: _(acronym)_ Please take another look.

.github/ISSUE_TEMPLATE/clarification-request.md
treyharris: Can I resolve this? If someone cuts-and-pastes from the comment into the issue text, GitHub-Flavored Markdown converts the `--` sequence into an em dash.

But in monospaced text such as in an editor where one’s likely to come across an HTML comment, an em dash in most fonts is fairly indistinguishable from other dashes. While it _can_ be distinguished from a hyphen if you’re looking for it, when seeing 

    this-text

vs.

    this—text

in isolation, it’s harder to tell.

As a prolific em dash user in my day-to-day writing, I don’t tolerate double-hyphens in final typeset (proportional) output, but in plain monospaced text such as this, I think it’s preferable for readability. (A totally separate topic, but I’d argue that in POD6, `--` and `—` should _both_ render as an em dash in proportionally-spaced output and as `--` in monospaced output.)

JJ:
I'm afraid I'm closing this for the time being. Not clear so many templates are needed, or if they are, that they should be done this way. Let's open a meta issue in this repo to discuss this. I thank @treyharris anyway for their disposition.

PR #2922, opened 2019-07-27
Some redactional changes and corrections

threadless-screw:
Attempt to clarify, prettify, and correct some matters in the regexes-section, most notably the incorrect reference to $?NL in the discussion of the backslashed character class \n.

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -228,7 +228,7 @@ Examples of word characters:
 
     Class    | Shorthand | Description
     =========+===========+=============
-    <alpha>  |           | Alphabetic characters including _
+    <alpha>  |           | Alphabetic characters plus underscore (_)
     <digit>  | \d        | Decimal digits
     <xdigit> |           | Hexadecimal digit [0-9A-Fa-f]
     <alnum>  | \w        | <alpha> plus <digit>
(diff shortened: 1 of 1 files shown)


doc/Language/regexes.pod6
```diff
@@ -39,25 +41,25 @@ Example of difference between C<m/ /> and C</ /> operators:
...
 
     if 'properly' ~~ / perl / {
         say "'properly' contains 'perl'";
     }
 
-Alphanumeric characters and the underscore C<_> are matched literally. All
-other characters must either be escaped with a backslash (for example, C<\:>
-to match a colon), or be within quotes:
+Alphanumeric characters, including the underscore C<_> which is considered
+alphabetic, are matched literally. All other characters must either be escaped
```
lizmat: This feels wrong, because it might give the impression that `dd "_" ~~ /<:alpha>/` would give a `Match`.  I would remove the "which is considered alphabetic" part.

threadless-screw: I understand what you are saying. At the same time, `"_".match(/ <alpha> /)' does yield a Match, which is why the current text confuses me.  But I see now that <alpha> is actually defined as "Alphabetic characters including _", so there is consistency. I'll revert the proposed change.

PR #2933, opened 2019-08-03
Proofread several documents

uzluisf:
about.pod6, community.pod6, faq.pod6 and tables.pod6

Changes include minor rewording, text reflow, replacing indented code samples
with =code directives-wrapped code samples, order book list in [faq.pod6](https://docs.perl6.org/language/faq#Can_I_get_some_books_about_Perl_6?) alphabetically, etc.

### DIFF
--- a/doc/Language/about.pod6
+++ b/doc/Language/about.pod6
@@ -4,18 +4,20 @@
 
 =SUBTITLE Meta-documentation
 
-This document collection represents the on-going effort to document the Perl 6 programming
-language with the goals of being: comprehensive; easy to use; easy to
-navigate; and useful to both newcomers and experienced Perl 6
-programmers.
+This document collection represents the on-going effort to document the
+Perl 6 programming language with the goals of being: comprehensive,
+easy to use, easy to navigate, and useful to both newcomers and experienced
+Perl 6 programmers.
 
 An HTML version of the documentation is located online at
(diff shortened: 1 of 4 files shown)


AlexDaniel:
Tests are failing:
```
t/02-pod-valid.t ........6/392# There's been this error in file: doc/Language/about.pod6
# ===SORRY!=== Error while compiling /test/doc/Language/about.pod6
# Expected "=end pod" to terminate "=begin pod
#
# There's been this error in file: doc/Language/about.pod6
# "; found "=end code" instead.
# at /test/doc/Language/about.pod6:97
# ------> =end code⏏<EOL>
```

JJ:
Yep, I fixed that before. You need to be careful when deindenting code that includes pod.

uzluisf:
> We don't force any kind of style there: they can use =for, =begin code or simple indentation; and those can use indentation too.

While proofreading some of the documents, I came across 4-space indented code samples that were rendered as code but still with four spaces and code samples wrapped in `=code` directives that were still 4-space indented. I figured that using `code` directives (both `for` and `=begin`/`=end`), instead of indentation, was the most sensible and meaningful thing (at least for someone reading the pod files) to do.

> Unless it makes the code longer than 80 characters, it's probably better to leave it alone.

I'll be more careful moving forward ;-).

PR #2934, opened 2019-08-03
Correct all methods' signatures in Metamodel::EnumHOW's page

Kaiepi:
None of its methods actually specify what type the object the method is
called on must be.

I just want to make sure this matches how the rest of the docs' signatures are written before pushing.


Kaiepi:
If the object's type should be included despite the method not actually specifying it, this still needs correction anyway since its type would never be `Metamodel::ClassHOW` without a custom meta class that extends both.

PR #2947, opened 2019-08-11
Improve readability.  Adjust token `letter` indentation.

gdonald:
## The problem
1) A few of the sentences on the page didn't read that well.
2) The token `letter` indentation was different than a previous example.

## Solution provided
1) Adjust a few words and phrases here and there.
2) Adjust token `letter` indentation to match previous example.


AlexDaniel:
Looks good to me, thanks!

PR #2952, opened 2019-08-14
Docs for \c, \C, \x and \X

ajs:
These sequences were only documented within enumerated character classes, but can appear just like `\w` and friends as top-level atoms, and so should be documented with them. Also, the documentation even within character classes was pretty thin, so this seeks to rectify that situation.

**Note** I'm having trouble with the build system, so I can verify that they pass through the parser, but can't actually get to the stage of writing html...

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -282,6 +282,51 @@ Examples of word characters:
     0409 Љ CYRILLIC CAPITAL LETTER LJE
     =end code
 
+=head3 X<C<\c> and C<\C>|regex,\c;regex,\C>
+
+C<\c> takes a parameter delimited by square-brackets which is the name
+of a Unicode character as it appears in the
+L<Unicode Character Database (UCD)|https://unicode.org/ucd/> and matches
+that specific character. For example:
+
+    'a.b' ~~ /\c[FULL STOP]/ and say ~$/;    # OUTPUT: «.»
+
+C<\C> matches a single character that is not the named Unicode character.
+
+Note that the word "character" is used, here, in the sense that the UCD
+does, but because Perl 6 uses L<NFG|/language/glossary#NFG>, combining
+code points and the base characters to which they are attached,
+will generally not match individually. For example if you compose
+C<"ü"> as C<"u\x[0308]">, that works just fine, but matching may surprise
+you:
+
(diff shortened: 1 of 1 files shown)


ajs:
Found a bug looking over my diffs. Re-submitted.

JJ:
I'm not really sure about this. They behave exactly [as should be expected from their literal use](http://docs.p6c.org/language/unicode#index-entry-%5Cc%5B%5D_unicode_name). Wouldn't this be shorter if we just said that?

doc/Language/regexes.pod6
```diff
@@ -282,6 +282,51 @@ Examples of word characters:
     0409 Љ CYRILLIC CAPITAL LETTER LJE
     =end code
 
+=head3 X<C<\c> and C<\C>|regex,\c;regex,\C>
+
+C<\c> takes a parameter delimited by square-brackets which is the name
+of a Unicode character as it appears in the
+L<Unicode Character Database (UCD)|https://unicode.org/ucd/> and matches
+that specific character. For example:
```
JJ: Maybe simply make a reference to the [literal description](http://docs.p6c.org/language/unicode#Entering_unicode_codepoints_and_codepoint_sequences)

ajs: I'd like to at least give an example, here. It would also be nice if one or both of these could link to the UCD, which is the canonical source of these names...

ajs:
In general, the issue is workflow. People come here to learn how regexes are used.

But there are three ways `\c` and friends are used:

* In character classes where they add a codepoint to the matched codepoints
* In general regex bodies where they match a codepoint
* In quoted strings where they insert the given codepoint

They each need to be documented, and each has their own odd issues. The NFG gotcha here is that matching combining codepoints with `\x` will ***almost never work***.

doc/Language/regexes.pod6
JJ: Please do that. Point is that in the case of \c and \x it's not acting in a way that's different from any other (maybe quoted) character.

JJ:
Maybe document it as a trap, then?

JJ:
Will you make any additional changes to this? I can merge it and do the changes myself, if that's what you prefer.

ajs:
If you are anxious to get it off the list, feel free. I keep inventing new things to do with my very few tuits, but this remains on the list.

PR #2960, opened 2019-08-20
Regexes

threadless-screw:
## The problem
The current Lexical Conventions section does not comprise a systematic discussion of anonymous and named regex definition syntax, and neither of common regex matching syntax.

## Solution provided
Extensive discussion of applicable syntax.


JJ:
This seems to be in conflict with latest changes by @tbrowder. Please
update your clone and update this accordingly.

PR #2968, opened 2019-08-25
Document Metamodel::Versioning

Kaiepi:

### DIFF
--- a/doc/Type/Metamodel/Versioning.pod6
+++ b/doc/Type/Metamodel/Versioning.pod6
@@ -1,0 +1,75 @@
+=begin pod :kind<Type> :subkind<role> :category<metamodel>
+
+=TITLE role Metamodel::Versioning
+
+=SUBTITLE Metaobjects that support versioning
+
+   role Metamodel::Versioning { ... }
+
+L<Metamodel|/language/mop> role for (optionally) versioning metaobjects.
+
+When you declare a type, you can pass it a version, author, and/or API and get
+them, like so:
+
+=begin code :lang<perl6>
+class Versioned:ver<0.0.1>:auth<github:Kaiepi>:api<1> { }
+
+say Versioned.^ver;  # OUTPUT: «v0.0.1␤»
+say Versioned.^auth; # OUTPUT: «github:Kaiepi␤»
+say Versioned.^api;  # OUTPUT: «1␤»
+=end code
+
+This is roughly equivalent to the following, which also sets them explicitly:
+
+=begin code :lang<perl6>
+BEGIN {
+    class Versioned { }
+    Versioned.^set_ver:  v0.0.1;
+    Versioned.^set_auth: 'github:Kaiepi';
+    Versioned.^set_api:  <1>;
+}
+
+say Versioned.^ver;  # OUTPUT: «v0.0.1␤»
+say Versioned.^auth; # OUTPUT: «github:Kaiepi␤»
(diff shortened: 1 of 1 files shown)


doc/Type/Metamodel/Versioning.pod6
```diff
@@ -0,0 +1,75 @@
...
+BEGIN {
+    class Versioned { }
+    Versioned.^set_ver:  '0,0,1';
+    Versioned.^set_auth: 'github:Kaiepi';
+    Versioned.^set_api:  IntStr.new: 0, '0';
+}
+
+say Versioned.^ver;  # OUTPUT: «0.0.1␤»
+say Versioned.^auth: # OUTPUT: «github:Kaiepi␤»
+say Versioned.^api:  # OUTPUT: «0␤»
```
JJ: This seems to error, and also output is different:
```
0,0,1
0
Too many positionals passed; expected 2 arguments but got 3
  in block <unit> at test-versioned.p6 line 8
```
You sure about the commas?

doc/Type/Metamodel/Versioning.pod6
```diff
@@ -0,0 +1,75 @@
...
+=end code
+
+This is roughly equivalent to the following, which also sets them explicitly:
+
+=begin code :lang<perl6>
+BEGIN {
+    class Versioned { }
+    Versioned.^set_ver:  '0,0,1';
+    Versioned.^set_auth: 'github:Kaiepi';
+    Versioned.^set_api:  IntStr.new: 0, '0';
```
JJ: Maybe use <0> here?

doc/Type/Metamodel/Versioning.pod6
```diff
@@ -0,0 +1,75 @@
...
+When you declare a type, you can pass it a version, author, and/or API and get
+them, like so:
+
+=begin code :lang<perl6>
+class Versioned:ver<0.0.1>:auth<github:Kaiepi>:api<0> { }
+
+say Versioned.^ver;  # OUTPUT: «0.0.1␤»
+say Versioned.^auth: # OUTPUT: «github:Kaiepi␤»
+say Versioned.^api:  # OUTPUT: «0␤»
+=end code
```
JJ: This is not working either.

Kaiepi:
Sorry, there's a couple issues with the unit tests that I haven't reported yet that prevent me from being able to run `xt/examples-compilation.t` and `t/02-pods-valid.t`. I'll comment again once Travis is happy

doc/Type/Metamodel/Versioning.pod6
```diff
@@ -0,0 +1,75 @@
...
+say Versioned.^api;  # OUTPUT: «1␤»
+=end code
+
+=head1 Methods
+
+=head2 method ver
+
+    method ver($obj)
+
+Returns the version of the metaobject, if any, otherwise returns L<Mu|/type/Mu>.
```
vrurg: With recent changes something like the following could be added to this:

> For CORE classes the method would report language version they were compiled with. For example:
>
> ```
> use v6.d; say PseudoStash.^ver; # 6.c because this is where the class is actually defined.
> ```
>
> ```
> use v6.e.PREVIEW; say PseudoStash.^ver; # 6.e
> ```

Actually, the version is set by `set_ver` method, but I'm not sure if it worth mentioning here because it is only happens when `CORE`s are compiled and thus is only relevant to core developers.

PR #2990, opened 2019-08-30
Ruby fixes

sshaw:
## The problem

Ruby examples are slightly off and/or incorrect.

### DIFF
--- a/doc/Language/rb-nutshell.pod6
+++ b/doc/Language/rb-nutshell.pod6
@@ -127,7 +127,7 @@ $n < 1; # Perl 6
 
 =end item
 
-=head2 C«.» Method calls, C<.send>
+=head2 C«.» Method calls, C<.public_send>
 
 Method call syntax uses a dot just like Ruby:
 
@@ -140,7 +140,7 @@ $person.name   # Perl 6
 To call a method whose name is not known until runtime:
 
 =for code :lang<ruby>
-object.send(methodname, args);  # Ruby
+object.public_send(methodname, args);  # Ruby
 =for code
 my $object; my Str $methodname; my @args; ...;
 $object."$methodname"(@args);   # Perl 6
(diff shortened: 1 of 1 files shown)


doc/Language/rb-nutshell.pod6
```diff
@@ -140,7 +140,7 @@ $person.name   # Perl 6
 To call a method whose name is not known until runtime:
 
 =for code :lang<ruby>
-object.send(methodname, args);  # Ruby
+object.public_send(methodname, args);  # Ruby
 =for code
 my $object; my Str $methodname; my @args; ...;
 $object."$methodname"(@args);   # Perl 6
```
sshaw: Ruby's `#send` works with private and public methods. New to Perl 6, but I'm pretty sure this only works with public methods.

doc/Language/rb-nutshell.pod6
```diff
@@ -263,7 +263,7 @@ say %calories« $keys ».join(','); # Perl 6, interpolated split
 I<Key/value-slicing>
 
 =for code :lang<ruby>
-say calories.slice('pear', 'plum').join(','); # Ruby, with ActiveRecord
+puts calories.slice('pear', 'plum')
```
sshaw: `#slice` returns a `Hash`, with key value pairs for `'pear'` and `'plum'` so calling `#join` results in an error. 

Not sure what the Perl 6 equivalent is –or if there is one. Let me know and I will make the appropriate updates.

JJ: Simply using the indexing operator <>: calories<pear plum>

sshaw: I don't follow, sorry. 

```
~ >ruby -e'p ({a:1, b:2, c:3}).slice(:a, :c)'
{:a=>1, :c=>3}
```

To get that, seems one must do:

```~ >perl6 -e'my %h = (a => 1, b => 2, c => 3); say %h<a c>:kv'
(a 1 c 3)
```

But, since that's a `List`, one must assign to a `Hash`: `my %h2 = %h<a c>:kv'`. Maybe there's a way without assigning the list to `%h2`? Maybe there's a way without having to add `:kv` which, if you're coming to Perl 6 from Ruby may not be a good selling point.

JJ: Add `.Hash`:

    perl6 -e'my %h = (a => 1, b => 2, c => 3); say %h<a c>:kv.Hash'

sshaw: Great, thanks! PR updated. 🦋

PR #3027, opened 2019-09-18
Fixing a very small error in Baggy

zostay:
## The problem

The example output shows `(b b b)`, but it ought to be `(a b b)`  or some permutation of that.

## Solution provided

I changed the example output to `(a b b)`. This is still slightly misleading as that's only one of 3 possible outcomes, but it's better than `(b b b)` which is none of the possible outcomes.


Altai-man:
What rakudo version do you use?

```
➜  ~ perl6 --version
This is Rakudo version 2019.07 built on MoarVM version 2019.07
implementing Perl 6.d.
➜  ~ perl6 -e 'my $n = ("a" => 0, "b" => 1, "b" => 2).BagHash; say $n.kxxv;'
(b b b)
➜  ~ 

```

JJ:
Also, `a` has weight 0, so it makes sense it does not show up.

coke:
Based on comments from September (and another test in my sandbox), rejecting patch, as current "b b b" seems correct.

PR #3048, opened 2019-10-08
Roles can also be mixed in using 'is'.

gdonald:
## The problem
The docs say a Role can be mixed into a class using "does" but fails to mention that "is" also works.

## Solution provided

Update the docs to clarify how "is" also works to mix a Role into a class.

### DIFF
--- a/doc/Language/objects.pod6
+++ b/doc/Language/objects.pod6
@@ -742,8 +742,8 @@ and code reuse within objects.
 
 X<|does>
 Roles use the keyword C<role> preceding the name of the role that is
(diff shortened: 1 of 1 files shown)


jnthn:
The semantics of `is` with a `role` are quite different from those offered by `does`. With `is`, a class is punned from the role, and then inherited from. Thus, there is no flattening composition, and none of the safeties it provides.

gdonald:
Thank you for the comment.  Relying heavily on your explanation I updated my PR to explain the difference a bit more.

lizmat:
s/none of the safeties C<does> provides./none of the safeties **which** C<does> provides./

??

Issue #3046, opened 2019-10-04
Path to Raku (Docs)


JJ:
## The problem

This issue has been created to discuss what needs to be done once, or when, the name has been changed to Raku, that is, once perl6/problem-solving#81 is accepted, perl6/problem-solving#89 is merged.

## Suggestions

#2951 has been proposed, however it's in conflict now and I'm not sure that direct approach is the way to go. I propose the following roadmap.

* [x] As soon as we get the docs.raku-lang.org domains going, we create a "perl6" branch of the documentation. 
* [x] Substitute Perl 6 for raku, perl6 for raku, and other related changes in the master branch of the documentation, that will be deployed to docs.raku.org
* [x] Changes in the `perl6` branch:
    * [x] Eliminate the possibility of editing, since those pages are not current any more.
    * [x] Add a footer with a deprecation notice in the "perl6" branch, as well as an automatically generated link to the new, equivalent page. These pages *will be deployed to docs.perl6.org*.
* [x] The "master" branch will include a note, maybe in non-visible parts like the header, indicating that Raku was formely called Perl 6. 
4. Let's monitor a series of search strings like "perl 6 documentation" and "raku documentation" to see where they lead. Ideally, they should both return docs.raku-lang.org eventually.
5. Change occurences of "Perl 5" to simply "Perl".
    1. Include file names here.
5. The tooling will have to be changed to accommodate the new extensions as well as the old ones, as well as any change that's done to the name of the classes (there are lots of Perl6:: and Pod:: )
6. Once tooling _and_ documentation does not fail, is findable, *then* change the *old* docs.perl6.org pages to a 301 redirection to the new URL.

There are several reasons for taking this cautious approach:
1. People will keep looking for Perl 6 for some time. We want them to be able to find the *old* documentation as well as the new one.
2. We need the two sets of pages for the new ones to gather SEO juice. We need to be extremely cautious about this, since we don't want people to lose the *old* perl6 pages and not be able to find either the new ones. 
3. Changing the tooling will definitely take some time, and we have still not finished the transition to the new version.

Also, if there's a good time to overhaul the appearance of the pages, this is it. So we might as well make the new pages more dashing, add more pizzazz, or something like that. New branding with old pages is probably not a good idea.

(I don't know if this is more adequate for the problem-solving repo, since it involves also infrastructure; if that's the case, just move it there or tell me so and I'll do it)

Altai-man:
`Substitute Perl 6 for raku, perl6 for raku`
[...]

PR #3056, opened 2019-10-15
Replace instances of Perl 6 with Raku and more

uzluisf:
* Perl 6 -> Raku
* perl6 -> raku
* .p6 -> raku
* .pod -> rakudoc
* .pm6 -> rakumod
* Variables containing PERL6/PERL ($*PERL6 -> $*RAKU, etc.)


AlexDaniel:
I'll close this to make sure nobody merges it ahead of time.

You can't change extensions because no tools support them yet. You can't change env vars for the same reason. You can't change emails because I don't believe there's forwarding set up already. You can't change the executable because most distro packages don't have raku executable.

All that will be possible soon, but not now. We have to fix the tooling first, and the changes in documentation will follow. It has been 1 day since the rename, so of course most raku pieces are not in place yet.

AlexDaniel:
Feel free to resubmit with just the plaintext changes (“Raku compiler”, “rakuish”, etc.)

JJ:
There's a roadmap here #3046 . I would really be grateful if we stuck to it. @AlexDaniel is right, for the time being just *the name of the language* has changed, and there's issue #3056 to solve that. Revising what has been done there, and finding other instances that haven't been substituted automatically is, for the time being, the next step.

uzluisf:
Sorry I couldn't get back to you earlier and you're totally right. I got a bit carried away 😅.

JJ:
@uzluisf thanks anyway, I appreciate the effort :+1:

PR #3066, opened 2019-10-20
Wrap inline code examples with C<>, remove unneeded capitalization, ...

uzluisf:
..., add example, etc.

### DIFF
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -890,7 +890,7 @@ for @array -> $first, $second {
 }
 =end code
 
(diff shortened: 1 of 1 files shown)


doc/Language/5to6-nutshell.pod6
```diff
@@ -890,6 +890,17 @@ for @array -> $first, $second {
 }
 =end code
 
+For instances in which the number of elements iterated over isn't a multiple
```
Altai-man: I am almost sure "For cases" is more natural than "For instances".
Maybe "For cases when the number of..."?

Altai-man:
Looks better with your changes. Thank you very much!

PR #3074, opened 2019-10-25
Pursue .WHAT removal

stoned:
- Sync output with code example
- Update code example description
- Advise .^name usage

## The problem
.WHAT was removed from an example in snippet in doc/Language/classtut.pod6 a year ago in favor of .^name: this is a followup

### DIFF
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -816,11 +816,10 @@ The output might look like this:
 =begin code :lang<output>
 It's an employee
 Not a geeky cook
-(Programmer)
+Programmer
 Programmer.new(known_languages => ["Perl", "Python", "Pascal"],
         favorite_editor => "gvim", salary => "too small")
 code_to_solve, known_languages, favorite_editor
-Programmer
 =end code
 
 The first two tests each smartmatch against a class name. If the object is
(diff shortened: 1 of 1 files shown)


doc/Language/classtut.pod6
```diff
@@ -816,20 +816,18 @@ The output might look like this:
...
 =end code
 
 The first two tests each smartmatch against a class name. If the object is
 of that class, or of an inheriting class, it returns true. So the object in
 question is of class C<Employee> or one that inherits from it, but not
 C<GeekCook>.
 
-The C<.WHAT> method returns the type object associated with the object
-C<$o>, which tells us the exact type of C<$o>: in this case C<Programmer>.
+The call C<$o.^name> tells us the type of C<$o>: in this case C<Programmer>.
```
jnthn: I think here it's useful to mention both of them.

AlexDaniel: @jnth The code doesn't use `.WHAT` so that's why the output and the description is being tweaked, I think you mean that we should revert the original commit?

AlexDaniel: See https://github.com/perl6/doc/commit/c4017865012453cd1efbd7a438b58d9e07a3cfdf

PR #3076, opened 2019-10-27
Fix broken links

tusindfryd:


doc/Language/create-cli.pod6
```diff
@@ -12,7 +12,7 @@ The default command line interface of Raku scripts consists of three parts:
 =head2 Parsing the command line parameters into a L<capture|/type/Capture>
 
 This looks at the values in L<@*ARGS|/language/variables#index-entry-@*ARGS>,
-interprets these according to some policy, and creates a L<Capture/type/Capture>
+interprets these according to some policy, and creates a L<capture|/type/Capture>
```
JJ: Can you please keep the capital for Capture here?

doc/Language/operators.pod6
```diff
@@ -1889,7 +1889,7 @@ is smaller than or equal to the second.
 
 Numeric less than or equal to operator.
 
-Equivalent to L«<=|routine/<=», at codepoint U+2264 (LESS-THAN OR EQUAL TO).
+Equivalent to L«<=/routine/<=», at codepoint U+2264 (LESS-THAN OR EQUAL TO).
```
JJ: I don't think this this should be changed actually

doc/Language/create-cli.pod6
tusindfryd: Done :)

doc/Language/operators.pod6
tusindfryd: Why? The first one is not working for me.

JJ: But you're changing a text-url separator (`|`) to a slash for directory separation. That one is not going to work no matter what. Maybe you should leave `|/` instead?

tusindfryd: Oh, that's right. My bad. Changed now

JJ:
Thanks for doing this, you have now permission to write in the repository. Congratulations!

PR #3100, opened 2019-11-25
Correct how you would imitate static fields in the class tutorial

Kaiepi:
## The problem
There are several problems with the "Static fields?" section of the class tutorial:
- the `Singleton` class in the example cannot be subclassed
- its `instance` method does not need to keep any state outside of the method
- there's no such thing as a static attribute; attributes do not exist on type objects

## Solution provided
Rewrite this section.

### DIFF
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -386,38 +386,27 @@ C<Str-with-ID>. Additionally, class variables declared with package scope are
 visible via their fully qualified name (FQN), while lexically scoped class
 variables are "private".
 
-=head1 Static fields?
+=head2 What about static fields?
 
-Raku has no B<static> keyword. Nevertheless, any class may declare anything
-that a module can, so making a scoped variable sounds like a good idea.
+Raku has no B<static> keyword, and attributes are exclusive to instances of
+types. Nevertheless, it is possible to imitate static fields using a stateful
+static method:
 
 =begin code
 class Singleton {
-    my Singleton $instance;
-    method new {!!!}
-    submethod instance {
-        $instance = Singleton.bless unless $instance;
-        $instance;
+    method instance(::?CLASS:U: --> ::?CLASS:D) {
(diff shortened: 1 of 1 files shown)


doc/Language/classtut.pod6
```diff
@@ -386,38 +386,27 @@ C<Str-with-ID>. Additionally, class variables declared with package scope are
...
-class HaveStaticAttr {
-    my Foo $.foo = some_complicated_subroutine;
+    method new(|) { !!! }
 }
 =end code
 
-Class attributes may also be declared with a secondary sigil – in a similar
-manner to instance attributes – that will generate read-only accessors if the
-attribute is to be public.
+C<Singleton> forbids creating instances of it using its C<new> method, and
```
JJ: Move the comma behind and,

doc/Language/classtut.pod6
```diff
@@ -386,38 +386,27 @@ C<Str-with-ID>. Additionally, class variables declared with package scope are
...
+using the C<bless> method it inherits from L<Mu|/type/Mu#method_bless>,
+creates an instance of itself once its C<instance> method gets called for the
```
JJ: it before creates

doc/Language/classtut.pod6
```diff
@@ -386,38 +386,27 @@ C<Str-with-ID>. Additionally, class variables declared with package scope are
...
+first time, which gets set to an anonymous state variable and ends up as its
+return value.
```
JJ: OK, so since state is a lexical of that method, that is inherited all across the hierarchy, they will always get the same value. While in the previous example it's a class variable that's only shared for instances of it; however, if it's changed to our, wouldn't that behave in the same way?
Point is, wouldn't this be better provided as an additional example instead of an alternative?

Kaiepi:
...not sure what I was thinking before, there's nothing wrong with the example here.

PR #3113, opened 2019-12-07
Some Raku<->Perl6 changes, some misc editing also

MasterDuke17:
Some "Perl 6"s which look like they had accidentally been changed to "Raku" changed back. Some "Perl 6"-ish words changed to their "Raku"-ish equivalent. Some misc editing fixes.

### DIFF
--- a/doc/Language/faq.pod6
+++ b/doc/Language/faq.pod6
@@ -308,8 +308,9 @@ X<|Data::Dumper (FAQ)>
 Typical options are to use the L<say|/routine/say> routine that uses
 the L<gist|/routine/gist> method which gives the "gist" of the object being
 dumped. More detailed output can be obtained by calling the
-L<perl|/routine/perl> method that typically returns an object's representation
-in L<EVAL|/routine/EVAL>-able code.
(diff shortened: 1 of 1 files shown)


doc/Language/faq.pod6
```diff
@@ -677,7 +677,7 @@ about the object deemed unimportant to understanding the essence of the object.
 
 Or phrased differently, C<$obj.Str> gives a string representation,
 C<$obj.gist> provides a short summary of that object suitable for
-fast recognition by a human, and C<$obj.perl> gives a Perlish representation
+fast recognition by a human, and C<$obj.perl> gives a Rakuish representation
```
lizmat: Perhaps mention the `.raku` method instead of `.perl` ?

MasterDuke17: I thought about it, but it's not documented yet. Guess mentioning can't hurt.

lizmat: Then we should think about documenting `.raku`, as 2019.11 does have it.

MasterDuke17: @lizmat think 8b0b7f1 resolves this?

PR #3128, opened 2019-12-17
closes #3105

### LINKED ISSUE #3105: Change the perl6 command line to raku
## The problem

Since 2019.11, raku is available as an alias, and can be used like that.

## Suggestions

Look for occurrences in tooling and documentation and change it.

veb7vmehra:


JJ:
Sorry, but you have simply performed automatic substitution, and that includes file names URLs, and so on. I am closing this issue for that reason, because I don't think it can be fixed; it's the approach that's wrong, not the individual changes.
Also, I would be grateful if in the future, PRs were more atomic, as in only one or a few files affected. It's impossible to review a PR with 311 files changed (although, if 311 files have changed, it's most probably less than awesome)

Issue #3107, opened 2019-12-03
Improve control structures tutorial


JJ:
## The problem

Check out [this suggestion by @lizmat](https://github.com/Raku/doc/commit/945c37eb9320e3f22da9eeae601909b95b615ebd#commitcomment-36244283)

## Suggestions

Try to provide an example, and change CAUTION to a note.

PR #3134, opened 2019-12-29
Fixes: #3107

threadless-screw:
Minor adaptations to the text describing the loop construct, and addition of an example illustrating use of a loop initializer variable in the outer scope after the loop has ended.

### DIFF
--- a/doc/Language/control.pod6
+++ b/doc/Language/control.pod6
@@ -890,17 +890,24 @@ C<given> can follow a statement to set the topic in the statement it follows.
 =head1 X<loop|control flow, loop>
 
 The C<loop> statement takes three statements in parentheses separated by C<;>
-that take the role of initializer, conditional and incrementer. The initializer
-is executed once and any variable declaration will spill into the surrounding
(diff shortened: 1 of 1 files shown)


doc/Language/control.pod6
```diff
@@ -890,18 +890,25 @@ C<given> can follow a statement to set the topic in the statement it follows.
...
+incrementer is executed after each iteration, and before the conditional is
+tested again.
+
+    loop (my $i = 0; $i < 10; $i++) {       # A typical loop
         say $i;
     }
 
+    my @str = "However Long".comb;          # Our very own .char routine:
+    loop (my $l = 0;;) {                    # Declare $l in outer scope
+        last if !@str[$i++]                 # and count chars until we hit an
```
lizmat: Shouldn't that be `$l` ?

threadless-screw: Sorry, my bad. Had already seen it, but apparently failed to correct it.

lizmat:
Thanks!

PR #3142, opened 2020-01-04
P6-101 text refinement + expansion of assignment operators section; closes: #2920

### LINKED ISSUE #2920: compound-assignment operations on type Any needs expounding
## The problem
compound-assignment operations permitted on Any variables are incompletely documented.

The only documentation currently is at: [https://docs.perl6.org/language/101-basics#Any_and_+=](https://docs.perl6.org/language/101-basics#Any_and_+=) and is insufficient, possibly leading an initiate through all levels of hell before s/he realizes that an unitialized value reports an error in a semantically identical but syntactically unique expression.

101 basics specifies that an uninitialized variable (Any) is treated as a value of zero when operated on with +=. So, it can be seen that
`{my $x; $x+=5}; # 5, since 0+5=5. 
`But this is not true for
`{my $x; $x = $x +5}; # error, but semantically equivalent!`

"* =" is also permitted but not addressed here or elsewhere. Further, it suffers from the same inconsistencies, i.e. "* =" defaults the value to 1 instead of 0, but, only on compound assignment. That is,
`{my $x; $x*=5}; # 5, because $x was Any and defaulted to 1`
but
`{my $x; $x = $x * 5}; # error, but looks semantically the same!`

## Suggestions
emphasize that compound-assignment is the exception to the rule coercing Any instead of producing an error and contrasting short-hand with long-hand in the same section.  [https://docs.perl6.org/language/101-basics#Any_and_+=](https://docs.perl6.org/language/101-basics#Any_and_+=)

document the behavior of  *= in the same page, distinguishing the coercion to the distinct Numeric value of 1 vs 0. [https://docs.perl6.org/language/101-basics#Any_and_+=](https://docs.perl6.org/language/101-basics#Any_and_+=)

this behavior was brought to the attention of the community via this discussion post:
[https://stackoverflow.com/questions/57210795/math-with-undefined-perl6-values](https://stackoverflow.com/questions/57210795/math-with-undefined-perl6-values)

~~~~~
irced
I am a hobbyist interested in the perl6 language. Thanks for perl6!
the following irc'ers have confirmed and contributed their observations
concerning this undocumented phenomena: jmerelo, lizmat, AlexDaniel

threadless-screw:
## The problem

The P6-101 page included imprecise statements about the operation of compound assignment operators, and the actual behavior was not documented anywhere else.

## Solution provided

The relevant text of the P6-101 page has been corrected, and the 'Assignment operators' section of the Operators page has been expanded with more detailed information about the relation between simple and compound assignment operators, and the implicit initializing behavior of compound assignment operators.

Closes #2920

### DIFF
--- a/doc/Language/101-basics.pod6
+++ b/doc/Language/101-basics.pod6
@@ -241,20 +241,33 @@ The program then counts the number of sets each player has won:
 %sets{$p2} += $r2;
 =end code
 
-The C<+=> assignment operator is a shortcut for:
+The above two statements involve the C<+=> compound assignment operator. They
+are a variant of the following two statements that use the simple assignment
+operator C<=> and that may be easier to understand at first sight:
 
 =begin code :preamble<my %sets;my $p1;my $p2;my $r1;my $r2>
 %sets{$p1} = %sets{$p1} + $r1;
 %sets{$p2} = %sets{$p2} + $r2;
 =end code
 
+For your program, the shorthand version using the C<+=> compound assignment
+operator is preferred over the longhand version using the simple assignment
+operator C<=>. This is so not only because the shorter version requires less
+typing, but also because the C<+=> operator silently initializes undefined
+values of the hash's key-value pairs to zero. In other words: by using C<+=>,
+there is no need to include a separate statement like C<%sets{$p1} = 0> before
+you can meaningfully add C<$r1> to it. We'll look at this is behavior in a bit
+more detail below.
+
+
 =head3 X<C<Any>| Any (Basics)> and X<C<+=>>
 
-C<+= $r1> means I<increase the value in the variable on the left by $r1>. In
-the first iteration C<%sets{$p1}> is not yet set, so it defaults to a special
-value called C<Any>. The addition and incrementing operators treat C<Any> as a
-number with the value of zero; the strings get automatically converted to
-numbers, as addition is a numeric operation.
+C<+= $r1> means I<increase the value in the variable on the left by $r1>. In the
+first iteration C<%sets{$p1}> is not yet defined, so it defaults to a special
+value called C<Any>. The C<+=> operator conveniently treats the undefined value
+C<Any> as a number with the value C<0>, allowing it to sensibly add some other
+value to it. To perform the addition, the strings C<$r1>, C<$r2>, etc. are
+automatically converted to numbers, as addition is a numeric operation.
 
 =head3 X<C<fat arrow>>,  X<C<pair>> and X<C<autovivification>>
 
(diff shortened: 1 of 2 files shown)


doc/Language/101-basics.pod6
```diff
@@ -241,20 +241,33 @@ The program then counts the number of sets each player has won:
...
 =head3 X<C<Any>| Any (Basics)> and X<C<+=>>
 
-C<+= $r1> means I<increase the value in the variable on the left by $r1>. In
-the first iteration C<%sets{$p1}> is not yet set, so it defaults to a special
-value called C<Any>. The addition and incrementing operators treat C<Any> as a
-number with the value of zero; the strings get automatically converted to
-numbers, as addition is a numeric operation.
+C<+= $r1> means I<increase the value in the variable on the left by $r1>. In the
+first iteration C<%sets{$p1}> is not yet defined, so it defaults to a special
+value called C<Any>. The C<+=> operator conveniently treats the undefined value
```
jonathanstowe: I appreciate this was like this before but this could be a bit misleading about the type system and the role of type objects as the  "undefined value", it might be worth introducing a bit more detail about how `Any`  will become the zero or the empty string or whatever depending on the type RHS of the operator.

doc/Language/101-basics.pod6
```diff
@@ -241,20 +241,33 @@ The program then counts the number of sets each player has won:
...
+operator C<=> and that may be easier to understand at first sight:
 
 =begin code :preamble<my %sets;my $p1;my $p2;my $r1;my $r2>
 %sets{$p1} = %sets{$p1} + $r1;
 %sets{$p2} = %sets{$p2} + $r2;
 =end code
 
+For your program, the shorthand version using the C<+=> compound assignment
+operator is preferred over the longhand version using the simple assignment
+operator C<=>. This is so not only because the shorter version requires less
```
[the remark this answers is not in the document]
JJ: maybe change it by "; this is needed..."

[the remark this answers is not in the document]
threadless-screw: I'll rephrase it...

doc/Language/101-basics.pod6
threadless-screw: I agree the current text isn't very instructive on the point of the type system, type objects and definedness. But I'm not sure how to clarify things in more detail without expanding the 101 too much, and possibly beyond its purpose. In addition I think your observation isn't necessarily related to issue # 2920, so perhaps it's better to deal with this point in a new issue.

The issue could, for instance, call attention to the fact that the 101 does not systematically introduce its reader to the Raku type system and the aforementioned concepts. I'd think that resolving the issue would require at least the following to be included to make these concepts understandable: that Raku is object oriented to the core; that everything in Raku is an object (or behaves like one), with objects being either "type objects" that define types, or "instances" that derive from type objects; that all objects have a property called definedness that indicates whether the object is an instance and represents a meaningful (initialized) value; that by default type objects are undefined while instances are defined; that (Any) is a type object that is the default value for i.a. hash values; and that compound assignment operators check whether their left operand is a type object/is undefined, and in case it is, implicitly initialize it with an instance having a value meaningful for the specific operator...

But even though this seems very useful info, it may be a bit too much for a 101 whose primary purpose seems to be to get the uninitiated up and running with Raku quickly. Or is the purpose of the 101 a different one?

JJ: No, it's not. And it's likely that there's too much information there already. Just leave it this way.

Issue #2205, opened 2018-07-22
Document overriden roles in Mixy


JJ:
## The problem

Some roles implemented in Baggy do not make sense in Mixy and are then overridden, bailing out with an error. For instance, [pick](https://github.com/rakudo/rakudo/blob/7629820461d9a9079289c04340bf8cf63a1548ca/src/core/Mixy.pm6)

## Suggestions

Translate that to the documentation to make it quite clear and avoid people using it and [being surprised](https://stackoverflow.com/questions/51464303/pick-not-supported-on-a-mix)

zoffixznet:
There's a draft in a branch https://github.com/perl6/doc/commit/c32b4b35b2 currently blocked by inconsistencies in how we fail and potential lack of spec for those failures: https://github.com/rakudo/rakudo/issues/2109

PR #3149, opened 2020-01-07
Document Mixy.pick/.grab/.kxxv

JJ:
Potential fix for D#2205 https://github.com/perl6/doc/issues/2205
pending resolution of https://github.com/rakudo/rakudo/issues/2109

## The problem

This was created by @zoffixznet some time ago. Before deleting, I guess it's fair to give it a chance to be reviewed.

## Solution provided

It's take it or leave it, we accept it as is (and then change whatever) or we reject it and the branch is deleted.


coke:
taking as is, we can review again once the rakudo issue is done.

PR #3158, opened 2020-01-13
Mention immutability of lists on the List page

uzluisf:
Issue: https://github.com/Raku/doc/issues/3156

### DIFF
--- a/doc/Type/List.pod6
+++ b/doc/Type/List.pod6
@@ -26,7 +26,7 @@ L<unshift|/routine/unshift>, L<push|/routine/push>, L<pop|/routine/pop>,
 L<splice|/routine/splice>
 and L<binding|/language/operators#index-entry-Binding_operator>.
 
-=begin code :lang<text>
+=begin code :skip-test<illustrates error>
 (1, 2, 3).shift;      # Error Cannot call 'shift' on an immutable 'List'
 (1, 2, 3).unshift(0); # Error Cannot call 'unshift' on an immutable 'List'
 (1, 2, 3).push(4);    # Error Cannot call 'push' on an immutable 'List'
@@ -40,7 +40,7 @@ A C«List» doesn't L«containerize|/language/containers» its elements, but if
 any element happens to be inside a L«Scalar|/type/Scalar» container then
 the element's contents can be replaced via an assignment.
 
-=begin code :lang<text>
+=begin code :skip-test<illustrates error>
 my $a = 'z';
 my $list = ($a, $, 'b');
 
--- a/doc/Type/List.pod6
+++ b/doc/Type/List.pod6
@@ -26,23 +26,32 @@ L<unshift|/routine/unshift>, L<push|/routine/push>, L<pop|/routine/pop>,
 L<splice|/routine/splice>
 and L<binding|/language/operators#index-entry-Binding_operator>.
 
-    (1, 2, 3).shift;      # Error Cannot call 'shift' on an immutable 'List'
-    (1, 2, 3).unshift(0); # Error Cannot call 'unshift' on an immutable 'List'
-    (1, 2, 3).push(4);    # Error Cannot call 'push' on an immutable 'List'
-    (1, 2, 3).pop;        # Error Cannot call 'pop' on an immutable 'List'
-    (1, 2, 3)[0]:delete;  # Error Cannot remove elements from a List
-    (1, 2, 3)[0] := 0;    # Error Cannot use bind operator with this left-hand side
-    (1, 2, 3)[0] = 0;     # Error Cannot modify an immutable Int
+=begin code :lang<text>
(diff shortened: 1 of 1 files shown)


doc/Type/List.pod6
```diff
@@ -17,6 +17,33 @@ Arrays to have every value of the list stored in a container.
...
 
+=head1 Immutability
+
+Lists are immutable objects, i.e., neither the number of elements in a list
+nor the elements themselves can be changed. Thus, it is not possible to use
+operations that change the list structure itself such as L<shift|/routine/shift>,
+L<unshift|/routine/unshift>, L<push|/routine/push>, L<pop|/routine/pop>,
+L<splice|/routine/splice>
+and L<binding|/language/operators#index-entry-Binding_operator>.
+
```
JJ: Maybe use =for code :skip-test<illustrates error> here

doc/Type/List.pod6
```diff
@@ -17,6 +17,33 @@ Arrays to have every value of the list stored in a container.
...
+    (1, 2, 3).shift;      # Error Cannot call 'shift' on an immutable 'List'
+    (1, 2, 3).unshift(0); # Error Cannot call 'unshift' on an immutable 'List'
+    (1, 2, 3).push(4);    # Error Cannot call 'push' on an immutable 'List'
+    (1, 2, 3).pop;        # Error Cannot call 'pop' on an immutable 'List'
+    (1, 2, 3)[0]:delete;  # Error Cannot remove elements from a List
+    (1, 2, 3)[0] := 0;    # Error Cannot use bind operator with this left-hand side
+    (1, 2, 3)[0] = 0;     # Error Cannot modify an immutable Int
+
+However, this immutability is not deep and thus any of its elements wrapped
```
JJ: Maybe rephrase? What would deep or shallow immutability be? And also, the list elements themselves are immutable. The content of those elements may vary.

uzluisf: I'm taking the definition from [core.c/List.pm6](https://github.com/rakudo/rakudo/blob/add2ec0d491218e59e6c7a517e37458e1a6c3daf/src/core.c/List.pm6#L1-L4). From my understanding, the immutability of list not being deep refers to the fact that this immutability doesn't extend to the elements of structures inside a list that containerize their elements (e.g., Arrays) or to single Scalar containers for that matter. 

Regarding the sentence "Except for the containerized elements, everything else will remain immutable though." what I intended to mean, as you state, was the content of those list elements may vary (if containerized), not necessarily that they can be changed. I was relying on the context provided by the preceding sentence but in hindsight, it's better to remove it since it might cause some confusion.

uzluisf: I've rephrased the paragraph, @JJ. Hopefully it's a little better now.

doc/Type/List.pod6
JJ: If you to this, it will not be highlighted. It's better to say that you're skipping the tests and the reason why.

doc/Type/List.pod6
```diff
@@ -17,6 +17,42 @@ Arrays to have every value of the list stored in a container.
...
+(1, 2, 3)[0]:delete;  # Error Cannot remove elements from a List
+(1, 2, 3)[0] := 0;    # Error Cannot use bind operator with this left-hand side
+(1, 2, 3)[0] = 0;     # Error Cannot modify an immutable Int
+=end code
+
+A C«List» doesn't L«containerize|/language/containers» its elements, but if
+any element happens to be inside a L«Scalar|/type/Scalar» container then
+the element's contents can be replaced via an assignment.
+
+=begin code :lang<text>
```
JJ: Same here. Please use the most adequate environment. If it's going to error, just skip-test

doc/Type/List.pod6
uzluisf: I don't think I'm understand. Do you mean something like `L<C<splice>|/routine/splice>`? This is just a list of links so I'm unsure why it would need to use `=for code :skip-test`.

JJ: Sorry, this referred to the lines below, not above...

doc/Type/List.pod6
```diff
@@ -17,6 +17,42 @@ Arrays to have every value of the list stored in a container.
...
+
+Lists are immutable objects, i.e., neither the number of elements in a list
+nor the elements themselves can be changed. Thus, it is not possible to use
+operations that change the list structure itself such as L<shift|/routine/shift>,
+L<unshift|/routine/unshift>, L<push|/routine/push>, L<pop|/routine/pop>,
+L<splice|/routine/splice>
+and L<binding|/language/operators#index-entry-Binding_operator>.
+
+=begin code :lang<text>
+(1, 2, 3).shift;      # Error Cannot call 'shift' on an immutable 'List'
```
JJ: My comment above referred to this line here, sorry.

PR #3171, opened 2020-01-19
Add NUL byte in NativeCall example

### LINKED ISSUE #3169: NativeCall and manually creating C strings
## The problem
The [NativeCall](https://docs.raku.org/language/nativecall#Passing_and_returning_values) documentation contains an example of how to manually turn a Raku string into a C string to pass it to a native function in case the lifetime of that C string must be prolonged until after the native call returns (so `Str is encoded` is not an option):

``` perl6
my $string = "FOO";
# The lifetime of this variable must be equal to the required lifetime of 
# the data passed to the C function. 
my $array = CArray[uint8].new($string.encode.list);
 
set_foo($array);
# ... 
use_foo();
# It's fine if $array goes out of scope starting from here.
```

Now, C libraries *customarily but not always* assume that a pointer ("to a string"[¹](https://www.nntp.perl.org/group/perl.perl6.users/2020/01/msg7824.html)) is terminated by a NUL character, which is not added by the above Raku code. (Even if it usually just appears after the CArray[uint8] by chance.[²](https://www.nntp.perl.org/group/perl.perl6.users/2020/01/msg7844.html))

## Suggestions

The example should be amended so that either:

- it is made clear that the function being called accepts a string that is not NUL-terminated, which is usually done by passing a pointer and a size, or
- the construction of the CArray[uint8] must include the NUL terminator, for example by using `CArray[uint8].new($string.encode.list, 0)`.

Best may be to show both and leave a small comment about the meaning of each variant, in informal terms, trusting the C experience of the reader.

## History

- <https://www.nntp.perl.org/group/perl.perl6.users/2019/12/msg7586.html>
- <https://www.nntp.perl.org/group/perl.perl6.users/2020/01/msg7822.html>

taboege:
Also clarify when/why this is necessary.

Closes #3169

### DIFF
--- a/doc/Language/nativecall.pod6
+++ b/doc/Language/nativecall.pod6
@@ -99,19 +99,33 @@ routine itself.
 Note that a C<NULL> string pointer can be passed by using the C<Str> type
 object; a C<NULL> return will also be represented by the type object.
 
-If the C function requires the lifetime of a string to exceed the function
-call, the argument must be manually encoded and passed as C<CArray[uint8]>:
+If the C function requires the lifetime of data, say some buffer of type
+C<CArray[uint8]>, passed by pointer to exceed the function call, you have
+to ensure that the backing C<CArray[uint8]> Raku object is not destroyed
+before the C function expects it.
+
+This is particularly important for strings, as the (usually preferable)
+C<is encoded>-way of passing strings creates only a temporary object that
+is automatically destroyed after the call. In this case, the argument
+must be manually encoded:
 
     use NativeCall;
-    # C prototype is void set_foo(const char *)
+    # void set_foo(const char *)
+    # Records a char pointer for later usage
     sub set_foo(CArray[uint8]) is native('foo') { * }
-    # C prototype is void use_foo(void)
-    sub use_foo() is native('foo') { * } # will use pointer stored by set_foo()
+    # void use_foo(void)
+    # Uses the pointer stored by set_foo()
+    sub use_foo() is native('foo') { * }
 
     my $string = "FOO";
-    # The lifetime of this variable must be equal to the required lifetime of
-    # the data passed to the C function.
-    my $array = CArray[uint8].new($string.encode.list);
+    # Manually marshal the $string into $array. Take care that $array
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks.  Although I think it might be an idea to add a `:NUL` named parameter to `CArray.new`, which would add the NUL-byte.

taboege:
I had a similar thought, except that in my version it would be an alternative `CArray.new` candidate:

``` perl6
# class CArray
multi new (Str $data , :$encoding = 'utf8') {
    self.new($data.encode($encoding).list, 0)
}
```

Adding the NUL is specific to strings I would say, so having the first positional argument of that type seems more self-documenting to me.

fecundf:
The problem with this
`self.new($data.encode($encoding).list, 0)`

is that at least one encoding, UTF-16, needs two 0 bytes to terminate the string.

Would this work?

`self.new(($data~"\0").encode($encoding).list)`

ToddAndMargo:
On 2020-01-20 00:25, fecundf wrote:
> The problem with this
> |self.new($data.encode($encoding).list, 0)|
> 
> is that at least one encoding, UTF-16, needs two 0 bytes to terminate 
> the string.
> 
> Would this work?
> 
> |self.new(($data~"\0").encode($encoding).list)|


Hi Fcundf,

This is the way I am currently doing it:

UTF8:  my $array = CArray[uint8].new($string.encode.list,  0);
UTF16: my $array = CArray[uint16].new($string.encode.list, 0);

Since UTF16 is a "word", meaning two bytes long, the "0"
at the end is writing 0x0000 to the word.  I have verified
this with printing out the .base(16) contents of the
resulting CArray.

And it is working perfectly with my UTF16 NativeCalls
to Windows' Kernel32.dll and user32.dll.

Does that help?

-T

ZzZombo:
> is that at least one encoding, UTF-16, needs two 0 bytes to terminate the string.

The quoted routine works only for UTF-8 though, because of `:$encoding = 'utf8'`. It doesn't get dispatched to for any other encoding.

fecundf:
I think you're misreading the signature, ":$encoding = 'utf8'" means the default is utf8.

PR #3172, opened 2020-01-19
Clarify when a Seq .is-lazy

dumarchie:
## The problem

Even though a cached `Seq` can be considered a lazy list in the common sense of the word, the `.is-lazy` predicate has a particular meaning in Raku: if true it _guarantees_ lazy evaluation, even if the
underlying iterator is used to create another `Iterable`.

## Solution provided

* Use "lazy" more precisely in subtitle and introduction.
* Move a more precise definition of `.is-lazy` to just below the definition of `.iterator`.
* In the definition of `.elems`, document why an infinite sequence should be declared lazy.

### DIFF
--- a/doc/Type/Seq.pod6
+++ b/doc/Type/Seq.pod6
@@ -2,15 +2,17 @@
 
 =TITLE class Seq
 
-=SUBTITLE An iterable, lazy sequence of values
+=SUBTITLE An iterable, potentially lazy sequence of values
 
     class Seq is Cool does Iterable does PositionalBindFailover { }
 
-A C<Seq> represents anything that can lazily produce a sequence of values. A
-C<Seq> is born in a state where iterating it will consume the values. However,
-calling C<.cache> on a C<Seq> will return a List that is still lazy, but stores the
-generated values for later access. However, assigning a C<Seq> to an array consumes the C<Seq>;
-alternatively, you can use the C<lazy> statement prefix to avoid it from being
+A C<Seq> represents anything that can produce a sequence of values. A
+C<Seq> is born in a state where iterating it will consume the values.
+Calling C<.cache> on a C<Seq> will make it store the generated values
+for later access.
+
+Assigning the values of a C<Seq> to an array consumes a C<Seq> that is
+not lazy. Use the C<lazy> statement prefix to avoid a C<Seq> from being
 iterated during the assignment:
 
 =begin code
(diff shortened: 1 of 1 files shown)


doc/Type/Seq.pod6
```diff
@@ -78,23 +80,26 @@ Returns the underlying iterator, and marks the invocant as consumed.
...
 
-Returns C<True> if the sequence is lazy and potentially infinite, and C<False>
-otherwise.
-If called on an already consumed sequence, throws an error of type
+Returns C<True> if and only if the underlying iterator guarantees lazy
+evaluation and has not been completely evaluated. If called on an
+already consumed sequence, throws an error of type
 L<X::Seq::Consumed|/type/X::Seq::Consumed>.
 
+=head2 method elems
```
JJ: Why was this moved? Did you want the is-lazy method to be as far up as possible?

doc/Type/Seq.pod6
```diff
@@ -2,15 +2,17 @@
...
+=SUBTITLE An iterable, potentially lazy sequence of values
 
     class Seq is Cool does Iterable does PositionalBindFailover { }
 
-A C<Seq> represents anything that can lazily produce a sequence of values. A
-C<Seq> is born in a state where iterating it will consume the values. However,
-calling C<.cache> on a C<Seq> will return a List that is still lazy, but stores the
-generated values for later access. However, assigning a C<Seq> to an array consumes the C<Seq>;
-alternatively, you can use the C<lazy> statement prefix to avoid it from being
+A C<Seq> represents anything that can produce a sequence of values. A
```
JJ: Instead of eliminating lazily, why not qualify it? Use "maybe", "potentially"...

doc/Type/Seq.pod6
dumarchie: The description of `.elems` refers to laziness, so I thought it better to document `.is-lazy` first. Also, I thought it better to document `.is-lazy` near the property that determines if a sequence is lazy.

doc/Type/Seq.pod6
dumarchie: I believe an introductory sentence should be short, describing the core properties. Laziness is a bit of a problematic concept:

* in the sense of `.is-lazy` returning `True`, it is an option that is not the default
* in the sense of [lazy evaluation](https://en.wikipedia.org/wiki/Lazy_evaluation), only a `Seq` that is cached can be considered lazy because values are not _shared_ if the sequence is not cached

Maybe it's a good idea to point this out after introducing `.cache` and to consistently use "`.is-lazy`" instead of "is lazy" in all cases where laziness refers only to the first sense.  What do you think?

doc/Type/Seq.pod6
JJ: These pages are not really intended to read as a narrative, but more to check individual methods, mainly on their own pages. But it's OK, I was just curious.

doc/Type/Seq.pod6
JJ: OK, fair enough.

PR #3173, opened 2020-01-19
Remove orphan code line

jmaslak:
## The problem

There is an orphan line of code in one of the examples in Exception.

## Solution provided

Removed that orphan line of code.


JJ:
Although deleting the line makes perfect sense, I think that the intention of the code was to actually use defined $e, maybe somethign like throw it in f() instead of saying "die 'Bad'", because it's the same wording. I don't know if @Altai-man will remember anything. Any case, as said, without it it illustrates the point pretty much in the same way and it makes sense deleting it. So you're good to go.

PR #3186, opened 2020-01-24
Complete documentation of role Sequence

dumarchie:
This should resolve issue #2918

### DIFF
--- a/doc/Type/Sequence.pod6
+++ b/doc/Type/Sequence.pod6
@@ -1,15 +1,57 @@
-=begin pod :kind("Type") :subkind("role") :category("basic")
+=begin pod :kind("Type") :subkind("role") :category("composite")
 
-=TITLE class Sequence
+=TITLE role Sequence
 
-=SUBTITLE Operations for higher-level sequences
+=SUBTITLE Common methods of sequences
 
-    class Sequence does PositionalBindFailover { }
+    role Sequence does PositionalBindFailover { }
 
-A C<Sequence> implements a series of methods for converting high-level
-C<Seq>s into different types of objects, from positional to other.
+This role implements a series of methods for converting sequences and
+accessing their elements. These methods have in common that they all
+access the underlying iterator if the sequence has not been
+L<cached|/type/PositionalBindFailover#method_cache> yet. Therefore they
+will throw an error of type L<X::Seq::Consumed|/type/X::Seq::Consumed>
+if called on a C<Seq> that was already consumed.
 
 =head1 Methods
+
+=head2 method Str
+
+    multi method Str(::?CLASS:D:)
+
+L<Stringifies|/type/List#method_Str> the
+L<cached|/type/PositionalBindFailover#method_cache> sequence.
+
+=head2 method Stringy
+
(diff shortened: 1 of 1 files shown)


jnthn:
May be worth a mention that this role extracts the commonalities of all sequences regardless of how parallel they are (serial Seq, ordered parallel HyperSeq, and unordered parallel RaceSeq).

doc/Type/Sequence.pod6
```diff
@@ -7,8 +7,13 @@
     role Sequence does PositionalBindFailover { }
 
 This role implements a series of methods for converting sequences and
-accessing their elements. These methods have in common that they all
-access the underlying iterator if the sequence has not been
+accessing their elements. They should work on all sequences, regardless
```
JJ: If "should work" refers to "series of methods" I think that that would be an "it", not a "they", but it would probably be better to rephrase it with a relative clause like "which should work on..."

doc/Type/Sequence.pod6
```diff
@@ -7,8 +7,13 @@
...
+of how parallel they are: serial L<Seq|/type/Seq>, ordered parallel
```
JJ: "how parallel" does not sound too good. To be precise, they are potentially concurrent, not really parallel.

jnthn: `RaceSeq` and `HyperSeq` are certainly about parallel processing, rather than concurrency.

jnthn: Probably "whether the operations are parallelized" is perhaps better than "how parallel", though.

treyharris: How about:

    They should work on all sequences, regardless of what (if any) parallelism they support:

doc/Type/Sequence.pod6
treyharris: The word “they” is correct for “a series of methods”; “series” can be either singular or plural, and in this case it’s plural. And “they” can refer back to “these methods” as well, which is very clearly plural. Switching to singular “it” is confusing to me; I can’t read its bound referent as either “these methods” or “series”.

“They” vs a relative clause is a stylistic matter, but if you want the predicate to refer to “series of methods” rather than “these methods”, it would have to move to an unnatural place in the passage. “They” pretty clearly can encompass both (and encompassing just one or the other doesn’t result in a bad reading).

(Sorry for dipping the linguistics oar in. But bound references in English was one of my research topics.)

dumarchie: I read "a series" as singular, so on that aspect I have to agree with JJ. But using "it" doesn't feel natural and using "which" after "a series of methods for converting sequences and accessing their elements" is ambiguous. So I will change "they" back to "these methods".

doc/Type/Sequence.pod6
dumarchie: Less is more. The listed classes have a serial/parallel adjective already, so there's no need to mention parallelism in the introduction of the list.

JJ: @jnthn can you please clarify? From my POV, concurrency provides an abstract layer to use parallelism *if the capacity is available*. It does not care if it's not.

PR #3191, opened 2020-01-29
Add boolean logical AND operator

stoned:


doc/Language/operators.pod6
```diff
@@ -1322,6 +1322,15 @@ and then performs a numeric bitwise right shift on the bits of the buffer.
...
 Please note that this has not yet been implemented.
 
+=head2 infix C«?&»
+
+    multi sub infix:<?&>(Mu $x = Bool::True)
+    multi sub infix:<?&>(Mu \a, Mu \b)
+
+X<Boolean logical AND operator>. Coerces the argument(s) to L<Bool|/type/Bool>
+and performs logical AND on them: it will return True if and only if both
+arguments are true. It returns identity on a single argument.
```
JJ: Minor quirks: use always or never capitals with True, although in this case it would be probably better to just eliminate the lower-cased and; follow the logic in the sentence: argument(s) should be followed on it(them). 
Also, returns identity → behaves as identity; it does not return the identity value for whatever type and operator it takes.

stoned:
OK understood. Thanks ! Here is a newer proposal.
I'll keep your comments in mind during my intended "review" of the other boolean operators :)

Issue #3187, opened 2020-01-28
Checklist for version 2020.01


JJ:
#  Changes:
 * [ ] A big overhaul of role applying rules and behavior or
      submethods, constructors and destructors
      -[38c72649](https://github.com/rakudo/rakudo/commit/38c72649)-[a8f61882](https://github.com/rakudo/rakudo/commit/a8f61882)-[cddcd46a](https://github.com/rakudo/rakudo/commit/cddcd46a)-[61d3aef2](https://github.com/rakudo/rakudo/commit/61d3aef2)-[27147ed0](https://github.com/rakudo/rakudo/commit/27147ed0)-[cbc4b839](https://github.com/rakudo/rakudo/commit/cbc4b839)
      -[3f7caf57](https://github.com/rakudo/rakudo/commit/3f7caf57)-[f242bcfa](https://github.com/rakudo/rakudo/commit/f242bcfa)-[cd74a3a7](https://github.com/rakudo/rakudo/commit/cd74a3a7)-[51055d0e](https://github.com/rakudo/rakudo/commit/51055d0e)-[9369c68e](https://github.com/rakudo/rakudo/commit/9369c68e)-[ad92abc3](https://github.com/rakudo/rakudo/commit/ad92abc3)
      -[00fec7c3](https://github.com/rakudo/rakudo/commit/00fec7c3)-[6f0dd1ea](https://github.com/rakudo/rakudo/commit/6f0dd1ea)-[c9ac15cb](https://github.com/rakudo/rakudo/commit/c9ac15cb)-[cb903667](https://github.com/rakudo/rakudo/commit/cb903667)-[d89a0c92](https://github.com/rakudo/rakudo/commit/d89a0c92)-[a08e1ff5](https://github.com/rakudo/rakudo/commit/a08e1ff5)
      -[c674a1d3](https://github.com/rakudo/rakudo/commit/c674a1d3)-[0654d3e5](https://github.com/rakudo/rakudo/commit/0654d3e5)-[fffb5c9c](https://github.com/rakudo/rakudo/commit/fffb5c9c)-[232382f8](https://github.com/rakudo/rakudo/commit/232382f8)-[b47be208](https://github.com/rakudo/rakudo/commit/b47be208)-[44648fbe](https://github.com/rakudo/rakudo/commit/44648fbe)
[...]

PR #3192, opened 2020-01-30
Mention that the colon syntax on $.foo method is now supported

uzluisf:
Refs: #3187

### DIFF
--- a/doc/Language/objects.pod6
+++ b/doc/Language/objects.pod6
@@ -449,14 +449,27 @@ will itemize; C<$.a> will be equivalent to C<self.a.item> or C<$(self.a)>.
 
 =begin code
 class A {
+    has Int @.numbers;
     has $.x = (1, 2, 3);
-    method b() { .say for self.x; .say for $.x }
+
+    method show-diff() { .say for self.x; .say for $.x }
+
+    method twice  { self.times: 2 }
+    method thrice { $.times: 3    }
+
+    method times($val = 1) { @!numbers.map(* * $val).list }
 };
-A.new.b; # OUTPUT: «1␤2␤3␤(1 2 3)␤»
+
+my $obj = A.new(numbers => [1, 2, 3]);
+$obj.show-diff;   # OUTPUT: «1␤2␤3␤(1 2 3)␤»
+say $obj.twice;   # OUTPUT: «(2 4 6)␤»
+say $obj.thrice;  # OUTPUT: «(3 6 9)␤»
 =end code
 
-The colon-syntax for method arguments is only supported for method calls
-using C<self>, not the shortcut.
+
+The colon-syntax for method arguments is supported for method calls
+using either C<self> or the shortcut, as illustrated with the methods
+C<twice> and C<thrice> in the example above.
 
 Note that if the relevant methods C<bless>, C<CREATE> of L<Mu|/type/Mu>
 are not overloaded, C<self> will point to the type object in those methods.


doc/Language/objects.pod6
```diff
@@ -449,14 +449,27 @@ will itemize; C<$.a> will be equivalent to C<self.a.item> or C<$(self.a)>.
...
+    has Int @.numbers;
     has $.x = (1, 2, 3);
-    method b() { .say for self.x; .say for $.x }
+
+    method show-diff() { .say for self.x; .say for $.x }
+
+    method twice  { self.times: 2 }
+    method thrice { $.times: 3    }
+
+    method times($val = 1) { @!numbers.map(* * $val).list }
```
JJ: Sorry, but I don't see the colon anywhere either here or in the previous example. Shouldn't it be something like method b( $a: )?

doc/Language/objects.pod6
```diff
@@ -449,14 +449,27 @@ will itemize; C<$.a> will be equivalent to C<self.a.item> or C<$(self.a)>.
...
 };
-A.new.b; # OUTPUT: «1␤2␤3␤(1 2 3)␤»
+
+my $obj = A.new(numbers => [1, 2, 3]);
+$obj.show-diff;   # OUTPUT: «1␤2␤3␤(1 2 3)␤»
+say $obj.twice;   # OUTPUT: «(2 4 6)␤»
+say $obj.thrice;  # OUTPUT: «(3 6 9)␤»
 =end code
```
JJ: It's nowhere to be seen here either. You are not using obj.twice: whatever, which is another type of colon syntax. So it would be better to clarify what kind of colon you're talking about, the one in signatures or in method calls, and then actually use it in the examples.

doc/Language/objects.pod6
vrurg: @JJ I think the point is in `$.times: 3` method call which wasn't possible before 2020.01.

uzluisf: @JJ From the documentation on [self](https://docs.raku.org/language/objects#self):

> The colon-syntax for method arguments is only supported for method calls using self, not the shortcut.

Given a method `meth(p1, p2, ..., p3)`, you've always been able to call it on `self` using the colon as precedence dropper (e.g., `self.meth: p1, p2, ..., p3)` but not on [`$`](https://docs.raku.org/language/variables#The_._twigil) (e.g., `$.meth: p1, p2, ..., p3`). With 2020.01, the shortcut form is now also supported.

> Sorry, but I don't see the colon anywhere either here or in the previous example.

```
self.times: 2 # self.times(2)
$.times: 3    # $.times(3)
```

>Shouldn't it be something like `method b( $a: )`?

That's the syntax for declaring an [explicit invocant](https://docs.raku.org/language/objects#Class_and_instance_methods).

doc/Language/objects.pod6
uzluisf: Since the sentence makes a reference to `self` (which is only available to a method's body), I thought it provided enough context but you're right about adding some clarification.

vrurg:
> Sorry, I missed that specific call. Thought you would be using it _outside_ the class.

There is no `self`, to which `$` is an alias, outside of a class. :)

JJ:
@vrurg you are right again. Sorry.

uzluisf:
@JJ No problem :-).

PR #3195, opened 2020-02-01
Merge pull request #3195 from Raku/markdown-fix-github-org

treyharris:
## The problem
Contributor Markdown refers to the perl6 org, not Raku

## Solution provided
Change README.md and STYLEGUIDE.md to refer to Raku.

Note that there are numerous cases of 'perl6[.]org' appearing in non-Markdown, such as in build scripts, SASS, etc. Not touching those, since those will affect doc site building. This is just for contributor-facing text.

I’m still sending this as PR so that I don’t step on others’ toes—I can’t find the umbrella “perl6(.org) → Raku(.org) migration plan” ticket if it exists...


AlexDaniel:
Just fix anything you notice. I think the only thing we tend not to change now is the binary name, everything else can and should be changed.

PR #3197, opened 2020-02-03
Document Str.Version

stoned:
Refs: #3187


lizmat:
This needs to mention that this became available with the 2020.01 release.  There are other places in the docs, something like with https://docs.raku.org/type/Buf#Methods_on_buf8_only_(6.d,_2018.12_and_later)

PR #3200, opened 2020-02-05
Disambiguate PERL6LIB in search with categories

Altai-man:
A fix for https://github.com/Raku/doc/issues/3199
*I didn't check it is fine locally*, please do before merge.

### DIFF
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -1509,7 +1509,7 @@ Perl modules is C<PERL5LIB>.
 $ PERL5LIB="/some/module/lib" perl program.pl
 
 In Raku this is similar, one merely needs to change a number!  As you
-probably guessed, you just need to use X<C<PERL6LIB>|environment variable, p5 nutshell>:
+probably guessed, you just need to use X<C<PERL6LIB>|p5 nutshell>:
 
 =for code :lang<shell>
 $ PERL6LIB="/some/module/lib" perl6 program.p6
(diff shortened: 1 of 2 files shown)


doc/Language/5to6-nutshell.pod6
```diff
@@ -1509,7 +1509,7 @@ Perl modules is C<PERL5LIB>.
 $ PERL5LIB="/some/module/lib" perl program.pl
 
 In Raku this is similar, one merely needs to change a number!  As you
-probably guessed, you just need to use X<C<PERL6LIB>|PERL6LIB>:
+probably guessed, you just need to use X<C<PERL6LIB>|p5 nutshell>:
```
JJ: Um, this looks weird. This is going to be indexed as what's to the right of the |; it might be better to use `PERL6LIB,p5 nutshell`. That creates a new category, so use whatever category is already been used there.

doc/Programs/03-environment-variables.pod6
```diff
@@ -16,7 +16,7 @@ page|https://github.com/rakudo/rakudo/wiki/dev-env-vars#moarvm>.
 
 =head2 Module loading
 
-X<|RAKUDOLIB>X<|PERL6LIB>
+X<|RAKUDOLIB, environment variable>X<|PERL6LIB, environment variable>
```
JJ: Ditto here. Use only if there's already a category under that name.

coke:
Updated files pass 'make xtest'

Altai-man:
Closing as it's useless without search categories established.

PR #3211, opened 2020-02-09
Add Supply.repeated

stoned:
Document Supply.repeated
Ref #3187

### DIFF
--- a/doc/Type/Supply.pod6
+++ b/doc/Type/Supply.pod6
@@ -634,7 +634,7 @@ L<C<unique>|/language/independent-routines#routine_unique>).
 See L<C<repeated>|/language/independent-routines#routine_repeated>
 for more examples that use its sub form.
 
-B<Note>: Available since version 6.e (2020.01 and later).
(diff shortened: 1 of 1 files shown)


doc/Type/Supply.pod6
```diff
@@ -620,6 +620,22 @@ C<:expires> parameter how long to wait (in seconds) before "resetting"
...
+by the optional C<:as> and C<:with> parameters (same as with
+L<C<unique>|/language/independent-routines#routine_unique>).
+
+    my $supply = Supply.from-list(<a A B b c b C>).repeated(:as(&lc));
+    $supply.tap(&say);           # OUTPUT: «A␤b␤b␤C␤»
+
+See L<C<repeated>|/language/independent-routines#routine_repeated>
+for more examples that use its sub form.
+
+B<Note>: Available since version 6.e (2020.01 and later).
```
Altai-man: Maybe we should specify it is "Rakudo 2020.01 and later"?

stoned: I pushed a new commit with 'Rakudo' specified. Thanks !

PR #3217, opened 2020-02-12
Reorganize and add Trait::Traced to the debugging programs page

Kaiepi:
## The problem
- The debugging programs page is a bit disorganized. More emphasis is placed on ecosystem modules than core features, and what paragraphs pertain to which section isn't entirely clear to me at first glance.
- Albeit a somewhat new module, `Trait::Traced` offers broader support for tracing than other ecosystem modules out there (to my knowledge). I may be a bit biased since it's my own module though 😉.

## Solution provided
- Reorganize the debugging programs page.
- Add `Trait::Traced` to the ecosystem modules section.

### DIFF
--- a/doc/Programs/01-debugging.pod6
+++ b/doc/Programs/01-debugging.pod6
@@ -4,35 +4,9 @@
 
 =SUBTITLE Modules and applications used to debug Raku programs
 
-There are at least two useful debuggers available for Rakudo, the Raku
-compiler:
+=head1 Core debugging features
 
-=head2 L<Debugger::UI::CommandLine|https://modules.raku.org/repo/Debugger::UI::CommandLine>
-
-A command-line debugger frontend for Rakudo. This module installs the
-C<perl6-debug-m> command-line utility, and is bundled with the Rakudo
-Star distributions. Please check
-L<its repository|https://github.com/jnthn/rakudo-debugger>
-for instructions and a tutorial.
-
-=head2 L<Grammar::Debugger|https://modules.raku.org/repo/Grammar::Debugger> (and C<Grammar::Tracer> in the same distribution)
-
-Simple tracing and debugging support for Raku grammars.
-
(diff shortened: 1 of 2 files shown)


doc/Programs/01-debugging.pod6
```diff
@@ -107,6 +79,43 @@ a string:
...
+line|https://github.com/rakudo/rakudo/wiki/Running-rakudo-from-the-command-line>
+for more information.
+
+=head1 Ecosystem debugging modules
+
+There are at least two useful debuggers and two tracers available for Rakudo
+(the Raku compiler).  For more information on these modules, please refer to
+their documentation.
+
+Historically other modules have existed and others are likely to be
```
JJ: This kinda interrupts the flow. You talk about the modules, but only start enumerating them later. Maybe put this apart as a footnote? It sounds like a footnote, really.

doc/Programs/01-debugging.pod6
```diff
@@ -107,6 +79,43 @@ a string:
...
+written in the future. Please check the L<Raku
+Modules|https://modules.raku.org/> website for more modules like these.
+
+=head2 L<Debugger::UI::CommandLine|https://modules.raku.org/repo/Debugger::UI::CommandLine>
```
JJ: Maybe surround it with the traditional C<>?

doc/Programs/01-debugging.pod6
Kaiepi: How are footnotes usually formatted?

doc/Programs/01-debugging.pod6
Kaiepi: Sounds good

doc/Programs/01-debugging.pod6
JJ: Just wrap them in N<>, and put them where the footnote will be referred to.

PR #3219, opened 2020-02-15
Fix index entries and descriptions of (cont), ∋ and ∌

stoned:
## The problem
The (cont), ∋ and ∌ sections look a lot like the (elem), ∈ and ∉ones and may need a few corrections.

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -2412,9 +2412,9 @@ C«∉» is codepoint U+2209 (NOT AN ELEMENT OF).
     multi sub infix:<(cont)>($a,$b --> Bool:D)
     multi sub infix:<∋>($a,$b --> Bool:D)
 
-X<Membership operator>.
+X<Containing operator>.
 
-Returns C<True> if C<$a> is an B<element> of C<$b>.
+Returns C<True> if C<$a> B<contains> C<$b>.
 
     =begin code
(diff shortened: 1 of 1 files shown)


stoned:
Well my fix try was to only change the description of (cont) to '...if $b is an element of $a' and the description of !(cont) to '...if $b is not an element of $a'. But reversing the order of arguments in the description with respect to the prototypes reads a bit confusing.

JJ:
It's perfect that way, thanks. I only wanted to point out that it wasn't
really incorrect before, just in case you wanted to, at least, leave the
indexing option. Losing indexing anchors might lose links and that's not
nice.

stoned:
Well, we did lose duplicate anchors (#index-entry-Membership_operator and #index-entry-Non-membership_operator) but there are still one of each left, for (elem) and ∉.
So we did not break existing links to these anchors and thus these operators. And anyway I am not sure that the duplicate anchors, to (cont) and ∌, would have been used by browsers as they appeared after in the document.

But we could bring back search results, with something like
`X<Containing operator|Membership operator,Containing operator>`, and `X<Non-containing operator|Non-membership operator,Non-containing operator>`, which would end up to the following: 

search result

How does that look to you ?

JJ:
If you lose the comma, and the text after it, that creates categories, it's
OK. Or if you keep the old anchor.

PR #3221, opened 2020-02-16
Add Supply.first

stoned:
Document Supply.first. Ref #3187

### DIFF
--- a/doc/Type/Supply.pod6
+++ b/doc/Type/Supply.pod6
@@ -374,17 +374,19 @@ Creates a "tail" supply with the same semantics as L<List.tail|/type/List#method
 
     method first(Supply:D: :$end, |c)
 
-This method is a frontend on the C<grep> method. Its creates a new "head"
-supply, or a "tail" supply if the optional named parameter C<:end>
-is truthy, from a supply created by calling the C<grep> method on the
-invocant, with any remaining arguments as parameters.
-If there are no remaining argument, this method is equivalent to calling
+This method creates a supply of the first element, or the last element if
+the optional named parameter C<:end> is truthy, from a supply created by
+calling the C<grep> method on the invocant, with any remaining arguments
+as parameters.
+If there is no remaining argument, this method is equivalent to calling
 on the invocant, without parameter, the C<head> or the C<tail> method,
 according to named parameter C<:end>.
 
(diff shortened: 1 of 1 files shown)


doc/Type/Supply.pod6
```diff
@@ -370,6 +370,22 @@ Creates a "tail" supply with the same semantics as L<List.tail|/type/List#method
     my $ts = $s.tail(2);
     $ts.tap(&say);           # OUTPUT: «3␤2␤»
 
+=head2 method first
+
+    method first(Supply:D: :$end, |c)
+
+This method is a frontend on the C<grep> method. Its creates a new "head"
```
JJ: Not clear what you mean here. If it means that it's like getting a supply of the first element that would be returned from grep, say so. Saying it's a frontend for something is like saying you've put a layer over that something, but that it does the same, maybe in a different way.

doc/Type/Supply.pod6
```diff
@@ -370,6 +370,22 @@ Creates a "tail" supply with the same semantics as L<List.tail|/type/List#method
...
+supply, or a "tail" supply if the optional named parameter C<:end>
+is truthy, from a supply created by calling the C<grep> method on the
```
JJ: Maybe you mean something like "bookend"? It returns either the first or the last element that meets that expression. Or ot emits the first of the remaining elements?

doc/Type/Supply.pod6
```diff
@@ -370,6 +370,22 @@ Creates a "tail" supply with the same semantics as L<List.tail|/type/List#method
...
+invocant, with any remaining arguments as parameters.
+If there are no remaining argument, this method is equivalent to calling
+on the invocant, without parameter, the C<head> or the C<tail> method,
+according to named parameter C<:end>.
+
+    my $rand = supply { emit (rand × 100).floor for ^∞ };
+    my $first-prime = $rand.first: &is-prime;
+    $first-prime.tap: &say; # output the first prime from the endless random number supply $rand
+
```
JJ: maybe add "and then stop... "

stoned:
Thanks JJ for the suggestions! I pushed a new version. Please let me know what you think of it.

JJ:
> Thanks JJ for the suggestions! I pushed a new version. Please let me know what you think of it.

Looks fine! Waiting for the CI result to merge.

PR #3224, opened 2020-02-17
Document how `when` can be used in `for` loops

Kaiepi:
This is a pattern I sometimes use that I figure would be worth documenting.


coke:
Looks good; passes xtest.

PR #3225, opened 2020-02-21
Spelling

jsoref:
## The problem

There appear to be some misspellings that could be corrected

## Solution provided

Correct misspellings. Remove stale whitelisting entries.


jsoref:
My tool identified that both spellings were used and the other spelling was dominant.

PR #3227, opened 2020-02-22
Document how to type enums

Kaiepi:
Enum values can be typed. Enum pairs are also typed, and may be given additional roles to add behaviour and state to them. How this is all done is undocumented though.

### DIFF
--- a/doc/Language/typesystem.pod6
+++ b/doc/Language/typesystem.pod6
@@ -817,6 +817,44 @@ Firstly, we read lines from C<config> file, split every line using
 C<words> method and return resulting pair for every line, thus
 creating a List of Pairs.
 
+=head3 Typing Enums
+
+When declaring enums with an explicit scope, a type may be provided,
+which will be used to typecheck the enum's values:
+
+=for code
+my Str enum Foo (foo => 'foo');
+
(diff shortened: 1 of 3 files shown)


doc/Language/typesystem.pod6
```diff
@@ -817,6 +817,44 @@ Firstly, we read lines from C<config> file, split every line using
...
+behavior and state. For example, to make it simpler to check if a number
+matches a flag in a bitmask enum, you could write a
+C<BitmaskEnumeration> role with an C<ACCEPTS> method to handle this via
+smartmatching:
+
+=begin code
+role BitmaskEnumeration {
+    multi method ACCEPTS(::?CLASS:D: Int:D $value --> Bool:D) {
+        so $value +& $.value
+    }
```
JJ: I don't think this use of the $. twigil has been documented. This works, but after acceptance I'd like to see an issue that helps with this (if it does not exist)

JJ: OK, I see now. $.value is working in the captured $?CLASS context. In that case I think that $value should be called something else, maybe $mask.

JJ: (I don't think that's documented either, BTW).

PR #3230, opened 2020-02-23
Star raku

stoned:
Complete what was already done with respect to $*RAKU. Ref #3187

### DIFF
--- a/doc/Language/variables.pod6
+++ b/doc/Language/variables.pod6
@@ -1532,6 +1532,9 @@ It stringifies to C<Raku>:
 
     $*RAKU.put; # OUTPUT: «Raku␤»
 
+B<Note:> Before Rakudo version 2020.1, this information was available
+through the C<$*PERL> variable.
+
 X<|$*PERL>
 =head4 C<$*PERL>
 
--- a/doc/Language/variables.pod6
+++ b/doc/Language/variables.pod6
@@ -1532,8 +1532,9 @@ It stringifies to C<Raku>:
 
     $*RAKU.put; # OUTPUT: «Raku␤»
 
-B<Note:> Before Rakudo version 2020.1, this information was available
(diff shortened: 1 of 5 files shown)


doc/Language/variables.pod6
```diff
@@ -1532,6 +1532,9 @@ It stringifies to C<Raku>:
 
     $*RAKU.put; # OUTPUT: «Raku␤»
 
+B<Note:> Before Rakudo version 2020.1, this information was available
+through the C<$*PERL> variable.
+
 X<|$*PERL>
 =head4 C<$*PERL>
```
JJ: It still is. We cam maybe discourage its use, but it's still there.

doc/Type/Compiler.pod6
```diff
@@ -31,7 +34,8 @@ It's empty, but it might contain the codename for specific releases.
 
     method build-date()
 
-Up to version 2019.03.1, it returns the date when it was built.
+Up to Rakudo version 2019.03.1, hence only with C<$*PERL.compiler>,
```
JJ: Don't really understand this sentence. The hence, mainly.

doc/Type/Compiler.pod6
```diff
@@ -31,7 +34,8 @@ It's empty, but it might contain the codename for specific releases.
...
+it returns the date when it was built.
```
JJ: *returned*?

stoned: indeed!

doc/Type/Compiler.pod6
stoned: yes it was a bit cryptic. I simply removed this :-)

doc/Language/variables.pod6
stoned: I mentioned both are now available, for the time being.

PR #3232, opened 2020-02-24
chg p6 to raku

tbrowder:

### DIFF
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -4,6 +4,12 @@
 
 =SUBTITLE How do I do what I used to do? (Raku in a nutshell)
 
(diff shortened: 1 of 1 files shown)


doc/Language/5to6-nutshell.pod6
```diff
@@ -4,6 +4,12 @@
 
 =SUBTITLE How do I do what I used to do? (Raku in a nutshell)
 
+Note: Some references to
+Perl 6 still apply in the Raku infrastructure, but the changeover
```
JJ: Can't check it from here, but remember that must still be a non-breaking space.

tbrowder: I tried to copy the nbsp, but I may have fumbled it.

PR #3240, opened 2020-02-24
chg p6 to raku

tbrowder:


doc/Language/haskell-to-p6.pod6
```diff
@@ -371,14 +371,14 @@ guard section of this page.
 
 TODO
 
-explain difference between perl6 Array, Sequence, List. Explain data shapes in regards
+explain difference between raku Array, Sequence, List. Explain data shapes in regards
```
JJ: "a Raku"?

tbrowder: Tidied text in that area.

PR #3252, opened 2020-02-26
proposed table of Raku vs. Perl 6 deprecations

tbrowder:


coke:
My preference would be to mark the original items as deprecated where they are defined and have them link to the new items.

doc/Language/intro.pod6
```diff
@@ -9,6 +9,20 @@ goals, such as being brief whilst being comprehensive, catering to professional
 developers with wide experience whilst also being accessible to newcomers to the
 language.
 
+Note the recent name change to Raku from Perl 6 has resulted in some objects
+having changed names and their former names, while still available, are
```
JJ, line 13: I don't know why this slipped by. Probably, instead of "recent", just quote the date. Also, "objects and methods"

JJ:
> My preference would be to mark the original items as deprecated where they are defined and have them link to the new items.

We can do both; still, the only thing that can be done now is discourage, since, AFAIK, they've not entered a deprecation cycle yet.

PR #3254, opened 2020-03-04
Add example for Callable/Whatever in splice

uzluisf:

### DIFF
--- a/doc/Type/Array.pod6
+++ b/doc/Type/Array.pod6
@@ -261,6 +261,27 @@ my @foo = <a b c d e f g>;
 say @foo.splice(6, 4, <M N O P>);       # OUTPUT: «[g]␤»
 say @foo;                               # OUTPUT: «[a b c d e f M N O P]␤»
 
+The following equivalences hold (assuming that C«@a.elems ≥ $i»):
+
+    @a.push($x, $y)      @a.splice: *  , *, $x, $y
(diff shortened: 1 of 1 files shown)


doc/Type/Array.pod6
```diff
@@ -261,6 +261,27 @@ my @foo = <a b c d e f g>;
 say @foo.splice(6, 4, <M N O P>);       # OUTPUT: «[g]␤»
 say @foo;                               # OUTPUT: «[a b c d e f M N O P]␤»
 
+The following equivalences hold (assuming that C«@a.elems ≥ $i»):
+
+    @a.push($x, $y)      @a.splice: *  , *, $x, $y
+    @a.pop               @a.splice: *-1,
+    @a.shift             @a.splice: 0  , 1,
+    @a.unshift($x, $y)   @a.splice: 0  , 0, $x, $y
+    @a[$i] = $y          @a.splice: $i , 1, $y,
```
Altai-man: This wont compile. Can you wrap it in a pod table, maybe? Or do a longer piece of code with e.g. `@a` and `@b` with different calls used, showing their contents is the same afterwards?

JJ: Right. As soon as it's formatted as code, it will be interpreted as code. Either use `:lang<text>` or format it as a table, with individual elements formatted as code with `C<>`

PR #3256, opened 2020-03-09
correct output in comment

sztanyi:
say ($first-instant, $second-instant) ~~ %intervals.keys.sort;  # OUTPUT: «False␤»
output should be True

## The problem
output in comment is False

## Solution provided
output in comment should be True


Altai-man:
Seems valid. Thanks!

PR #3265, opened 2020-03-15
Buf.list() now returns a List

stoned:

### DIFF
--- a/doc/Type/Buf.pod6
+++ b/doc/Type/Buf.pod6
@@ -116,7 +116,7 @@ Defined as:
 
     multi method list(Buf:D:)
 
-Returns a C<Seq> of codepoints.
+Returns a C<List> of codepoints.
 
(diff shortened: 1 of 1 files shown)


doc/Type/Buf.pod6
```diff
@@ -116,7 +116,7 @@ Defined as:
 
     multi method list(Buf:D:)
 
-Returns a C<Seq> of codepoints.
+Returns a C<List> of codepoints.
```
lizmat: Hmmm...  I wonder why it is a List, and why the documentation mentions codepoints here?  They're just bytes!

stoned:
About Seq vs List, see https://github.com/rakudo/rakudo/commit/7ba50d8647acaa9b5e2ae7f31f95caa3d22b7e3c

And about the codepoints, see #3266 :)

Issue #3266, opened 2020-03-15
Buf.list() and Blob.list() return type


stoned:
The documentation for Buf.list() and Blob.list() says that these methods return a List (previously a Seq) of codepoints. But it seems to me that a List of integers would be more correct. What do you think of this ?

lizmat:
Indeed.  They could be codepoints semantically, but they're just integers realistically.

JJ:
Just mention the type; maybe you can additionally say "that can be treated as codepoints"

PR #3267, opened 2020-03-15
clarify Buf.list() and Blob.list() return value's type

stoned:
As discussed in #3266


lizmat:
FWIW, feels to me the mention of codepoints here is superfluous and even confusing, because one could also say that of any integer value.

stoned:
I tend to agree with you. So, should I just drop the codepoints mention ?

JJ:
> I tend to agree with you. So, should I just drop the codepoints mention ?

Yes, please do that.

PR #3269, opened 2020-03-20
Explicitly mention min/max/minmax()'s :by argument

stoned:


coke:
Looks good - for anyone reviewing via the diff, note that there's a :by example just after the diff section.

Issue #3091, opened 2019-11-17
Change Perl 5 to Perl


JJ:
## The problem

It does not make sense any more to use Perl 5 for Perl.

## Suggestions

Search and substitute all occurrences.

pmurias:
I don't see the benefit of the substitution.

JJ:
All instances of Perl 6 have been renamed to Raku. There's no source of confusion.

Altai-man:
If you grep current docs on `Perl 5` occurrences, there are none, thus the original issue is resolved. Re-open if disagree.

PR #3272, opened 2020-03-22
Change the perl6 command line to raku

tinmarino:
Issue #3105 <- See what more todo for renaming process there (1/4 done by this commit)

Only  __command line__ : `perl6` -> `raku`

Found by hand with `ag '(?<!ft=)perl6'`


AlexDaniel:
Isn't it a bit too early?

tinmarino:
> Isn't it a bit too early?

Yes it is, compared to some others. But if everybody says that ...
@bduggan did a full rename in [p6-jupyter-kernel](https://github.com/bduggan/p6-jupyter-kernel/commit/36ca275216f85d9f6e054a7e6adae660030d5b09) -> I had to get my version up to date, so I change others lol.

tinmarino:
The only requirement here is to get `raku` executable.
I don't know who is also changing the urls: `github.com/Raku` and `https://docs.raku.org/`
Nice issues for a newcomer : getting me to the the file tree

AlexDaniel:
@tinmarino URLs is one thing, but changing the executable name in the docs assumes that users already can type `raku` in their terminal and get something running. Most distros probably don't provide `raku` executable yet, so it's a bit too early I think.

JJ:
El dom., 22 mar. 2020 a las 17:40, Aleks-Daniel Jakimenko-Aleksejev (<

> @tinmarino <https://github.com/tinmarino> URLs is one thing, but changing
> the executable name in the docs assumes that users already can type raku
> in their terminal and get something running. Most distros probably don't
> provide raku executable yet, so it's a bit too early I think.
>

That, too. perl6 works all over, and it's not even deprecated. raku might
not.

tinmarino:
@AlexDaniel, I did not think about end users, which are actually the clients of the doc ... you are right: some years are needed for Debian repo to be up to date (not in Raku's scope).

@JJ I did it by hand but then I could separate PR for each file type, at least directory ...

@Both, should I close (I'll keep a branch in my local for the changes not to be garbage collected) ?

tinmarino:
@JJ : I make a separate for `Perl5 -> Perl` (issue #3091 -> commit 8da4c685de3c)

JJ:
@tinmarino sorry, you lost me here. Are you saying that you are going to make a separate one for that? Or that you have done it already?

tinmarino:
@JJ : I did not do more already. Asking if I should work.

This PR has 2 enormous commits.
1. Perl6 -> Raku) is too early
2. Perl5 -> Perl too big.

I am asking if I should divide 2 in let's say 10 commits and make another PR ?

PS : I was not clear at all (my bad)

JJ:
Right. Do it little by little, not more than 5 files at a time.

PR #3277, opened 2020-03-22
Rename: Perl 5 -> Perl (about) (5/6)

tinmarino:
Issue #3091

### DIFF
--- a/doc/Language/traps.pod6
+++ b/doc/Language/traps.pod6
@@ -530,7 +530,7 @@ in the list, which will be numified separately.
 In some languages one could reference the last element of an array by
 asking for the "-1th" element of the array, e.g.:
 
-=for code :lang<perl5>
+=for code :lang<Perl>
 my @array = qw{victor alice bob charlie eve};
 say @array[-1];    # OUTPUT: «eve␤»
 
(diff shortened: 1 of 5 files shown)


doc/Language/traps.pod6
```diff
@@ -530,7 +530,7 @@ in the list, which will be numified separately.
 In some languages one could reference the last element of an array by
 asking for the "-1th" element of the array, e.g.:
 
-=for code :lang<perl5>
+=for code :lang<Perl>
```
JJ: You've used lower-case perl elsewhere. You should probably stick to that.

doc/Language/traps.pod6
```diff
@@ -819,7 +819,7 @@ not be what you expected.
 
 In some languages, using strings as range end points, considers the entire
 string when figuring out what the next string should be; loosely treating the
-strings as numbers in a large base. Here's Perl 5 version:
+strings as numbers in a large base. Here's Perl version:
```
JJ: _the_ Perl version.

doc/Language/traps.pod6
tinmarino: Fixed: 1291bd1

PR #3282, opened 2020-03-24
change perl method to raku method

sztanyi:
change 1e0.perl to 1e0.raku

## The problem
example use .perl method

## Solution provided
change them to .raku

### DIFF
--- a/doc/Language/numerics.pod6
+++ b/doc/Language/numerics.pod6
@@ -86,13 +86,13 @@ L<denormals|https://en.wikipedia.org/wiki/Denormal_number> (also known as
 Keep in mind that output routines like L<say|/routine/say> or L<put|/routine/put> do not try very hard to
 distinguish between how L<Numeric|/type/Numeric> types are output and may choose to display
 a L<Num|/type/Num> as an L<Int|/type/Int> or a L<Rat|/type/Rat> number. For a more definitive string to
-output, use the L<perl|/routine/perl> method:
+output, use the L<raku|/routine/raku> method:
 
 =begin code
 say  1e0;      # OUTPUT: «1␤»
 say .5e0;      # OUTPUT: «0.5␤»
-say  1e0.perl; # OUTPUT: «1e0␤»
-say .5e0.perl; # OUTPUT: «0.5e0␤»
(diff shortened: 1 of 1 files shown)


coke:
Yes, we should be using .raku instead of .perl.

.perl is an alias that is there for backward compatibility.

If this were removing the documentation on .perl, that would be bad, yes, but to suggest .raku instead of .perl when is the correct approach.

JJ:
> Yes, we should be using .raku instead of .perl.
> 
> .perl is an alias that is there for backward compatibility.
> 
> If this were removing the documentation on .perl, that would be bad, yes, but to suggest .raku instead of .perl when is the correct approach.

Yes, but not hiding .perl, since it's still there and is being used. My point is that while .raku should be encouraged, we should still mention .perl, maybe discouraging it, not substituting one for the other. Just a sentence, maybe even without examples, mentioning that would be in order here.

coke:
> > Yes, we should be using .raku instead of .perl.
> > .perl is an alias that is there for backward compatibility.
> > If this were removing the documentation on .perl, that would be bad, yes, but to suggest .raku instead of .perl when is the correct approach.
> 
> Yes, but not hiding .perl, since it's still there and is being used. My point is that while .raku should be encouraged, we should still mention .perl, maybe discouraging it, not substituting one for the other. Just a sentence, maybe even without examples, mentioning that would be in order here.

This isn't the documentation for the .perl method (that's over in https://docs.raku.org/routine/perl). It's using the functionality, and in that case, we should be using the preferred method consistently. We can definitely update documentation for the raku method and perl method to refer to each other, but at any other point in the docs, we should be using .raku, not .perl.

For a related issue, we've already approached this sort of thing with .^name versus .WHAT (yes, I know these are not identical but they are similar to each other); in the docs, we pick one, and consistently use that in all the cases except where we explicitly want to mention the other.

So to follow that precedent, we wouldn't delete every reference to .perl, but we'd consistently use .raku unless we were specifically documenting something related to .perl (which would be the method definition itself, and the relationship between the two method) - but for references that just wanted to get .perl like output? Those should use .raku - the explanatory sentence you mention should be done once in .raku, echo'd once in .perl - not every time we mention .raku

coke:
See #3309

PR #3286, opened 2020-03-27
Document how to write named arguments that do not create any symbols

Kaiepi:
## The problem
Named arguments can be used like sentinels. In this case, there's no point in there being any symbols for them, as they're unused.

## Solution provided
See title.

### DIFF
--- a/doc/Type/Signature.pod6
+++ b/doc/Type/Signature.pod6
@@ -885,6 +885,16 @@ say alias-named(color => "red", style => "A");
 say alias-named(colour => "green", variety => "B");
 say alias-named(color => "white", class => "C");
 
+Using argument aliases, you can create named arguments that do not
+create any symbols. This is useful when using named arguments as
+sentinels:
+
+=for code
+multi sub trait_mod:<is>(Routine:D $r is raw, :timestamped($)!) {
+    $r does role timestamped { has Instant:D @.timestamps };
+    $r.wrap: -> | { ENTER $r.timestamps.push: now; callsame };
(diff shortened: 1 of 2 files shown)


doc/Type/Signature.pod6
```diff
@@ -885,6 +885,16 @@ say alias-named(color => "red", style => "A");
...
+Using argument aliases, you can create named arguments that do not
+create any symbols. This is useful when using named arguments as
+sentinels:
+
+=for code
+multi sub trait_mod:<is>(Routine:D $r is raw, :timestamped($)!) {
+    $r does role timestamped { has Instant:D @.timestamps };
+    $r.wrap: -> | { ENTER $r.timestamps.push: now; callsame };
+}
+
```
Altai-man: To be honest, I am not a raku beginner, but having some hard time grasping the meaning it wants to convince. Maybe it can be explained in a simpler style?

>not introducing symbol

A symbol in this context is what?

>sentinel

A guard expression?

JJ: It took some time for me too, but I think it's better to improve it once accepted. I understood that it's simply a way of invoking the function, no need to use a variable attached to it.

Kaiepi: > A symbol in this context is what?

A variable, basically.

> A guard expression?

Sort of. What I mean is a named parameter that's used to select a multi candidate, which is why I used `trait_mod:<is>` in the example, but it's not very clear why that's used on second thought.

PR #3292, opened 2020-03-28
Faq: string common operations

tinmarino:
Add common operations as specified here #3109 (1/5)

### DIFF
--- a/doc/Language/faq.pod6
+++ b/doc/Language/faq.pod6
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
 through stubbing. Therefore another possible solution is to move
 classes into the same compilation unit.
 
+
+=head1 Common operations
+
+=head2 String: How can I parse and get a L<number|/language/numerics> from a L<string|/language/type/Str>
+
+Use the L<+ prefix|/language/operators#prefix_+>:
+
+=begin code :ok-test<dd>
+say "42.123456789123456789";  # OUTPUT: «42.123456789123456789␤»
+say +"42.4e-2";               # OUTPUT: «0.424␤»
+=end code
+
+This method can numify any string you could enter as a L<litteral number|/language/syntax#Number_literals>.
+L<val|/routine/val> routine converts it to L<allomorph|/language/glossary#Allomorph>.
+L<unival|/routine/unival> routine converts one unicode codepoint.
+
+=head2 String: How can I check if a string contains a substring and if so, how can I get indices of matches
+
+Use L<.contains|/type/Str#method_contains> or L<.indices|/type/Str#method_indices>:
+
+=begin code :ok-test<dd>
+"az and az and az again".contains("az"); # OUTPUT: «True␤»
+"az and az and az again".indices("az");  # OUTPUT: «(0 7 14)␤»
+=end code
+
+=head2 String: How can I get the hexadecimal representation of a string
+
+First convert it to a L<Blob|/type/Blob> with L<.encode|/routine/encode>.
+
+=begin code :ok-test<dd>
+say "I ❤ 🦋".encode>>.base(16);  OUTPUT: «(49 20 E2 9D A4 20 F0 9F A6 8B)␤»
+=end code
+
+Note that L<.gist|/routine/gist> or L<.raku|/routine/perl> methods are your friends when L<debugging|/programs/01-debugging>:
+
+=begin code
+say "I ❤ 🦋".encode.raku;  # OUTPUT: «utf8.new(73,32,226,157,164,32,240,159,166,139)␤»
+say "I ❤ 🦋".encode.gist;  # OUTPUT: «utf8:0x<49 20 E2 9D A4 20 F0 9F A6 8B>␤»
+=end code
+
+=head2 String: How can I remove from a string some characters by index
+
+Use L<.comb|/routine/comb> to transform it to a L<Seq|/type/Seq>, then the L<(-) infix|/language/operators#infix_(-),_infix_\\> to remove the unwanted indices:
+
+=begin code :ok-test<dd>
+say '0123456789'.comb[(^* (-) (1..3, 8).flat).keys.sort].join;  # OUTPUT: «045679␤»
+=end code
+
+If the string is large, L<.comb|/routine/comb> can take time. In which case, L<.substr-rw|/routine/substr-rw> is faster:
+
+=begin code :ok-test<dd>
+multi postcircumfix:<[- ]> (Str:D $str is copy, +@indices) {
+    for @indices.reverse {
+        when Int   { $str.substr-rw($_,1) = '' }
+        when Range { $str.substr-rw($_  ) = '' }
+    }
+    return $str;
+}
+
+say '0123456789'[- 1..3, 8 ];  # OUTPUT: «045679␤»
+=end code
+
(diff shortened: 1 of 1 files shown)


doc/Language/faq.pod6
```diff
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
 through stubbing. Therefore another possible solution is to move
 classes into the same compilation unit.
 
+
+=head1 Common operations
+
+=head2 String: How can I parse and get a L<number|/language/numerics> from a L<string|/language/type/Str>
+
+Use the L<+ prefix|/language/operators#prefix_+>:
```
JJ: This is an example of contextualizing; it's mentioned explicitly in the document on contexts. It's OK to add new examples, but referring to the existing content is even better.

doc/Language/faq.pod6
```diff
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
...
+
+=begin code :ok-test<dd>
+say "42.123456789123456789";  # OUTPUT: «42.123456789123456789␤»
```
JJ: why the ok-test<dd>?

doc/Language/faq.pod6
```diff
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
...
+say +"42.4e-2";               # OUTPUT: «0.424␤»
+=end code
+
+This method can numify any string you could enter as a L<litteral number|/language/syntax#Number_literals>.
```
JJ: typo

doc/Language/faq.pod6
```diff
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
...
+L<val|/routine/val> routine converts it to L<allomorph|/language/glossary#Allomorph>.
+L<unival|/routine/unival> routine converts one unicode codepoint.
+
+=head2 String: How can I check if a string contains a substring and if so, how can I get indices of matches
+
+Use L<.contains|/type/Str#method_contains> or L<.indices|/type/Str#method_indices>:
+
+=begin code :ok-test<dd>
+"az and az and az again".contains("az"); # OUTPUT: «True␤»
```
JJ: again, why the `<dd>` You're not using it.

doc/Language/faq.pod6
```diff
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
...
+"az and az and az again".indices("az");  # OUTPUT: «(0 7 14)␤»
+=end code
+
+=head2 String: How can I get the hexadecimal representation of a string
+
+First convert it to a L<Blob|/type/Blob> with L<.encode|/routine/encode>.
```
JJ: Not really needed:

    "aå«".ords.map: *.base(16) # OUTPUT: (61 E5 AB)

doc/Language/faq.pod6
```diff
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
...
+
+=begin code :ok-test<dd>
+say "I ❤ 🦋".encode>>.base(16);  OUTPUT: «(49 20 E2 9D A4 20 F0 9F A6 8B)␤»
+=end code
+
+Note that L<.gist|/routine/gist> or L<.raku|/routine/perl> methods are your friends when L<debugging|/programs/01-debugging>:
+
```
JJ: Er, what?

doc/Language/faq.pod6
```diff
@@ -299,6 +299,77 @@ dependencies within a single compilation unit (e.g., file) are possible
...
+    return $str;
+}
+
+say '0123456789'[- 1..3, 8 ];  # OUTPUT: «045679␤»
+=end code
+
+=head2 String: How can I split a string in equal parts
+
+L<.comb|/routine/comb> is accepting a L<Int|/type/Int> or a L<Regex|/type/Regex>:
+
```
JJ: ... but then the regex is not answering the question, right?

doc/Language/faq.pod6
tinmarino: 200% agree. All that is an excuse to link to the doc. I did not know this page.

doc/Language/faq.pod6
tinmarino: I thought it was used for automatic testing. What is its use ?

doc/Language/faq.pod6
tinmarino: litteral -> literal

doc/Language/faq.pod6
tinmarino: I must be explicit better that the desired output is one number per byte:  `"I ❤ 🦋".ords.map:  *.base(16) # OUTPUT: (49 20 2764 20 1F98B)` like an hexadecimal decoder

doc/Language/faq.pod6
tinmarino: "Are useful for variable introspection" better ?

doc/Language/faq.pod6
JJ: It's in the [EXAMPLES guide](https://github.com/Raku/doc/blob/master/writing-docs/EXAMPLES.md)
There're no automatic tests of examples, it would take too long, but you can use author tests to test the new examples.

doc/Language/faq.pod6
JJ: Why not one number per codepoint? If you want an hexadecimal decoder, that's what you should say in the question

doc/Language/faq.pod6
JJ: I don't see it related to the question. If there's anything implicit in the question about debugging, maybe you should say so explicitly.

doc/Language/faq.pod6
tinmarino: __Short:__ For each byte to take the same space on output (=> alignment)

__Long:__ To align the byte list (20 bytes per line) and get the offset of the byte so you can __reverse engineer__ a binary with the offset on the side or count by hand how mush space is occupying some string, if it is zero terminated ...

Like `hexdump` command on linux 
hex-intro

doc/Language/faq.pod6
tinmarino: Just ask and I remove it. I have a reverse engineer background and this `.perl` is giving a fast view on the hex dump I wanted. Usually people looking for hexadecimal representation are analyzing binaries (Blobs).

tinmarino: > Usually people

I admit, I take my personal case as a generality ...

JJ: Not a friend of removing stuff; it's good work and it's done and why would you want to throw it away? My point is that maybe you should re-formulate the question a bit more pointedly... Or just craft another question for which that phrase is an answer. And also remember this is a FAQ, and F is *frequent*.

tinmarino: Thanks for the review.

> I don't see it related to the question.

True, its VERY implicit but all your doc is very dense and I'm trying to follow this pattern. The secret relation is that `"I ❤ 🦋".encode.raku; # OUTPUT utf8.new(73,32,226,157,164,32,240,159,166,139)` so you can strop after `.encode`, if you are interested in bytes. The rest is for display.

PR #3293, opened 2020-03-28
Faq: list common operations

tinmarino:
Add common operations as specified here #3109 (2/5)

### DIFF
--- a/doc/Language/faq.pod6
+++ b/doc/Language/faq.pod6
@@ -302,7 +302,7 @@ classes into the same compilation unit.
 
 =head1 Common operations
 
-=head2 String: How can I parse and get a L<number|/language/numerics> from a L<string|/language/type/Str>
+=head2 String: How can I parse and get a L<number|/language/numerics> from a L<string|/language/type/Str>?
 
 Use the L<+ prefix|/language/operators#prefix_+>:
 
@@ -315,7 +315,7 @@ This example of L<contextualization|/language/contexts> can numify any string yo
 L<val|/routine/val> routine converts it to L<allomorph|/language/glossary#Allomorph>.
 L<unival|/routine/unival> routine converts one unicode codepoint.
 
-=head2 String: How can I check if a string contains a substring and if so, how can I get indices of matches
+=head2 String: How can I check if a string contains a substring and if so, how can I get indices of matches?
 
 Use L<.contains|/type/Str#method_contains> or L<.indices|/type/Str#method_indices>:
 
@@ -324,7 +324,7 @@ Use L<.contains|/type/Str#method_contains> or L<.indices|/type/Str#method_indice
 "az and az and az again".indices("az");  # OUTPUT: «(0 7 14)␤»
 =end code
 
-=head2 String: How can I get the hexadecimal representation of a string
+=head2 String: How can I get the hexadecimal representation of a string?
 
 To get an hexadecimal list of each byte of a string (i.e. hex encoder),
 first convert the string to a L<Blob|/type/Blob> with L<.encode|/routine/encode>.
@@ -340,7 +340,7 @@ say "I ❤ 🦋".encode.raku;  # OUTPUT: «utf8.new(73,32,226,157,164,32,240,159
 say "I ❤ 🦋".encode.gist;  # OUTPUT: «utf8:0x<49 20 E2 9D A4 20 F0 9F A6 8B>␤»
 =end code
 
-=head2 String: How can I remove from a string some characters by index
+=head2 String: How can I remove from a string some characters by index?
 
 Use L<.comb|/routine/comb> to transform it to a L<Seq|/type/Seq>, then the L<(-) infix|/language/operators#infix_(-),_infix_\\> to remove the unwanted indices:
 
@@ -362,7 +362,7 @@ multi postcircumfix:<[- ]> (Str:D $str is copy, +@indices) {
 say '0123456789'[- 1..3, 8 ];  # OUTPUT: «045679␤»
 =end code
 
-=head2 String: How can I split a string in equal parts
+=head2 String: How can I split a string in equal parts?
 
 L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
 
@@ -370,10 +370,21 @@ L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
 .say for 'abcdefghijklmnopqrstuvwxyz'.comb: 8;  # OUTPUT: «abcdefgh␤ijklmnop␤qrstuvwx␤yz»
 =end code
 
+=head2 List: How can I show index while looping through a list?
+
+Use L<.kv|/routine/kv> or L<.pairs|/routine/pairs>:
+
+=begin code
+my @aa = 8 .. 12;
+for @aa.kv -> $i, $_ { say "$i: $_" };     # OUTPUT: «0: 8␤1: 9␤2: 10␤3: 11␤4: 12␤»
+say "$_.key(): $_.value()" for @aa.pairs;  # OUTPUT: «0: 8␤1: 9␤2: 10␤3: 11␤4: 12␤»
+=end code
+
+
 =head1 Language features
 
 X<|Data::Dumper (FAQ)>
-=head2 How can I dump Raku data structures
+=head2 How can I dump Raku data structures?
 (like Perl Data::Dumper and similar)?
 
 Typical options are to use the L<say|/routine/say> routine that uses


doc/Language/faq.pod6
```diff
@@ -299,6 +299,98 @@ dependencies within a single compilation unit (e.g., file) are possible
...
+
+Use L<.unique|/routine/unique>:
+
+=begin code
+(1, 2, 3, 2, 3, 1, 1, 0).unique;  # OUTPUT: «(1 2 3 0)␤»
+=end code
+
+=head2 List: How can I delete an element from a list
+
+To delete by index, use L<.splice|//routine/splice>:
```
treyharris: Is the doubled `//routine` intentional?

tinmarino: Tk, it was a (dangerous) tipo.

doc/Language/faq.pod6
```diff
@@ -370,6 +370,99 @@ L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
 .say for 'abcdefghijklmnopqrstuvwxyz'.comb: 8;  # OUTPUT: «abcdefgh␤ijklmnop␤qrstuvwx␤yz»
 =end code
 
+=head2 List: How can I remove duplicates items in a list
+
```
JJ: "duplicate", also question mark at the end.
And... do you really think this is a FAQ? Woudln't it be better served by simply adding more information to the [existing description of `unique`](https://docs.raku.org/routine/unique)

doc/Language/faq.pod6
```diff
@@ -370,6 +370,99 @@ L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
...
+Use L<.unique|/routine/unique>:
+
+=begin code
+(1, 2, 3, 2, 3, 1, 1, 0).unique;  # OUTPUT: «(1 2 3 0)␤»
+=end code
+
+=head2 List: How can I delete an element from a list
```
JJ: As above, if you want people to find it, just googling it takes you to [this stackoverflow question](https://stackoverflow.com/questions/60404000/delete-an-element-from-a-list-or-array-without-the-any-artefact-in-raku), which is good enough, and besides, provides another way of doing the same, using `:delete`

doc/Language/faq.pod6
```diff
@@ -370,6 +370,99 @@ L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
...
+Note that the L<delete subscript|/language/subscripts#index-entry-:delete_(subscript_adverb)> replaces the element with its natural (i.e. default) state. Here (Any):
+
+=begin code
+my @a = <3 18 4 8 92 14 30>;
+@a[3]:delete;
+say @a;  # OUTPUT: «[3 18 4 (Any) 92 14 30]␤»
+=end code
+
+To delete by value, use L<.grep|/routine/grep>:
+
```
JJ: *and* .= to do it in place.

doc/Language/faq.pod6
```diff
@@ -370,6 +370,99 @@ L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
...
+Use L<.kv|/routine/kv> or L<.pairs|/routine/pairs>:
+
+=begin code
+my @aa = 8 .. 12;
+for @aa.kv -> $i, $_ { say "$i: $_" };     # OUTPUT: «0: 8␤1: 9␤2: 10␤3: 11␤4: 12␤»
+say "$_.key(): $_.value()" for @aa.pairs;  # OUTPUT: «0: 8␤1: 9␤2: 10␤3: 11␤4: 12␤»
+=end code
+
+=head2 List: How can I concatenate lists
```
JJ: I don't really think this is a FAQ. Come this point, I think you should really justify adding new questions to the FAQ. There's no issue in this particular area, these are simple operations that could be better explained or used as an example somewhere else.

doc/Language/faq.pod6
```diff
@@ -370,6 +370,99 @@ L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
...
+=head2 List: How can I generate a list of random numbers
+
+Use L<.roll|/routine/roll>:
+
+=begin code
+say (1..100).roll(100);
+=end code
+
+Prints 50 numbers between 1 and 100.
+
```
JJ: It prints 100. Also, that's a simple example of using roll. Just add another example to roll.

doc/Language/faq.pod6
tinmarino: Removed.

I am sure it is one of the 10 most frequent questions on code operation. In the top 10 view for python and top 15 for [algorithm](https://stackoverflow.com/questions/tagged/algorithm?sort=MostVotes&edited=true)

doc/Language/faq.pod6
tinmarino: Removed

doc/Language/faq.pod6
tinmarino: Removed

Agreed, like the head of pages before method declaration with some code examples : https://docs.raku.org/type/List#Immutability. I think they deserve more code example, wich is what people are looking for (I don't remember the post laughing on that in blog.perl).

doc/Language/faq.pod6
tinmarino: Added:

lazy example with `*` ([strangely consistent](http://strangelyconsistent.org/blog/6-builtins-in-perl-6-that-you-never-knew-you-needed) has also some nice one)

For your information:
What I was looking for actually with this question is that making a list of 100M random integer is faster than generating 100M random integers. But this time is finished apparently because the C++ libs are keeping the state and optimization results in the same.
So the only clever performance trick is with [RDRAND](https://stackoverflow.com/questions/10484164) with [benchmark](https://stackoverflow.com/questions/26771329).

doc/Language/faq.pod6
```diff
@@ -362,49 +362,15 @@ multi postcircumfix:<[- ]> (Str:D $str is copy, +@indices) {
...
-=begin code
-my @letters = <a b c>;
-@letters .= grep: * ne "b";
-say @letters;  # OUTPUT: «[a c]␤»
-=end code
-
-=head2 List: How can I show index while looping through a list
+=head2 List: How can I show index while looping through a list?
 
 Use L<.kv|/routine/kv> or L<.pairs|/routine/pairs>:
```
JJ: Again, this is a direct use of the .kv method. Increase visibility by adding another example, but it's not a FAQ.

doc/Language/faq.pod6
```diff
@@ -414,46 +380,33 @@ for @aa.kv -> $i, $_ { say "$i: $_" };     # OUTPUT: «0: 8␤1: 9␤2: 10␤3:
...
-my @a = <a b c>;
-my @b = <d e f>;
-my @ab = |@a, |@b; say @ab;  # OUTPUT: «[a b c d e f]␤»
-say (@a, @b).flat;           # OUTPUT: «(a b c d e f)␤»
-=end code
-
-Note that flat recursively flattens L<non-containerized|/language/containers> L<iterables|/type/Iterable>.
-
-=head2 List: How can I generate a list of random numbers
+=head2 List: How can I generate a list of random numbers?
```
JJ: Not a FAQ either. If it's not clear enough, or not searchable through that question, just add the key words to the description of "roll". In general, if the question is answered with "use x", just improve the description of X so that it can be found that way.

doc/Language/faq.pod6
```diff
@@ -414,46 +380,33 @@ for @aa.kv -> $i, $_ { say "$i: $_" };     # OUTPUT: «0: 8␤1: 9␤2: 10␤3:
...
-say (1..100).roll(100);
+say (1..100).roll(50);   # prints 50 numbers between 1 and 100
+say [+] (1..6).roll(4);  # prints 1 number = 4d6
+for (1..10).roll(*) { .say; last if $_ == 9 };  # * is creating a lazy list
 =end code
 
-Prints 50 numbers between 1 and 100.
-
-=head2 List: How can I get a list of repeated item
+=head2 List: How can I get a list of repeated item?
```
JJ: Not a FAQ.

doc/Language/faq.pod6
```diff
@@ -414,46 +380,33 @@ for @aa.kv -> $i, $_ { say "$i: $_" };     # OUTPUT: «0: 8␤1: 9␤2: 10␤3:
...
 
 Use the L<xx infix|/routine/xx> operator:
 
 =begin code
 say 42 xx 3;  # OUTPUT: «(42 42 42)␤»
 =end code
 
-=head2 List: How can I get the sum of a list of numbers
+=head2 List: How can I get the sum of a list of numbers?
```
JJ: Definitely not a FAQ.

doc/Language/faq.pod6
```diff
@@ -414,46 +380,33 @@ for @aa.kv -> $i, $_ { say "$i: $_" };     # OUTPUT: «0: 8␤1: 9␤2: 10␤3:
...
 
 Use the C<[+]> L<reduction metaoperator|/operators#Reduction_metaoperators>:
 
 =begin code
 say [+] 1..10;  # OUTPUT: «55␤»
 =end code
 
-=head2 List: How can I add two lists element wise
+=head2 List: How can I add two lists element wise?
```
JJ: Not a FAQ.

tinmarino:
> If there's anything left when those are removed, please go ahead.

Nothing left: there were 5 and all 5 aren't FAQ.

JJ:
I'm sorry, but by all means check out the corresponding functions and improve them all you want; please raise issues first so that we can work out the best option.

treyharris:
I’d just quibble with one small thing here: while I won’t summon the individual by at-mentioning them, one of T&M’s most valid points before I muted them was that many topics weren’t very “reverse-lookupable” — and I think that `.kv` is a case of this; if you know it exists, then you probably don’t need to refer to its reference material except for fine detail on its semantics. But the `routine/kv` page isn’t very discoverable if you don’t know that’s what the functionality is called.

Googling for this at this moment unfortunately first turns up [an answer](https://stackoverflow.com/a/46698358) suggesting creating a custom iterator. (No disrespect to Brad intended; it’s a great answer, it’s just unfortunately not the best answer to on question that Google surfaces it for.) While this answer does eventually get to mentioning `.kv`, it’s not until at the very end.

So I’d humbly suggest this one does deserve its place in the FAQ.

treyharris:
(To be clear, I re-opened the PR to suggest the `.kv` part belongs back, so the PR is not vacuous anymore.)

tinmarino:
I added the .kv (looping with index (`enumerate` in python) and added some `?` last string PR.

doc/Language/faq.pod6
```diff
@@ -362,18 +362,29 @@ multi postcircumfix:<[- ]> (Str:D $str is copy, +@indices) {
...
-=head2 String: How can I split a string in equal parts
+=head2 String: How can I split a string in equal parts?
 
 L<.comb|/routine/comb> is accepting an optional L<Int|/type/Int>:
 
 =begin code
 .say for 'abcdefghijklmnopqrstuvwxyz'.comb: 8;  # OUTPUT: «abcdefgh␤ijklmnop␤qrstuvwx␤yz»
 =end code
 
+=head2 List: How can I show index while looping through a list?
```
JJ: Unless I don't understand this properly, you've added everything back, not just this question. And again I'm sorry @treyharris , my main argument is that if it involves a single method it belongs better with that method, not in the FAQ. In [this case](https://docs.raku.org/routine/kv), the documentation for `.kv` is probably LTA, mostly repeating "it interleaves indexes and whatnot", so the way to proceed here is
1. Raise an issue about it, as in "the documentation for `.kv` does not mention that it can be used to loop through a list explicitly".
2. Talk it out to see where that improvement would belong better. As raised, it's probably the best to just change the method description, but it's described in like 6 classes, so where would it go better? Maybe in a `language` tutorial?
3. Propose a change where there's a consensus. 

"Single source of truth" also applies here. Methods, classes, are better described in the best possible way in a single place. Someone checking out the reference shouldn't be like "Oh, wait, there's more in this other page called *faq*". FAQs and tutorial are mostly devoted to patterns and actions involving stuff that can't possibly go in any individual page. It's not this case.

JJ:
And sorry, I'm closing again. First, because the PR seems to involve old code that was already rejected, and then because of what I've explained above. I strongly encourage you to to that, and I really appreciate your efforts to improve the documentation; but please, follow those steps there.

treyharris:
Understood; I have opened #3299 to consider this further.

Note that I was pretty clear that it was never my intention to reintroduce the entirety of @tinmarino ’s edits, simply the single Q&A referring to `.kv` and `.pairs`, and their commit message sounds like it was not their intention, either; I suspect it was a Git error.

JJ:
That's pretty clear, don't worry.

tinmarino:
Yes I gitted badly : In the git, I also added `?` to some string question because I realized they should have been there. I had troubles with `reset` but I think the result (file changed) is what I wanted.

Issue #3220, opened 2020-02-16
Index `<{ ... }>` and other missing regex interpolatives


treyharris:
## The problem
If one is attempting to discover what the `<{ … }>` syntax in a regex does, it is not indexed:
![IMG_1075](https://user-images.githubusercontent.com/97895/74611670-f3609e00-50cb-11ea-9c7c-42d4c38b224f.jpeg)

## Suggestions
Add indices for `<{ … }>` and other interpolative syntax found in [Regex Interpolation](https://docs.raku.org/language/regexes#index-entry-regex__Regex_Interpolation-Regex_interpolation).

I would do it myself but I can’t currently htmlify for some reason, and I don’t want to submit a patch that I don’t know for certain works. (I need to setup the Docker version so my own install doesn’t affect this in the future…)

JJ:
htmlify no longer works; you have to use documentable now as indicated here https://github.com/Raku/doc/blob/master/README.md#building-the-html-documentation This is all part of the effort to produce a 2.0 version of the documentation #2983 ... for Christmas

PR #3295, opened 2020-03-28
Feature: Index regex interpolation tokens

tinmarino:
Issue: #3220
Add the following to index

```raku
$variable
$(code)
<$variable>
<{code}>
<?{}>
<!{}>
<?>
<!>
<?before>
<?after>
```


tinmarino:
It is indexes but the __?__ disappear in the search bar

PR #3305, opened 2020-04-03
Add type smiley to the glossary

### LINKED ISSUE #3302: Define Smiley in Glossary
Recent commit used "smiley" to describe ```:D``` signature modifier.

Perfectly reasonable, but should add smiley to the glossary.

_I think another option would be to link to smiley in type signatures every time we use it, but we'll miss it, need tests, etc. - glossary seems easier._

Kaiepi:
Fixes #3302


coke:
Very good. I'd add a sentence about why it's a called smiley,  in addition to the definition. (that is, because it looks like an emoticon)

PR #3316, opened 2020-04-07
Index :api and add bit of documentation

uzluisf:

### DIFF
--- a/doc/Language/typesystem.pod6
+++ b/doc/Language/typesystem.pod6
@@ -536,22 +536,6 @@ attributes is not yet implemented. Please note that adding a multi candidate
 that differs only in its named parameters will add that candidate behind the
 already defined one and as such it won't be picked by the dispatcher.
 
-=head3 Versioning, authorship, and API
-
-Versioning, authorship, and API can be applied via the adverbs
-X«C«:ver<>»|:ver<> (class)», X«C«:auth<>»|:auth<> (class)», and
-X«C«:api<>»|:api<> (class)» respectively. All of them take a string as argument;
-for C<:ver> the string is converted to a L<Version|/type/Version> object,
-and for C<:api> the string is converted into the allomorph
-L«C«IntStr»|/type/IntStr». To query a class about its version, author, and API
-use C<.^ver>, C<.^auth>, and C<.^api> respectively.
-
-    class C:ver<4.2.3>:auth<me@here.local>:api<1> {}
-    say C.^ver;       # OUTPUT: «v4.2.3␤»
-    say C.^ver.parts; # OUTPUT: «(4 2 3)␤»
-    say C.^auth;      # OUTPUT: «me@here.local␤»
-    say C.^api;       # OUTPUT: «1␤»
-
 =head2 C<role>
 
 X<|declarator,role (typesystem)>
@@ -689,22 +673,6 @@ say [75kg, N(75kg)];
 say [(75kg).^name, N(75kg).^name];
 # OUTPUT: «[Int+{SI-kilogram} Rat+{SI-Newton}]␤»
 =end code
-
-=head3 Versioning, authorship, and API
-
-Versioning, authorship, and API can be applied via the adverbs
-X«C«:ver<>»|:ver<> (role)», X«C«:auth<>»|:auth<> (role)», and
-X«C«:api<>»|:api<> (role)» respectively. All of them take a string as argument;
-for C<:ver> the string is converted to a L<Version|/type/Version> object,
-and for C<:api> the string is converted into the allomorph
-L«C«IntStr»|/type/IntStr». To query a role's version, author, and API
-use C<.^ver>, C<.^auth>, and C<.^api> respectively.
-
-    class R:ver<4.2.3>:auth<me@here.local>:api<1> {}
-    say R.^ver;       # OUTPUT: «v4.2.3␤»
-    say R.^ver.parts; # OUTPUT: «(4 2 3)␤»
-    say R.^auth;      # OUTPUT: «me@here.local␤»
-    say R.^api;       # OUTPUT: «1␤»
 
 X<|Enumeration; Enums; enum>
 =head2 C<enum>
@@ -908,21 +876,6 @@ another Raku program.
 
 For a full explanation see L<Modules|/language/modules>.
 
-=head3 Versioning, authorship, and API
-
-Versioning, authorship, and API can be applied via the adverbs
-C«:ver<>», C«:auth<>», and C«:api<>» respectively. All of them take a string as
-argument; for C<:ver> the string is converted to a L<Version|/type/Version> object,
-and for C<:api> the string is converted into the allomorph
-L«C«IntStr»|/type/IntStr». To query a module's version, author, and API
-use C<.^ver>, C<.^auth>, and C<.^api> respectively.
-
-    class M:ver<4.2.3>:auth<me@here.local>:api<1> {}
-    say M.^ver;       # OUTPUT: «v4.2.3␤»
-    say M.^ver.parts; # OUTPUT: «(4 2 3)␤»
-    say M.^auth;      # OUTPUT: «me@here.local␤»
-    say M.^api;       # OUTPUT: «1␤»
-
 =head2 C<package>
 
 Packages are nested namespaces of named program elements. Modules, classes and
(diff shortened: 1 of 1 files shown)


doc/Language/typesystem.pod6
```diff
@@ -536,17 +536,21 @@ attributes is not yet implemented. Please note that adding a multi candidate
...
-    class C:ver<4.2.3>:auth<me@here.local> {}
-    say [C.^ver, C.^auth];
-    # OUTPUT: «[v4.2.3 me@here.local]␤»
+=head3 Versioning, authorship, and API
+
+Versioning, authorship, and API can be applied via the adverbs
+X«C«:ver<>»|:ver<> (class)», X«C«:auth<>»|:auth<> (class)», and
+X«C«:api<>»|:api<> (class)» respectively. All of them take a string as argument;
+for C<:ver> the string is converted to a L<Version|/type/Version> object,
+and for C<:api> the string is converted into the allomorph
```
ugexe, line 545: If true this is incorrect behvior. `:api` should act like a version to provide expected sorting behavior, and indeed this is how it is treated in CUR land.

doc/Language/typesystem.pod6
```diff
@@ -536,17 +536,21 @@ attributes is not yet implemented. Please note that adding a multi candidate
...
-L<Version|/type/Version> object. To query a class version and author use
-C<.^ver> and C<^.auth>.
-
-    class C:ver<4.2.3>:auth<me@here.local> {}
-    say [C.^ver, C.^auth];
-    # OUTPUT: «[v4.2.3 me@here.local]␤»
+=head3 Versioning, authorship, and API
+
+Versioning, authorship, and API can be applied via the adverbs
+X«C«:ver<>»|:ver<> (class)», X«C«:auth<>»|:auth<> (class)», and
```
JJ, line 542: API version, maybe?

doc/Language/typesystem.pod6
```diff
@@ -686,17 +690,21 @@ say [(75kg).^name, N(75kg).^name];
...
 
-=head3 Versioning and authorship
+=head3 Versioning, authorship, and API
 
-Versioning and authorship can be applied via the adverbs
-X«C«:ver<>»|:ver<> (role)» and X«C«:auth<>»|:auth<> (role)».
-Both take a string as argument, for C<:ver> the string is converted to a
-L<Version|/type/Version> object. To query a role's version and author use
-C<.^ver> and C<^.auth>.
+Versioning, authorship, and API can be applied via the adverbs
```
JJ, line 695: again, "API versioning" wouldn't be better?

doc/Language/typesystem.pod6
```diff
@@ -900,16 +908,20 @@ another Raku program.
...
+use C<.^ver>, C<.^auth>, and C<.^api> respectively.
 
-    module M:ver<4.2.3>:auth<me@here.local> {}
-    say [M.^ver, M.^auth];
-    # OUTPUT: «[v4.2.3 me@here.local]␤»
+    class M:ver<4.2.3>:auth<me@here.local>:api<1> {}
+    say M.^ver;       # OUTPUT: «v4.2.3␤»
+    say M.^ver.parts; # OUTPUT: «(4 2 3)␤»
+    say M.^auth;      # OUTPUT: «me@here.local␤»
+    say M.^api;       # OUTPUT: «1␤»
```
JJ, line 924: If API is a version, it would admit the same methods as ^ver

doc/Language/typesystem.pod6
```diff
@@ -926,17 +938,21 @@ are classes.
...
-L<Version|/type/Version> object. To query a grammars version and author use
-C<.^ver> and C<^.auth>.
-
-    grammar G:ver<4.2.3>:auth<me@here.local> {}
-    say [G.^ver, G.^auth];
-    # OUTPUT: «[v4.2.3 me@here.local]␤»
+=head3 Versioning, authorship, and API
+
+Versioning, authorship, and API can be applied via the adverbs
+X«C«:ver<>»|:ver<> (grammar)», X«C«:auth<>»|:auth<> (grammar)», and
```
JJ, line 944: Don't really see a point in repeating things over for everything that can be versioned... I get this is the way it was before, but maybe as part of the overhaul we should centralize that information...

uzluisf: A new section (right at the end of type system page) might the appropriate way of going about it. Like this:

```
=head1 Versioning, authorship, and API version.

When you declare a type, you can pass it a version, author, and/or API,
all of which you can subsequently introspect. The versioning, authorship, 
and/or API version of a type can be applied via the adverbs C«:ver<>»,
C«:auth<>», and C«:api<>» respectively.
All of them take a string as argument; for C<:ver> the string is converted to
a L«C«Version»|/type/Version» object, and for C<:api> the string is converted 
into an L<allomorph|/language/glossary#index-entry-Allomorph>
L«C«IntStr»|/type/IntStr» object. To query the version, author, and
API version of a type use C<.^ver>, C<.^auth>, and C<.^api> respectively,
as illustrated down below.

X«|:ver<> (class)»X«|:auth<> (class)»X«|:api<> (class)»
=head2 C<class>

    class R:ver<4.2.3>:auth<me@here.local>:api<1> {}
    say R.^ver;       # OUTPUT: «v4.2.3␤»
    say R.^ver.parts; # OUTPUT: «(4 2 3)␤»
    say R.^auth;      # OUTPUT: «me@here.local␤»
    say R.^api;       # OUTPUT: «1␤»

X«|:ver<> (role)»X«|:auth<> (role)»X«|:api<> (role)»
=head2 C<role>

    role R:ver<4.2.3>:auth<me@here.local>:api<1> {}
    say R.^ver;       # OUTPUT: «v4.2.3␤»
    say R.^ver.parts; # OUTPUT: «(4 2 3)␤»
    say R.^auth;      # OUTPUT: «me@here.local␤»
    say R.^api;       # OUTPUT: «1␤»

X«|:ver<> (module)»X«|:auth<> (module)»X«|:api<> (module)»
=head2 C<module>

    module M:ver<4.2.3>:auth<me@here.local>:api<1> {}
    say M.^ver;       # OUTPUT: «v4.2.3␤»
    say M.^ver.parts; # OUTPUT: «(4 2 3)␤»
    say M.^auth;      # OUTPUT: «me@here.local␤»
    say M.^api;       # OUTPUT: «1␤»

X«|:ver<> (grammar)»X«|:auth<> (grammar)»X«|:api<> (grammar)»
=head2 C<grammar>

    grammar G:ver<4.2.3>:auth<me@here.local>:api<1> {}
    say G.^ver;       # OUTPUT: «v4.2.3␤»
    say G.^ver.parts; # OUTPUT: «(4 2 3)␤»
    say G.^auth;      # OUTPUT: «me@here.local␤»
    say G.^api;       # OUTPUT: «1␤»

```

doc/Language/typesystem.pod6
uzluisf: From what I gather, `:api` doesn't use a `Version` object like  `:ver` does:

```
class Versioned:ver<0.0.1>:auth<github:Kaiepi>:api<1> { }
 
given Versioned {
    say .^ver.^name;  # Version
    say .^auth.^name; # Str
    say .^api.^name;  # IntStr
}
```

ugexe: Ah. Well, thats a bug...  `:api` should be a version. Wish that change had been submitted for a review so the reasoning could have been clarified for everyone.

lizmat: Why should `:api` be a `Version`?  An `identifier` I could understand.  But why a `Version`?

lizmat: An API value in my book, is like an "auth" value.  At allows you to select a specific API.  There should not be any sorting on API, like there is no sorting on "auth".

ugexe: The only reason `auth` exists is to have a non-sortable trait to influence module loading. `api` is a version because in practice it is used as and expected to sort the same way as a version. I would ask what value allowing `api` to be anything but a version allows... any string you want to use as an identifier can be turned into a version, and it will sort in a consistent way when users don't use `use Foo:api<...>` (almost always).

`auth` being non-sortable is what makes it special, and yes it can make things not-consistent. But its an escape hatch -- its not meant to encourage further inconsistency.

doc/Language/typesystem.pod6
JJ: This is probably much better... You could use different calls, also, not clear the value in repeating the same calls over and over...

doc/Language/typesystem.pod6
JJ: I can see the merit in both sides of the argument, but unfortunately, getting to the point, @uzluisf was right to start with, it's an allomorph, so the text is correct and I'm resolving this conversation.

doc/Language/typesystem.pod6
JJ: Ot API numbering, since it's not a version?

doc/Language/typesystem.pod6
JJ: Also, about indexing... Not a lot of value putting so many indices close together, and pointing to examples. A single index without categories would probably do.

doc/Language/typesystem.pod6
ugexe: What exactly do you gain by marking this conversation as resolved immediately instead of waiting an extra couple days? My original comment even outlines the reason why rushing to merge stuff is LTA

JJ: Please feel free to continue the conversation here or wherever you want. Unfortunately, from the point of view of the context, the matter is settled. The text says that api is an allomorph, you said that it's wrong if true, but @uzluisf has proved that it's true, so the text is correct in the way it's written now and the matter, strictly within the boundaries of the documentation and this PR, is setteld.
I'm not rushing to merge because there are outstanding issues. But unless what says @uzluisf here is wrong, I don't see any further problem with this sentence.

JJ:
Can you please check the comments here?

PR #3317, opened 2020-04-07
Document @(code) interpolation in regex

chloekek:
## The problem

$(code) interpolation was documented, but @(code) was not.

## Solution provided

Document @(code) interpolation, with an example that uses the keys  of a hash and substitutes the corresponding values.

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -1894,6 +1894,9 @@ pattern, which may be summarized as follows:
 
 =end table
 
+Instead of the C<$> sigil, you may use the C<@> sigil for array interpolation.
+See below for how this works.
+
 X<|regex, $variable>X<|regex, $(code)>
 Let's start with the first two syntactical forms: C«$variable» and C«$(code)».
 These forms will interpolate the stringified value of the variable or the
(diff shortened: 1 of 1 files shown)


doc/Language/regexes.pod6
```diff
@@ -2017,6 +2020,11 @@ regexes. Just as with ordinary C<|> interpolation, the longest match succeeds:
     my @a = '2', 23, rx/a.+/;
     say ('b235' ~~ /  b @a /).Str;      # OUTPUT: «b23»
 
+You may use C<@(code)> to interpolate any expression that evaluates to a list:
+
+    my %h = a => 1, b => 2;
+    say S:g/@(%h.keys)/%h{$/}/ given 'abc';    # OUTPUT: «12c>
+
```
JJ, line 2027: What would they expand to? Values separated by |?

chloekek: Yes, this is mentioned in the previous paragraph. I clarified that @(code) is the same as first assinging code to @a and then interpolating @a.

chloekek:
Btw, usually want to squash fixup! commits before merging. ;)

With `git rebase -i --autosquash`.

PR #3325, opened 2020-04-11
Clarify how control flow is handled in CATCH blocks

Kaiepi:
## The problem

The documentation on catching exceptions is rather unclear:
- A paragraph explaining how a role is used in an example is placed after what's presumably another example added later on.
- How exceptions are handled in `CATCH` is explained in terms of `given`/`when`, even though it's not particularly special when it comes to blocks.
- What happens when control flow is allowed to reach the end of a `CATCH` block isn't very clear, despite it being useful not to use `default` or `when` in one in some cases.

## Solution provided

- Move that paragraph to the correct place.
- Describe `CATCH` in terms of blocks.
- Explain what happens when you allow the control flow to reach the end of a `CATCH` block and provide an example of when this can be useful.

### DIFF
--- a/doc/Language/exceptions.pod6
+++ b/doc/Language/exceptions.pod6
@@ -69,11 +69,11 @@ to match in the same manner, they must implement the given role. Just
 existing in the same namespace will make them look alike but won't match
 in a C<CATCH> block.
 
-A X<C<CATCH>|CATCH> block places any exception thrown in its topic
-variable (C<$_>), thus it's possible to catch and handle various
-categories of exceptions inside a C<when> block. To handle all
-exceptions, use a C<default> statement. This example prints out almost
-the same information as the normal backtrace printer:
+A C<CATCH> block places any exception thrown in its topic variable
+(C<$_>), thus it's possible to catch and handle various categories of
+exceptions inside a C<when> block. To handle all exceptions, use a
+C<default> statement. This example prints out almost the same
+information as the normal backtrace printer:
 
     CATCH {
          default {
--- a/doc/Language/exceptions.pod6
+++ b/doc/Language/exceptions.pod6
@@ -66,8 +66,8 @@ will probably be the console by default.
 
 Note that the match target is a role. To allow user defined exceptions
 to match in the same manner, they must implement the given role. Just
-existing in the same namespace will look alike but won't match in a
-C<CATCH> block.
+existing in the same namespace will make them look alike but won't match
+in a C<CATCH> block.
 
 A X<C<CATCH>|CATCH> block places any exception thrown in its topic
 variable (C<$_>), thus it's possible to catch and handle various
--- a/doc/Language/exceptions.pod6
+++ b/doc/Language/exceptions.pod6
@@ -47,7 +47,8 @@ problem.
 
 =head1 Catching exceptions
 
-It's possible to handle exceptional circumstances by supplying a C<CATCH> block:
+It's possible to handle exceptional circumstances by supplying a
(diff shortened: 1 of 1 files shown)


doc/Language/exceptions.pod6
```diff
@@ -57,18 +58,22 @@ It's possible to handle exceptional circumstances by supplying a C<CATCH> block:
...
+I<stderr>, which is what C<$*ERR.say> does, getting displayed on
+whatever constitutes the standard error device in that moment, which
+will probably be the console by default.
 
-A X<C<CATCH>|CATCH> block uses smartmatching similar to how C<given/when>
-smartmatches on options, thus it's possible to catch and handle various
-categories of exceptions inside a C<when> block.
+Note that the match target is a role. To allow user defined exceptions
+to match in the same manner, they must implement the given role. Just
+existing in the same namespace will look alike but won't match in a
```
JJ, line 69: ... "make them look alike"?

doc/Language/exceptions.pod6
```diff
@@ -57,18 +58,22 @@ It's possible to handle exceptional circumstances by supplying a C<CATCH> block:
...
+C<CATCH> block.
 
-To handle all exceptions, use a C<default> statement. This example prints out
-almost the same information as the normal backtrace printer.
+A X<C<CATCH>|CATCH> block places any exception thrown in its topic
```
JJ, line 72: The anchor to `CATCH` will be right here. Are you sure this is the best place? Or wouldn't it better to use CATCH block for indexing?

doc/Language/exceptions.pod6
```diff
@@ -81,9 +86,18 @@ almost the same information as the normal backtrace printer.
...
 
-Note that the match target is a role. To allow user defined exceptions to match
-in the same manner, they must implement the given role. Just existing in the
-same namespace will look alike but won't match in a C<CATCH> block.
+While this is a very common pattern, it is not strictly necessary to use
+C<default> or C<when> in a C<CATCH> block. This is done to prevent the
+control flow in one from reaching its end where, unless the exception
+has been L<resumed|#Resuming_of_exceptions>, the exception will continue
+to be thrown.  Allowing this can be used for logging purposes, for
+instance:
```
JJ, line 94: up to where? The end of the catch block?

doc/Language/exceptions.pod6
Kaiepi: I think this belongs more on the phasers page.

doc/Language/exceptions.pod6
Kaiepi: Yes. "the end of the block" might be clearer than "its end" here.

doc/Language/exceptions.pod6
Kaiepi: This is a strange paragraph... I think what it's trying to say that other exception types won't match simply because they're in the `X` namespace. I don't think explaining how roles can be used with exceptions really fits on this part of the page.

JJ: I was saying "will look alike" would be better as "will make them look alike". Not getting into anything else.

doc/Language/exceptions.pod6
JJ: Then it's better if you move the X thingie

coke:
Converting to draft - been a long time since comments were made, not addressed, need to finalize or reject.

coke:
Applied, manually, finally.

Many thanks for all the effort Kaiepi had contributed to the Raku community.

PR #3327, opened 2020-04-12
Also split TEST_FILES on newlines

stoned:
This allows multiple files to be set in TEST_FILES environment variable
from a shell command substitution, hence with embedded newlines


coke:
This doesn't quite work. the whitespace is ignored by default, so you're just breaking on newlines here.

```split(/ \s /)```

works

stoned:
Oh right! Fixed! Thanks.

PR #3341, opened 2020-04-15
how to flatten itemized sublists on an Iterable

kbucheli:
## The problem
I looked up the method [flat on Iterable](https://docs.raku.org/type/Iterable#method_flat) which, unlike on [Any](https://docs.raku.org/type/Any#method_flat), does not show how to flat itemized sublists

## Solution provided
Copy and adapt that part from the documentation of `Any.flat`.

### DIFF
--- a/doc/Type/Iterable.pod6
+++ b/doc/Type/Iterable.pod6
@@ -76,8 +76,8 @@ itemized sublists:
 
     say ($('a', 'b'), 'c').perl;    # OUTPUT: «($("a", "b"), "c")␤»
 
-You can use L<hyper method call|/language/operators#index-entry-postfix_».> to
-call L«C<.List>|/routine/List» method on all the inner itemized sublists
+You can use the L<hyper method call|/language/operators#index-entry-postfix_».> to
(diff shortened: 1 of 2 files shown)


doc/Type/Iterable.pod6
```diff
@@ -76,6 +76,12 @@ itemized sublists:
 
     say ($('a', 'b'), 'c').perl;    # OUTPUT: «($("a", "b"), "c")␤»
 
+You can use L<hyper method call|/language/operators#index-entry-postfix_».> to
```
JJ, line 79: _the_ hyper method call

doc/Type/Iterable.pod6
```diff
@@ -76,6 +76,12 @@ itemized sublists:
...
+call L«C<.List>|/routine/List» method on all the inner itemized sublists
+and so de-containerize them, so that C<flat> can flatten them:
```
JJ, line 81: _the_ .List method

kbucheli:
It makes sense to add the article. So I added them also in the Any documentation.

PR #3349, opened 2020-04-23
Add textwidth=72 vim setting for all POD6 documentation files

Kaiepi:
This allows documentation to be formatted to have the correct line
length as per the style guide with gq in vim.

### DIFF
--- a/doc/404.pod6
+++ b/doc/404.pod6
@@ -19,4 +19,4 @@
 
 =end Html
 
-# vim: expandtab softtabstop=4 shiftwidth=4 ft=perl6
+# vim: expandtab softtabstop=4 shiftwidth=4 textwidth=72 ft=perl6
--- a/doc/HomePage.pod6
+++ b/doc/HomePage.pod6
@@ -65,4 +65,4 @@ Raku compiler developers may also be interested in
 
 =end Html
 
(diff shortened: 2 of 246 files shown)


Kaiepi:
Some files were missing a `softtabstop=72` setting, and some files use `ft=raku` instead of `ft=perl6`. Which `ft` setting should be used? Should there be a test of some sort to ensure vim settings exist in documentation?

Altai-man:
See https://github.com/Raku/doc/issues/3280
Why not just configure vim to apply certain properties to files with particular extension?

Kaiepi:
> Why not just configure vim to apply certain properties to files with particular extension?
I already do. I made this PR because the docs are the exception WRT style for text width when it comes to the Raku files I work with. Removing the vim config entirely is something I'm against doing, but why and how it can better be handled is something I'll explain in the issue.

Kaiepi:
...oh, there is a way to configure vim for specific directories that's fairly painless, making this PR kind of pointless.

JJ:
Thanks, Ben, anyway, and for understanding this. Also, I am not really keen on PRs that span hundreds of files. You _do_ have to review them one by one anyway, just in case, and it's a pain.

PR #3352, opened 2020-04-30
Document Metamodel::TypePretense, Metamodel::MethodDelegation, and Metamodel::CoercionHOW

Kaiepi:
## The problem
`Metamodel::TypePretense`, `Metamodel::MethodDelegation`, and `Metamodel::CoercionHOW` are undocumented. While documentation for the latter isn't *terribly* important IMO, the former two are very useful when implementing your own HOWs and really need documentation. Documenting them also helps explain some behaviours of roles that may not be entirely clear.

## Solution provided
Document the metaroles and metaclasses mentioned, and update roles' HOWs' documentation.

### DIFF
--- a/doc/Type/Metamodel/CoercionHOW.pod6
+++ b/doc/Type/Metamodel/CoercionHOW.pod6
@@ -1,0 +1,99 @@
+=begin pod :kind<Type> :subkind<class> :category<metamodel>
+
+=TITLE class Metamodel::CoercionHOW
+
+=SUBTITLE Metaclass for type coercions
+
+I<Warning>: this class is part of the Rakudo implementation, and is not
+a part of the language specification.
+
+In signatures, I<type coercions> may be used to change the types of
+arguments or return values from one type to another:
+
+=begin code
+sub sum-length(Numeric:D(Iterable:D) $a, Numeric:D(Iterable:D) $b --> Numeric:D) {
+    $a + $b
+}
+
+say sum-length [1,2,3], gather { .take for 4..6 }; # OUTPUT: «6␤»
+=end code
+
+These coercions are simply type objects. Since types in Raku are first
+class, type coercions such as this can be used anywhere, just like any
+other type object:
+
+=for code
+say Int:D(Str:D).^name; # OUTPUT: «Int:D(Str:D)␤»
+
+C<Metamodel::CoercionHOW> is the HOW for such coercion type objects.
+Internally, the metamethods this metaclass provides are used to handle
+coercing objects from one type to another.
+
+A type coercion like C<Int:D(Str:D)> may also be written like this:
+
+=for code
+my constant IntStr = Metamodel::CoercionHOW.new_type: Int:D, Str:D;
+say IntStr.^name; # OUTPUT: «Int:D(Str:D)␤»
+
+Coercions L<pretend to be|/type/Metamodel::TypePretense> of types C<Any>
+and C<Mu>, and L<delegate|/type/Metamodel::MethodDelegation> methods to
+C<Any>.
+
+=head1 Methods
+
+=begin comment
+
+This method's documentation would make more sense once archetypes are
+documented.
+
+=head2 method archetypes
+
+    method archetypes()
+
+Returns the archetypes this HOW has. This HOW simply has the I<coercive>
+archetype.
+
+=end comment
+
+=head2 method new_type
+
+    method new_type($target, $constraint)
+
+Creates a new coercion type. C<$target> is the target type to which an
+object should be coerced to, while C<$constraint> is the constraint on
+the source type from which an object should be coerced. When no type is
+specified in a type coercion, this type constraint is C<Any>.
+
+=head2 method name
+
+    method name($coercion_type)
+
+Returns the name for the coercion type. Refer to
+L<C<Metamodel::Naming>|/type/Metamodel::Naming> for more information.
+
+=head2 method shortname
+
+    method shortname($coercion_type)
+
+Returns the shortname for the coercion type. Refer to
+L<C<Metamodel::Naming>|/type/Metamodel::Naming> for more information.
+
+=head2 method target_type
+
+    method target_type($coercion_type)
+
+Returns the target type for a type coercion:
+
+=for code
+say Int:D(Str:D).^target_type.^name; # OUTPUT: «Int:D␤»
+
+=head2 method constraint_type
+
+    method constraint_type($coercion_type)
+
+Returns the constraint type for a type coercion:
+
+=for code
+say Int:D(Str:D).^constraint_type.^name; # OUTPUT: «Str:D␤»
+
+=end pod
--- a/doc/Type/Metamodel/CurriedRoleHOW.pod6
+++ b/doc/Type/Metamodel/CurriedRoleHOW.pod6
@@ -61,6 +61,10 @@ a single parameter an instantiated role will also be of the same type:
     role Zape[::T] {};
     say Zape[Int].HOW; #: «Perl6::Metamodel::CurriedRoleHOW.new␤»
 
+Curried roles L<pretend to be|/type/Metamodel::TypePretense> of types
+C<Mu>, C<Any>, and C<Cool>, and
+L<delegate|/type/Metamodel::MethodDelegation> methods to C<Any>.
+
 I<Note>: As most of the C<Metamodel> classes, this class is here mainly for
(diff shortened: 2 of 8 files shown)


doc/Type/Metamodel/CoercionHOW.pod6
```diff
@@ -0,0 +1,99 @@
...
+=TITLE class Metamodel::CoercionHOW
+
+=SUBTITLE Metaclass for type coercions
+
+I<Warning>: this class is part of the Rakudo implementation, and is not
+a part of the language specification.
+
+In signatures, I<type coercions> may be used to change the types of
+arguments or return values from one type to another:
+
```
JJ, line 12: Maybe link the corresponding explanation?

doc/Type/Metamodel/CoercionHOW.pod6
```diff
@@ -0,0 +1,99 @@
...
+=begin comment
+
+This method's documentation would make more sense once archetypes are
+documented.
+
+=head2 method archetypes
+
+    method archetypes()
+
+Returns the archetypes this HOW has. This HOW simply has the I<coercive>
```
JJ, line 53: Um, I'm kind of weary of these calls forward. Wouldn't it make more sense to submit it _once_ those archetypes have been documented?

doc/Type/Metamodel/CoercionHOW.pod6
```diff
@@ -0,0 +1,99 @@
...
+A type coercion like C<Int:D(Str:D)> may also be written like this:
+
+=for code
+my constant IntStr = Metamodel::CoercionHOW.new_type: Int:D, Str:D;
+say IntStr.^name; # OUTPUT: «Int:D(Str:D)␤»
+
+Coercions L<pretend to be|/type/Metamodel::TypePretense> of types C<Any>
+and C<Mu>, and L<delegate|/type/Metamodel::MethodDelegation> methods to
+C<Any>.
+
```
JJ, line 41: Maybe provide an use case for this?

doc/Type/Metamodel/CurriedRoleHOW.pod6
```diff
@@ -61,6 +61,10 @@ a single parameter an instantiated role will also be of the same type:
     role Zape[::T] {};
     say Zape[Int].HOW; #: «Perl6::Metamodel::CurriedRoleHOW.new␤»
 
+Curried roles L<pretend to be|/type/Metamodel::TypePretense> of types
```
JJ, line 64: I really appreciate your work here, and in general. The problem is that it takes a long time to review a PR with 8 files. So, this is not really something you should react to, but in this specific case it would have been much better if files like this would have gone to a separate PR, or just submitted directly. After all, they're essentially a different issue. In that case, I have to check that the files you're linking are the ones that have been already included in this issue, so it's additional work...

doc/Type/Metamodel/MethodDelegation.pod6
```diff
@@ -0,0 +1,67 @@
+=begin pod :kind<Type> :subkind<class> :category<metamodel>
+
```
JJ, line 2: I guess this would rather be `subkind<role>`

doc/Type/Metamodel/MethodDelegation.pod6
```diff
@@ -0,0 +1,67 @@
...
+This metarole is commonly used in combination with
+L<C<Metamodel::TypePretense>|/type/Metamodel::TypePretense>.
+
+=head1 Methods
+
+=head2 method delegate_methods_to
+
+    method delegate_methods_to($type)
+
+Delegates methods to a type object.
```
JJ, line 52: This is literally correct, according to the source. But what happens if I don't use a type object here?

doc/Type/Metamodel/TypePretense.pod6
```diff
@@ -0,0 +1,75 @@
+=begin pod :kind<Type> :subkind<class> :category<metamodel>
+
```
JJ, line 2: Also role here

doc/Type/Metamodel/TypePretense.pod6
```diff
@@ -0,0 +1,75 @@
...
+=TITLE role Metamodel::TypePretense
+
+=SUBTITLE Metarole for type pretenses
+
+    role Metamodel::TypePretense { }
+
+I<Warning>: this role is part of the Rakudo implementation, and is not
+a part of the language specification.
+
+Roles can type-check as C<Mu>, C<Any>, and C<Cool>, but don't actually
```
JJ, line 12: Not clear if you mean that all roles do that, using this mechanism. Maybe you could clarify.

doc/Type/Metamodel/CurriedRoleHOW.pod6
```diff
@@ -61,6 +61,10 @@ a single parameter an instantiated role will also be of the same type:
...
+C<Mu>, C<Any>, and C<Cool>, and
+L<delegate|/type/Metamodel::MethodDelegation> methods to C<Any>.
+
 I<Note>: As most of the C<Metamodel> classes, this class is here mainly for
 illustration purposes and it's not intended for the final user to instantiate.
```
JJ, line 70: Know it's not part of the PR, but I don't see a point in this sentence any more. Metamodels are being used extensively in the wild now, from RED to other modules...

doc/Type/Metamodel/CoercionHOW.pod6
Kaiepi: If I did things right, this part is commented out so it can be put in the documentation once those are documented.

doc/Type/Metamodel/MethodDelegation.pod6
Kaiepi: That's correct...

doc/Type/Metamodel/TypePretense.pod6
Kaiepi: All roles do. "Any role will" would probably be clearer than "Roles can" here

doc/Type/Metamodel/MethodDelegation.pod6
Kaiepi: Since `delegate_method_to` is only used with type objects in Rakudo, `delegating_methods_to` seems to only exist for reflection purposes, and `find_method` is the only metamethod in Rakudo that actually uses the value `delegate_methods_to` is given and it uses the value's HOW only, it shouldn't matter whether or not `$type` is actually a type object or not. IMO, it should be one though

doc/Type/Metamodel/CurriedRoleHOW.pod6
Kaiepi: I agree. `Metamodel::ClassHOW` in particular is something that's useful to subclass and is something I subclass in `Data::Record`. One of the module's kinds, `MetamodelX::RecordTemplateHOW`, behaves similarly to role groups, but I don't subclass `Metamodel::ParametricRoleGroupHOW` for those since they produce `MetamodelX::RecordHOW`s, not roles. I can see it being useful to subclass the role HOWs in certain cases, though.

doc/Type/Metamodel/CoercionHOW.pod6
JJ: Yep, but then you have to remember to do that... It's really better if you create an issue.

Kaiepi: Made one at https://github.com/Raku/doc/issues/3355

Kaiepi:
`Metamodel::TypePretense` and `Metamodel;:MethodDelegation` haven't changed since writing this. There are a ton of changes to `Metamodel::CoercionHOW` that have happened between then and now though. Is it alright if I just remove it to revisit another time?

coke:
> `Metamodel::TypePretense` and `Metamodel::MethodDelegation` haven't changed since writing this. There are a ton of changes to `Metamodel::CoercionHOW` that have happened between then and now though. Is it alright if I just remove changes to do with that type to revisit another time?

Yes that would be fine; maybe create a ticket to track any remaining work.

coke:
Of the three, only Metamodel::CoercionHOW seems to be spec'd.

Kaiepi:
`Metamodel::CoercionHOW`-related changes have been removed. There don't look to be references to coercions in `Metmoadel;:TypePretense` or `Metamodel::MethodDelegation`.

coke:
Applied, manually, finally.

Many thanks for all the effort Kaiepi had contributed to the Raku community.

Issue #3309, opened 2020-04-05
Change .perl to .raku in examples


JJ:
## The problem

As part of the #3046 , .raku has superseded .perl. The latter still works perfectly (although it will be deprecated in the next vesion), but it would probably be better to stop using it in examples.

## Suggestions

List below examples that use it and start changing it.
* [x] [Shell environment variables used by Raku](https://docs.raku.org/programs/03-environment-variables)
* [x] [class Set](https://docs.raku.org/type/Set)
* [x] [class List](https://docs.raku.org/type/List)
* [x] [class Code](https://docs.raku.org/type/Code)
* [x] [role Iterable](https://docs.raku.org/type/Iterable)
* [x] [class Metamodel::ClassHOW](https://docs.raku.org/type/Metamodel/ClassHOW)
* [x] [role Metamodel::MethodContainer](https://docs.raku.org/type/Metamodel/MethodContainer)
* [x] [role Metamodel::MultipleInheritance](https://docs.raku.org/type/Metamodel/MultipleInheritance)
* [x] [class Method](https://docs.raku.org/type/Method)
* [x] [class Routine](https://docs.raku.org/type/Routine)
* [x] [class Range](https://docs.raku.org/type/Range)
* [x] [class Signature](https://docs.raku.org/type/Signature)
* [x] [class Any](https://docs.raku.org/type/Any)
* [x] [class Scalar](https://docs.raku.org/type/Scalar)
* [x] [class Attribute](https://docs.raku.org/type/Attribute)
* [x] [class Complex](https://docs.raku.org/type/Complex)
* [x] [class Junction](https://docs.raku.org/type/Junction)
* [x] [class Date](https://docs.raku.org/type/Date)
* [x] [class Int](https://docs.raku.org/type/Int)
* [x] [class Mu](https://docs.raku.org/type/Mu)
* [x] [class Rat](https://docs.raku.org/type/Rat)
* [x] [class Slip](https://docs.raku.org/type/Slip)
* [x] [class Str](https://docs.raku.org/type/Str)
* [x] [role Buf](https://docs.raku.org/type/Buf)
* [x] [class Proc](https://docs.raku.org/type/Proc)
* [x] [class Backtrace](https://docs.raku.org/type/Backtrace)
* [x] [class Array](https://docs.raku.org/type/Array)
* [ ] [role Numeric](https://docs.raku.org/type/Numeric)
[...]

PR #3359, opened 2020-05-05
Test for discouraged usage of .perl in sample code

stoned:
Following #3309 we could consider a test like the following.

But it may be considered a bit drastic. :-}

### DIFF
--- a/writing-docs/EXAMPLES.md
+++ b/writing-docs/EXAMPLES.md
@@ -56,6 +56,15 @@ You can allow it with ok-test:
         dd 42;
     =end
 
+### Allow .perl
+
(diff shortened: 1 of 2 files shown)


writing-docs/EXAMPLES.md
```diff
@@ -56,6 +56,15 @@ You can allow it with ok-test:
         dd 42;
     =end
 
+### Allow .perl
+
+One of the checks is to dissuade using `.perl` in tests. If needed you
```
JJ, line 61: Probably discourage is better than dissuade, and also mention that .raku should be used instead.

stoned: Sure ! Done :-)

PR #3367, opened 2020-05-07
Meaningless entry in Glossary

Prince213:
## The problem
[The entry 'Raku'](https://github.com/Raku/doc/blob/master/doc/Language/glossary.pod6#L906) is looking strange. The renaming process probably caused this.

## Solution provided
Remove the entry 'Raku' from glossary.

Maybe we should rename this entry to Perl 6 and explaing the relation between Raku and Perl 6. What do you think?

### DIFF
--- a/doc/Language/glossary.pod6
+++ b/doc/Language/glossary.pod6
@@ -838,8 +838,8 @@ X<|Perl>
 
 The Perl programming language in its many forms.
 
-X<|Perl 6>
-=head1 Perl 6
+X<|Perl 6>
+=head1 Perl 6
 
 The name used for Raku before October 2019.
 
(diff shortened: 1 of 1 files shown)


JJ:
... Or change it back to Perl 6. it's the glossar, after all... Thanks!

doc/Language/glossary.pod6
```diff
@@ -838,6 +838,11 @@ X<|Perl>
 
 The Perl programming language in its many forms.
 
+X<|Perl 6>
+=head1 Perl 6
```
JJ, line 842: There's a quirk, though... We always use non-breaking space between Perl and 6. It looks exactly the same from here, but if you didn't use that, it will not pass tests.

PR #3372, opened 2020-05-07
Add EditorConfig file

stoned:
ref [EditorConfig](https://editorconfig.org/)
followup or may help #3349 #3280 ;-)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+
+[*.pod6]
+indent_style = space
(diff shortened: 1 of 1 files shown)


stoned:
As you wish! That is only a proposal. Feel free to close it.  :-)

JJ:
Well, I said I was not sure, and I'm not one to throw away other people's work... If you think it's better in the repo, so be it (you'll have to maintain it anyway, and document elsewhere that we have it there). But, as I said, I'd like to hear some more opinions...

PR #3375, opened 2020-05-07
Fix reference to hash assignment documentation

stoned:


coke:
Looks good; as long as you're in here, I think the comma *after* "except" is not needed.

PR #3386, opened 2020-05-10
Clarify behavior of raku method on type objects

stoned:


doc/Type/Mu.pod6
```diff
@@ -186,7 +186,7 @@ near future. Use C<.raku> instead.
     multi method raku(Mu:D:)
 
 For type objects, returns its name if C<.perl> has not been redefined from
```
JJ, line 188: Er. Maybe it should say C<.raku>...

stoned: It is my understanding that this description is accurate with respect to current rakudo's implementation. So I'm not sure what should be done here: keep an accurate description of the implementation or ... ?

lizmat:
Before the rename, all classes had a .perl method, either directly in the class, or inherited from Mu.perl.

After the rename, all of these got renamed to .raku, and a special Mu.perl method was created that would either make that calling .perl on any object would still work (by calling .raku internally).

A special Mu.raku method was also created to make sure that calling .raku on user classes that only had a .perl method, would also just work.

Hope this made sense: it was quite a puzzle to get all of this working correctly :-)

doc/Type/Mu.pod6
JJ: In general, we shouldn't take implementation details into account, but the intent of the class. Probably what would make sense is to describe the behavior without really referring to how the implementation is really made.

stoned: This description had been meanwhile updated by 0143a7cb90ae0c8f75fca01028fb630a4931ee6f
Closing this for now.

PR #3387, opened 2020-05-10
Mention or use .raku method instead of .perl

stoned:
ref #3309

### DIFF
--- a/doc/Type/Date.pod6
+++ b/doc/Type/Date.pod6
@@ -28,7 +28,7 @@ an object the current day according to the system clock.
     say $n - $d;                    # OUTPUT: «7␤», 7 days between New Years/Christmas Eve
     say $n + 1;                     # OUTPUT: «2016-01-01␤»
 
-B<Note> since version 6.d, .perl can be called on C<Date>. It will also
+B<Note> since version 6.d, .raku can be called on C<Date>. It will also
 reject synthetic numerics such as 7̈ .
 
 =head1 Methods
--- a/doc/Type/Complex.pod6
+++ b/doc/Type/Complex.pod6
@@ -196,7 +196,7 @@ Defined as:
 Returns an implementation-specific string that produces an L<equivalent|/routine/eqv> object
 when given to L<EVAL|/routine/EVAL>.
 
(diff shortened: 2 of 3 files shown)


doc/Type/Date.pod6
```diff
@@ -28,7 +28,7 @@ an object the current day according to the system clock.
     say $n - $d;                    # OUTPUT: «7␤», 7 days between New Years/Christmas Eve
     say $n + 1;                     # OUTPUT: «2016-01-01␤»
 
-B<Note> since version 6.d, .perl can be called on C<Date>. It will also
+B<Note> since version 6.d, .raku can be called on C<Date>. It will also
```
JJ, line 31: Not sure about this one. To be precise, raku was introduced a year later. Maybe ".perl (and .raku since version 2019.22"?

nxadm: I would turn it around: the present as main statement and the past between parenthesis:
".raku (.perl before version 2019.22)".

stoned:
Actually I'm a bit confused.
It seems that Date had had a perl method since a long time.
See 
https://github.com/rakudo/rakudo/commit/5ab1c03d4ded2bc90f5ccf3dec656fd8c4973449#diff-33f30fdc536b61e50d61fe6d55cf8630R112
or
https://gist.github.com/Whateverable/7ec244c0ed36fdd9c8741307268ad45d

So I wonder why this is mentioned as new in #2632 and even in 
https://github.com/Raku/roast/blame/master/docs/announce/6.d.md#L173

lizmat:
Before the rename, all classes had a `.perl` method, either directly in the class, or inherited from `Mu.perl`.

After the rename, all of these got renamed to `.raku`, and a special `Mu.perl` method was created that would either make that calling `.perl` on any object would still work (by calling .raku internally).

A special `Mu.raku` method was also created to make sure that calling `.raku` on user classes that only had a `.perl` method, would *also* just work.

Hope this made sense: it was quite a puzzle to get all of this working correctly  :-)

stoned:
Thanks @lizmat, it makes perfect sense!
Would you care to make the same comment in #3386 ;-)

Issue #3394, opened 2020-05-11
Better document use of '$' in forma


tbrowder:

alabamenhu:
Do you mean in `sprintf`?

tbrowder:
Yes, thanks.

PR #3398, opened 2020-05-12
Add warning info for sprintf use

tbrowder:
Addresses issue #3394.


doc/Type/independent-routines.pod6
```diff
@@ -634,6 +634,34 @@ token. Directives guide the use (if any) of the arguments. When a directive
 other than C<%> is used, it indicates how the next argument passed is to be
 formatted into the string to be created.
 
+=begin head3 Warning
```
Altai-man, line 637: Have you tested rendering of this? I mean, should it really be begin/end of head3, instead of simple `=head3 Warning` to make a sub-paragraph? Also, "Warning" is maybe not the best subsection title. Any other ideas?

JJ: Well, it's correct, but certainly unusual...

doc/Type/independent-routines.pod6
```diff
@@ -634,6 +634,34 @@ token. Directives guide the use (if any) of the arguments. When a directive
...
+string either escape it or use the single-quoted form.  For example,
+either form works without error:
+
+=begin code
+sprintf("2 x \$20 = \$%d", 2*20).put; # OUTPUT: «2 x $20 = $40␤»
+sprintf('2 x $20 = $%d', 2*20").put;  # OUTPUT: «2 x $20 = $40␤»
+=end code
+
+=end head3
+
```
JJ, line 664: Also, this is going to probably mess with indexing. If what you want to achieve is somehow to highlight it, It's probably better to do it some other way...

tbrowder: Yeah, I knew better. Suggestions welcome!

doc/Type/independent-routines.pod6
tbrowder: No testing, but I knew better. Suggestions welcome!

doc/Type/independent-routines.pod6
JJ: Big bold italics sign at the top saying "Dangerous waters ahead"?

JJ:
But can't you just use a normal =head3 ? That would fix it...

tbrowder:
On Tue, May 12, 2020 at 13:20 Juan Julián Merelo Guervós <
notifications@github.com> wrote:

> But can't you just use a normal =head3 ? That would fix it...
>

Yes, but what title should we use?

JJ:
Something along the lines of "Warning about using single or double quotes in formats"

PR #3402, opened 2020-05-13
Add more details of routine sprintf and its format string and *@args

tbrowder:
Addresses Raku doc GH issue #3394

The `$` symbol has a special meaning in the `sprintf` format string and
must be escaped in a double-quoted format string if intended to be
used as a literal character. That fact is very briefly described in an
offhand manner, and it may result in a hard-to-find error due to an
error message that, up to Rakudo release 2020.05, does not mention it.

This commit should help by warning of a possible interpolated `$`
symbol in the format string upon input errors when the reported `%`
directives in the format string exceed the number of elements in the
`*@args` array.

The commit also expands information on the directive `parameter index` as well as
adding a little more textual structure with `=head3` Pod directives.

In addition, some other cleanups are introduced.

File modified:

+ doc/doc/Type/independent-routines.pod6

### DIFF
--- a/doc/Type/independent-routines.pod6
+++ b/doc/Type/independent-routines.pod6
@@ -639,29 +639,27 @@ below.
 
 The C<$format> may be defined enclosed in single or double quotes. The
 double-quoted C<$format> string is interpolated before being scanned
-and any embedded string whose interpolated value contains a '%'
-character will cause an exception. N<The invalid sprintf
-formatexception messages have been improved for Rakudo releases after
-I<2020.05.1>.> For example:
+and any embedded string whose interpolated value contains a C<%>
+character will cause an exception. For example:
 
 =begin code
 my $prod = "Ab-%x-42";
 my $cost = "30";
 sprintf("Product $prod; cost: \$%d", $cost).put;
-Your printf-style directives specify 2 arguments, but 1 argument was supplied
-  in block <unit> at <unknown file> line 1
+# OUTPUT: «Your printf-style directives specify 2 arguments, but 1 argument was supplied␤»
+          «  in block <unit> at <unknown file> line 1␤»
 =end code
 
 When handling unknown input you should avoid using such syntax by
-putting all variables in the C<*@args> array and have one '%' for each
-in C<$format>. If you need to include a '$' symbol in the format
+putting all variables in the C<*@args> array and have one C<%> for each
+in C<$format>. If you need to include a C<$> symbol in the format
 string (even as a I<parameter index>) either escape it or use the
 single-quoted form.  For example, either of the following forms works
 without error:
 
 =begin code
 sprintf("2 x \$20 = \$%d", 2*20).put; # OUTPUT: «2 x $20 = $40␤»
-sprintf('2 x $20 = $%d', 2*20").put;  # OUTPUT: «2 x $20 = $40␤»
+sprintf('2 x $20 = $%d', 2*20).put;   # OUTPUT: «2 x $20 = $40␤»
 =end code
 
 In summary, unless you need something very special, you will have
@@ -705,6 +703,8 @@ Compatibility:
 
 =end table
 
+=head2 Modifiers
+
 Modifiers change the meaning of format directives, but are largely
 no-ops (the semantics are still being determined).
 
@@ -724,11 +724,11 @@ order, these are:
 
 =head3 Format parameter index using the '$' symbol
 
-An explicit format parameter integer index (ranging from 1 to N args)
-before the directive, such as C<2$>. By default, C<sprintf> will
+An explicit format parameter index (ranging from 1 to N args)
+before the directive, such as C<%2$d>. By default, C<sprintf> will
 format the next unused argument in the list, but the parameter index
-allows you to take the arguments out of order (note that the single quotes
-are needed unless you escape the '$'):
+allows you to take the arguments out of order (note single quotes
+are required unless you escape the C<$>):
 
 =head4 Without index:
 
@@ -766,6 +766,8 @@ One or more of:
    #     |  ensure the leading "0" for any octal,
          |  prefix non-zero hexadecimal with "0x" or "0X",
          |  prefix non-zero binary with "0b" or "0B"
+----------------------------------------------------------------
+   v     |  NYI vector flag, see description below
 =end table
 
 For example:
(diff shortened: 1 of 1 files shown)


doc/Type/independent-routines.pod6
```diff
@@ -629,16 +629,50 @@ functions.  The only difference between the two functions is that C<sprintf>
...
+and any embedded string whose interpolated value contains a '%'
+character will cause an exception. N<The invalid sprintf
+formatexception messages have been improved for Rakudo releases after
+I<2020.05.1>.> For example:
+
+=begin code
+my $prod = "Ab-%x-42";
+my $cost = "30";
+sprintf("Product $prod; cost: \$%d", $cost).put;
+Your printf-style directives specify 2 arguments, but 1 argument was supplied
```
JJ, line 651: `#` missing here?

doc/Type/independent-routines.pod6
```diff
@@ -629,16 +629,50 @@ functions.  The only difference between the two functions is that C<sprintf>
...
+When handling unknown input you should avoid using such syntax by
+putting all variables in the C<*@args> array and have one '%' for each
+in C<$format>. If you need to include a '$' symbol in the format
+string (even as a I<parameter index>) either escape it or use the
+single-quoted form.  For example, either of the following forms works
+without error:
+
+=begin code
+sprintf("2 x \$20 = \$%d", 2*20).put; # OUTPUT: «2 x $20 = $40␤»
+sprintf('2 x $20 = $%d', 2*20").put;  # OUTPUT: «2 x $20 = $40␤»
```
JJ, line 664: Extra " after 20?

doc/Type/independent-routines.pod6
```diff
@@ -688,13 +722,32 @@ Between the C<%> and the format letter, you may specify several
 additional attributes controlling the interpretation of the format. In
 order, these are:
 
-=head3 Format parameter index
+=head3 Format parameter index using the '$' symbol
 
-An explicit format parameter index, such as C<2$>. By default,
-C<sprintf> will format the next unused argument in the list, but this
-allows you to take the arguments out of order:
+An explicit format parameter integer index (ranging from 1 to N args)
```
JJ, line 727: I can't seem to parse this sentence. Do you mean it's an integer index *for* the format parameter?

doc/Type/independent-routines.pod6
```diff
@@ -688,13 +722,32 @@ Between the C<%> and the format letter, you may specify several
...
+before the directive, such as C<2$>. By default, C<sprintf> will
```
JJ, line 728: _that goes_ before the directive? Also, show the directive in the example.

doc/Type/independent-routines.pod6
```diff
@@ -629,16 +629,50 @@ functions.  The only difference between the two functions is that C<sprintf>
...
+argument passed is to be formatted into the string to be
+created. I<Parameter indexes> may also be used in the format
+tokens. They take the form C<N$> and are explained in more detail
+below.
+
+The C<$format> may be defined enclosed in single or double quotes. The
+double-quoted C<$format> string is interpolated before being scanned
+and any embedded string whose interpolated value contains a '%'
+character will cause an exception. N<The invalid sprintf
+formatexception messages have been improved for Rakudo releases after
```
Altai-man, line 644: >formatexception

No space here? Also, this note seems redundant to me. If the user uses a new compiler, they will see a warn. Otherwise, no. No point to document this implementation detail in docs.

doc/Type/independent-routines.pod6
Altai-man: "integer" is reundant here, yes.

doc/Type/independent-routines.pod6
JJ: Um, why not? If they see the message, they might look it up in Google and they'll be happy to find this.

doc/Type/independent-routines.pod6
JJ: Well, an index is always an integer, that's true. But I guess the intention was to express that it was an integer, as in a non-zero UInt. Probably it would be clearer to just say it starts at 1.

doc/Type/independent-routines.pod6
Altai-man: 1)If the warning message does not give a clear meaning to the user and they google it up, maybe better to fix the message? We usually have nice ones.
2)Docs describe API user can use in their code: types, routines, syntax, generic techniques etc. Compiler warnings are not API. I mean, technically we have e.g. glossary page and some other places which are not language API, but there is really no need to introduce it here imo.

JJ: OK, I see your point.

doc/Type/independent-routines.pod6
tbrowder: Yes, I should have shown it as # RESULT: ... (which should be changed to OUTPUT throughout the file in a new cmmit).

doc/Type/independent-routines.pod6
tbrowder: Yes!

doc/Type/independent-routines.pod6
tbrowder: Will fix.

doc/Type/independent-routines.pod6
tbrowder: Good catch, will fix.

doc/Type/independent-routines.pod6
tbrowder: I would like to keep the note for now (or file a new doc issue). My intent was to cover this short period before my Rakudo error message change is seen in the next release.

The message shown in the example will have to be changed after the next release. Maybe keep the note for now AND file an issue?

JJ: > I would like to keep the note for now (or file a new doc issue). My intent was to cover this short period before my Rakudo error message change is seen in the next release.
> 
> The message shown in the example will have to be changed after the next release. Maybe keep the note for now AND file an issue?

OK, your call. I don't think it's superfluous, but I see what @Altai-man says.

tbrowder: I am working on the changes. I will commit them in an hour or so after I check them more closely in context.

PR #3407, opened 2020-05-15
Change .perl to .raku in documents

Prince213:
ref #3309

### DIFF
--- a/doc/Type/IO/Path/Win32.pod6
+++ b/doc/Type/IO/Path/Win32.pod6
@@ -27,7 +27,7 @@ Defined as:
 Returns a string that, when given passed through L«C<EVAL>|/routine/EVAL»
 gives the original invocant back.
 
-=for code ok-test<perl>
+=for code :ok-test<perl>
 IO::Path::Win32.new("foo/bar").perl.say;
 # OUTPUT: IO::Path::Win32.new("foo/bar", :CWD("C:\\Users\\camelia"))
 
--- a/doc/Type/IO/Path/Win32.pod6
+++ b/doc/Type/IO/Path/Win32.pod6
@@ -27,7 +27,7 @@ Defined as:
 Returns a string that, when given passed through L«C<EVAL>|/routine/EVAL»
 gives the original invocant back.
 
-    IO::Path::Win32.new("foo/bar").perl.say;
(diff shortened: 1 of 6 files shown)


doc/Type/Seq.pod6
```diff
@@ -57,9 +57,10 @@ B<Caution:> No program should ever assume a C<Seq> may only be iterated once
 even if not cached by the program. Caching is a volatile state exposed to the
 developer as an optimization. The C<Seq> may become cached by many operations,
 including calling C<perl> on the C<Seq> (if called prior to a non-cached
-iteration). From version 6.d, you C<.perl> can be called on consumed <Seq>. If a
-program assumes a Seq can only iterate once, but then is later changed to call
-one of these operations during the loop, that assumption will fail.
+iteration). From version 6.d, you C<.raku> (C<.perl> before version 2019.22) can
```
JJ, line 60: there's something missing between you and `.raku`

JJ: Also version 2019.11, not 22

Prince213: I think the word 'you' can probably be removed, what do you think? Also, I got this piece of information on version number from https://github.com/Raku/doc/pull/3387#discussion_r422661364 . Can you change the version in that comment to the right version?

JJ: That was my bad. Done.

doc/Type/IO/Path/Win32.pod6
```diff
@@ -27,8 +27,9 @@ Defined as:
 Returns a string that, when given passed through L«C<EVAL>|/routine/EVAL»
 gives the original invocant back.
 
-    IO::Path::Win32.new("foo/bar").perl.say;
-    # OUTPUT: IO::Path::Win32.new("foo/bar", :CWD("C:\\Users\\camelia"))
+=for code :ok-test<perl>
```
JJ, line 30: Sorry, what is this `:ok-test<perl>`? If it's OK to test, you don't need to write anything.

coke:
Ok-test indicated it's ok for the test to have that item. So Perl means the test won't complain if .perl appears in the code

JJ:
@coke I know, that's why I deleted that comment.

PR #3420, opened 2020-05-23
Fix -f routine lookup in Type/independent-routines

softmoth:
## The problem

Because the pseudo-`Type` name 'independent-routines' contains the string
'routine', it was being filtered out of the results list when looking up
a routine in it, such as `p6doc -f sprintf`.

```
$ p6doc -f sprintf
No such type ''
```

## Solution provided

Be less cavalier about filtering the `@types` array.


softmoth:
Ah, I didn't realize it was going away. What is replacing it?

JJ:
A new and better p6doc... I think I'll merge this anyway.

PR #3429, opened 2020-05-25
Minor rewording with few examples

uzluisf:

### DIFF
--- a/doc/Language/control.pod6
+++ b/doc/Language/control.pod6
@@ -24,7 +24,7 @@ also be written as:
 =head1 X<Blocks|control flow, blocks>
 
 Like many other languages, Raku uses C<blocks> enclosed by C<{> and C<}> to
-turn a sequence of statements into a single one. It is OK to skip the semicolon
+turn a sequence of statements into a single one. It is OK to ommit the semicolon
 between the last statement in a block and the closing C<}>.
 
     { say "Hello"; say "World" }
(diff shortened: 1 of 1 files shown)


doc/Language/control.pod6
```diff
@@ -24,7 +24,7 @@ also be written as:
 =head1 X<Blocks|control flow, blocks>
 
 Like many other languages, Raku uses C<blocks> enclosed by C<{> and C<}> to
-turn a sequence of statements into a single one. It is OK to skip the semicolon
+turn a sequence of statements into a single one. It is OK to ommit the semicolon
 between the last statement in a block and the closing C<}>.
```
JJ, line 28: why the change? Skip is OK.

doc/Language/control.pod6
```diff
@@ -677,20 +693,30 @@ X<|lazy list,gather>X<|lazy list,take>
...
+
+    sub factors( Int:D \n ) {
+        my $k = 1;
+        gather {
+            while $k**2 < n {
+                if n %% $k {
+                    take $k;
+                    take n div $k;
+                }
+                $k += 1;
```
JJ, line 708: $k++ is probably better.

doc/Language/control.pod6
uzluisf: For me, "skip" denotes that the semicolon is being passing over, however it doesn't suggest that it's being literally left out (as far the person writing is concerned) from the code as "omit" does. I might be wrong though so no biggie.  To top it up, I even misspelled the word. 😅

JJ: Either one works, as long as the spelling is OK.

PR #3434, opened 2020-05-26
Document few additions introduced with 2020.05

uzluisf:
* `Str` now has `.Date` and `DateTime` coercers
* `Date.yyyy-mm-dd` now takes a separator parameter, defaulting to `-`
* Add `Date.dd-mm-yyyy` and `Date.mm-dd-yyyy` methods
* Add `Date.last-date-in-month` and `Date.first-day-in-month` methods


coke:
We can probably update the "dd" check in the examples compilation to ignore mm-dd and dd-mm so we don't need to tag these methods.

PR #3438, opened 2020-05-27
Add few examples for named arguments as adverbs

uzluisf:

### DIFF
--- a/doc/Language/functions.pod6
+++ b/doc/Language/functions.pod6
@@ -179,16 +179,30 @@ C<Capture> container. Details about the syntax and use of these containers
 can be found in the L<documentation on the C<Capture> class|/type/Capture>.
 
 When using named arguments, note that normal list "pair-chaining" allows
-one to skip commas between named arguments. If positional arguments are also
-passed, then either parens must be used or the comma immediately after the
-last positional must be kept when skipping commas between named arguments.
+one to skip commas between named arguments.
 
     sub f(|c){};
     f :dest</tmp/foo> :src</tmp/bar> :lines(512);
     f :32x :50y :110z;   # This flavor of "adverb" works, too
-    f('a', 'i') :1x :2y;
-    f 'a', 'i', :1x :2y;
     f :a:b:c;            # The spaces are also optional.
+
+If positional arguments are also passed, then either they must be passed
+within parentheses placed I<immediately> after the function's name or the comma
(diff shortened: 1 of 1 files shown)


doc/Language/functions.pod6
```diff
@@ -170,19 +170,24 @@ argument list of a function call, some special syntax is supported:
...
 
 Arguments passed to a function are conceptually first collected in a
 C<Capture> container. Details about the syntax and use of these containers
 can be found in the L<documentation on the C<Capture> class|/type/Capture>.
 
-When using named arguments, note that normal List "pair-chaining" allows
-one to skip commas between named arguments.
+When using named arguments, note that normal list "pair-chaining" allows
+one to skip commas between named arguments. If positional arguments are also
+passed, then either parens must be used or the comma immediately after the
```
JJ, line 183: parentheses

JJ: Clarify a bit where the parentheses can be placed. Can they be placed, for instance, in the middle of two positional arguments? Can you chain adverbs like this? :1a:2b?

uzluisf: This is what I've, hopefully it's less ambiguous now:

```
When using named arguments, note that normal list "pair-chaining" allows
one to skip commas between named arguments.

    sub f(|c){};
    f :dest</tmp/foo> :src</tmp/bar> :lines(512);
    f :32x :50y :110z;   # This flavor of "adverb" works, too
    f :a:b:c;            # The spaces are also optional.

If positional arguments are also passed, then either they must be passed
within parentheses placed I<immediately> after the function's name or the comma
after the last positional argument must be kept when chaining named arguments in
abbreviated adverb form. The spaces between chained named arguments and 
the list of positional arguments is optional.

    sub p($x, $y, :$translate, :$rotate) {};
    p(1, 1, :translate, :rotate); # normal way
    p 1, 1, :translate, :rotate;  # also normal way
    
    p(1, 1) :translate  :rotate;  # parentheses + chained named arguments
    p(1, 1) :translate:rotate;
    p(1, 1):translate:rotate;
    
    p 1, 1, :translate  :rotate;  # dangling comma + chained named arguments
    p 1, 1, :translate:rotate;
    p 1, 1,:translate:rotate;
```

JJ: Just push it here and we're done.

Issue #3353, opened 2020-04-30
Roles' HOWs deserve better documentation


Kaiepi:
## The problem
Roles are complex. This is not a bad thing per se, but they need good documentation in order for people to completely understand how they work. While the objects page does a good job of explaining them in the kind of terms people tend to work with them with, the documentation on their HOWs is lacking.

## Suggestions
Expand the documentation on `Metamodel::ParametricRoleGroupHOW`, `Metamodel::ParametricRoleHOW`, `Metamodel::CurriedRoleHOW`, and `Metamodel::ConcreteRoleHOW`.

This is something I'm willing to work on.

JJ:
Please do that by all means. But make sure to eliminate the note saying "Don't use this at home, kids" of the ones that are already documented. And thanks!

Issue #2460, opened 2018-11-13
Parameters for .^attributes are missing


Juerd:
## The problem
https://docs.perl6.org/routine/attributes doesn't list `:local`, `:excl`, and `:all`.

JJ:
I see [`local` and `tree`](https://github.com/perl6/roast/blob/6e5273432213bc3bec52048e94422c8ac8115a6c/S12-introspection/attributes.t) tested in roast. I don't see the others...

TopherCooper:
Note for clarification: The Rakuda 2019.03 release includes :$all (default behavior) and :$excl (says to exclude the specified superclasses' attributes from the list) but since these are non-standard any documentation (flagged as Rakuda specific of course) is completely optional.  Rakuda's lack of an implementation for :$tree is noted in the roast.

While documenting :$tree would be needed if completeness were desired, documenting :$local seems to me to have a higher priority since it is used extensively without explanation in https://docs.perl6.org/type/Attribute.  That means that the lack of documentation for :$local effectively weakens the clarity that document as well.

PR #3441, opened 2020-05-28
Revise and add to Attribute's documentation

Kaiepi:
## The problem
`Attribute` is documented as something you normally shouldn't use directly. I think this is useful to do when working with the MOP though, and I use `Attribute` like this in an example in the revision of one of the role HOWs' type pages (#3353).

## Solution provided
- Reorganize `Attribute`'s type page somewhat and remove some unnnecessary "Defined as:" before each method signature.
- Show how `Attribute` can be used when working with the MOP.
- Document `Attribute.new`, along with `Attribute.is_bound`, and `Attribute.is_built`.
- Document use of `:bind` with the `is built` trait, as this is referred to by the new documentation for `Attribute.new`.

### DIFF
--- a/doc/Type/Attribute.pod6
+++ b/doc/Type/Attribute.pod6
@@ -10,8 +10,42 @@ In Raku lingo, an I<attribute> refers to a per-instance/object storage slot.
 An C<Attribute> is used to talk about classes' and roles' attributes at the
 metalevel.
 
-Normal usage of attributes does not require the user to use this class
-explicitly.
+C<Attribute> is typically useful when working with the L<MOP|/language/mop>.
+For instance, the names of all the attributes of a type can be introspected
+using its L<name|#method_name> method:
+
+=begin code :solo
+class WithAttributes {
+    has $.attribute;
+    has $.attribute-two-electric-boogaloo;
+    has $.yet-another-attribute;
+}
+
+.say for WithAttributes.^attributes(:local).map(*.name);
+# OUTPUT:
+# $!attribute
+# $!attribute-two-electric-boogaloo
+# $!yet-another-attribute
+=end code
+
+Because of C<Attribute>, a type containing attributes, such as this:
+
+=for code :solo
+class WithAttribute {
+    has $.attribute;
+}
+
+Is not something only the compiler knows how to generate. This class in
+particular can be generated manually like so:
+
+=for code :solo
+BEGIN {
+    constant WithAttribute = Metamodel::ClassHOW.new_type: :name<WithAttribute>;
+    WithAttribute.^add_attribute: Attribute.new:
+        :name<$!attribute>, :type(Any), :package(WithAttribute),
+        :1has_accessor;
+    WithAttribute.^compose;
+}
 
 =head1 Traits
 
@@ -38,8 +72,6 @@ argument of C<is default> will set the default item value or hash value.
     say $foo; # OUTPUT: «Foo.new(bar => [42])␤»
 
 =head2 X<Trait is required|trait,is required (Attribute)>
-
-Defined as:
 
     multi sub trait_mod:<is> (Attribute $attr, :$required!)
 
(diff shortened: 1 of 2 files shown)


doc/Type/Attribute.pod6
```diff
@@ -10,8 +10,42 @@ In Raku lingo, an I<attribute> refers to a per-instance/object storage slot.
 An C<Attribute> is used to talk about classes' and roles' attributes at the
 metalevel.
 
-Normal usage of attributes does not require the user to use this class
-explicitly.
+C<Attribute> is typically useful when working with the L<MOP|/language/mop>.
+For instance, the names of all the attributes of a type can be introspected
+using its L<name|#method_name> method:
+
+=begin code :solo
```
JJ, line 17: Why :solo? This would compile correctly, right? And it does not need its own file.

doc/Type/Attribute.pod6
```diff
@@ -10,8 +10,42 @@ In Raku lingo, an I<attribute> refers to a per-instance/object storage slot.
...
+    has $.yet-another-attribute;
+}
+
+.say for WithAttributes.^attributes(:local).map(*.name);
+# OUTPUT:
+# $!attribute
+# $!attribute-two-electric-boogaloo
+# $!yet-another-attribute
+=end code
+
```
JJ, line 30: Maybe explain a bit? ^attributes is not very well documented #2460.

doc/Type/Attribute.pod6
```diff
@@ -10,8 +10,42 @@ In Raku lingo, an I<attribute> refers to a per-instance/object storage slot.
...
+Because of C<Attribute>, a type containing attributes, such as this:
+
+=for code :solo
```
JJ, line 33: Again, not needed.

doc/Type/Attribute.pod6
```diff
@@ -124,47 +156,72 @@ default accessor for the attribute will return a writable value.
 
 =head2 X<trait is built|trait,is built (Attribute)>
 
-Defined as:
-
     multi sub trait_mod:<is>(Attribute:D $a, :$built!)
```
JJ, line 159: Why did you eliminate `Defined as`?

doc/Type/Attribute.pod6
```diff
@@ -124,47 +156,72 @@ default accessor for the attribute will return a writable value.
...
-    say $a.raku;            # OUTPUT: «Attribute.new␤»
-    say $a.name;            # OUTPUT: «@!things␤»
-    say $a.package;         # OUTPUT: «(Useless)␤»
-    say $a.has_accessor;    # OUTPUT: «False␤»
-
-Modifying a private attribute from the outside is usually not possible, but
-since Attribute is at the level of the metaclass, all is fair game.
+=head2 method new
+
+=for code :skip-test<too long a method signature for one line>
```
JJ, line 186: In principle, there should be no problem with so many lines, as long as you use =for code or =begin code

doc/Type/Attribute.pod6
```diff
@@ -124,47 +156,72 @@ default accessor for the attribute will return a writable value.
...
+method new(Attribute:_:
+    :$name!, :$type!, :$package!,
+    :$inlined = 0, :$has_accessor = 0, :$is_built = $has_accessor, :$is_bound = 0,
+    :$positional_delegate = 0, :$associative_delegate = 0, *%other)
+
+Creates a new attribute. The following named arguments are required:
+
+- C<$name> contains the attribute's name, which should always be a
```
JJ, line 194: Maybe use =item here?

doc/Type/Attribute.pod6
```diff
@@ -192,8 +264,6 @@ Returns the package (class/grammar/role) to which this attribute belongs.
 
 =head2 method has_accessor
 
-Defined as:
-
     method has_accessor(Attribute:D: --> Bool:D)
```
JJ, line 267: again, why?

Kaiepi:
1. Tests are skipped in this because of symbol clashes, which looks to allow typos like `:1has_accessor` already, so those examples need redoing.
2. I thought it was common style in the docs to put the code block under the header without the `Defined as:` before, but it looks to be [mixed](https://github.com/Raku/doc/blob/e2e6e179b5d247a0974c95d13a5555bd2a8b4d54/doc/Type/Mu.pod6#L126-L151). Is there any preference?

coke:
I don't think we need "Declared as:" going forward - having the signature as the first line seems obvious enough.

PR #3446, opened 2020-05-30
Add section on tail call optimization/elimination.

Michael-S:
Someone coming from Haskell, Common Lisp, and
several other functional languages might want to know
if Raku supports tail call optimization.

I can't find any recent detailed discussions on it, but
my understanding is that it's not in the Raku language spec
or in the Rakudo implementation.

## The problem
Per issue #3393  I wanted to add a section on tail call optimization.

## Solution provided

A short section mentioning the feature and that Raku does not require it, and would get smaller benefit from it than Haskell or many other functional programming languages.

### DIFF
--- a/doc/Language/haskell-to-p6.pod6
+++ b/doc/Language/haskell-to-p6.pod6
@@ -537,12 +537,11 @@ compare it to let/in and where constructs maybe?
 
 TODO
 
-=head1 Tail Call Optimization
+=head1 Tail Call Optimization or Tail Call Elimination
 
-=head2 Tail Call Optimization or Tail Call Elimination
-
-Haskell and many other functional programming languages use tail call optimization to remove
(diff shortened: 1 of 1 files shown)


doc/Language/haskell-to-p6.pod6
```diff
@@ -537,6 +537,21 @@ compare it to let/in and where constructs maybe?
 
 TODO
 
+=head1 Tail Call Optimization
+
+=head2 Tail Call Optimization or Tail Call Elimination
+
```
JJ, line 543: Hum, why create a level-1 section with nothing buth a level-2 section? Wouldn't it be more convenient to just have one of them?

Michael-S:
That makes sense to me, thanks.  How about this version?

Issue #1748, opened 2018-01-28
Intent, not syntax


JJ:
This is something I have found quite repeatedly through the documentation, and for that matter throughout almost everything that is written for every single language: Examples usually focus on syntax, and not on intent.  This is the last one I have found, but I'm sure you can think of many others: `lazy 1..5` when documenting  `lazy`. This is an absolutely correct and serviceable example. The problem is that when we are dealing with relatively uncommon features such as this one, we would avoid cargo-culting by learners by using examples that could be actually used in real-life code.
In this particular case, I guess that a sequence that needs heavy-weight computation and particularly that needs to compute every member based in the last one would be the case. This might be hard to think about, and focusing on syntax is generally OK because, after all, reference docs are not tutorials. However, in some cases, making the intent of the feature clearer would help newcomers to better understand the language and put it to use. 
I'm sure you can come up with many other examples you have bumped upon; if that's the case and you can work on then (or assign them to me), that would be fine. If you think this is not a priority now or simply don't agree, I'd like to hear your opinion.

zoffixznet:
In my experience, most "real world" examples are often convoluted and pack a lot of cruft completely unrelated to the construct presented. That's fine in blog articles and tutorials that go over each example in detail, but in docs, trying to get through 1000% more content, when all you're after is a sample of how to use the particular feature being discussed, can be daunting.

I briefly looked at [Signature docs](https://docs.perl6.org/type/Signature) and they example showing referencing a variable from the `where` clause looks like this:

```perl-6
sub one-of-them(:$a, :$b, :$c where { $a.defined ^^ $b.defined ^^ $c.defined }) {
    $a // $b // $c
};
say one-of-them(c=>42); # OUTPUT: «42␤» 
```

Not only it requires knowing and understanding a rather obscure `^^` operator (which reader unfamiliar with variables-in-wheres is unlikely to know), it abuses the `where` clause feature (it's an `only` sub... just stick a `die` at the end of the body instead of duplicating the checks), and doesn't actually demonstrate what it does (e.g. failing typecheck).

On the other hand, a completely useless and made up example below clearly shows I can "just use" a variable in a where clause and have its value work. And even if a person isn't familiar with WhaterCode yet, they can surmise the meaning of `*` in the code, thanks to the simple math operators used and explanatory sentence printed by the routine.

```perl-6
sub foo($a, $b where * == $a²) { say "$b is a square of $a" }
[...]

PR #3447, opened 2020-05-30
Clarify description of reduce sub/method for Any and List

uzluisf:

### DIFF
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -1051,24 +1051,28 @@ Defined as:
     multi method reduce(Any:D: &with)
     multi sub reduce (&with, +list)
 
-This routine combines the elements in a list-y object together, and thus
-produces a single result, by using a binary subroutine. The subroutine must be
-either an
+This routine combines the elements in a list-y object, and produces a single
+result, by applying a binary subroutine. It applies its argument (or first
+argument for the sub form) as an operator to all the elements in the object (or
+second argument for the sub form), producing a single result. The subroutine
+must be either an
 L<infix operator|https://docs.raku.org/language/operators#index-entry-infix_operator>
-or take two positional arguments.
+or take two positional arguments. When using an infix operator, we must provide
+the code object of its subroutine version, i.e., the operator category, followed
+by a colon, then a list quote construct with the symbol(s) that make up the
+operator (e.g., C«infix:<+>»). See L<Operators|/language/operators>.
 
-    say (1..4).reduce( &[+] );     # OUTPUT: «10␤»
-    say reduce &[+], 1..4;         # OUTPUT: «10␤»
+    say (1..4).reduce(&infix:<+>);   # OUTPUT: «10␤»
+    say reduce &infix:<+>, 1..4;     # OUTPUT: «10␤»
+    say reduce &min, 1..4;           # OUTPUT: «1␤»
 
-    say reduce &min, 1..4;         # OUTPUT: «1␤»
-
-    sub concat( Str() \s, Str() \t ) { s ~ t };
-    say reduce &concat, 'a'..'c';  # OUTPUT: «abc␤»
+    sub hyphenate(Str \a, Str \b) { a ~ '-' ~ b }
+    say reduce &hyphenate, 'a'..'c'; # OUTPUT: «a-b-c␤»
 
 Applied to a class, the routine will always return C<Nil>.
 
-    say Range.reduce( &[+] );       # OUTPUT: «Nil␤»
-    say Str.reduce( &[~] );         # OUTPUT: «Nil␤»
+    say Range.reduce(&infix:<+>);    # OUTPUT: «Nil␤»
+    say Str.reduce(&infix:<~>);      # OUTPUT: «Nil␤»
 
 See L<List.reduce|/type/List#routine_reduce> for a more thorough discussion.
 
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -1051,13 +1051,26 @@ Defined as:
     multi method reduce(Any:D: &with)
     multi sub reduce (&with, +list)
 
-Applying it to a class will always produce Nil.
-Applies its argument (or first argument, in case it's a sub) as an operator to
-all the elements in the object (or second argument), producing a single result.
-The argument must be an infix operator or take, in any case, two positional
-arguments.
+This routine combines the elements in a list-y object together, and thus
+produces a single result, by using a binary subroutine. The subroutine must be
+either an
+L<infix operator|https://docs.raku.org/language/operators#index-entry-infix_operator>
+or take two positional arguments.
 
-    (1..13).reduce( &[*] ).say; # OUTPUT: «6227020800␤»
+    say (1..4).reduce( &[+] );     # OUTPUT: «10␤»
+    say reduce &[+], 1..4;         # OUTPUT: «10␤»
+
+    say reduce &min, 1..4;         # OUTPUT: «1␤»
+
+    sub concat( Str() \s, Str() \t ) { s ~ t };
+    say reduce &concat, 'a'..'c';  # OUTPUT: «abc␤»
+
+Applied to a class, the routine will always return C<Nil>.
+
+    say Range.reduce( &[+] );       # OUTPUT: «Nil␤»
+    say Str.reduce( &[~] );         # OUTPUT: «Nil␤»
+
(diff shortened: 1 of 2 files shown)


doc/Type/Any.pod6
```diff
@@ -1051,13 +1051,26 @@ Defined as:
...
-Applying it to a class will always produce Nil.
-Applies its argument (or first argument, in case it's a sub) as an operator to
-all the elements in the object (or second argument), producing a single result.
-The argument must be an infix operator or take, in any case, two positional
-arguments.
+This routine combines the elements in a list-y object together, and thus
+produces a single result, by using a binary subroutine. The subroutine must be
+either an
+L<infix operator|https://docs.raku.org/language/operators#index-entry-infix_operator>
+or take two positional arguments.
```
JJ, line 1059: You've dropped the statement about what it's going to apply, that is, the reference to `with`. Also, the method is explained, but not the `sub`, which takes the list-y thing as a second argument.

doc/Type/Any.pod6
```diff
@@ -1051,13 +1051,26 @@ Defined as:
...
 
-    (1..13).reduce( &[*] ).say; # OUTPUT: «6227020800␤»
+    say (1..4).reduce( &[+] );     # OUTPUT: «10␤»
+    say reduce &[+], 1..4;         # OUTPUT: «10␤»
+
```
JJ, line 1062: This is not exactly what you describe, and I think it might be some subproduct of the hyper operator, which is what you're applying here. If you're going to use `+`, use the `infix<+>` definition, not the hyper. Also mention somewhere that it's probably better to use that latter.

doc/Type/Any.pod6
```diff
@@ -1051,13 +1051,26 @@ Defined as:
...
+    say reduce &min, 1..4;         # OUTPUT: «1␤»
+
+    sub concat( Str() \s, Str() \t ) { s ~ t };
+    say reduce &concat, 'a'..'c';  # OUTPUT: «abc␤»
```
JJ, line 1066: Try to reveal intent, not syntax #1748. You're making a trivial synonym of ~. Just think about some binary sub that could be used as an example.

doc/Type/Any.pod6
```diff
@@ -1051,13 +1051,26 @@ Defined as:
...
+
+Applied to a class, the routine will always return C<Nil>.
+
+    say Range.reduce( &[+] );       # OUTPUT: «Nil␤»
```
JJ, line 1070: Again, weird example. It's going to return Nil no matter what, try to use a stand alone operator.

doc/Type/List.pod6
```diff
@@ -884,24 +884,40 @@ Defined as:
     multi method reduce(Any:D: &with)
     multi sub reduce (&with, +list)
 
-The first form is obviously a no-op. The second form generates a single
+The first form is obviously a no-op. The remaining forms generate a single
```
JJ, line 887: Well, these methods are actually not defined in [List](https://github.com/rakudo/rakudo/blob/88d98225a4df61473ff5b935561719c56b04d1c2/src/core.c/List.pm6). Not in your PR, but as long as you're working with this, maybe the first could be eliminated, since it quite obviously does not belong here.

doc/Type/List.pod6
```diff
@@ -884,24 +884,40 @@ Defined as:
...
 "combined" value from a list of arbitrarily many values, by iteratively
-applying a function which knows how to combine I<two> values.
+applying a function which knows how to combine I<two> values. In addition to
```
JJ, line 889: function -> routine or operator?

doc/Type/List.pod6
```diff
@@ -884,24 +884,40 @@ Defined as:
...
-If C<@values> contains just a single element, the operator is applied to that
+    result0 = init
+    result1 = f(result0, list[0])
+    result2 = f(result1, list[1])
+    ...
+    resultn = f(resultn-1, list[n-1])
+
+C<resultn> is the final result for an n-element list.
+
+    say reduce &[+], [1, 2, 3];     # OUTPUT: «6␤»
```
JJ, line 903: Again, weird example. In this case, it's better to just apply [+], without reduce.

doc/Type/List.pod6
```diff
@@ -884,24 +884,40 @@ Defined as:
...
+    say [1, 2, 3].reduce: &[+];     # OUTPUT: «6␤»
```
JJ, line 904: This is an array. While it's totally legal, maybe we should circumscribe ourselves to talk about arrays here. Above you're using arrays too.

doc/Type/List.pod6
```diff
@@ -884,24 +884,40 @@ Defined as:
...
+    say reduce &max, [5, 9, 12, 1]; # OUTPUT: «12␤»
+
+If C<list> contains just a single element, the operator is applied to that
 single element if possible; if not, it returns the element itself.
 
-    say [-] <10 5 3>; #OUTPUT: 2␤
-    say [-] 10;       #OUTPUT: 10␤
+    say reduce &[+], <10 5 3>; # OUTPUT: «2␤»
```
JJ, line 910: The initial example was bad, because it was using the hyperoperator instead of the reduce routine, but it made sense by itself. Again, this one is not really a good example.

doc/Type/List.pod6
```diff
@@ -919,25 +935,49 @@ for operators which aren't left-associative:
...
+user-defined subroutine instead, provide an additional layer of square brackets
+around the subroutine's code object:
 
-    # The following all do the same thing...
-    my @numbers = (1,2,3,4,5);
+    say [*] [1, 2, 3, 4];       # OUTPUT: «24␤»
+    say [min] [4, 2, 1, 3];     # OUTPUT: «1␤»
+
+    sub mult { $^a * $^b };
+    say [[&mult]] [1, 2, 3, 4]; # OUTPUT: «24␤»
```
JJ, line 949: Explain why you're using double hyper here...

doc/Type/List.pod6
```diff
@@ -919,25 +935,49 @@ for operators which aren't left-associative:
...
+
+    my @numbers = 1, 2, 3, 4, 5;
     say reduce { $^a + $^b }, 0, |@numbers;
     say reduce * + *, 0, |@numbers;
-    say reduce &[+], @numbers; # operator does not need explicit identity
-    say [+] @numbers;          # most people write it this way
+    say reduce &[+], @numbers; # operator does not need explicit identity value
+    say [+] @numbers;
+
+Since C<reduce> is an implicit loop, it responds to C<next>, C<last> and C<redo>
```
JJ, line 959: "is an" was already there, but again, maybe express it better. "Iterates over its reducing function?"

doc/Type/List.pod6
```diff
@@ -919,25 +935,49 @@ for operators which aren't left-associative:
...
+with the L<C«is assoc»|/language/functions#index-entry-is_assoc_(trait)>.
 
-    # Generate a random-ish math formula like "(4 + ((3 * x) + 11) / 6))"
+    sub infix:<foo>($a, $b) is assoc<right> { "($a, $b)" }
+    say [foo] 1, 2, 3, 4; # OUTPUT: «(1, (2, (3, 4)))␤»
+
+    sub infix:<bar>($a, $b) is assoc<left> { "($a, $b)" }
+    say [bar] 1, 2, 3, 4; # OUTPUT: «(((1, 2), 3), 4)␤»
+
+B<Practical example 1:> In this example, we generate a random-ish math
```
JJ, line 979: This is nice. :100:

doc/Type/Any.pod6
uzluisf: >If you're going to use +, use the infix<+> definition, not the hyper. Also mention somewhere that it's probably better to use that latter.

I left it out from this page since I think it's better discussed in the `List` page where it's introduced. However, I didn't know how to word it so I didn't mention there either. Probably better for another PR.

doc/Type/List.pod6
uzluisf: I did away with the `Array`s and instead used `List`s almost everywhere.

doc/Type/List.pod6
uzluisf: I mention it in [the paragraph preceding the example](https://github.com/Raku/doc/pull/3447/commits/9ef5e5e452bd163871d35db986c8fc51ccb069d2#diff-cd75d8f66462222cb39d1d41c2823e9cR951-R953).

PR #3453, opened 2020-06-01
Add example's output

stoned:


doc/Language/containers.pod6
```diff
@@ -107,8 +107,8 @@ omitted when executing the C<Callable>. The default type constraint is
 L<Callable|/type/Callable>.
 
     my &callable = -> $ν { say "$ν is", $ν ~~ Int??" whole"!!" not whole" }
-    callable(⅓);
-    callable(3);
+    callable(⅓);   # OUTPUT: «0.333333 is not whole␤»
+    callable(3);   # OUTPUT: «3 is whole␤»
```
Altai-man, line 111: Cool! While we at it, I think adding spaces around ??!! will make it better too, can you please edit it as well?

stoned: Sure. Done!

Issue #3320, opened 2020-04-11
Some problems in 101 basics


JJ:
## The problem

The heading "array, methods and invocant" seem to be in no man's land
![Captura de pantalla de 2020-04-11 10-13-38](https://user-images.githubusercontent.com/500/79038880-42fbab80-7bdd-11ea-851d-1aabda29c98a.png)

## Suggestions

Check it; besides, it includes an `array` index, although it seems to refer to non-native arrays.

In general, I don't really get that page. It looks like things that should be just indexes have become actual text. In general, it kinda makes sense, but text in one section refers to examples in the previous section, so we kinda get lost in the middle...

coke:
Looks like this was resolved at some point, I can’t find the text above in that page anymore!

PR #3457, opened 2020-06-02
Revise 'Raku by example 101' page

uzluisf:
* Organize the page such that h1 headers introduce a few lines of code
that might be further discussed by subheaders. This way the examples
are self-contained to their own section (and subsections).

* Remove meta-commentary from the sample code. Instead, leave it for the
  discussion text.

* Index terms more clearly.

Refs: #3320

### DIFF
--- a/doc/Language/101-basics.pod6
+++ b/doc/Language/101-basics.pod6
@@ -77,7 +77,7 @@ Beth has won 1 match and 4 sets
 
 =head1 Pragma X<C<v6>|v6 (Basics)>
 
-=begin code :preamble<use v6>
+=begin code
 use v6;
 =end code
 
@@ -339,7 +339,7 @@ have both won two matches. That simple sort doesn't account for the number of
 sets won, which is the secondary criterion to decide who has won the
 tournament.
 
-=head2 X<Stable sort|stable sort (Basics)>
+=head2 Stable sort
 
 When two array items have the same value, C<sort> leaves them in the same order
 as it found them. Computer scientists call this a I<stable sort>. The program
@@ -356,9 +356,6 @@ C<sort> sorts in ascending order, from smallest to largest. This is the
 opposite of the desired order. Therefore, the code calls the C<.reverse> method
 on the result of the second sort, and stores the final list in C<@sorted>.
 
-X<|say (Basics)>
-X<|print (Basics)>
-X<|put (Basics)>
 =head1 Standard output
 
 =begin code :preamble<my @sorted;my %matches;my %sets>
@@ -446,7 +443,7 @@ followed by a postcircumfix E<mdash> a bracketing pair that follows a
 statement. It's also ok to have a method call between the variable name and
 the postcircumfix.
 
-=head1 X<Zen slices|Zen slices (Basics)>
+=head1 Zen slices
 
 =begin code
 my @flavors = <vanilla peach>;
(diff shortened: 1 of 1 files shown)


doc/Language/101-basics.pod6
```diff
@@ -74,37 +66,55 @@ for @sorted -> $n {
...
 Ana has won 2 matches and 8 sets
 Dave has won 2 matches and 6 sets
 Charlie has won 1 match and 4 sets
 Beth has won 1 match and 4 sets
+=end code
 
-=head3 X<C<v6>|v6 (Basics)>
+=head1 Pragma X<C<v6>|v6 (Basics)>
+
+=begin code :preamble<use v6>
```
JJ, line 80: Er. You add the preamble to the code itself? Wouldn't this be :solo?

doc/Language/101-basics.pod6
```diff
@@ -325,13 +339,13 @@ have both won two matches. That simple sort doesn't account for the number of
 sets won, which is the secondary criterion to decide who has won the
 tournament.
 
-=head3 X<C<stable sort>>
+=head2 X<Stable sort|stable sort (Basics)>
```
JJ, line 342: You are adding (basics) all over, or only those that have another entry somewhere? Because I don't remember this to be indexed. Also, not so sure it should be indexed...

doc/Language/101-basics.pod6
```diff
@@ -342,6 +356,11 @@ C<sort> sorts in ascending order, from smallest to largest. This is the
 opposite of the desired order. Therefore, the code calls the C<.reverse> method
 on the result of the second sort, and stores the final list in C<@sorted>.
 
+X<|say (Basics)>
```
JJ, line 359: Again, why index this? Or, for that matter, mostly anything here if it's not explained anywhere? It's supposed to be a tutorial, if we add all kinds of entries we might end up with too many of them...

doc/Language/101-basics.pod6
```diff
@@ -429,7 +446,7 @@ followed by a postcircumfix E<mdash> a bracketing pair that follows a
 statement. It's also ok to have a method call between the variable name and
 the postcircumfix.
 
-=head3 X<C<Zen slice>| Zen slice (Basics)>
+=head1 X<Zen slices|Zen slices (Basics)>
```
JJ, line 449: This is very well covered elsewhere, really no need to index it again.

doc/Language/101-basics.pod6
uzluisf: I guess this configuration information is special to Documentable. Is there some documentation on how it uses them?

doc/Language/101-basics.pod6
uzluisf: [This was already indexed](https://github.com/Raku/doc/blob/master/doc/Language/101-basics.pod6#L353).

doc/Language/101-basics.pod6
JJ: Well, the preamble thing is common to all documentation. What I'm saying is that it makes no sense to repeat the code in the preamble. And that in these case, IARA, you can use :solo so that it's compiled in a single file, although I guess it's OK if you simply elminate the preamble.

doc/Language/101-basics.pod6
JJ: Yep, as I say above, maybe too many things were indexed here. And if it was, what did you do precisely here? Change the anchor?

PR #3471, opened 2020-06-10
$_ defaults to OUTER::<$_>

gfldex:
no reason to hide this fact

### DIFF
--- a/doc/Type/Block.pod6
+++ b/doc/Type/Block.pod6
@@ -10,8 +10,8 @@ A C<Block> is a code object meant for small-scale code reuse.  A block is
 created syntactically by a list of statements enclosed in curly braces.
 
 Without an explicit signature or placeholder arguments, a block has C<$_>
-as a positional argument. Which defaults to the outer scope's $_. Thus it
-will inherit the L<topic|/language/variables#index-entry-topic_variable> if possible.
+as a positional argument, which defaults to the outer scope's C<$_>. Thus it
+will inherit the L<topic|/language/variables#index-entry-topic_variable> if there is any.
 
     my $block = { uc $_; };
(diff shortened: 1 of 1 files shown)


doc/Type/Block.pod6
```diff
@@ -10,11 +10,13 @@ A C<Block> is a code object meant for small-scale code reuse.  A block is
 created syntactically by a list of statements enclosed in curly braces.
 
 Without an explicit signature or placeholder arguments, a block has C<$_>
-as a positional argument
+as a positional argument. Which defaults to the outer scope's $_. Thus it
```
JJ, line 13: ", which" instead of ". Which" maybe?

doc/Type/Block.pod6
```diff
@@ -10,11 +10,13 @@ A C<Block> is a code object meant for small-scale code reuse.  A block is
...
+will inherit the L<topic|/language/variables#index-entry-topic_variable> if possible.
```
JJ, line 14: Maybe clarify "if possible". You probably mean "unless it's called with an argument".

doc/Type/Block.pod6
```diff
@@ -10,11 +10,13 @@ A C<Block> is a code object meant for small-scale code reuse.  A block is
...
 
     my $block = { uc $_; };
     say $block.^name;           # OUTPUT: «Block␤»
     say $block('hello');        # OUTPUT: «HELLO␤»
+    say {;}.signature;          # OUTPUT: «(;; $_? is raw = OUTER::<$_>)␤»
```
JJ, line 20: Maybe add some example here?
```perl6
$_ = "hello";
my &block = { uc $_; };
say block # HELLO
```

You probably need to use the & form so that you can invoke it without any argument.

PR #3486, opened 2020-06-19
show that roles are objects

gfldex:
And note the gotcha with parameterized roles as objects.

Discussion: https://colabti.org/irclogger/irclogger_log/raku-dev?date=2020-06-18#l384


doc/Language/typesystem.pod6
```diff
@@ -629,6 +629,20 @@ Default parameters can be provided.
     my $i = 1 does R;
     CATCH { default { say .^name, ': ', .Str} }
     # OUTPUT: «X::AdHoc: Could not instantiate role 'R':␤Please provide a parameter to role R␤»
+    
+Roles can be stored in a scalar and so can parameterized roles. Specializing them does not work
```
JJ, line 633: I'm not positive we support British spelling. Let me check.

PR #3487, opened 2020-06-20
Roles are not closures but traps

gfldex:
discussion: https://colabti.org/irclogger/irclogger_log/raku-dev?date=2020-06-18#l366

### DIFF
--- a/doc/Language/traps.pod6
+++ b/doc/Language/traps.pod6
@@ -1992,7 +1992,7 @@ say @a.grep( 2 ~~ *.Int ); # OUTPUT: «(2)␤»
 
 =head1 Closures
 
-Raku accepts closures syntax in many places. Not all of them are supposed to work.
+Raku accepts closure syntax in many places. Not all of them are supposed to work.
 
 =head2 Role definition
 
--- a/doc/Language/traps.pod6
+++ b/doc/Language/traps.pod6
@@ -1992,20 +1992,21 @@ say @a.grep( 2 ~~ *.Int ); # OUTPUT: «(2)␤»
 
 =head1 Closures
 
-Raku supports closures in many places. Not all of them are supposed to work.
+Raku accepts closures syntax in many places. Not all of them are supposed to work.
 
 =head2 Role definition
 
 Roles can take a variable from the outer scope to initialize an attribute with a default value.
 
 =begin code
-my $seamingly-closure-var = 'will vanish';
-my $s = "don't do this" but role :: { has $.attr = $seamingly-closure-var }
+my $apparently-closure-var = 'will vanish';
+my $s = "don't do this" but role :: { has $.attr = $apparently-closure-var }
 =end code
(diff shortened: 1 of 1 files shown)


gfldex:
There will be a follow up PR that provides links to this trap.

doc/Language/traps.pod6
```diff
@@ -1990,5 +1990,22 @@ my @a = <1 2 3>;
 say @a.grep( 2 ~~ *.Int ); # OUTPUT: «(2)␤»
 =end code
 
+=head1 Closures
+
+Raku supports closures in many places. Not all of them are supposed to work.
+
```
JJ, line 1996: Hum, what do you mean exactly here? If it "supports" them, shouldn't it be supposed to work?

doc/Language/traps.pod6
```diff
@@ -1990,5 +1990,22 @@ my @a = <1 2 3>;
...
+=head2 Role definition
+
+Roles can take a variable from the outer scope to initialize an attribute with a default value.
+
+=begin code
+my $seamingly-closure-var = 'will vanish';
+my $s = "don't do this" but role :: { has $.attr = $seamingly-closure-var }
```
JJ, line 2003: Maybe you mean "seemingly" here?

JJ: Even better `$apparently-a-closure-var`

doc/Language/traps.pod6
```diff
@@ -1990,5 +1990,22 @@ my @a = <1 2 3>;
...
+=end code
+
+The attribute will turn into C<Any> eventually. This trap is particular nasty because this may not
+happen before the program ends. Concurrency will trigger this behaviour often. At the time of writing
+Rakudo does not provide a warning either.
+
```
JJ, line 2009: But the example does not provide an actual working version of the trap. If it's going to turn into Any, it would be better if you provided an example where it does.

gfldex: We already got an example in /language/typesystem . Link will be provided.

JJ: Not sure the link above is the one you intended to provide, but anyway it's much clearer now.

doc/Language/traps.pod6
JJ: Please take a look at this thing.

raiph:
"Raku accepts closure syntax in many places. Not all of them are supposed to work."

Fwiw, imo that's a confusing way to express it. Raku accepts *block* syntax in many places. They all work. Not all [blocks](https://en.wikipedia.org/wiki/Block_(programming)) are [closures](https://en.wikipedia.org/wiki/Closure_(computer_programming)). For example, role blocks aren't closures. (Likewise class, grammar, phaser blocks, etc.)

gfldex:
I give up. The docs are all yours.

PR #3490, opened 2020-06-20
Document is impl trait

Altai-man:
Resolves a point for https://github.com/Raku/doc/issues/3360 and closes https://github.com/Raku/doc/issues/3485

### DIFF
--- a/doc/Language/traits.pod6
+++ b/doc/Language/traits.pod6
@@ -108,7 +108,7 @@ would be a way of adding functionality which cuts across class and role
 hierarchies, or can even be used to add behaviors to independently defined
 routines.
 
-=head2 X<C<is implementation-detail>,traits> trait
+=head2 X<C<is implementation-detail>, traits> trait
 
 Note: this trait has been available in Rakudo compiler starting from 2020.05 release.
 
--- a/doc/Language/traits.pod6
+++ b/doc/Language/traits.pod6
@@ -108,7 +108,7 @@ would be a way of adding functionality which cuts across class and role
 hierarchies, or can even be used to add behaviors to independently defined
 routines.
 
-=head2 X<C<is implementation-detail>, traits> trait
+=head2 X<C<is implementation-detail>|is implementation-detail, traits> trait
 
 Note: this trait has been available in Rakudo compiler starting from 2020.05 release.
 
--- a/doc/Language/traits.pod6
+++ b/doc/Language/traits.pod6
@@ -90,7 +90,7 @@ On the other hand, X<P6opaque> is the default representation used for
 all objects in Raku.
 
     class Thar {};
-    say Thar.REPR;    #OUTPUT: «P6opaque␤»
+    say Thar.REPR;    # OUTPUT: «P6opaque␤»
 
 The L<metaobject protocol|/language/mop> uses it by default for every object
 and class unless specified otherwise; for that reason, it is in general not
@@ -110,14 +110,22 @@ routines.
 
 =head2 X<C<is implementation-detail>,traits> trait
 
-Note: this trait is available in Rakudo compiler starting from 2020.05 release.
+Note: this trait has been available in Rakudo compiler starting from 2020.05 release.
 
-The C<is> with C<implementation-detail> argument is used by Raku
-language implementations and module authors to mark particular
-routines (including methods) as not meant to be a public API.
+This trait is used by Raku language implementations and module authors
+to mark particular routines (including methods) as not meant to be a
+part of public API. While such routines can be found when looked up
+directly, they will not appear in results of introspection:
 
-Applying this trait makes the C<is-implementation-detail> method on
-L<Code|/type/Code> to return C<True>.
+    my &do-not-use-routine = CORE::<&DYNAMIC>;
+    say CORE::.keys.grep(* eq '&DYNAMIC'); # OUTPUT: «()␤»
+
+Such routines are not meant for use by users and their behavior and availability can be changed
+anytime.
+
+Applying this trait makes the C<is-implementation-detail> method called on
+L<Code|/type/Code> to return C<True>, thus giving a hint to the user not to use it
+if they are not willing to maintain this code in case of changes for years to come:
 
     my &fail-routine = &fail;
     unless &fail-routine.is-implementation-detail {
--- a/doc/Language/traits.pod6
+++ b/doc/Language/traits.pod6
@@ -129,7 +129,7 @@ if they are not willing to maintain this code in case of changes for years to co
 
     my &fail-routine = &fail;
     unless &fail-routine.is-implementation-detail {
-        say "It is safe to use &fail routine in your modules!";
+        say "&fail is not an implementation detail, can expect backward compatibility";
     }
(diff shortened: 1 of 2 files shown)


doc/Language/traits.pod6
```diff
@@ -108,6 +108,27 @@ would be a way of adding functionality which cuts across class and role
 hierarchies, or can even be used to add behaviors to independently defined
 routines.
 
+=head2 X<C<is implementation-detail>,traits> trait
+
```
JJ, line 112: You sure about this syntax? I would say X<C<is implementation-detail>|is implementation-detail,traits> is the way I've usually seen it.

doc/Language/traits.pod6
```diff
@@ -108,6 +108,27 @@ would be a way of adding functionality which cuts across class and role
...
+Note: this trait is available in Rakudo compiler starting from 2020.05 release.
+
```
JJ, line 114: "has been", instead of "is"

doc/Language/traits.pod6
```diff
@@ -108,6 +108,27 @@ would be a way of adding functionality which cuts across class and role
...
+The C<is> with C<implementation-detail> argument is used by Raku
+language implementations and module authors to mark particular
```
JJ, line 116: Maybe cut it short to "This trait is used..."

doc/Language/traits.pod6
```diff
@@ -108,6 +108,27 @@ would be a way of adding functionality which cuts across class and role
...
+routines (including methods) as not meant to be a public API.
```
JJ, line 117: "part of", maybe. Also clarify what's exactly meant by this, that is, that you can use them and they are public-facing, but they can be removed or changed at any time in the future, so future-proof related code.

doc/Language/traits.pod6
```diff
@@ -108,6 +108,27 @@ would be a way of adding functionality which cuts across class and role
...
+
+Applying this trait makes the C<is-implementation-detail> method on
+L<Code|/type/Code> to return C<True>.
+
+    my &fail-routine = &fail;
+    unless &fail-routine.is-implementation-detail {
+        say "It is safe to use &fail routine in your modules!";
+    }
```
JJ, line 125: Why the reference to safety? Shouldn't you clarify that before, implying that implementation details can be removed at any time from the API?

doc/Language/traits.pod6
```diff
@@ -108,6 +108,27 @@ would be a way of adding functionality which cuts across class and role
...
+
+    sub PRIVATE-CALCULATION is implementation-detail { #`(Not safe to rely on this) }
+    if &PRIVATE-CALCULATION.is-implementation-detail {
+        say "You better not to rely on &PRIVATE-CALCULATION unless you really know what you are doing";
+    }
+
 =end pod
```
JJ, line 133: Maybe also add something about what it implies; for instance, not showing up in method lookups (if that's the case)

doc/Language/traits.pod6
Altai-man: No, I am totally not sure about this syntax.
I wonder if I am the only one who founds the indexing syntax very prone to errors, which resulted in our troubles with categories and stuff.
OTOH, I did grep the docs and saw usages of such style, so don't think it hurts.
I think writing out a clear rules explaining the indexing syntax will help a lot.

As I understand it, we have:
`X< $displayed-on-page-part | $a-text-by-that-you-can-search , $category; $another-text-for-search , $category2 ...>`
Maybe. I am not really sure. However:

```
12:=head2 X<C<lazy>|lazy (statement prefix)>
436:=head2 X<C<self>>
871:=head3 X<C<Package>, C<Module>, C<Class>, C<Role>, and C<Grammar> declaration|declarator,unit;declarator,module;declarator,package>
```
^ we have those in docs already, which sounds matching the grammar given above.

JJ: The thing is categories are *behind* the `|` always, never close to the indexed term. Theoretically, indexing is covered by the Synopsis and should be tested by Documentable. Let's put the other one I've accepted to test and see what happens, but my money is on no.

doc/Language/traits.pod6
Altai-man: The example tries to say the `fail` routine is ok to use in your modules, no? Extended the explanation, though.

Altai-man:
Another round, please?

doc/Language/traits.pod6
JJ: The problem is what you mean by "is ok". An implementation detail can be used as long as you test for it, for instance, or have an easy way out if it's changed. So "safe" or "is OK" are not really shorthand for that...

doc/Type/Code.pod6
```diff
@@ -187,6 +187,15 @@ Returns the line number in which the code object was declared.
 
     say &infix:<+>.line;
 
+=head2 method is-implementation-detail
+
+    method is-implementation-detail(--> False)
+
+Note: this trait has been available in Rakudo compiler starting from 2020.05 release.
+
+Returns C<True> if code object was marked with C<is implementation-detail> trait,
```
JJ, line 196: *the* code object.

doc/Language/traits.pod6
Altai-man: Just to clarify a possible misunderstanding...
This condition tests if the routine (`fail`) is *not* an implementation detail, and then says "If it is not (and it is not), then it is safe to use it". Is it wrong?

doc/Language/traits.pod6
Altai-man: Added.

doc/Language/traits.pod6
JJ: Not really wrong, but I'm not sure "safe" is the right word for it. It's "safe" as long as it does not change, but it might change; again, it's "safe" if you future-proof your code by testing for those features and having a B plan in case they change.

Altai-man: What about "&fail is not not an implementation detail, can expect backward compatibility"?

doc/Language/traits.pod6
Altai-man: So do we change it? I am not sure anymore.

JJ: Yes, lease.

doc/Language/traits.pod6
JJ: except for the double "not", I guess it's better.

doc/Language/traits.pod6
JJ: This still to go. It will fail anyway in space-after-comma.

JJ: I would *really* prefer you to use the common syntax. This is not going to work, and it will show up in text as anything inside the `X<>`

Altai-man: Apparently, my copy-pasting of your first comment somehow failed... Sorry for the noise.

Altai-man: Please, select squashing as an option for this PR.

PR #3499, opened 2020-06-27
Make existing info on contributing more visible

finanalyst:
## The problem
During the recent Perl conference session on docs, it seemed something on Contributing was missing. The documentation is available but lost downpage

## Solution provided
Move the document to the 'Beginning section', change the Subtitle to something more immediately apparent than 'metadocumentation'


doc/Language/about.pod6
```diff
@@ -1,8 +1,8 @@
-=begin pod :kind("Language") :subkind("Language") :category("reference")
+=begin pod :kind("Language") :subkind("Language") :category("beginning")
 
 =TITLE About the docs
 
-=SUBTITLE Metadocumentation
+=SUBTITLE How to contribute to Raku and to the documents, how to generate the docs locally
```
JJ, line 6: Why not "and" instead of comma?

PR #3500, opened 2020-06-29
Expand on the use of {*} with proto

finanalyst:
## The problem

The documentation on proto was not as clear as I needed. I tried several times to understand it, but failed. 
I got it nearly right when I asked on perl6-users.  
Crucially '{*}' is a special token without whitespace - this needs to be noted.  
The extra example combines the proto declaration with methods that use it.

## Solution provided

Added a note and a longer example


coke:
In addition to the curlies comment above, please add "newclass" to xt/code.pws (sorted) so the spellcheck test passes. There is also some trailing whitespace that should be removed.

Thanks!

finanalyst:
@coke amended as suggested

coke:
(I accidentally pushed a commit to this branch post-merge, instead of master, apologies)

Issue #1808, opened 2018-03-02
Mention somewhere that TWEAK should not be used in roles (trap?)


AlexDaniel:
Context: https://github.com/perl6/whateverable/blob/177b77cb2ebc045736b8e7a1cf6eb8e25fdce7b6/lib/Whateverable.pm6#L77
IRC discussion: https://irclog.perlgeek.de/perl6-dev/2018-02-24#i_15854365

To be honest I'm not even sure if this is something that should be doc-ed. I consider it as a bug. But still, that's a too major thingie for it to be mentioned nowhere at all.

JJ:
Maybe in the Roles page? Is that the only thing that can't be used in Roles?

JJ:
OK, I read the discussion. The situation right now is that TWEAK (or any submethod) overwrites that of the class it's mixed into. The ideal situation would be one in which all mixed-in TWEAKs are "glommed" together (whatever that means). I don't think this is really a NOTSPECCED, it's rather a wishlist. The situation now is that it's better not to TWEAK in roles. Can I document that as a trap?

JJ:
Ref #2601

Issue #1796, opened 2018-02-23
add an example of TWEAK


tbrowder:
also explain its advantages over other methods of object contruction

JJ:
Where? For instance, here: https://docs.perl6.org/language/objects ? Right now there is this example:

```
class RectangleWithCachedArea {
    has ($.x1, $.x2, $.y1, $.y2);
    has $.area;
    submethod TWEAK() {
        $!area = abs( ($!x2 - $!x1) * ( $!y2 - $!y1) );
    }
}
 
say RectangleWithCachedArea.new( x2 => 5, x1 => 1, y2 => 1, y1 => 0).area;
```
It's also mentioned several times. I don't think it really deals with its advantages...

tbrowder:
I didn't see the example because it is far below the search  reference point for TWEAK. I added a note there that there is an example later in the text.  Closed with  1b723a26bc4ea3c09a983f5e351e697860c56da8.

JJ:
Reopen with slightly changed title.

tbrowder:
This is still troubling me for the paucity of a real-world example. I asked for help for a practical class construction on #raku a couple of days ago and got some info that is helpful but I don't find it in the docs. I'm looking for the discussion reference now and will add it here.

JJ:
Please do.

tbrowder:
Wrong, TWEAK enables me to do exactly what I need for my use case.

tbrowder:
Submethod TWEAK is missing on the Type system page (typesystem.pod6)

PR #3502, opened 2020-07-05
Adding more information about TWEAK

finanalyst:
## The problem
There is almost nothing about TWEAK in the documentation. Even when the TWEAK is mentioned, it is not clear why it is needed, until you get a compile error like the one shown in the example.  
 Further, the documentation does not contain any reference to the signature of TWEAK.  
A small test with `submethod TWEAK(|c) {say $c.raku}` shows that it is passed the same variables as are passed to BUILD.

Added documentation (and fixed a typo in the BUILD description)

### DIFF
--- a/doc/Language/typesystem.pod6
+++ b/doc/Language/typesystem.pod6
@@ -239,7 +239,7 @@ class C {
 
 The L<submethod|/type/Submethod> C<BUILD> is (indirectly) called by
 L<.bless|/type/Mu#method_bless>. It
-is meant to set private and public attributes of a class and receives all names
+is meant to set private and public attributes of a class and receives all named
 attributes passed into C<.bless>. The default
 constructor L<.new|/type/Mu#method_new> defined in C<Mu> is the method that
 invokes it. Given that public accessor methods are not available in C<BUILD>,
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
     C.new.say; C.new('answer').say;
     # OUTPUT: «C.new(attr => 42)␤
     #          C.new(attr => "answer")␤»
+    
+=head4 submethod TWEAK
+
+The L<submethod|/type/Submethod> C<TWEAK> is (indirectly) called by
+L<.bless|/type/Mu#method_bless>. It
+is meant to set private and public attributes of a class after the BUILD submethod has been called. The default
+constructor L<.new|/type/Mu#method_new> defined in C<Mu> is the method that
+invokes it. Given that public accessor methods are not available in C<TWEAK>,
+you must use private attribute notation instead.
+
+The difference between BUILD and TWEAK is that not all the attributes in an object have been built.
+In addition, when there is inheritance the BUILD submethods of the child classes are called after
+the BUILD submethods of the parent classes. The TWEAK submethods are called after all the BUILD 
+submethods.
+
+Both C<BUILD> and C<TWEAK> receive all named
+attributes passed into C<.bless>
+
+    class C {
+        has $.m = 'nothing';
+        has $.attr = 42;
+        submethod BUILD {
+            # this will fail because $!attr has not been given its default value yet
+            $!message = 'A very special number' if $!attr == 42;
+        }
+    };
+    class D {
+        has $.m = 'nothing';
+        has $.attr = 42;
+        submethod TWEAK {
+            $!message = 'A very special number' if $!attr == 42;
+        }
+    };
+
+    C.new.m.say;
+    # OUTPUT: 
+    # Use of uninitialized value of type Any in numeric context
+    # in submethod BUILD at <unknown file> line 1
+    # in block <unit> at <unknown file> line 1
+    #
+    # nothing
+    D.new.m.say;
+    # OUTPUT: A very special number
 
 =head4 Fallback method
 X<|FALLBACK (method)>


doc/Language/typesystem.pod6
```diff
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
     C.new.say; C.new('answer').say;
     # OUTPUT: «C.new(attr => 42)␤
     #          C.new(attr => "answer")␤»
+    
+=head4 submethod TWEAK
+
+The L<submethod|/type/Submethod> C<TWEAK> is (indirectly) called by
+L<.bless|/type/Mu#method_bless>. It
```
JJ, line 265: It's indirectly called when an object is created, even if it does not call explicitly `bless`

doc/Language/typesystem.pod6
```diff
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
...
+is meant to set private and public attributes of a class after the BUILD submethod has been called. The default
```
JJ, line 266: meant to compute attributes that can't be known at build time. You can set them in the `BUILD` phaser.

doc/Language/typesystem.pod6
```diff
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
...
+constructor L<.new|/type/Mu#method_new> defined in C<Mu> is the method that
+invokes it. Given that public accessor methods are not available in C<TWEAK>,
```
JJ, line 268: Not clear this is that way, or that actually means something. It's a phaser, called after BUILD. Whether it's called from Mu.new or not does not seem too relevant.
Also, are you sure public methods are not available? Can you provide an example?

doc/Language/typesystem.pod6
```diff
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
...
+you must use private attribute notation instead.
+
+The difference between BUILD and TWEAK is that not all the attributes in an object have been built.
+In addition, when there is inheritance the BUILD submethods of the child classes are called after
```
JJ, line 272: Not really true either. You can build as many as you want in BUILD, or bind them in the call. But `self` is not available, so computed attributes might need to be defined later, during TWEAK.

doc/Language/typesystem.pod6
```diff
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
...
+the BUILD submethods of the parent classes. The TWEAK submethods are called after all the BUILD 
+submethods.
+
```
JJ, line 275: All TWEAKs after all BUILDs or every TWEAK after its own BUILD?

doc/Language/typesystem.pod6
```diff
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
...
+Both C<BUILD> and C<TWEAK> receive all named
+attributes passed into C<.bless>
+
+    class C {
+        has $.m = 'nothing';
+        has $.attr = 42;
+        submethod BUILD {
+            # this will fail because $!attr has not been given its default value yet
+            $!message = 'A very special number' if $!attr == 42;
```
JJ, line 284: Because `self` does not yet exist.

doc/Language/typesystem.pod6
```diff
@@ -258,6 +258,49 @@ you must use private attribute notation instead.
...
+
+    C.new.m.say;
+    # OUTPUT: 
+    # Use of uninitialized value of type Any in numeric context
+    # in submethod BUILD at <unknown file> line 1
+    # in block <unit> at <unknown file> line 1
+    #
+    # nothing
+    D.new.m.say;
+    # OUTPUT: A very special number
```
JJ, line 304: This example is OK because it shows when to use attributes and when they are not available; but still, I'm not too convinced about the rest of the explanation.

doc/Language/typesystem.pod6
finanalyst: I used EXACTLY the same words as for submethod BUILD. If the explanation for TWEAK is wrong, then so is the explanation for BUILD. 

My understanding is exactly as documented. 'new' calls 'bless' which calls all BUILDs and then all TWEAKs.

doc/Language/typesystem.pod6
finanalyst: submethod BUILD and BUILD phaser are different. The first is for object instantiation and can occur many times. The second is for code compilation of a program, which probably should only occur once per run.

doc/Language/typesystem.pod6
finanalyst: JJ: I have run into the problem quite a lot. I am creating a class. I BUILD an object (not the phaser). But the attribute has a default that has not been applied in BUILD. But it is there in TWEAK. That's what my example shows.

doc/Language/typesystem.pod6
finanalyst: Now that is something I do not know. Be interesting to find out. Or to have my text edited to show.

doc/Language/typesystem.pod6
finanalyst: Nothing to do with self. $!attr can be manipulated in BUILD (see example in submethod BUILD explanation in the section about where I have inserted this text.

doc/Language/typesystem.pod6
finanalyst: The example shows exactly why TWEAK is needed. Naturally its a bit simple, but a more complex situation needs more code. 

Re #1808 - I can add that no problem.
Re #1796 - I think my text resolves the issue. It explains what TWEAK is, and why it is needed in addition to BUILD.

doc/Language/typesystem.pod6
JJ: Yeah... maybe it's not the best explanation either... My point anyway was that there was no need in the explanation to involve `bless`. Quite enough if you just say when it happens.

doc/Language/typesystem.pod6
JJ: Well, they're "phasers" when building an object. Changed it to submethod, the main point still holds; you can set private and public attributes in BUILD.

doc/Language/typesystem.pod6
JJ: You can apply the default in the BUILD signature, but the point is that since self has not been built, no defaults are available. It's not the values, it's the object. The whole point of TWEAK is that the object is already built. Same thing when an attribute is required: you need to assign a value during BUILD; TWEAK is too late.

doc/Language/typesystem.pod6
JJ: Or just change to what we know is true: every object's TWEAK happens after every object's BUILD.

doc/Language/typesystem.pod6
JJ: It can be bound, it can't be really be manipulated. $!attr is self.attr, and just try and use it during BUILD.

doc/Language/typesystem.pod6
finanalyst: I think there is confusion here. Either you or me. I think: BUILD phaser and object submethod BUILD (and TWEAK) are not the same.

doc/Language/typesystem.pod6
finanalyst: Just run a small program (see below). It's clear: each TWEAK follows directly after the BUILD, from parent to child.
```
#!/usr/bin/env perl6
class A {
    has $.a;
    submethod BUILD {
        say 'In A BUILD'
    }
    submethod TWEAK {
        say 'In A TWEAK'
    }
}
class B is A {
    has $.a;
    submethod BUILD {
        say 'In B BUILD'
    }
    submethod TWEAK {
        say 'In B TWEAK'
    }
}
class C is B {
    has $.a;
    submethod BUILD {
        say 'In C BUILD'
    }
    submethod TWEAK {
        say 'In C TWEAK'
    }
}

C.new;
# OUTPUT: 
#In A BUILD
#In A TWEAK
#In B BUILD
#In B TWEAK
#In C BUILD
#In C TWEAK
```

doc/Language/typesystem.pod6
finanalyst: I did change the TWEAK text to add what AlexDaniels wanted. But it has got separated from this patch.

doc/Language/typesystem.pod6
JJ: Fair enough, just clarify it in the text.

doc/Language/typesystem.pod6
JJ: This is a branch, the other is another. If we merge the other, it will conflict with this one. Please close the other one, or just make an atomic commit with the small change that's requested (I don't know where, BTW)

doc/Language/typesystem.pod6
finanalyst: bless is mentioned several times in the documentation with respect to BUILD, see for example the page on Classes. 

I am not enough of an expert to say whether an object can only be instantiated using bless. But I do not see how to get rid of bless in my explanation of TWEAK if it exists in the explanation of BUILD. They are closely linked submethods.

doc/Language/typesystem.pod6
finanalyst: So far as I can recall, the documents have never referred to submethod BUILD as a phaser. I seem to remember playing around with submethod BUILD before Phasers were available. And I certainly remember not being able to do things with BUILD alone that can be done easily with TWEAK.

doc/Language/typesystem.pod6
finanalyst: Yes it is possible to put the default in the BUILD signature, and also in the has declaration. 

The whole point is that BUILD and TWEAK offer different flexibililties. I know I could not do some things with BUILD alone. And there is no documentation for TWEAK.

lizmat:
FWIW, I don't consider BUILD and TWEAK as phasers.  They are part of the internal, semi-official object building API.  Just like FETCH and STORE.

doc/Language/typesystem.pod6
JJ: Let's please stick to this sentence.

> The difference between BUILD and TWEAK is that not all the attributes in an object have been built.

First "the difference... is that not all the attributes" should be preceded by a subject that said *where* they have not been built. But even so, it's not true. All attributes can be built after BUILD. It's not named BUILD for no reason. And TWEAK is named TWEAK precisely for that reason: can be used for after-build tweaks, checks, whatever. The emphasis in in after-**build**: the object has already been built. You can tweak it afterwards, but what TWEAK receives is a perfectly fine object.

doc/Language/typesystem.pod6
JJ: Again, I'm not saying that this is incorrect, just that it's irrelevant. If it's mentioned related to BUILD, that's quite enough, we don't need to go all over again there. TWEAK happens after BUILD, which obviously means it's happened when called from bless. Been there, done that. No need to mention it all over again.

finanalyst: This seems resolved

doc/Language/typesystem.pod6
finanalyst: i'm trying to resolve each matter

tbrowder:
The lack of documentation is still with us.  I agree with @finanalyst and use TWEAK a lot because it's much easier to use when you don't want to burden the user (or author) with tedious details of required inputs when the TWEAK method can deal with it behind the scenes. I may be designing my classes wrongly, but they suit me perfectly.  **TMTOWTDI** thank the Lord and Larry.

And one giant shortcoming of the docs is the paucity of simple, real-world examples for Raku's public API. Instead, we generally lead off with a thesis on the zillion ways  to constrain arguments.

JJ:
I'll reopen #1796 then and note it's not totally been taken care of. The problem with this PR is not that it was not timely or addressed to it, is that we couldn't agree in some aspects of it.

PR #3503, opened 2020-07-05
Update typesystem.pod6

finanalyst:
## The problem
There is almost nothing about TWEAK in the documentation. Even when the TWEAK is mentioned, it is not clear why it is needed, until you get a compile error like the one shown in the example.  
 Further, the documentation does not contain any reference to the signature of TWEAK.  
A small test with `submethod TWEAK(|c) {say $c.raku}` shows that it is passed the same variables as are passed to BUILD.

Added documentation (and fixed a typo in the BUILD description)


JJ:
Sorry, is this a duplicate? You've opened another PR?

PR #3508, opened 2020-07-08
update method start for Proc::Async

interlab:
https://github.com/rakudo/rakudo/blob/master/src/core.c/Proc/Async.pm6#L307


coke:
Thanks!

PR #3510, opened 2020-07-10
Correct associativity for infix:<Z>

codesections:
## The problem

The docs currently list the associativity for the `infix:<Z>` operator as `assoc<chain>`, but it's actual associativity is `assoc<list>`.  `Z` has list associativity because `$a Z $b Z $c == infix:<Z>($a, $b, $c)`.  It does not have chain associativity: `($a Z $b Z $c) ≠ ($a Z $b) and ($b Z $c)`.

## Solution provided

This PR  corrects the associativity for the `infix:<Z>` operator from `assoc<chain>` to `assoc<list>`.  The chart at the top of the page already correctly indicated that `Z` has list associativity and thus does not need to be updated.

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -3020,7 +3020,7 @@ other positions, it's a syntax error.
 
 =head2 infix C«Z»
 
-    sub infix:<Z>(**@lists --> Seq:D) is assoc<chain>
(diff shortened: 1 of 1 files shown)


coke:
Thanks, merged!

Issue #1948, opened 2018-04-21
Change "Perl 6 Pod" in the page title to "Pod 6"


JJ:
## The problem

We're kind of standardizing the name of the documentation DSL in Perl 6 to Pod 6. It's going to be known that way in GitHub when the PR github/linguist#4083 is eventually accepted 

## Suggestions

We should maybe change the title of the page to Pod 6. Besides, an additional problem is that it shows up second on web search, after docs.perl6.org.

AlexDaniel:
Uh, actually, maybe not yet. At least not without a deep discussion.

Some *random* thoughts:

The thing is that we will have [an alias](https://github.com/perl6/6.d-prep/tree/master/TODO#language-extended-naming) eventually. Let's say it'll be “AAA”. With current naming of pod, it'll then be “Perl 6 Pod” and “AAA Pod” depending on context. Yes, two different names for the same thing, but that's what we want (sort of).

So if we rename it to POD6, then I guess it'll stay “POD 6” in both cases, which is nice… But then, what is “6”?

I guess “pod6” will work nicely with “6lang”, but then why not “6pod”? :)

JJ:
But we're using Pod 6 pretty much everywhere else.

JJ:
OK, I'm closing this then. If anyone feels strongly about changing the page name, just open it up again.

PR #3534, opened 2020-07-25
Call it "Pod", not "POD" or "Pod6"

softmoth:
It's not Plain Ol' Documentation in Raku. It's just Pod. From S26:

> Pod is an evolution of Perl 5's Plain Ol' Documentation (POD) markup.

## The problem

_POD_ and _Pod6_ are Perlisms that don't really fit after the name change to Raku.

## Solution provided

Following [S26](https://design.raku.org/S26.html), call Raku's documentation _Pod_, and leave _POD_ to Perl. Remove as many references to `Pod6` and `POD` as possible.

**TODO**
- Rename all docs files to `*.rakudoc` instead of `*.pod6`
- Rename `:lang<pod6>` in code blocks


JJ:
OK, several problems here.
1. We really don't like PRs with more than 4-5 files, every one with a few changes. 22 files takes a long time to review, and the whole PR might be stuck because of a single comma.
2. See #1948. That change was already made, in the other direction. In general, it's better to achieve a consensus on a big change before committing to it. It was called pod6, which is simply an abbreviation for Perl 6 Pod. 
3. If we are going to make some change now, it will be to rakudoc, probably, not pod. 
4. Changing all file names to rakudoc is in the roadmap, but not something that we will do inmediately.

Thanks anyway for the work, but I would really prefer that you keep it small and to any other grammar fixes or any other kind that you might have bump into when doing it.

softmoth:
It's a proposal. I didn't spend much time on it. If people hate it, it can be canned. I think discussing things can be more efficient when there's already a diff to talk about.

Using "Rakudoc" instead of "Pod" is OK by me. Using "Pod 6" is not, I think, appropriate.

JJ:
> It's a proposal. I didn't spend much time on it. If people hate it, it can be canned. I think discussing things can be more efficient when there's already a diff to talk about.
> 

Yep, but that change goes in a particular direction, so instead of arriving to a consensus of what to do, it's a binary decision: take it or leave it.

> Using "Rakudoc" instead of "Pod" is OK by me. Using "Pod 6" is not, I think, appropriate.

Nor we are using it. It's pod6, short for Perl 6 Pod. Rakudoc, in the medium term, is probably the way to go.

tbrowder:
I'm working on a couple of modules that involve using pod as a markup language to produce PDF output. What should I call **pod**?

The first module I have named **Pod::To::PDF** and a follow-on to be called **DocWriter**, so I do need some guidance, *por favor, senor*.

JJ:
Pod::to is the admitted way of calling those modules, so no problem there. In the documentation, we generally call it pod6, the same as the extensions.

tbrowder:
Thank. BTW, I just read fully @finanalyst's 20th Advent article and saw he wrote of what I was thinking but so much better and more thoroughly. The literate programming aspect of pod is what I had a eureka moment over a while back, and his references were greatly appreciated.

softmoth:
I think this issue is more confusing than helpful, so I'll close it. Thanks for the discussion and review.

PR #3535, opened 2020-07-25
Tweak names of two variables

usev6:
## The problem
I was confused with the original names. Also, the text after the code example talks about strings being interpreted in numeric context, so ```$not-a-string``` looks off.

## Solution provided
I guess the new names match the original author's intention.

### DIFF
--- a/doc/Language/contexts.pod6
+++ b/doc/Language/contexts.pod6
@@ -66,9 +66,9 @@ I<Number context> is called whenever we need to apply a numerical operation on a
 variable.
 
 =begin code
-my $not-a-string = "1                 ";
-my $neither-a-string = "3                        ";
-say $not-a-string + $neither-a-string; # OUTPUT: «4␤»
+my $stringone = "1                 ";
+my $stringthree = "3                        ";
+say $stringone + $stringthree; # OUTPUT: «4␤»
 =end code
 
 In the code above, strings will be interpreted in numeric context as long as


pmichaud:
I suggest removing the negatives altogether -- use something like ```$stringone```    and ```$stringthree``` to make it clear that the variables contain strings.

I somewhat agree with OP that naming a variable ```$not-a-string``` when it clearly _does_ contain a string value is a bit confusing.

Pm

JJ:
> I suggest removing the negatives altogether -- use something like `$stringone` and `$stringthree` to make it clear that the variables contain strings.
> 
> I somewhat agree with OP that naming a variable `$not-a-string` when it clearly _does_ contain a string value is a bit confusing.
> 

Well, I was aiming for "surrealist". But it it gets in the way, a neutral name might be best.

usev6:
I'd be totally fine with a neutral name, too. (It's not a big problem anyway, I just stumbled over it and my brain went 'WAT'.)

JJ:
> I'd be totally fine with a neutral name, too. (It's not a big problem anyway, I just stumbled over it and my brain went 'WAT'.)

Can you please change them to that? We'll accept the PR then.

usev6:
I've force pushed with that names suggested by pmichaud++. (I hope that's okay.)

For the record: This was the original patch (what was discussed above).
```
diff --git a/doc/Language/contexts.pod6 b/doc/Language/contexts.pod6
index eb71aaa..32702ce 100644
--- a/doc/Language/contexts.pod6
+++ b/doc/Language/contexts.pod6
@@ -66,9 +66,9 @@ I<Number context> is called whenever we need to apply a numerical operation on a
 variable.
 
 =begin code
-my $not-a-string = "1                 ";
-my $neither-a-string = "3                        ";
-say $not-a-string + $neither-a-string; # OUTPUT: «4␤»
+my $not-a-number = "1                 ";
+my $neither-a-number = "3                        ";
+say $not-a-number + $neither-a-number; # OUTPUT: «4␤»
 =end code
 
 In the code above, strings will be interpreted in numeric context as long as
```

Issue #3517, opened 2020-07-18
Boolean should be capitalized


JJ:
## Problem or new feature

Since its origin is a proper name, Boole, [it should be capitalized](https://english.stackexchange.com/questions/4481/should-the-word-boolean-be-capitalized).

## Suggestions

There are more `boolean` than `Boolean` in the documentation.

If applicable, suggest something towards the solution. For starters, eliminate it from the dictionary, and looking forward add it to a test.

coke:
Additionally, when referring to the Raku type, it should probably be changed to a link to Bool, the enum.

paultcochrane:
See #3542.  There were some instances in the docs where a "Boolean value" was mentioned; I don't think that this was the Boolean type, hence I made no linkes to the Bool enum.  The PR is submitted in the hope that it is useful, so please let me know if there is anything I could improve.

uzluisf:
Should this issue be closed since it looks like it was addressed already?

JJ:
@uzluisf maybe add a test also.

coke:
Should be possible to add this requirement to xt/word-variants.t

coke:
Merged @stoned's PR, ran the test, fixed the one item remaining.

PR #3542, opened 2020-07-30
Capitalise the word "Boolean"

paultcochrane:
... since it comes from a person's name.  This should address issue #3517; although I'm not 100% sure if I've satisfied the requirements of @coke's comment about linking to the Bool enum, as I didn't find (what I thought) were references to the Raku Boolean type, however I might have misread the docs in those places.

### DIFF
--- a/doc/Language/traps.pod6
+++ b/doc/Language/traps.pod6
@@ -797,7 +797,7 @@ for other, more common, things:
 
 The C<^>, C<|>, and C<&> are I<not> bitwise operators, they create
 L<Junctions|/type/Junction>. The corresponding bitwise operators in Raku are:
-C<+^>, C<+|>, C<+&> for integers and C<?^>, C<?|>, C<?&> for booleans.
+C<+^>, C<+|>, C<+&> for integers and C<?^>, C<?|>, C<?&> for Booleans.
 
 =head2 Exclusive sequence operator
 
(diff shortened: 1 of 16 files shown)


doc/Language/traps.pod6
```diff
@@ -797,7 +797,7 @@ for other, more common, things:
 
 The C<^>, C<|>, and C<&> are I<not> bitwise operators, they create
 L<Junctions|/type/Junction>. The corresponding bitwise operators in Raku are:
-C<+^>, C<+|>, C<+&> for integers and C<?^>, C<?|>, C<?&> for booleans.
+C<+^>, C<+|>, C<+&> for integers and C<?^>, C<?|>, C<?&> for Booleans.
```
JJ, line 801: This is one case where `Bool` would suit better than Booleans, since you're referring to specific types. Anyway, that was not in the original, so capitalization is OK (but a fix further down the line would be better)

doc/Type/Pair.pod6
```diff
@@ -152,8 +152,8 @@ be different:
     say :10z ~~ :42a; # OUTPUT: «False␤»
 
 If C<$topic> is any other value, the invocant C<Pair>'s key is treated as a method name.
-This method is called on C<$topic>, the L«boolean|/routine/Bool» result of which is compared
-against the invocant C<Pair>'s L«boolean|/routine/Bool» value. For example, primality can
+This method is called on C<$topic>, the L«Boolean|/routine/Bool» result of which is compared
+against the invocant C<Pair>'s L«Boolean|/routine/Bool» value. For example, primality can
 be tested using smartmatch:
```
JJ, line 157: Couple of cases where changing it to Bool would also make sense.

PR #3544, opened 2020-07-30
Add `update-html` make target

paultcochrane:
This target name is easier to remember than the `documentable update` command as it uses the make infrastructure which is already present.  It also uses the same general interface to the various build processes provided by make (i.e. one can use `make <target>` as opposed to having to remember what command is necessary to perform a particular operation).  Also, this wraps the `update` process behind an interface; the details of which can be changed in the future without changing the interface.

I've also updated the README to mention the new target name.

Also, and somewhat unrelated to the above changes, I've simplified the docs in the README for HTML generation as well as have removed superfluous vertical whitespace.  These changes have been put into separate commits so that they can be cherry picked as desired.

This PR is submitted in the hope that it is useful.  If any changes are required I'll happily update the PR and resubmit as necessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,9 @@ help:
 html: for-documentable
 	documentable start -a -v --highlight
 
+update-html:
+	documentable update
+
 init-highlights highlights/package-lock.json:
 	ATOMDIR="./highlights/atom-language-perl6";  \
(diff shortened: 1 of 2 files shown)


paultcochrane:
I don't understand the Travis error here: I didn't change any of that code.  I'm going to see if I can trigger a new build.

PR #3554, opened 2020-08-11
Add short description for a NativeCall helper

tbrowder:
## The problem

Because of its name, the subject module is hard to find. Additionally, with its link shown early on in the very lengthy NativeCall description, it could encourage more authors to contribute Raku wrappers for their favorite libraries since they could get help starting with framework code.

## Solution provided

Short descrription and a link to the parent repo.


coke:
I'd remove the comment about "unusual name".

tbrowder:
> I'd remove the comment about "unusual name".

Good point, so done.

PR #3558, opened 2020-08-13
More on { } and < > operators for Mix and MixHash

stoned:
- mention <> postcircumfix
- add ref to operators' description


stoned:
> I don't really like cut and paste, even if they are describing the self same thing, but anyway it's a good start and it can be modified later. Thanks!

I'm still pondering how the setbagmix pod can be modified to factor more of the Set, SetHash, Bag, BagHash, Mix and MixHash pods :-)

PR #3565, opened 2020-08-19
Current and explicit Baggy.default signature

stoned:


JJ (approved):
Surprisingly, this is true. Not clear why you'd want that, anyhow, but good catch. Thanks.

PR #3568, opened 2020-08-19
Uniformize examples' output

stoned:
And while here, if we would like an example to produce an output,
we might as well call `say` :)


JJ (approved):
We should maybe have author tests for these. But thanks again, much appreciated.

PR #3570, opened 2020-08-19
Rework List.join examples and description

stoned:
- Remove duplicate note about the optional separator for the method form
- It's the subroutine form that is slurpy and flattening
- Group related examples
- Correct a typo


JJ (approved):
Thanks again!

Issue #3529, opened 2020-07-21
Improvements to util/list-missing-methods.p6 [tracking issue]


codesections:
I very much like the idea behind the `list-missing-methods.p6` program – Raku provides extremely rich introspection abilities, and we should be able to automatically detect any undocumented areas.  Additionally, the same program should be able to list methods we document that do *not* currently exist (something that occurred to me after working on #3528)

However, while attempting to add that functionality to `list-missing-methods.p6`, I realized that it currently has a few other issues (e.g., it was confusing `Pod::Block` with `Block`).  I'm currently working on a revision that both adds that functionality and addresses several minor issues; I'm opening this as a tracking issue for that work.

JJ:
Thanks!

coke:
missing methods relies on bin/p6doc which is going away (See #3483) - if still needed, will need to rely on https://github.com/raku/rakudoc

codesections:
Unless I'm missing something, that's true of [`missing-methods.p6`](https://github.com/Raku/doc/blob/master/util/missing-methods.p6) but not [`list-missing-methods.p6`](https://github.com/Raku/doc/blob/master/util/list-missing-methods.p6), which is the script I'm working on revising.

(I'm not at all clear on why there are two so similarly named utilities or if one of them should be deleted in favor of the other.  I started with `list-missing-methods` because it's the one mentioned in [CONTRIBUTING](https://github.com/Raku/doc/blob/9ba24606b02ed7662450c5c60958f9fa17af3bc7/CONTRIBUTING.md#testing-method-completeness))

coke:
My bad - if we can remove the other one, that's great.

codesections:
Closed by #3571

PR #3571, opened 2020-08-19
Overhaul the list-missing-methods.p6 script

codesections:
This PR grew significantly as I was working on it.  A while ago, I opened #3529, and said that I would add the ability to detect methods that exist in the doc but that have been removed from Raku and "address several minor issues" with the script.  Somehow, that's not quite what ended up happening.

Now, 48 commits later, I have a significantly overhauled version of the script.  This version does significantly more to screen out false positives – specifically, there are many methods that the `.^methods(:local)` method (and the old script) consider to be local to a Type but that are actually implemented on a closely related Role.  The previous script listed these as methods missing documentation, even though they are properly documented on the Role.  Correcting this should make the script significantly more helpful for targeting our documentation efforts.

The new script also adds two new categories of information for each Type: over-documented methods that no longer exist in Raku (my original goal!) and methods without enough introspection to be able to tell, or where the introspection results appear to be inconsistent, which caused the false-positives issue described above.

I've also added the ability to print summary information for all scanned types, which should be useful both in keeping track of our overall progress and in targeting our efforts.  When run on the full `doc/Type` directory, this summary info shows that the script was able to check 3,522 methods.  Of these, 8% were over-documented, 57% were under-documented and 35% had no detected issues.

 That 35% number is perhaps artificially low, however, as additional summary info shows.  Of the under-documented methods, over a third come from methods that are missing from 20+ types.  These are not false positives – they really are implemented on the Type – but they may not be in need of documentation.  Examples include `gist`, `message`, `raku`, etc.  The previous version of the script included support for an `ignored-methods.txt` file that could ignore methods from specific types; I've expanded that feature to also allow ignoring methods from all types at once, and we may want to add many of these don't-need-to-be-documented methods to that file.  (I added `BUILD_ALL`, which was previously "missing" from 184 types but is clearly an implementation detail, but have mostly refrained from adding ignored methods because I didn't want to make that call without discussing the issue – although the script _does_ now allow specifying a location for the ignored-methods file, which would let us use more than one.)

I've also added several new CLI options for filtering or otherwise limiting output; please consult the `--help` message for detailed info.

The cost of all of this added functionality is that the script is now much larger – nearly 700 lines.  Many of those are comments, and much of the rest is devoted to the unavoidably verbose output/formatting; the script's core logic is only ~200 lines.  Nevertheless, I would normally split a program of this size up into multiple modules, and strongly considered doing so here (including for performance reasons).  However, I've kept this as a single file, at least for now.  That's how it was, and I didn't want to add sub-directories to the `docs/util` directory without a clear need.

Finally, I want to mention that this is the first non-trivial Raku program I've written and that I greatly appreciate all the help I got from kind people on the #raku IRC channel.  I ended up diving pretty deep into Raku's introspection abilities – as @lizmat put it when she was helping me with a particularly hairy bit, I got down to where the turtles are showing.  And that was quite a way to learn the language.  So thanks, all!  And I'm definitely open to any feedback others might have on this script.

### DIFF
--- a/util/ignored-methods.txt
+++ b/util/ignored-methods.txt
@@ -1,9 +1,15 @@
-# This file is loaded by `list-missing-methods.p6` (unless a different file is # specified a CLI
-# option).  It list methods that should *not* be listed as 'missing' from the docs.  It is read as a
-# standard Raku hash, so you can use any valid Raku syntax.  When adding items to this list, please
-# do your best to add a line-comment explaining why that method should not be counted as 'missing'.
-%( Any =>     ( 'lazy-if'),
-   Hash =>    ( 'categorize-list',
+# This file is loaded by `list-missing-methods.p6` (unless a different file is specified
+# as a CLI option).  It list methods that should *not* be listed as 'undocumented'.  It is
+# read as a standard Raku hash, so you can use any valid Raku syntax.  Each key should be
+# either 1) a Type name, or the special key `GLOBAL`; each value should be a list of
+# methods that don't need to be checked for that Type.  The list for the `GLOBAL` key
+# applies to all Types.  When adding items to this list, please do your best to add a
+# line-comment explaining why that method should not be checked.
+
+%( GLOBAL  => ( 'BUILDALL',            # Implementation detail listed as local in 100+ types
+              ),
+   Any     => ( 'lazy-if', ),
+   Hash    => ( 'categorize-list',
                 'dynamic',
                 'name', 
                 'BIND-KEY',
@@ -11,13 +17,13 @@
                 'clone', ),
    Dateish => ( 'IO',
                 'earlier'),
-   Str =>     ( 'codes',
+   Str     => ( 'codes',
                 'simplematch',
                 'BUILD',
                 'Int',
                 'Num',
                 'word-by-word', ),
-   Array =>   ( 'dynamic',
+   Array   => ( 'dynamic',
                 'STORE',
                 'default',
                 'name',
--- a/util/ignored-methods.txt
+++ b/util/ignored-methods.txt
@@ -28,5 +28,6 @@
                 'default',
                 'name',
                 'from-iterator',
-                'reification-target',)
+                'reification-target',),
+   Regex   => ( 'clone', 'gist', 'raku' )
 )
--- a/util/ignored-methods.txt
+++ b/util/ignored-methods.txt
@@ -1,6 +1,26 @@
-# This file can have comments.
-Any: lazy-if
-Hash: categorize-list dynamic name BIND-KEY classify-list clone
-Dateish: IO earlier
-Str: codes simplematch BUILD Int Num word-by-word
-Array: dynamic STORE default name from-iterator reification-target
+# This file is loaded by `list-missing-methods.p6` (unless a different file is # specified a CLI
+# option).  It list methods that should *not* be listed as 'missing' from the docs.  It is read as a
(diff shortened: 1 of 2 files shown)


JJ (approved):
On one hand, this is just amazing and I love it. On the other hand, as it happens with every script, we have no tests to check that it's working exactly as it should (because we don't know what should). I accept it anyway because push comes to shove we can always retrieve the older version, and it's really a user script. Will leave it open for some time, anyway, just in case anyone wants to pitch in. And again, much appreciated.

codesections:
> On one hand, this is just amazing and I love it.

Thanks, I'm glad you like it!

> On the other hand, as it happens with every script, we have no tests to check that it's working

Yeah, I also had that thought.  I attempted to write the code to be _testable_, even though there aren't currently any tests – there are lots of small, pure functions, which should be pretty easy to write tests for.  After this is merged, I'd be happy to work on another PR adding some tests, though I do think that would mean splinting the code up into a module rather than staying as a single script.  Does that seem worthwhile to you?

JJ:
> > On one hand, this is just amazing and I love it.
> 
> Thanks, I'm glad you like it!
> 
> > On the other hand, as it happens with every script, we have no tests to check that it's working
> 
> Yeah, I also had that thought. I attempted to write the code to be _testable_, even though there aren't currently any tests – there are lots of small, pure functions, which should be pretty easy to write tests for. After this is merged, I'd be happy to work on another PR adding some tests, though I do think that would mean splinting the code up into a module rather than staying as a single script. Does that seem worthwhile to you?

Of course. And on that note, it would probably also be worth the while to spin it off, as a separate module, because it probably deserves the full attention of a project maintainer. In order to do that, please try to keep the history of the file and all related ones, because credit where credit is due.
Also, welcome to the team! Your commit bit is served.

codesections:
> And on that note, it would probably also be worth the while to spin it off, as a separate module, because it probably deserves the full attention of a project maintainer. In order to do that, please try to keep the history of the file and all related ones, because credit where credit is due.

Are you saying that it should be spun off into a separate _repo_ ?  I was thinking that it could just be a (Raku) module that lived in a subdirectory inside the `doc/util/` directory (people would still get it when they `git clone` the docs).  But if you think it should leave the docs repo entirely, I can do that instead.  Just let me know :)

> Also, welcome to the team! Your commit bit is served.

Thanks very much – I'm glad to be here :100:  Based on past projects I've contributed to, I assume that the etiquette is generally not to merge my own PRs unless they are truly trivial.  Am I right about that?

JJ:
El jue., 20 ago. 2020 a las 17:34, Daniel Sockwell (<

> And on that note, it would probably also be worth the while to spin it
> off, as a separate module, because it probably deserves the full attention
> of a project maintainer. In order to do that, please try to keep the
> history of the file and all related ones, because credit where credit is
> due.
>
> Are you saying that it should be spun off into a separate *repo* ? I was
> thinking that it could just be a (Raku)
>

Yes.

> module that lived in a subdirectory inside the doc/util/ directory
> (people would still get it when they git clone the docs). But if you
> think it should leave the docs repo entirely, I can do that instead. Just
> let me know :)
>

I'd appreciate if you did that.

> Also, welcome to the team! Your commit bit is served.
>
> Thanks very much – I'm glad to be here 💯 Based on past projects I've
> contributed to, I assume that the etiquette is generally not to merge my
> own PRs unless they are truly trivial. Am I right about that?
>

Whatever you feel more comfortable with.

Issue #3569, opened 2020-08-19
Document the .add/.remove BagHash methods and the .set/.unset SetHash methods 


codesections:
The mutable types `BagHash` and `SetHash` both have methods to add and remove items from the collection, but none of them are currently documented.

I will submit a PR documenting these methods.  I don't believe I have sufficient permission on this repo to self-assign this issue, so I would appreciate it if someone could assign it to me.

Issue #3209, opened 2020-02-09
How to document protos ({*


JJ:
## The problem

Some roles and classes have method stubs, interface-only definitions that use the yadda, yadda, yadda block or the WhateverCode. They are totally different, one of them forces the consumer to implement an interface, the other offers a default implementation, maybe with code, as a base for a set of multis.

We have no standard way of documenting these. In many cases, as in [`Scheduler`](https://docs.raku.org/type/Scheduler) (`cue`) or in [`QuantHash`]8https://docs.raku.org/type/QuantHash#method_Mixy) (see #3208: both, Baggy, Mixy and Setty are protos  and hash/Hash are forward declaration) they are documented as an actual implementation in subclasses instead of as such protos or

JJ:
There's not been a lot of discussion... So I'll go ahead with the suggestion for the time being.

Issue #392, opened 2016-02-11
Sort Table of Contents


zoffixznet:
As reported in https://github.com/perl6/user-experience/issues/10, the method [names in the table of contents](https://doc.perl6.org/type/Backtrace) are in no coherent order, which makes it hard to spot the one you want to read about. I'd suggest them to be sorted alphabetically.

![untitled](https://cloud.githubusercontent.com/assets/5747918/12966230/54e42baa-d02a-11e5-9b94-3b8c718d2b92.png)

AlexDaniel:
Are you saying that ｢combinations｣ and ｢permutations｣ should be far away from each other? I don't think so. In fact, I always loved how it was more or less sorted in “what I need is usually on top” order.

tbrowder:
Then maybe there ought to be clear category subparts under which such things could be grouped.  Closing this without more discussion isn't helpful IMHO.

Skarsnik:
Yep. For me an example of a good documentation is Qt. it's alphabetical
order but methods that do thing related thing are mentioned in each other.

2016-02-11 13:49 GMT+01:00 Tom Browder <notifications@github.com>:

> Then maybe there ought to be clear category subparts under which such
> things could be grouped. Closing this without more discussion isn't helpful
> IMHO.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/perl6/doc/issues/392#issuecomment-182848899>.
>



-- 
Sylvain "Skarsnik" Colinet

Victory was near but the power of the ring couldn't be undone

zoffixznet:
@tbrowder would you have any examples of how such a category layout look like?

In the screenshot above, I see `keys`, `values`, and `kv` as one group, `pairs` is somewhat similar to `kv`, but doesn't really fit into that group entirely. To me `classify` more tightly fits into a group with `sort`, but you can equally argue the `grep` group is appropriate too.

There is some intelligence with the current grouping; trying to break it up into more refined subcategories would result in a lot of bikeshed IMO, especially since many method are inherited from other classes.

Also, does anyone know how the current sorting is made up? It is all just hand-ordered?

tbrowder:
On Thursday, February 11, 2016, Skarsnik <notifications@github.com> wrote:

> Yep. For me an example of a good documentation is Qt. it's alphabetical
> order but methods that do thing related thing are mentioned in each other.


Sounds like a good solution to me (maybe a bit of pain for doc writers, but
only for a brief while).

-Tom

>

[...]

PR #3572, opened 2020-08-20
Add undocumented SetHash methods

codesections:
This PR adds the following 16 methods to `SetHash`: `clone`, `grab`, `grabpairs`, `iterator`, `kv`, `set`, `unset`, `values`, `Baggy`, `Mixy`, `Set`, `SetHash`, `Setty`, `AT-KEY`, `DELETE-KEY`, and `STORE`.  Adding these methods means that `SetHash` currently has no undocumented methods.  The PR also adds a bit of text to the overview describing how the contents of a `SetHash` can be altered with the `set` and `unset` methods.

I wasn't entirely sure whether to add all methods.  Many methods are specialized implementations of methods already provided/stubbed higher in the dependency graph.  Please let me know if there are any that you think should be left undocumented, I will delete them.   However, I decided to document all methods for reasons basically analogous to those brought up in #3209: these methods really are implemented in `SetHash` (i.e., they both appear in the source code and live in its local methods table), and anyone subclassing/augmenting/etc `SetHash` may want to know that.

Given the number of added methods, however, I decided to depart from strict case-intensive ordering (which seemed to be the tentative consensus emerging from #392).  Instead, I put the three ALL-CAPS methods that are called automatically via `{ }` at the end; above those, I put the IntialCapital coercion methods, and then put all the other methods above that (the methods are alphabetical within their respective groups).  I believe this keeps methods clearly organized while also helping keep less user-facing methods away from the top of the list.  But I'm open to other thoughts on this as well.

This PR goes halfway to resolving #3569.

### DIFF
--- a/doc/Type/SetHash.pod6
+++ b/doc/Type/SetHash.pod6
@@ -231,7 +231,7 @@ Defined as:
 
     multi method kv(SetHash:D: --> Seq)
 
-Returns a C<Seq> of the C<SetHash>'s keys alternating with C<True>.
+Returns a C<Seq> of the keys in the C<SetHash> key alternating with C<True>.
 
     SetHash.new(1, 2, 3).kv.say       # OUTPUT: «(3 True 1 True 2 True)␤»
 
@@ -313,11 +313,9 @@ Defined as:
 
     multi method AT-KEY(SetHash:D: \k --> Bool)
 
-Returns a C<Bool> (when called with a single key) or a C<List> of
-C<Bool>s (when called with a C<List> of keys) that indicate whether
-the provided key or keys are present in the C<Set>.  This method is
-automatically called by indexing with L<C«postcircumfix { }»
-|/routine/{ }#(Operators)_postcircumfix_{_}>.
+Returns a C<Bool> indicating whether the provided key is present in
+the C<SetHash>.  This method is automatically called by indexing with
+L<C«postcircumfix { }» |/routine/{ }#(Operators)_postcircumfix_{_}>.
 
 =head2 method DELETE-KEY
 
@@ -325,12 +323,10 @@ Defined as:
 
     multi method DELETE-KEY(SetHash:D: \k --> Bool)
 
-Deletes a key or list of keys from the C<SetHash>.  Returns a C<Bool>
-(when called with a single key) or a C<List> of C<Bool>s (when called
-with a C<List> of keys) that indicate whether the provided key or keys
-were present in the C<Set>.  This method is what L<C«postcircumfix { }»
-|/routine/{ }#(Operators)_postcircumfix_{_}> calls when invoked
-like C<$set-hash<k>:delete>.
+Deletes a key and returns a C<Bool> indicating whether the key was
+present in the C<SetHash>.  This method is what L<C«postcircumfix { }»
+|/routine/{ }#(Operators)_postcircumfix_{_}> calls when invoked like
+C<$set-hash<k>:delete>.
 
 =head2 method STORE
 
--- a/doc/Type/SetHash.pod6
+++ b/doc/Type/SetHash.pod6
@@ -20,22 +20,53 @@ say $fruits.elems;      # OUTPUT: «3␤»
 say $fruits.keys.sort;  # OUTPUT: «apple orange peach␤»
 =end code
 
-C<SetHash>es can be treated as object hashes using the C<{ }> postcircumfix
-operator, which returns the value C<True> for keys that are elements of the set,
-and C<False> for keys that aren't. Assigning a value that boolifies to C<True>
-or C<False>, respectively, can be used to add or remove a set element:
+Just like C<Set>s, C<SetHash>es can be treated as object hashes using the C<{ }>
+postcircumfix operator, which returns the value C<True> for keys that are
+elements of the set, and C<False> for keys that aren't.
 
 =begin code
 my $fruits = <peach apple orange apple apple>.SetHash;
 
 say $fruits<apple>;     # OUTPUT: «True␤»
 say $fruits<kiwi>;      # OUTPUT: «False␤»
+=end code
+
+Unlike C<Set>s, C<SetHash>es are mutable.  You can add an item or list of items
+to the C<SetHash> with the C<set> method and can remove an item or list of items
+with the C<unset> method:
+
+=begin code
+my $fruits = <peach>.SetHash;
+$fruits.set('apple');
+say $fruits;            # OUTPUT: «SetHash(apple peach)␤»
+
+$fruits.unset('peach');
+say $fruits;            # OUTPUT: «SetHash(apple)␤»
+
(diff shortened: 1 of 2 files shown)


lizmat:
Feels to me that `clone`, `iterator`, `AT-KEY`, `DELETE-KEY` and `STORE` would not need to be documented here.

codesections:
> Feels to me that `clone`, `iterator`, `AT-KEY`, `DELETE-KEY` and `STORE` would not need to be documented here.

Ok, I'm definitely happy to take them out.  What's the principle that lets you pull out those five for exclusion?  They all do what a user would expect them to do, so I agree that documenting them doesn't add a whole lot.  But the same could be said of `Set`, so I'm not quite sure where we're drawing the line.  I'm asking  I plan to add missing methods to more types over the coming days/weeks, and I'd like to understand which methods I should be including.

lizmat:
Well, if `AT-KEY` and `DELETE-KEY` are document, so should at least `EXISTS-KEY` be as well as potentially `BIND-KEY`.

I'm not sure where the boundary is really.

lizmat:
Having test subs return their status, has the advantage that you can easily add a line before them with diagnostic info.  So suppose you have:

    ok Foo ~~ Bar;

you can do something like:

    diag "WTF is happening here" unless
    ok Foo ~~ Bar;

to get extra, non-standard diagnostics.

codesections:
> Well, if `AT-KEY` and `DELETE-KEY` are documented, so should at least `EXISTS-KEY` be as well as potentially `BIND-KEY`.
> I'm not sure where the boundary is really.

Right now (in the PR as it exists), the boundary is "document anything that exists _in the type_ being documented" (in the current Rakudo implementation).  `EXISTS-KEY` is not implemented in `SetHash`: running `SetHash.^lookup('EXISTS-KEY').candidates».package` shows that `SetHash` has `EXIST-KEY` methods from `Any` and `Setty` in its MRO, and grepping over the Rakudo source code confirms that it's not in the `SetHash.pm6` file.  Similarly, `SetHash` gets `BIND-KEY` from `Any`.  In contrast, [`STORE`](https://github.com/rakudo/rakudo/blob/master/src/core.c/SetHash.pm6#L206-L231), [`AT-KEY`](https://github.com/rakudo/rakudo/blob/master/src/core.c/SetHash.pm6#L233-L258), and [`DELETE-KEY`](https://github.com/rakudo/rakudo/blob/master/src/core.c/SetHash.pm6#L260-L271) are all implemented in `SetHash` itself.

I'm not _sure_ that "exists in the Rakudo source" is the right place to draw the line – it's a bit implementation dependent, for one thing and, at least arguably, the choice to locally implement `AT-KEY` but not `EXISTS-KEY` could be viewed as an implementation detail.

So I'm open to drawing the line somewhere else.  But I'd like to have a clear way to draw the line, and right now the only idea I've come up with is "is it in the source/local to the type".  If we can come up with a better line, though, I'll be all for it – especially if it means we have fewer methods that need documenting!

lizmat:
`SetHash.EXISTS-KEY` lives in the `Setty` role, **not** in `Any`.

My opinion wrt documentation of classes, is to document what is user-facing.  In that sense, all of the `UPPER-CASE` methods are implementation details.  But where to draw the line, I'm not sure.  I mean, do we document the `Bool` method on a `SetHash` ?  Or the `Int` method?  Or the `defined` method?  Etc. Etc.  In a world without inheritance or mixins, this would all be so much simpler :-)

codesections:
> `SetHash.EXISTS-KEY` lives in the `Setty` role, **not** in `Any`.

I agree.  That's why I said that "`SetHash` has `EXIST-KEY` methods from `Any` and `Setty` in its MRO" – and it's certainly correct that the one from `Setty` is the one that's called with a basic call to `SetHash.EXIST-KEY`.

> My opinion wrt documentation of classes, is to document what is user-facing.

I also agree with that – but I guess the real question is what do we mean by "user"?  I guess my thought was that users of Raku are, by definition, Raku programmers and might be inclined to do things implement complex type inheritance and/or mess with meta objects.  But maybe documenting `UPPER-CASE` methods is carrying that too far.

So, how about this as a general rule: for each Type, we document a method if it is: 1) public, 2) defined in that class (rather than inherited), 3) not an `UPPER-CASE` method, and 4) not flagged with the `is implementation detail` trait.  That would mean that we would **not** document `Bool`, `Int`, or `defined` in `SetHash`, since none of those methods is defined in the `SetHash` class (they're inherited from `Mu` or `QuantHash`).

Does that four-part rule seem like a good place to draw the line to you/anyone else following this conversation?

JJ:
El jue., 20 ago. 2020 a las 23:41, Daniel Sockwell (<

> SetHash.EXISTS-KEY lives in the Setty role, *not* in Any.
>
> I agree. That's why I said that "SetHash has EXIST-KEY methods from Any
> and Setty in its MRO" – and it's certainly correct that the one from Setty
> is the one that's called with a basic call to SetHash.EXIST-KEY.
>
> My opinion wrt documentation of classes, is to document what is
> user-facing.
>
> I also agree with that – but I guess the real question is what do we mean
> by "user"? I guess my thought was that users of Raku are, by definition,
> Raku programmers and might be inclined to do things implement complex type
> inheritance and/or mess with meta objects. But maybe documenting
> UPPER-CASE methods is carrying that too far.
>
> So, how about this as a general rule: for each Type, we document a method
> if it is: 1) public, 2) defined in that class (rather than inherited), 3)
> not an UPPER-CASE method, and 4) not flagged with the is implementation
> detail trait. That would mean that we would *not* document Bool, Int, or
> defined in SetHash, since none of those methods is defined in the SetHash
> class (they're inherited from Mu or QuantHash).
>
> Does that four-part rule seem like a good place to draw the line to
> you/anyone else following this conversation?
>

That might or might not seem a good place, but *this PR* is not. While we
can agree in some of them, sometimes lines are not so clear-cut.
Definitely, "is implementation-detail" should not be documented. But about
the rest, YMMV. In general, if they are public they are game.
So I'd suggest we stick to this specific PR for the time being, and maybe
draft an issue to modify the CONTRIBUTING document with these details.

doc/Type/SetHash.pod6
```diff
@@ -133,6 +164,188 @@ say $a ⊖ $b;  # OUTPUT: «SetHash(1 3 4)␤»
...
+
+=head2 method AT-KEY
+
+Defined as:
+
+    multi method AT-KEY(SetHash:D: \k --> Bool)
+
+Returns a C<Bool> (when called with a single key) or a C<List> of
+C<Bool>s (when called with a C<List> of keys) that indicate whether
+the provided key or keys are present in the C<Set>.  This method is
```
JJ, line 318: If it returns a list, how come it be defined with a single Bool as return type?

JJ (changes requested):
Please chck out the comments.

PR #3584, opened 2020-08-25
Split spell

### LINKED ISSUE #2639: split spellcheck for code
## The problem

Spellcheck is run over all the pod, including code samples - as a way to keep the "words" needed for code separate from those for the text, there are two .pws files.

However, all the word fragments and odd combinations needed for the code to pass spellcheck are then allowed in the text.

We do still want to spell check the code because there is often English text present in the code blocks, either in output, variable names, etc.

## Suggestions

Split the spell check so that that both lexicons are used when checking code, but only the (mostly) english words are used when testing the text.

coke:
This resolves #2639 - the lexicon for code is now actually used only in code blocks, while the words lexicon is used for everything. This prevents non-words needed in code blocks from leaking out into the rest of the text accidentally (though if it does intentionally, that's fine.)

Reworked the test to take advantage of native concurrency, with a lot of help from @ugexe. It now uses race, and completes faster than it did before.

It runs 2 tests for each file, one to check the code blocks, one to check everything else.

Added a new test which also verifies that the lexicon files themselves meet certain standards.


coke:
the travis failure is in a file I didn't touch (and which passes in this branch locally)

JJ:
> the travis failure is in a file I didn't touch (and which passes in this branch locally)

It's the flapper we've been carrying over since I started around here.

JJ (approved):
Great work. Thanks.

xt/aspell.t
```diff
@@ -33,40 +36,81 @@ if $proc.exitcode {
...
-    for %output{$file}.lines -> $line {
-        FIRST next; # dump first line
-        next if $line eq '';
-        diag $line;
-        $count++;
+    # Process the text so that aspell understands it.
+    # Every line starts with a ^
+    # turn \n and \t into spaces to avoid \nFoo being read as "nFoo" by aspell
+    for $input-file.IO.slurp.lines -> $line {
+        my Bool $is-code = $line.starts-with(' ');
```
Altai-man, line 65: What about
```
=for code
say 42;
=end
```
? There is no indent.

xt/aspell.t
```diff
@@ -24,7 +27,7 @@ text that is part of a code example)
 
 my @files = Test-Files.documents.grep({not $_ ~~ / 'README.' .. '.md' /});
 
-plan +@files;
+plan +@files *2;
```
Altai-man, line 30: A tiny nitpick, but it should be either `+@files * 2` or `+@files*2`?

xt/aspell.t
Altai-man: I suspect it should be a bit smarter and work on pod structure directly, not on unparsed string representation.

xt/aspell.t
```diff
@@ -33,40 +36,81 @@ if $proc.exitcode {
...
+            $body = $text;
+            $dict = $*PROGRAM.parent.child("words.pws").absolute;
+        }
+
+        react {
+            my $proc = Proc::Async.new(:w, ['aspell', '-a', '-l', 'en_US', '--ignore-case', "--extra-dicts=$dict", '--mode=url']);
+
+            whenever $proc.stdout.lines {
+                $lock.protect: {
+                    %output{$file}{$type}<output> = (%output{$file}{$type}<output> // '') ~ $_ ~ "\n";
```
Altai-man, line 89: Won't auto-vivification take care of this?
```
> my %h
{}
> %h<foo> ~= 'hehe'
hehe
> %h
{foo => hehe}
```
Just append `"$_\n"`, no?

xt/aspell.t
```diff
@@ -33,40 +36,81 @@ if $proc.exitcode {
...
-    $proc.bind-stdin: $fixer.stdout: :bin;
-    %output{$file}="";
-    $proc.stdout.tap(-> $buf { %output{$file} = %output{$file} ~ $buf });
-    $proc.stderr.tap(-> $buf {});
-    test-it([$fixer.start, $proc.start], $file);
+for %output.keys.sort -> $file {
+    for %output{$file}.keys -> $type {
+        my $spelling-error-count = 0;
+        for %output{$file}{$type}<output>.lines.tail(*-1) -> $line {
+            next if $line eq '';
```
Altai-man, line 108: Just `next unless $line`. Even better with `for %output{$file}{$type}<output>.lines.tail(*-1).grep(so *)`

xt/words.t
```diff
@@ -0,0 +1,51 @@
...
+ok(sorted(@code), "xt/code.pws is sorted");
+
+my @dupes = @words.Set ∩ @code.Set;
+
+is(~@dupes, "", "No duplicates between xt/words.pws and xt/code.pws");
+
+# are all the words lower case?
+# (ignore some unicode that aspell doesn't case fold as well as we do.
+sub get-uppers(@lexicon) {
+    @lexicon.grep({.lc ne $_}).grep({ ! $_.contains('Þ')})
```
Altai-man, line 41: I think it's easier to read when there is either a space between curlies or not.

xt/aspell.t
coke: Yes, updating to work directly against pod instead of the rendered pod in the cache would be an improvement, but this is unchanged from the previous version of the test.

xt/words.t
coke: f6d325843

xt/aspell.t
coke: f6d325843

xt/aspell.t
coke: f6d325843

xt/aspell.t
coke: f6d325843

PR #3589, opened 2020-08-25
Make `make update-html` generate syntax highlighting

### LINKED ISSUE #3586: `make html` and `make update-html` should handle highlighting in the same way
## Problem

The Makefile contains these two targets:
``` make
html: for-documentable
	documentable start -a -v --highlight

update-html:
	documentable update
```

This means that `make html` builds the site with code highlightling, but `make update-html` builds it without highlighting.  This is confusing to new contributors, especially since it isn't mentioned in the README.  (I personally first thought that `make update-html` was broken and spent several minutes troubleshooting the issue before thinking to check the README).

## Suggestions

Either both targets should include the `--highlight` flag, or neither should.  I plan to submit a PR adding the `--highlight` flag to `make update-html`, but if anyone knows of a reason the flag shouldn't be there, let me know and I'll instead remove the `--highlight` flag from `make html`.

codesections:
This passes the `--highlight` flag to Documentable via the `update-html` target, which is consistent with how the `html` target 
invokes Documentable.

Closes #3586

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ html: for-documentable
 	documentable start -a -v --highlight
 
 update-html:
-	documentable update
+	documentable update --highlight
 
 init-highlights highlights/package-lock.json:
 	ATOMDIR="./highlights/atom-language-perl6";  \


JJ (approved):
Totally makes sense. Thanks.

PR #3590, opened 2020-08-25
Remove gratuitous Perl mention from Tests

codesections:
The Test module previously included the language "as with any Perl project", which outdatedly implied that Raku projects are Perl projects.


coke:
Thanks!

Issue #3593, opened 2020-08-27
aspell shouldn't handle affixes, should it?


codesections:
## Problem

When commenting on #3587, @JJ  raised an interesting question: should "apostrophes be excluded from spell check"?  I looked into aspell's functionality a bit, and the docs indicate that aspell supports a feature called [affix compression](http://aspell.net/man-html/Affix-Compression.html), which _seems_ related to automatically handling apostrophes and other word-endings.  After reading a bit more about the feature, though, I think it's not a good fit for us – it looks like it's mostly about reducing the size-on-disk for the dictionary files and would at least slightly slow processing time.

However, I'm not at all familiar with aspell.  I've relied on it for years, but today may be the first time I've opened its man page.  @coke, I know you just spent some time with the aspell tests – do you (or anyone else) know whether my understanding of affix compression is correct?  And/or do you know if aspell has any other feature that would automatically exclude possessives or other variants of a word when we add the base word to our word list?

JJ:
We can preprocess the word and eliminate apostrophes; after all, we don't
use aspell directly but through a script.

Also, this is an example of a functionality that should be spun off to a
general module for Pod spell checking.

coke:
I don't see anything obvious in the aspell options for handling possessives.

As JJ suggests, we could get rid of any ```'s``` that appears as the end of a word. That'd leave 2 entries in ```code.pws``` (after first making sure that the possessive is actually warranted and it shouldn't be a plural).

coke:
Gave a little more thought - I'd rather leave the ```'s``` explicitly in the lexicon because then we can verify the grammar when they are added. If we strip them, we then allow possessives on things that shouldn't be.

codesections:
That makes sense to me.  Plus, stripping them out would mean adding them back via the script – which would slow things down at least a little.  It seems like it has too many costs for not enough gain.

PR #3594, opened 2020-08-27
Provide a length hint for aspell

codesections:
When looking into aspell for #3593, I discovered that the header in an aspell word list  can optionally have a hint about the length of the list.  Giving aspell this hint allows it to perform very slightly (but measurably) faster.  This appears to improve the speed of our spelling test by ~2%.  The hint does not need to be perfectly accurate to be helpful, so it won't be a problem if the hint isn't updated as contributors add more words to the aspell word list.

See [the relevant aspell docs](http://aspell.net/man-html/Format-of-the-Personal-and-Replacement-Dictionaries.html#Format-of-the-Personal-and-Replacement-Dictionaries).

### DIFF
--- a/xt/pws/words.pws
+++ b/xt/pws/words.pws
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 0 utf-8
+personal_ws-1.1 en 2172 utf-8
 abab
 ababab
 abcd


coke:
This same header is used both in the single version of words, and the combined (dynamic) list - If this speeds things up, great - if it makes things faster to use the right # for both versions, we can do that separately later.

codesections:
> This same header is used both in the single version of words, and the combined (dynamic) list

Yeah, I noticed that.  I the number I added to `words.pws` header is actually the sum of the words in `words.pws` and `code.pws` – which results in `aspell.pws` having the correct header when it's dynamically generated.  I _think_ that's the best outcome, anyway.

Issue #3578, opened 2020-08-21
Add *%_ to documentation for *all* method signatures 


codesections:
After discussing #3576 in the #raku IRC channel, I've decided to close that issue in favor of a new approach.

## Problem

As documented at [type/Method](https://docs.raku.org/type/Method)
> Methods will ignore extra named arguments where other types of `Routine` will throw at runtime. Extra arguments will be forwarded by `nextsame` and friends.

Another way to say this is that each method declaration implicitly adds `*%_` to the method's signature – which can be confirmed by calling `.signature` on any `Method`.  Indeed, [Roast requires](https://github.com/Raku/roast/blob/master/S12-class/interface-consistency.t) that `.signature` display the `*%_` parameter.

However, _none_ of the method signatures the documentation list include this parameter.

## Suggestions

Add `*%_` to all method signatures (through a combination of automated tooling and manual review) and add a brief note to the /type/Method page clarifying that methods receive an implicit `*$_`  parameter. 

As jdv79 pointed out on IRC, this would help new users avoid confusion from looking at the signature listed in the docs and being confused by the fact that a method's signature doesn't exactly match the signature displayed in the docs.  It's possible that it could _create_ confusion for a different class of new user – one so new that they are confused by the `*$_` syntax.  Hopefully, however, adding a clear note in the Method docs will prevent that confusion (and give very new users a chance to learn about some of Raku's cooler features!)

Because this change will impact a significant number of files in `doc/Type` (er, well, _all_ of them), I won't do any work on this issue for a few days in case anyone has objections to this change.

JJ:
Just to be clear, the suggestions do *not* look fine to me. In some cases where it's actually used (say, the infamous `lines` where those arguments are passed through) it might be helpful, but on a case by case basis. Please revert #3601 and let's talk about the best solution for this.

codesections:
Again, my apologies for the miscommunication; I explained how it happened a bit in a comment #3601, and have updated this issue's title to clarify the scope of the proposed changes.

Since there was a miscommunication, let me back up and explain where I'm coming from with this change.

* The documentation currently lists the signature for each documented method
* Roast requires that `Method`s have a `signature` method, which also lists the signature for each method
* In an **ideal** world (one with a 100% complete Roast, a 100% accurate implementation, and a 100% accurate set of docs), these two ways of viewing the signature for the method would show the same signature
[...]

PR #3596, opened 2020-08-28
Normalize files to end with newline

codesections:
This commit normalizes all files in /doc to end with a newline (in conformity with POSIX standards).  This isn't a big deal, but will
allow processing files with automated tooling without generating misleading diffs in the future (e.g, for #3578)


JJ:
Thanks, @coke , that's what I was going to say...

Issue #1981, opened 2018-04-30
Convert 1-file docs to epub with TOC


evb0110:
## The problem
It would be very convenient to have the whole docs file in epub (or mobi, or fb2) format to use it on Kindles and other e-ink devices. The crucial thing is to convert the documents multilevel structure into epub/mobi's bookmarks, otherwise it's impossible to navigate a file of several thousand kindle screens.

## Suggestions
To have epub/mobi versions of the docs with TOC, which should be kept up-to-date to reflect current updates to the docs.

JJ:
Converting HTML to epub is relatively straightforward; generating a table of contents is not big deal either. However, creating a concept index could be a problem, because the generated HTML links do not seem to be compatible with ePub. Would the first be OK?

W4anD0eR96:
Treat index as an item of toc. Be an appendix.

JJ:
But pandoc generates TOC automatically. And I think the way the index is
generated now (using 'id' attribute) does not work in ePub.

JJ:
I don't know how useful is that kind of index in an ePub, though. You can
simply search for whatever you're looking for. Maybe for navigation, but
will have to see...

evb0110:
I always navigate an epub with TOC, and when I make epub with calibre from html, there are usually problems with multi-level TOC.

JJ:
So word and concept index might not be necessary, right?

evb0110:
For me — no, I can live without them, especially on Kindle.

FCO:
I don’t know if it’s my iphone’s Issue, but the ePub version is all blank for me...

JJ:
Well, I've opened it in Firefox and it overlaps a bit, but other than that,
it looks OK...

JJ:
Since the bigpage target has been eliminated because it does not work now, I'm reopening this issue to get it to work again...

mendel:
I managed to generate the "bigpage" HTML with a patched `pod2bigpage`: https://github.com/perl6/perl6-pod-to-bigpage/pull/36. Apparently in the meantime the `epub` and `bigpage` Makefile targets were removed as broken, so those need to be restored, too: #3599.

coke:
https://github.com/Raku/doc-website/issues/75

PR #3599, opened 2020-08-29
Restore epub and bigpage

mendel:
The Makefile targets for generating EPUB and "bigpage" (one-page HTML) were removed as they were broken. Now they may be restored - but only after (or if) https://github.com/perl6/perl6-pod-to-bigpage/pull/36 is merged & released.

This should solve #1981.

### DIFF
--- a/META6.json
+++ b/META6.json
@@ -9,7 +9,7 @@
   "auth": "perl6",
   "depends": [
     "File::Temp",
-    "Pod::To::BigPage:ver<0.5.1+>",
+    "Pod::To::BigPage:ver<0.5.2+>",
     "Pod::To::HTML:ver<0.6.1+>",
     "Test::META",
     "Perl6::TypeGraph"
--- a/META6.json
+++ b/META6.json
@@ -9,7 +9,6 @@
   "auth": "perl6",
   "depends": [
     "File::Temp",
-    "Pod::To::BigPage:ver<0.5.2+>",
     "Pod::To::HTML:ver<0.6.1+>",
     "Test::META",
     "Perl6::TypeGraph"
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ help:
 	@echo "docker-testall:      run all tests (in container)"
 	@echo "docker-run:          run the development webserver (in container)"
 
-html: for-documentable bigpage
+html: for-documentable
 	documentable start -a -v --highlight
 
 update-html:
--- a/Makefile
+++ b/Makefile
@@ -9,17 +9,19 @@ else
 SELINUX_OPT          :=
 endif
 
-.PHONY: html init-highlights assets \
+.PHONY: html init-highlights assets bigpage epub \
 	test xtest ctest help run clean-html clean-images \
 	clean-search clean test-links push \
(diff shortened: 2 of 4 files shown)


META6.json
```diff
@@ -9,7 +9,7 @@
   "auth": "perl6",
   "depends": [
     "File::Temp",
-    "Pod::To::BigPage:ver<0.5.1+>",
+    "Pod::To::BigPage:ver<0.5.2+>",
```
JJ, line 12: Probably the best would be to remove this dependency. At the end of the day, it's only used to generate a single page, and this is only done infrequently. So please consider just eliminating it, and use documentation to warn about installation.

Makefile
```diff
@@ -34,7 +36,7 @@ help:
 	@echo "docker-testall:      run all tests (in container)"
 	@echo "docker-run:          run the development webserver (in container)"
 
-html: for-documentable
+html: for-documentable bigpage
 	documentable start -a -v --highlight
```
JJ, line 40: This should probably be better left removed. bigpage can simply be an independent target.

doc/HomePage.pod6
```diff
@@ -3,7 +3,7 @@
 Welcome to the official documentation of the <a href="https://raku.org">Raku</a>
 programming language!
 Besides online browsing and searching, you can also
-<del><a href="/perl6.html">view everything in one file</a> or</del>(See <a href="https://github.com/Raku/doc/issues/1981">Raku/doc#1981</a>)
+<a href="/perl6.html">view everything in one file</a> or
```
JJ, line 6: While we're at this, we should probably call this page raku.html...

JJ (approved):
This looks fine, but please consider making the indicated changes.

doc/HomePage.pod6
mendel: > While we're at this, we should probably call this page raku.html...

Nice catch - I renamed it in the Makefile, but missed this one. Fixed in 54b36607d.

Makefile
mendel: I agree that `bigpage` should be a completely independent target. OTOH I think it would be nice if the bigpage version of docs.raku.org is kept updated (and in sync with the other pages) and I guessed (possibly incorrectly) that whatever automation re-generates it uses the `html` Makefile target. Ideally the automation should build both, independent targets (possibly via a `docs.rakudo.org` target, which would also build the `epub` target besides `html` and `bigpage`) - WDYT?

So I removed the dependency in 7779e9a14. Do you know what would be the best way to accomplish the goal of keeping the bigpage version updated?

META6.json
mendel: Makes sense. I did so, see 68ea5e97c.

Makefile
JJ: Well, so far updating is done by hand (by YT). I'll do that by hand, from time to time, when there's a sufficient amount of change and it's in a stable state. Also, I don't use make, but do that directly. So not automation here.
If you see the roadmap, web generation should be totally independent. When that's done, it'll probably be automated along with the rest, with a huge amount of tests; not for the time being, I'm afraid.

PR #3601, opened 2020-08-29
Update method signatures to include *%_

codesections:
This PR partially implements #3578 by updating method signatures to include the `*%_` (indicating that they can optionally take named parameters) and, where necessary, updating the signatures to include the invocant.

Despite the large apparent size of this PR, it does not substantively alter the content of any documentation, but should make the documentation as a whole significantly more consistent – both with itself and with the output of the `.signature` introspection method.

This PR only partly implements #3578 because it does not yet include multi methods; I plan to submit a follow-up PR with those changes.

### DIFF
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -223,7 +223,7 @@ In the case of C<Associative>s, it will be applied to its values:
 
 Defined as:
 
-    method duckmap(&block) is rw is nodal
+    method duckmap(&block, *%_) is rw is nodal
 
 C<duckmap> will apply C<&block> on each element that behaves in such a way that
 C<&block> can be applied. If it fails, it will descend recursively if possible,
@@ -679,7 +679,7 @@ don't really have keys.
 
 Defined as:
 
-    method flatmap(&block, :$label)
+    method flatmap(&block, :$label, *%_)
 
 B<DEPRECATION NOTICE>: This method is deprecated in 6.d and will be removed in
 6.e. Use C<.map> followed by C<.flat> instead.
@@ -941,7 +941,7 @@ list containing an empty list or a list with a single element.
 
 Defined as
 
-    method join($separator = '') is nodal
+    method join($separator = '', *%_) is nodal
 
 Converts the object to a list by calling
 L<C<self.list>|/type/Any#routine_list>,
--- a/doc/Type/Any.pod6
+++ b/doc/Type/Any.pod6
@@ -36,7 +36,7 @@ Many built-in types override this for more specific comparisons.
 
 Defined as:
 
-    method any(--> Junction:D)
+    method Any(Any: *%_ --> Junction:D)
 
 Interprets the invocant as a list and creates an
 L<any|/routine/any>-L<Junction|/type/Junction> from it.
@@ -48,7 +48,7 @@ L<any|/routine/any>-L<Junction|/type/Junction> from it.
 
 Defined as:
 
-    method all(--> Junction:D)
+    method Any(Any: *%_ --> Junction:D)
 
 Interprets the invocant as a list and creates an
 L<all|/routine/all>-L<Junction|/type/Junction> from it.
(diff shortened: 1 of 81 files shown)


JJ:
TBH, I really don't like this. At least, you should have waited for someone else to review a PR involving such a major reversal of (possibly implicit) policy. In general, the policy here is to make definitions as close as possible, or the same, as they appear in the source. First, because, well, that's what there is. Second, because when there's a change in the signature (or a new multi appears) you can be sure that it's something that's actually changed, and keep them in sync. Third, tests, one of which you created yourself. The signature of the method now will be different to its definition, so all of them will appear different from the original.

Third, it really does not solve what you're addressing. In #3578 the OP is "add it to method signature documentation", and I said "fine", because I thought that you were actually adding it to the documentation of Signature, or some other general place where it should be clarified. Not to *every* method! Another reason why you should have waited for someone to approve this PR before merging it.

About this, I gather that this might have been done, in part, automatically. Errors occur, and there might be some insidious one. So I would really ask you to please revert this PR so that we can discuss it at length.

doc/Type/Any.pod6
```diff
@@ -115,7 +115,7 @@ scalar in a list context by calling C<.list> and turning it into an C<Array>.
 
 Defined as:
 
-    method push(|values --> Positional:D)
+    method push(Any:U \SELF: |values --> Positional:D)
```
JJ, line 118: This is not related to the rest of the PR. On top of all, I really *don't* like PRs that involve 80 files. If you want PRs to be reviewed, please do not make more than a bunch of changes, on a (small) bunch of files or a single file. Since you've changed the definition, you might want to revise or change the explanation, too.

doc/Type/Any.pod6
```diff
@@ -203,7 +203,7 @@ the C<Seq> will be iterating over its elements, calling C<.take> on them.
 
 Defined as:
 
-    method deepmap(&block --> List) is nodal
+    method deepmap(Any: &block, *%_ --> List) is nodal
```
JJ, line 206: In this case, for instance, it does not make any sense to add *%. The code is simply:

https://github.com/rakudo/rakudo/blob/8134470538457c68164eb64f90e8d3e946d235a0/src/core.c/Any.pm6#L168

this means that that argument is going to be discarded. As said in my comment, we *really* need to think about this on a method-by-method basis.

doc/Type/Any.pod6
```diff
@@ -921,7 +921,7 @@ examples.
 
 Defined as:
 
-    method permutations(|c)
+    method permutations(Any: |c)
```
JJ, line 924: Well, and I'm surprised here. You add 'Any:' which is really implicit, but not the tail? What gives?

JJ (commented):
Well, I was going to `reject` this, but it's merged. So, again, please revert it (and consider simply closing it).

codesections:
Sorry about this! I've reverted the changes and will discuss further in the issue.

> In #3578 the OP is "add it to method signature documentation", and I said "fine", because I thought that you were actually adding it to the documentation of Signature, or some other general place where it should be clarified. Not to every method!

I think we had a miscommunication.  I ended the OP in that issue by saying

> Because this change will impact a significant number of files in `doc/Type` (er, well, all of them), I won't do any work on this issue for a few days in case anyone has objections to this change.

so I thought I was clear that I was suggesting a change to **all** signatures, not just a change to the `Signature` documentation.  Looking back, though, I can see that the _title_ of #3578 wasn't clear on that point.  I know that you/others reading issues are busy, and should have made the scope of the proposed changes clearer in the title.   

> At least, you should have waited for someone else to review a PR involving such a major reversal of (possibly implicit) policy.

I normally would have.  The only reason I didn't here is that I thought #3578 had already established making changes to _all of_ the files in `doc/Type` was OK – and, because this PR changed a lot of lines but 99% just in the way we had (I thought) already agreed on, I didn't want anyone to spend time reviewing a large (but not that informative) diff.

PR #3603, opened 2020-08-30
Update 02-reading-docs.pod6 - s/p6doc/rakudoc/g

mendel:
## The problem

https://github.com/Raku/doc/blob/master/doc/Programs/02-reading-docs.pod6 still recommends using `p6doc` to read POD from the command-line. Unfortunately the latest version of the `p6doc` distro (as installed by `zef install p6doc`) does contain neither the `p6doc` nor the `rakudoc` executable (I tried to `zef install p6doc`, but I found neither `p6doc` nor `rakudoc` executable after successfully installing it).

It looks to me that the command-line POD reader tool, which used to live in the https://github.com/Raku/doc repo and called `p6doc` now lives in a separate repo, https://github.com/Raku/rakudoc, and it's called `rakudoc` now; and now the `p6doc` distro (from the https://github.com/Raku/doc repo) only contains the POD, but not the command-line reader tool. To make things even worse, the `rakudoc` tool (https://github.com/Raku/rakudoc) is currently broken, tests fail: https://github.com/Raku/rakudoc/issues/7

## Solution provided

In this PR I just updated the name of the command-line POD reader tool (and the name of the distro to be installed with zef) to `rakudoc` - but that does not fix the problem on its own.

If all my above assumptions and facts are correct, it looks to me that these steps would be needed to fix the original problem:
* merging https://github.com/Raku/rakudoc/pull/8
* possibly adding a symlink (or copy) of the `rakudoc` executable to also keep the `p6doc` command work for the time being
* uploading a new release of https://github.com/Raku/rakudoc
* merging this PR

### DIFF
--- a/doc/Programs/02-reading-docs.pod6
+++ b/doc/Programs/02-reading-docs.pod6
@@ -2,59 +2,59 @@
 
 =TITLE Reading the docs
 
-=SUBTITLE p6doc - the Raku pod reader
+=SUBTITLE rakudoc - the Raku pod reader
 
-=head1 X<INTRODUCTION|p6doc>
+=head1 X<INTRODUCTION|rakudoc>
 
-Program C<p6doc> is a command-line-interface (CLI) program that reads
+Program C<rakudoc> is a command-line-interface (CLI) program that reads
 Raku pod from B<installed> modules' source code, in contrast to
-running C<perl6 --doc=MODULE programfile> which reads Raku pod from
+running C<raku --doc=MODULE programfile> which reads Raku pod from
 the named source file.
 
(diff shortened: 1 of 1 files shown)


JJ (approved):
This looks fine, but I have yet to test extensively the new rakudoc. I'm confident it works fine, so I'm approving this. Thanks!

mendel:
> This looks fine, but I have yet to test extensively the new rakudoc. I'm confident it works fine, so I'm approving this. Thanks!

Thanks for the review! When you have time, could you please review https://github.com/Raku/rakudoc/pull/8, too, and cut a new release of `rakudoc` (currently it's not installable b/c tests fail)?

PR #3607, opened 2020-08-31
Expand Code.line documentation

### LINKED ISSUE #3606: document behavior of `line` method for auto-generated accessor methods
## Problem 
When the `line` method is invoked on a `Method` that was automatically generated by the `has $.a` syntax, it points to the _class_ definition.  We don't mention this, but I'll add it to the `line` docs.

codesections:
Code.line's documentation previously just stated that it returns the line of the code object's declaration.  This is *usually* true, but
isn't true for auto-generated code objects, which don't have a declaration.  This commit explains how the method works in that case.  Closes #3606

### DIFF
--- a/doc/Type/Code.pod6
+++ b/doc/Type/Code.pod6
@@ -184,15 +184,32 @@ Defined as
     method line(Code:D: --> Int:D)
 
 Returns the line number in the source code at which the code object's
-declaration begins.  If the code object was generated automatically
-(and thus I<not> declared in the source code), then C<line> returns
-the line on which the enclosing scope's declaration begins.  For
-example, when called on an L<automatically generated accessor
-method|/language/objects#Attributes> produced by
-the C<has $.name> syntax, C<line> returns the line on which the
-method's class's declaration begins.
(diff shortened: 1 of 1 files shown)


doc/Type/Code.pod6
```diff
@@ -183,7 +183,14 @@ Defined as
...
 
-Returns the line number in which the code object was declared.
+Returns the line number in the source code at which the code object's
+declaration begins.  If the code object was generated automatically
+(and thus I<not> declared in the source code), then C<line> returns
+the line on which the enclosing scope's declaration begins.  For
+example, when called on an L<automatically generated accessor
+method|/language/objects#Attributes> produced by
+the C<has $.name> syntax, C<line> returns the line on which the
+method's class's declaration begins.
```
JJ, line 193: Adding examples of both would be just great.

JJ (approved):
As usual, thanks, address if you think it's necessary comments.

codesections:
Thanks for the suggestion.  Example added!

PR #3608, opened 2020-09-01
Add xt/check-signatures

### LINKED ISSUE #3604: Add xtest(s) to detect method signatures that are inconsistant with source/Roast
## Problem

I have noticed a number of places where there are slight typos/errors in the way we document method signatures.  For example, there are several methods that are missing a `:` after the invocant (e.g., [`Str.unival`](https://docs.raku.org/routine/unival#(Str)_method_unival) (which I haven't yet fixed so that I can use it as an example here – I'll fix it soon, though)

## Suggestions

We should have one or more tests that help detect these errors.  Some of these may be traditional xtests that, in general, should not fail unless there's a problem in the docs.

Others, however, might be more along the lines of test diagnostics that could identify _potential_ errors, but that might be a bit over-inclusive and thus would be expected to have at least a few failures (e.g., we could detect all methods that don't declare an invocant as their first parameter.  The _vast_ majority of methods do declare an invocant, so it would be helpful to know about the ones that that don't.  Some of those might be errors, but some might be the rare case where we intentionally don't declare an invocant.  These tests could also be useful, but might not belong in `doc/xt`.

codesections:
This commit adds a new xtest, check-signatures.t and closes #3604.  Here is the description of the test from its documentation:

> For each method documented in a .pod6 file in doc/doc/Type/, this test compares the documented signature against the signature in the Rakudo source code.  Because we are documenting _Raku_ and not _Rakudo_, detecting a difference in the signature does not automatically cause the test to fail: some differences in signature (such as different names for positional parameters) represent different implementation choices rather than an error in the docs.
>
> To ensure that implementation details don't cause failing tests, we check only for certain discrepancies that are guaranteed to indicate a substantive/non-implementation detail mismatch between Rakudo and the docs.
>
> Currently, we only test for one category of discrepancies: Methods that are defined with a specific invocant in Rakudo but not in the docs (this helpfully also catches the situation where a documentation signature was _intended_ to have an invocant, but where someone forgot to end the invocant with a `:`)

Running this test currently detects 81 errors -- and the **vast** majority of those errors are places where we forgot a `:` in a method documentation.

I plan to submit a companion pull request that fixes those errors, which will involve very minor changes to 29 files.

### DIFF
--- a/xt/check-signatures.t
+++ b/xt/check-signatures.t
@@ -1,0 +1,146 @@
+use Test;
+use Telemetry; # so we can check it
+grammar TypeDocumentation {...}
+
+=begin SYNOPSIS
+Check the signatures of documented methods against the Rakudo source code.
+
+For each method documented in a .pod6 file in doc/doc/Type/, this test compares
+the documented signature against the signature in the Rakudo source code.
+Because we are documenting I<Raku> and not I<Rakudo>, detecting a difference in
+the signature does not automatically cause the test to fail: some differences in
+signature (such as different names for positional parameters) represent
+different implementation choices rather than an error in the docs.
+
+To ensure that implementation details don't cause failing tests, we check only
+for certain discrepancies that are guaranteed to indicate a
+substantive/non-implementation detail mismatch between Rakudo and the docs.
+
+Currently, we only test for one category of discrepancies:
+
+=item Methods that are defined with a specific invocant in Rakudo but not in the
+docs (this helpfully also catches the situation where a documentation signature
+was I<intended> to have an invocant, but where someone forgot to end the
+invocant with a C<:>)
+
+In the future, this test could be expanded to also check for:
+
+=item arity mismatches
+=item different names for named parameters
+=item times when the Rakudo code specifies a return constraint but the docs do
+not (the inverse situation -- where we specify a return constraint that Rakudo
+does not -- would not necessarily represent an error in the docs so long as
+the function always does return that type)
+
+=end SYNOPSIS
+
+=begin pod
+To check against the Rakudo source code, we need to know have a path to a
+directory containing a git repository with the source code.  By default, this
+test looks for a repository in the same directory that contains the 'docs'
+directory; you can specify an alternate path with the RAKUDO_SRC environmental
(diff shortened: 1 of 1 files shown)


xt/check-signatures.t
```diff
@@ -0,0 +1,146 @@
...
+
+=item Methods that are defined with a specific invocant in Rakudo but not in the
+docs (this helpfully also catches the situation where a documentation signature
+was I<intended> to have an invocant, but where someone forgot to end the
+invocant with a C<:>)
+
+In the future, this test could be expanded to also check for:
+
+=item arity mismatches
+=item different names for named parameters
```
JJ, line 29: This can happen. I don't think this should be really checked. In some cases, names can be changed for clarification.

xt/check-signatures.t
```diff
@@ -0,0 +1,146 @@
...
+    diag($src-line);
+    diag("documented method: $sig")
+}
+
+run <git checkout ->, :out, :err;
+
+grammar TypeDocumentation {
+    token TOP { [<method> || <other-line>]+ }
+
+    token method { ^^ '=head2 method' \N* \n
```
JJ, line 134: Besides method, other declarations can also appear here: sub, infix, possibly prefix and postcircumfix. Those are, at least, the ones that are recognized by the indexing routine. I see that "other line" can also be recognized, but on that case I'm not sure the actual name of the sub/whatever will be recognized as such.

JJ (approved):
This is an xt, so not really critical. I think it's fine the way it is, maybe it can be improved in the future. Great work and thanks.

xt/check-signatures.t
codesections: Thanks.  I agree that names shouldn't be checked for _positional_ parameters (in fact, I mentioned that as the example of something _not_ to check).  But if we give the wrong name for a _named_ parameter, then a user following the docs wouldn't be able to use the method (e.g., if they passed in `:key` to `grep` instead of `:k`, the method won't work)

JJ: Ah, sorry, yes, of course.

PR #3609, opened 2020-09-01
Fix typos detected by xt/check-signatures.t

codesections:
This PR fixes the errors detected by xt/check-signatures.t.  Most of these were simple typos (typically forgetting `:`).  Nearly all others were places where Rakudo constrains the invocant but we did not.

One change was slightly more substantive: we had inadvertently documented the `match` method twice in `Cool`, so I merged the two definitions.

### DIFF
--- a/doc/Type/Baggy.pod6
+++ b/doc/Type/Baggy.pod6
@@ -15,7 +15,7 @@ L<Mixy|/type/Mixy>.
 
 Defined as:
 
-    method new-from-pairs(*@pairs --> Baggy:D)
+    method new-from-pairs(Baggy: *@pairs --> Baggy:D)
 
 Constructs a Baggy objects from a list of L«C<Pair> objects|/type/Pair»
 given as positional arguments:
--- a/doc/Type/Buf.pod6
+++ b/doc/Type/Buf.pod6
@@ -90,7 +90,7 @@ Invokes the C<subbuf-rw> method on the specified C<Buf>:
 
 =head2 method reallocate
 
-    method reallocate($elems)
+    method reallocate(Buf:D: Int:D $elems)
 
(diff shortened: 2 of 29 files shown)


doc/Type/Baggy.pod6
```diff
@@ -15,7 +15,7 @@ L<Mixy|/type/Mixy>.
 
 Defined as:
 
-    method new-from-pairs(*@pairs --> Baggy:D)
+    method new-from-pairs(Baggy: *@pairs --> Baggy:D)
```
JJ, line 18: is this really necessary? It's implicit, right?

doc/Type/Buf.pod6
```diff
@@ -90,7 +90,7 @@ Invokes the C<subbuf-rw> method on the specified C<Buf>:
 
 =head2 method reallocate
 
-    method reallocate($elems)
+    method reallocate(Buf:D: Int:D $elems)
```
JJ, line 93: In this case it might correspond to the actual definition in the source, but in that case you need to explain that reallocate will throw if it's not defined.

codesections: I don't think we need to explicitly call out that `Buf.reallocate` will throw a type error if called with an undefined `Buf` for two reasons:  First, that's just what it _means_ to require a defined invocant, so mentioning it again would be redundant.  Second, it's inconsistent with how we handle defined invocants everywhere else – we never call out that they throw a type error for undefined invocants.  (Some random examples: [Pair.invert](https://docs.raku.org/type/Pair#method_invert), [Capture.list](https://docs.raku.org/type/Capture#method_list), [Baggy.hash](https://docs.raku.org/type/Baggy#method_hash), [Real.sign](https://docs.raku.org/type/Baggy#method_hash))

PR #3610, opened 2020-09-02
Add documentation for recursive Regex <~~>

### LINKED ISSUE #3533: Recursive/self-referential regex match <~~> not documented
According to the spectest, "a leading ~~ indicates a recursive call back into some or all of the current rule."  [S05–metasyntax](https://github.com/Raku/roast/blob/master/S05-metasyntax/angle-brackets.t#L285).  This behavior is not currently described in the docs.

I'm happy to submit a PR – I assume it belongs in the [language/regexes](https://docs.raku.org/language/regexes) page?  However, I'd appreciate a careful review when I submit the PR; I didn't even know about the feature until a few minutes ago, so I might not be the best person to document it!

codesections:
The `<~~>` recursive regex was previously undocumented; now it is, with examples.  Closes #3533

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -1862,6 +1862,37 @@ The order of the regex capture is original:
     say $0; # OUTPUT: «｢c｣␤»
     say $1; # OUTPUT: «｢b｣␤»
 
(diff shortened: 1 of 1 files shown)


JJ (approved):
Great. Thanks!

PR #3622, opened 2020-09-07
Add example's output

stoned:


JJ (approved):
Since you have permissions, you can probably just commit these kind of changes. I'll be happy to review them, anyway, if you decide not to do so.

PR #3623, opened 2020-09-07
- Document the use of a Whatever to delegate all methods from a class.

Xliff:
See this discussion: https://colabti.org/irclogger/irclogger_log/raku?date=2020-09-07#l201

The use of a Whatever was mentioned, but not documented in the examples. This PR adds such an example.

An example of the use of a HyperWhateverable would also be nice, but I don't understand the concept well enough to add one. Maybe in a future PR.


JJ (approved):
It would have probably been better with a comment, but it's perfect as it is. Thanks.

PR #3625, opened 2020-09-08
Revisite (is-)implementation-detail

stoned:
- xref implementation-detail trait
- fix pasto (trait -> method)
- give is-implementation-detail its own section in implementation-detail
trait's description, hence nicely grouped in search results and routine
page


JJ (approved):
Thanks!

PR #3627, opened 2020-09-11
Update modules.pod6

rakool:
Three small text tunings: 

1)  _include_ _path_ appears to not be a Raku term.   This is the only usage I find on [docs.Raku.org](url)
2) "searched ... when Rakudo is started" seems misleading at best.  I think it is just wrong  as a command is needed to place an order for a module.  I don't know what Raku does with core modules but that seems off-topic or arcana here--doubtfully worth a mention or link.
3) 'recursively' seems wrong and just an implementation detail.  Isn't the list linearly searched and then mangling or mapping grabs a module.

If someone knows offhand where this RAKULIB search code is,  please give me a pointer.

### DIFF
--- a/doc/Language/modules.pod6
+++ b/doc/Language/modules.pod6
@@ -520,9 +520,9 @@ export RAKULIB=/path/to/my-modules,/path/to/more/modules
 
 Note that the comma (',') is used as the directory separator.
 
-The include path will be searched recursively for any modules when Rakudo is
-started. Directories that start with a dot are ignored and symlinks are
-followed.
+The directories in the C<RAKULIB> path will be searched for modules when
+Raku C<need>s,C<use>s or C<require>s them. Directories that start with a
+dot are ignored and symlinks are followed.
 
 =head1 Distributing modules
 


doc/Language/modules.pod6
```diff
@@ -520,9 +520,9 @@ export RAKULIB=/path/to/my-modules,/path/to/more/modules
 
 Note that the comma (',') is used as the directory separator.
 
-The include path will be searched recursively for any modules when Rakudo is
-started. Directories that start with a dot are ignored and symlinks are
-followed.
+The directories in the C<RAKULIB> path will be searched for modules when
```
JJ, line 523: `RAKULIB` is a environment variable, which holds several paths at the same time, so maybe change path by "variable"? Or environment variable?

JJ (approved):
Great catch. Thanks a lot for correcting so much stuff in so little space. If you want to fix the small nit before it's merged, that would be great.

rakool:
Will address your nit--your good point.

rakool:
I recant:  I think the environment variable-ness is well addressed just before the diff.   Let me know if you disagree.

My recheck found another possible issue that may need fixing or expansion, at the start of the section in lines 502-503.

**There will be a location provided by the L<distribution|/routine/distribution> and in the current home directory.**

Really, is there?  Here, I tend to think **$HOME** directory, but when I try to broaden my mind, I am still empty headed!

JJ:
> I recant: I think the environment variable-ness is well addressed just before the diff. Let me know if you disagree.

Maybe "variable" in any shape wasn't a good suggestion. What about simply "paths"? It can hold several paths, not just one.
> 
> My recheck found another possible issue that may need fixing or expansion, at the start of the section in lines 502-503.
> 
> **There will be a location provided by the L<distribution|/routine/distribution> and in the current home directory.**
> 
> Really, is there? Here, I tend to think **$HOME** directory, but when I try to broaden my mind, I am still empty headed!

Well, what I do in these cases is to find out the commit that created that change to find a bit of context. I can't really figure out what it really means, but the context (if it's addressing an issue, which might have, for instance, links to discussions in the IRC channel) will probably help. What I understand from this is that the location will be in the current *user*'s home directory, and that you might find out about it using the distribution routine. That's a falsable statement, so it can be checked. At any rate, this clearly needs some clarification, so by all means have a go at it.

rakool:
Regarding illustrating what **RAKULIB**:  I think that is done well enough  in the few prior lines.  If you disagree I will implement something in that vein.

Regarding the finding of intention behind the **in the current home** phrase:  I will look into that.   Have patience, I am a beginner with most of the tool chain I am using, GH included.

JJ:
Please refer to the discussion in this PR when you do.

PR #3633, opened 2020-09-13
Fix examples' output

stoned:


JJ (approved):
Good one, but really, you don't need to create a pull request for this. Just commit to master, it's going to be checked anyway for errors.

PR #3634, opened 2020-09-13
Fix example's output

stoned:


doc/Type/Proc.pod6
```diff
@@ -165,8 +165,9 @@ if the process it ran exited unsuccessfully.
 
 =for code :skip-test<illustrates exception>
 shell 'ls /qqq';
-# OUTPUT: «(exit code 1) ls: cannot access '/qqq': No such file or directory␤The spawned command 'ls /qqq' exited unsuccessfully (exit code: 2)␤
-n block <unit> at /tmp/3169qXElwq line 1␤␤»
+# OUTPUT: «(exit code 1) ls: cannot access '/qqq': No such file or directory␤
```
JJ, line 168: Well... No big deal, but you shouldn't really mix the ␤ symbol along with what it represents, a new line. If you actually produce a new line, just delete the character.

stoned: Indeed! I opted for a multiline output style already used in other documents.

PR #3639, opened 2020-09-13
Sync example's code and its output

stoned:


JJ (changes requested):
In this case, it's the output that was incorrect, not the verb. `put` does actually call the right context:

    put 333444777 ~~ /(3+)/; # OUTPUT: «333␤»

stoned:
> In this case, it's the output that was incorrect, not the verb. `put` does actually call the right context:
> 
> ```
> put 333444777 ~~ /(3+)/; # OUTPUT: «333␤»
> ```

Understood.Thanks. Corrected!

PR #3640, opened 2020-09-13
xref operators

stoned:


doc/Language/contexts.pod6
```diff
@@ -193,7 +193,9 @@ In general, non-zero, non-empty will be converted to C<True>; zero or empty
 will be equivalent to C<False>. But C<.so> can be defined to return any Boolean
 value we want, so this is just a rule of thumb.
 
-C<?> and C<!> are the Boolean context operator and its negation respectively.
+L«C<?>|/language/operators#prefix_?» and
```
JJ, line 196: Well, as long as you're changing this, I don't like the way it's phrased (and I probably wrote it). It's not "the negation" of the operator, it's the negation operation. Both operators are Boolean context operators; one checks truthiness, the other falsiness.

JJ (approved):
It's OK, but you can also address other concerns.

PR #3645, opened 2020-09-19
Simpler examples

stoned:


doc/Type/Seq.pod6
```diff
@@ -155,11 +155,11 @@ C<$n> count as 0.  Also can take a WhateverCode to indicate how many values
 to skip from the end.  Will block on lazy Seqs until the requested
 number of values have been discarded.
 
-    say (1..5).map({$_}).skip;      # OUTPUT: «(2,3,4,5)␤»
-    say (1..5).map({$_}).skip(3);   # OUTPUT: «(4,5)␤»
-    say (1..5).map({$_}).skip(5);   # OUTPUT: «()␤»
-    say (1..5).map({$_}).skip(-1);  # OUTPUT: «(1,2,3,4,5)␤»
-    say (1..5).map({$_}).skip(*-3); # OUTPUT: «(3,4,5)␤»
+    say (1..5).skip;      # OUTPUT: «(2,3,4,5)␤»
+    say (1..5).skip(3);   # OUTPUT: «(4,5)␤»
```
JJ, line 159: It's not a Seq if you don't include .map

```
say (1..5).map({$_}).^name; #OUTPUT: «Seq␤» 
say (1..5).^name; #  OUTPUT: «Range␤»              
```

Works all the same, but it's not an example for this specific feature.
You can simplify it, however, with 

```
say (1...5).^name
```

which *is* a Seq.

JJ (changes requested):
Appreciate the effort, but please check the comment.

doc/Type/Seq.pod6
stoned: oh yes, indeed! I'll close this PR and come back on this after I'll something. Thanks!

Issue #3360, opened 2020-05-05
Checklist for 2020.05


JJ:
These are the changes to Rakudo published in the last release

#  Additions:

 * [x] Implement support of environment variables `RAKULIB` and `RAKUDOLIB`
      to eventually replace deprecated `PERL6LIB`
      -[b3bb0d38](https://github.com/rakudo/rakudo/commit/b3bb0d38)-[cc165a05](https://github.com/rakudo/rakudo/commit/cc165a05)-[1fd660a9](https://github.com/rakudo/rakudo/commit/1fd660a9)-[00c0dec3](https://github.com/rakudo/rakudo/commit/00c0dec3)-[be9bb4b1](https://github.com/rakudo/rakudo/commit/be9bb4b1)-[30a3186b](https://github.com/rakudo/rakudo/commit/30a3186b)-[e8f7eb56](https://github.com/rakudo/rakudo/commit/e8f7eb56)-[61124c35](https://github.com/rakudo/rakudo/commit/61124c35)-[013b47d2](https://github.com/rakudo/rakudo/commit/013b47d2)
 * [x] Implement support for environment variables `RAKU_TEST_DIE_ON_FAIL` and `RAKU_TEST_TIMES`
      to replace `PERL6_TEST_DIE_ON_FAIL` and `PERL6_TEST_TIMES` -[30e7a67c](https://github.com/rakudo/rakudo/commit/30e7a67c)-[afff3b0b](https://github.com/rakudo/rakudo/commit/afff3b0b)
 * [x] Show timing information when installing core modules in build process -[58e61356](https://github.com/rakudo/rakudo/commit/58e61356)
 * [x] Add multiple needles feature to `Str.index/rindex` -[4cec44a3](https://github.com/rakudo/rakudo/commit/4cec44a3)-[cf02c2e9](https://github.com/rakudo/rakudo/commit/cf02c2e9)
 * [x] `dd` can now also handle `BOOTContext` objects -[8b70bfb8](https://github.com/rakudo/rakudo/commit/8b70bfb8)
 * [x] `Str` now has `.Date` and `DateTime` coercers -[d95d218f](https://github.com/rakudo/rakudo/commit/d95d218f)
 * [x] `Date.yyyy-mm-dd` now takes a separator parameter, defaulting to `-` -[3017cfea](https://github.com/rakudo/rakudo/commit/3017cfea)
[...]

PR #3647, opened 2020-09-19
document sequence operator variants ^... and ^...^ ref #3360

stoned:
while here add output to an example for this operator

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -3167,8 +3167,8 @@ knows about arithmetic and geometric sequences.
 
 If the endpoint is not C<*>, it's smartmatched against each generated
 element and the sequence is terminated when the smartmatch succeeded.
-The final element is included in the sequence according to the operator
-variant used.
+The final element is excluded of the sequence if a sequence operator
+variant with a final caret is used, it is included otherwise.
 
 This allows you to write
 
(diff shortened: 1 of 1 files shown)


doc/Language/operators.pod6
```diff
@@ -3148,11 +3155,12 @@ knows about arithmetic and geometric sequences.
 If the endpoint is not C<*>, it's smartmatched against each generated
 element and the sequence is terminated when the smartmatch succeeded.
 For the C<...> operator, the final element is included, for the C<...^>
-operator it's excluded.
+and the C<^...^> operators it's excluded.
```
JJ, line 3158: I guess that, for symmetry, you should also mention the ^… operator, right?

JJ (approved):
It's OK, please address my comment before merging if you consider it necessary.

doc/Language/operators.pod6
stoned: Hum... I think I see what you mean. What do you think about the commit f8fd3cc I pushed on this branch ?

JJ: Essentially yes, but check out again my comments, sorry :-)

stoned: no worries. :-)

PR #3649, opened 2020-09-20
mention the sequence operator

stoned:
while here document the various way to build a Seq at the
start of the document.

### DIFF
--- a/doc/Type/Seq.pod6
+++ b/doc/Type/Seq.pod6
@@ -21,7 +21,7 @@ L«sequence operator C<...>|/language/operators#infix_...» or one of its
 variants.
 
     my $s = (1...5);
-    say $s;              # OUTPUT: «1 2 3 4 5)␤»
+    say $s;              # OUTPUT: «(1 2 3 4 5)␤»
     say $s.^name;        # OUTPUT: «Seq␤»
 
 Assigning the values of a C<Seq> to an array consumes a C<Seq> that is
--- a/doc/Type/Seq.pod6
+++ b/doc/Type/Seq.pod6
@@ -11,6 +11,19 @@ C<Seq> is born in a state where iterating it will consume the values.
 Calling C<.cache> on a C<Seq> will make it store the generated values
 for later access.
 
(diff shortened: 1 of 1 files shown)


doc/Type/Seq.pod6
```diff
@@ -11,6 +11,19 @@ C<Seq> is born in a state where iterating it will consume the values.
...
+A high-level construct to generate a C<Seq> is L«C<gather/take>|/syntax/gather%20take»,
+as well as many built-in methods like L«C<map>|/routine/map» and
+L«C<grep>|/routine/grep», low-level constructors to create a
+C<Seq> from an iterator or from looping constructs are available too.
+
+A C<Seq> can also be constructed with the
+L«sequence operator C<...>|/language/operators#infix_...» or one of its
+variants.
+
+    my $s = (1...5);
```
coke, line 23: the parens are not needed here, right? can we remove them?

doc/Type/Seq.pod6
```diff
@@ -11,6 +11,19 @@ C<Seq> is born in a state where iterating it will consume the values.
...
+    say $s;              # OUTPUT: «1 2 3 4 5)␤»
```
coke, line 24: Mismatched )

coke (commented):
In general, fine, some minor nits on the example code.

doc/Type/Seq.pod6
stoned: indeed! Fixed!

doc/Type/Seq.pod6
stoned: AFAIU to assign a Seq to a scalar it is needed. Or else only the LHS of the sequence operator is assigned to the scalar.

``` 
$ cat /tmp/e.raku 
my $s = 1 ... 5;
($s.^name, $s).say;
$ raku /tmp/e.raku
Potential difficulties:
    Useless use of ... in sink context
    at /tmp/e.raku:1
    ------> my $s = 1 ...⏏ 5;
(Int 1)
```

JJ (approved):
Thanks!

PR #3656, opened 2020-09-28
Updating working on the TWEAK documentation based on confusion here :…

Scimon:
## The problem

There was a stack overflow question that raised some issues with the wording of the order of calling for BUILD and TWEAK methods. 

https://stackoverflow.com/questions/64103610/does-anyone-know-why-the-tweak-routine-gets-hit-before-the-build-routine

## Solution provided

Small change to the wording to make it's meaning more explicit based on my understanding of what the object build process is doing.

### DIFF
--- a/doc/Language/objects.pod6
+++ b/doc/Language/objects.pod6
@@ -773,9 +773,9 @@ default; these are available only during C<TWEAK> (see below), which can
 class.
 
 X<|TWEAK>
-After the C<BUILD> methods have been called, methods named C<TWEAK> are
-called, if they exist, again with all the named arguments that were passed
-to C<new>. See an example of its use below.
+After the C<BUILD> methods have been called for a class, methods named
+C<TWEAK> are called for that class, if they exist, again with all the
+named arguments that were passed to C<new>. See an example of its use below.
 
 Due to the default behavior of C<BUILD> and C<TWEAK> submethods, named
 arguments to the constructor C<new> derived from C<Mu> can


lizmat:
Please, let's make sure we're not actually documenting buggy behaviour here.  I'm not 100% sure a bug wasn't uncovered.

Scimon:
I can look into it further. The code I used to check what was going on was this : 
```
 #!/usr/bin/raku

class   Widget {
    submethod BUILD() {
        say "bw";
    }
    
    submethod TWEAK(:$content, :$styles) {
        say "tw";
    }
 }

 class   File is Widget {
     submethod BUILD() {
         say "bf";
     }

     submethod TWEAK() {
         say "tf";
     }
 }

 my $xml = File.new();
```

lizmat:
This also uncovered a bug in the BUILDPLAN module: I've fixed that and added the package information with https://github.com/rakudo/rakudo/commit/a26beb66e0

alabamenhu:
> So the current code path seems to be `Parent::BUILD -> Parent::TWEAK -> Child::BUILD -> Child::TWEAK`

FWIW, This *feels* right to me.  A parent should be able to get first dibs on making itself, and without interference from a child.  Otherwise, the child could call methods on `Parent` that it's not expecting in between the parent's `BUILD` and `TWEAK` and (potentially) cause problems. 

That siad, `TWEAK` doesn't appear in the original Perl 6 design docs.  But determining whether `TWEAK` should be intercolated with (or fully subsequent to) `BUILD` calls can probably be gleaned from the following two passages in Synopsis 12:

> The bless method automatically calls all appropriate BUILD routines for the current class, which initializes the object in least-derived to most-derived order. (DESTROY submethods work the same way, only in reverse.)

> Within a submethod the $.foo form may only be used on attributes from parent classes, because only the parent classes' part of the object is guaranteed to be in a consistent state (because bless calls the BUILD routines of the parent classes before the child classes).

The key factor for me is the mention that the parent classes' part of the object is *guaranteed to be in a consistent state* as a result of the build order, and `TWEAK` is very likely to be a part of how a class attains that state.  I don't know where the discussion on adding `TWEAK` is, but it seems that it should be read as a part of the "BUILD routines" in the Synopses, and indeed that's how its implemented, so the intercolated order (`Parent::BUILD`, `Parent::TWEAK`, `Child::BUILD`, `Child::TWEAK`) seems to be consistent with both actual behavior and the design docs.

AlexDaniel:
Please document bugs also. It is easy and in fact much more useful to later refactor something to say “Before Rakudo 20XX.XX, … …”. Otherwise users can spend hours trying to figure out why TWEAK doesn't behave in a certain way, which I did in the past a few times.

JJ (approved):
This documents better current behavior than the previous sentence does. So, buggy or not, I would say this should go in.

lizmat (approved):

Scimon:
Should I merge this? Do we have any process on Pull Requests?

PR #3657, opened 2020-09-28
race/hyper are methods not operators

stoned:
while here format {Race,Hyper}Seq word


Altai-man:
I would honestly ask someone's opinion here, because `.race` is not only a method, but also a statement prefix: https://docs.raku.org/language/statement-prefixes#hyper,_race

JJ:
I think it would probably be better to just not give it any qualification, neither method nor operator. Just erase "the method|operator" here.

stoned:
I did just that. Thanks for the feedback !

PR #3660, opened 2020-10-05
Fix output for proto example

jjatria:
## The problem

The listed output for the extended proto example did not match the code, and two of the functions had the same message, making following the code more difficult.

## Solution provided

The messages printed by the methods have been modified so they are all different, and the output listing has been updated after executing the code.


JJ (approved):
Thanks!

Issue #3579, opened 2020-08-23
Fix "pick" use in docs.raku.org/routine/is%20pure


fecundf:
## Problem or new feature

The docs as generated at [is-pure](https://docs.raku.org/routine/is%20pure) say

... `syllables().pick(2..5)` ...

But `pick(2..5)` is the same as `pick(4)`. Seeing this in the docs had me running to the "pick" documentation, and trying things out in the REPL.

## Suggestions

Simplify the example, use `pick(4)` instead.

PR #3661, opened 2020-10-05
Simplify use of pick in is-pure documentation

jjatria:
## The problem

As stated in #3579:

> But pick(2..5) is the same as pick(4). Seeing this in the docs had me running to the "pick" documentation, and trying things out in the REPL.

## Solution provided

As suggested on that ticket, the use of `pick` has been simplified.

### DIFF
--- a/doc/Type/Routine.pod6
+++ b/doc/Type/Routine.pod6
@@ -187,7 +187,7 @@ To see it an action with a particular compiler you can try this example:
     =begin code :preamble<sub syllables {}>
     BEGIN { say ‘Begin’ }
     say ‘Start’;
-    say (^100).map: { syllables().pick(2..5).join("") };
+    say (^100).map: { syllables().pick(4).join("") };
 
 
     # Example output:


Altai-man:
Thanks!

JJ:
It would have been great if you had revised also the output. While this fix is OK, the example seems to have drifted away from the "output", just below... It might have been a good idea to fix that too. Thanks anyway.

jjatria:
@JJ I did try the code before submitting it, but I don't think I could spot a signifiant change in the output? I mean, the words themselves have a different set of randomly selected syllables, but they are still a set of 100 4-syllable words.

Or did I miss something else?

JJ:
Well, "output" is quoted because it's "example" output. I haven't said there was a change; the problem is that it was wrong to start with, including "Generating..." which is not in this piece of code.

jjatria:
Well, no, not in _this_ bit of code. But it is in the "preamble" that is imported at the top, and defined just above it, in line 173:

```
sub syllables() is pure {
    say "Generating syllables";
    my @vowels = <a e i o u>;
    return  @vowels.append: <k m n sh d r t y> X~ @vowels;
}
```

JJ:
Well, example code don't exactly carry over from one to the next, and you can never be sure of where the reader is going to start. Also, not a good practice to print from a sub. So, all in all, examples should be as self-contained as possible... I see your point (and your context), and again, your PR ls great and much appreciated. But just erasing "Generating syllables" would probably make a lot of sense. I can do it now, no problem :-)

jjatria:
I tried to keep my changes to a minimum in this case, since I don't know what lead to the current structure of the code. But by all means, feel free to make changes to this if the current structure needs to be modified. For what it's worth, the code I ran from the docs was

```
sub syllables() is pure {
    say "Generating syllables";
    my @vowels = <a e i o u>;
    return  @vowels.append: <k m n sh d r t y> X~ @vowels;
}

BEGIN { say ‘Begin’ }
say ‘Start’;
say (^100).map: { syllables().pick(4).join("") };
```

and the output 

```
$ raku ~/is-pure.raku
Begin
Generating syllables
Start
(yuromene shatuedi shutisheno yadotika rakayoya ...
```

so I think "just erasing 'Generating syllables'" (without making any more changes) is probably not the right move here.

JJ:
Oh, but it is. That code you run is not the code in the example, and as I said, not a good practice to include output in a routine, so not good code to start with, so we might want to work on that too.
So, there are these options in decreasing order of (subjective) correctness.
* Make the example totally self-contained, by including all needed routines, erasing the preamble and so on. Just to be clear, this is the preferred option.
* Avoid strings that are not defined in the example itself, because that might cause some confusion and do not help illustrate the point. "Begin" and "Start" are the only outputs I can see in the example.
* Avoid "example outputs" that are not real outputs. If they are not generated by the code itself, you might as well skip it totally. 

While "minimal changes" are OK, I tend to slightly widen the angle to have a look at the surroundings of what was changed, and usually suggest changes in that too, once we've spotted some problem there. However, such as the change has been done now, it's perfectly good.

PR #3662, opened 2020-10-11
Pair.keys/kv/values/pairs/antipairs produce a Seq

stoned:
This was changed in rakudo/rakudo@30584dac2fe231038c5bea557946a41310e9fd0f hence since rakudo 2017.08. Should the previous behavior be mentioned ?

### DIFF
--- a/doc/Type/Pair.pod6
+++ b/doc/Type/Pair.pod6
@@ -234,11 +234,11 @@ For more about format strings, see L<sprintf|/routine/sprintf>.
 
 Defined as:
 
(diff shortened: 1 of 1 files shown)


JJ:
You *can*, but definitely not mandatory

JJ (approved):
You've changed list for Seqs, and that's not in the defined method signature. So it might be in roast (haven't checked), but that means anyway it's checking for current behavior, not for spec. So my approval means means your changes are most likely Ok, but inconsistency between method definition and behavior might merit further research

PR #3666, opened 2020-10-18
Fix typo in haskell example

olorin37:
## The problem

To make literal behaving like regular function sorround operator with parens. To get opposite sorround function with backticks.

## Solution provided
Proper syntax provided.


JJ (approved):
Thanks!

PR #3669, opened 2020-10-19
Correct typo

olorin37:
## The problem

Just a typo.

## Solution provided

Just corrected.

But by the way, I see no reason for Tail call elimination section if there is nothing to say about it. Maybe just to remove the section instead of typo fix?


coke (approved):

coke:
People coming from other languages that have often ask if we do as well, so it seems on topic.

olorin37:
@coke I know Haskell and yes, it is important topic there. But tail recursion is rather programming pattern than language feature. So this is why doesn't fits here as a whole section, maybe a note in function section.

coke:
More detailed PRs of course welcome if you have the tuits. Thanks for this one!

Issue #2077, opened 2018-05-29
Improve bless docs


JJ:
## The problem

The [`bless`](https://docs.perl6.org/routine/bless) does not clarify that `bless` calls the superclass constructor. It also says:

> (Though each time you write a custom constructor, remember that it makes subclassing harder).

Which is not exactly clear what it means. Why is it harder?

## Suggestion

Add example of calling the superclass constructor and change or clarify that sentence.

JJ:
Sorry, I will reopen this. It does not show how to call the superclass constructor using `bless`.
And while knowing about the peculiar `new` conventions might be a bit more difficult than knowing about class properties, which are initialized using their names, it's not such a big deal. You can also have as many  `new` as you want...

fluca1978:
You cannot call the superclass constructor using bless: bless does return a new instance and invokes buildall on that, that in turn walks the parent class hierarchy to call other constructors.
It is not clear to me what you mean with "improve" in this context, I suspect you should come up with a working small example of code.

JJ:
El vie., 1 jun. 2018 a las 9:28, Luca Ferrari (<notifications@github.com>)
escribió:

> You cannot call the superclass constructor using bless: bless does return
> a new instance and invokes buildall on that, that in turn walks the parent
> class hierarchy to call other constructors.
>

I think you actually can. I'll try to create an example

> It is not clear to me what you mean with "improve" in this context, I
> suspect you should come up with a working small example of code.
>

I'll try; if I can't, I will leave it closed. Thanks for understanding.

fluca1978:
This is my understanding:

1) `Mu.bless` does create the object and than calls BUILDALL

```
    method bless(*%attrinit) {
        nqp::create(self).BUILDALL(Empty, %attrinit);
    }
```

2) `Mu.new` does the same in the case no `bless` has been defined, or calls `bless` it it has been defined
```
    multi method new(*%attrinit) {
        nqp::if(
          nqp::eqaddr(
            (my $bless := nqp::findmethod(self,'bless')),
            nqp::findmethod(Mu,'bless')
          ),
          nqp::create(self).BUILDALL(Empty, %attrinit),
          nqp::invokewithcapture($bless,nqp::usecapture)
        )
    }
```

that means to me that `bless` is invoked by `new`, not viceversa. The `BUILDALL` is the key in constructor inheritance chain.

JJ:
[...]

PR #3671, opened 2020-10-23
Rework object construction documentation

patrickbkr:
This is mostly a rewording and restructuring of the previous material with the aim of being clearer and easier to understand.

Also several statements that I believe to be false have been corrected:

- There is no need to pass arguments to superclasses during construction.
- `BUILD` methods accessing attributes via `self.^attributes` are not a prerequisite for implementing custom `new` methods.
- In an inheritance hierarchy each `TWEAK` method is called after the corresponding `BUILD` method not after all `BUILD` methods in all classes have run.

### DIFF
--- a/doc/Language/objects.pod6
+++ b/doc/Language/objects.pod6
@@ -760,91 +760,50 @@ say "y: ", $p.y;
 C<Mu.new> calls method L<bless|/routine/bless> on its invocant, passing all
 the named L<arguments|/language/functions#Arguments>. C<bless> creates the new
 object, and then walks all subclasses in reverse method resolution order
-(i.e. from L<Mu|/type/Mu> to most derived classes) and in each class checks for
-the existence of a method named C<BUILD>. If the method exists, the
-method is called with all the named arguments from the C<new> method. If
-not, the public attributes from this class are initialized from named
-arguments of the same name. In either case, if neither C<BUILD> nor the
-default mechanism has initialized the attribute, default values are
-applied. This means that C<BUILD> may change an attribute, but it does
-not have access to the contents of the attribute declared as its
-default; these are available only during C<TWEAK> (see below), which can
-'see' the contents of an attribute initialized in the declaration of the
-class.
+(i.e. from L<Mu|/type/Mu> to most derived classes). In each class C<bless>
+executes the following steps in the order given here:
 
+=item It checks for the existence of a method named C<BUILD>. If the method
+exists, the method is called with all the named arguments it received (from the
+C<new> method).
+
+=item If no C<BUILD> method was found, the public attributes from this class
+are initialized from named arguments of the same name.
+
+=begin item
+All attributes that have not been touched in any of the previous steps have
+their default values applied:
+
+C<has $.attribute = 'default value';>
+=end item
+
+=begin item
 X<|TWEAK>
-After the C<BUILD> methods have been called for a class, methods named
-C<TWEAK> are called for that class, if they exist, again with all the
-named arguments that were passed to C<new>. See an example of its use below.
+C<TWEAK> is called should it exist. It will receive the same arguments
+C<BUILD> receives.
+=end item
 
-Due to the default behavior of C<BUILD> and C<TWEAK> submethods, named
-arguments to the constructor C<new> derived from C<Mu> can
+This object construction scheme has several implications:
+
+=begin item
+Named arguments to the default C<new> constructor (inherited from C<Mu>) can
 correspond directly to public attributes of any of the classes in the method
 resolution order, or to any named parameter of any C<BUILD> or C<TWEAK>
 submethod.
+=end item
 
-This object construction scheme has several implications for customized
-constructors. First, custom C<BUILD> methods should always be submethods,
-otherwise they break attribute initialization in subclasses. Second,
-C<BUILD> submethods can be used to run custom code at object construction
-time. They can also be used for creating aliases for attribute
-initialization:
+=begin item
+Custom C<BUILD> methods should always be submethods, otherwise they are
+inherited to subclasses and prevent default attribute initialization (item two
+in the above list) should the subclass not have its own C<BUILD> method.
+=end item
 
-=begin code
-class EncodedBuffer {
-    has $.enc;
-    has $.data;
-
-    submethod BUILD(:encoding(:$enc), :$data) {
-        $!enc  :=  $enc;
-        $!data := $data;
-    }
-}
-my $b1 = EncodedBuffer.new( encoding => 'UTF-8', data => [64, 65] );
-my $b2 = EncodedBuffer.new( enc      => 'UTF-8', data => [64, 65] );
-#  both enc and encoding are allowed now
-=end code
-
-Since passing arguments to a routine binds the arguments to the parameters,
-a separate binding step is unnecessary if the attribute is used as a
-parameter. Hence the example above could also have been written as:
-
-=begin code :preamble<has $!enc; has $!data>
-submethod BUILD(:encoding(:$!enc), :$!data) {
-    # nothing to do here anymore, the signature binding
-    # does all the work for us.
-}
-=end code
-
-However, be careful when using this auto-binding of attributes
-when the attribute may have special type requirements, such as an C<:$!id>
-that must be a positive integer. Remember, default values will be assigned
-unless you specifically take care of this attribute, and that
-default value will be C<Any>, which would cause a type error.
-
-The third implication is that if you want a constructor that accepts
-positional arguments, you must write your own C<new> method:
-
-=begin code
-class Point {
-    has $.x;
-    has $.y;
-    method new($x, $y) {
-        self.bless(:$x, :$y);
-    }
-}
-=end code
-
-However this is considered poor practice, because it makes correct
-initialization of objects from subclasses harder.
-
-Another thing to note is that the name C<new> is not special in Raku. It is
-merely a common convention, one that is followed quite thoroughly in
-L<most Raku classes|/routine/new>. You can call C<bless> from any method at
-all, or use C<CREATE> to fiddle around with low-level workings.
-
-The C<TWEAK> submethod allows you to check things or modify attributes after
-object construction:
+=begin item
+C<BUILD> may set an attribute, but it does not have access to the contents of
+the attribute declared as its default as they are only applied later.
+C<TWEAK> on the other hand is called after default values have been
+applied and will thus find the attributes initialized. So it can be used to
+check things or modify attributes after object construction:
 
 =begin code
 class RectangleWithCachedArea {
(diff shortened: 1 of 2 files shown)


doc/Language/objects.pod6
```diff
@@ -784,46 +787,135 @@ resolution order, or to any named parameter of any C<BUILD> or C<TWEAK>
...
-    # nothing to do here anymore, the signature binding
+submethod BUILD(:$!x, :$!y) {
+    # Nothing to do here anymore, the signature binding
     # does all the work for us.
 }
 =end code
+=end item
+
+=begin item
+To use default values together with a `BUILD()` method one can't use parameter
```
JJ, line 847: "In order to"...

doc/Language/objects.pod6
```diff
@@ -835,30 +927,36 @@ class Point {
...
-class RectangleWithCachedArea {
-    has ($.x1, $.x2, $.y1, $.y2);
-    has $.area;
-    submethod TWEAK() {
-        $!area = abs( ($!x2 - $!x1) * ( $!y2 - $!y1) );
+class Vector {
+    has $.x;
+    has $.y;
+    has $.length;
+        method new($x, $y) {
```
JJ, line 939: indentation here.

doc/Language/objects.pod6
```diff
@@ -835,30 +927,36 @@ class Point {
...
+        self.bless(:$x, :$y, length => sqrt($x**2 * $y**2));
     }
 }
 
-say RectangleWithCachedArea.new( x2 => 5, x1 => 1, y2 => 1, y1 => 0).area;
-# OUTPUT: «4␤»
+class NamedVector is Vector {
+    has $.name;
+        method new($name, $x, $y) {
```
JJ, line 946: indentation here too. Maybe there's a tab somewhere?

doc/Type/Mu.pod6
```diff
@@ -343,86 +343,9 @@ case we are declaring this C<new> method simply to avoid the extra syntax of
...
-            }
-        }
-    }
-}
-
-my $p = Point-with-ID.new(1,2);
-say $p.raku;
-# OUTPUT: «Point-with-ID.new(ID => "*1-2", x => 1, y => 2)␤»
-=end code
-
```
JJ, line 382: is there a point in deleting all this example? Is there something somewhere else that exemplifies the use of *%args? Or introspection?

JJ (changes requested):
The first file is OK, except if you want to address the two or three quirks. However, deleting a big chunk of documentation without a very good reason (like it being totally wrong, in which case it would be better to fix it) is something we don't really like.
What I would suggest is to leave the first file in this PR (with changes, if you want) and open an issue to discuss options about the second one. It's true that all that in Types/Mu might not seem ideal, but in that case, it would be better to move it somewhere else.

doc/Language/objects.pod6
```diff
@@ -784,46 +787,135 @@ resolution order, or to any named parameter of any C<BUILD> or C<TWEAK>
...
+    }
+}
+my $p1 = Point.new( x => 10, y => 5 );
+=end code
 
+The following C<BUILD> method is equivalent to the above:
 =begin code :preamble<has $!enc; has $!data>
-submethod BUILD(:encoding(:$!enc), :$!data) {
-    # nothing to do here anymore, the signature binding
+submethod BUILD(:$!x, :$!y) {
```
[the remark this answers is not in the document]
patrickbkr: I didn't remove it. It just moved down (see lines 897).

doc/Type/Mu.pod6
patrickbkr: The introductory text for this example states that the complex way of setting the `$!ID` attribute is necessary when implementing a custom `new()`. The following code seems to behave the same and has none of the complexity though:

```
class Point-with-ID is Point {
    has Str $.ID  is rw = '0';

    submethod BUILD( :$x, :$y ) {
        $!ID = "*" ~ $x ~ "-" ~ $y;
    }
}
```

If the above code is actually equivalent, then the text and example are wrong in that the complexity is in any way necessary. That's the reason why I deleted them. They state and suggest a wrong fact. (Please correct me if I just missed the point of the example and there actually is a reason for doing it this way!)

I see your point about not removing parts of the documentation that provide insight that is not better explained elsewhere. I searched a bit and found the following.

There is a bit about `.^attributes` in <https://docs.raku.org/language/classtut#Introspection>.
The `*%args` slurpy hash is explained in <https://docs.raku.org/type/Signature#Slurpy_(A.K.A._variadic)_parameters>.
Maybe we can add the interesting bits of this example there?

doc/Language/objects.pod6
```diff
@@ -760,22 +760,25 @@ say "y: ", $p.y;
...
+exists, the method is called with all the named arguments from the C<new>
+method.
+
+=item If no C<BUILD> method was found, the public attributes from this class
+are initialized from named arguments of the same name.
+
+=begin item
+All attributes that have not been touched in any of the previous steps have
+their default values applied:
+
```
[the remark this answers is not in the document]
patrickbkr: I now added the anchor a few lines down. I've never used `X<|STUFF>` before. I hope I did it right.

doc/Language/objects.pod6
patrickbkr: Done!

doc/Language/objects.pod6
patrickbkr: Done!

doc/Language/objects.pod6
patrickbkr: Done!

doc/Language/objects.pod6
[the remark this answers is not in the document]
patrickbkr: > The first file is OK, except if you want to address the two or three quirks. However, deleting a big chunk of documentation without a very good reason (like it being totally wrong, in which case it would be better to fix it) is something we don't really like.
> What I would suggest is to leave the first file in this PR (with changes, if you want) and open an issue to discuss options about the second one. It's true that all that in Types/Mu might not seem ideal, but in that case, it would be better to move it somewhere else.

I'd like to keep the changes together, as I tried to add the information I removed in `doc/type/Mu.pod6` to `doc/Language/objects.pod6` (Mostly the bit about the danger of using `new`). Separating the removal and addition part hides this a bit.
I'm not in a hurry with this PR, so I'm fine discussing the changes before merging. Is this OK with you?

doc/Type/Mu.pod6
JJ: Well, main point was and is and deleting something needs a bit of discussion. I take your point, but let me do a bit of research to find the reason why that was written that way.

JJ: The explanation for this code is in #2077. It's related to `bless`, the whole point being that that `BUILD` is going to be called from the superclass' `new`. Your code is probably a better choice. But in that case, the best way would have been to reopen #2077 and link this change to that issue.

doc/Type/Mu.pod6
```diff
@@ -343,86 +343,9 @@ case we are declaring this C<new> method simply to avoid the extra syntax of
...
-        }
-    }
-}
-
-say Str-with-ID.new("1.1,2e2").ID;                  # OUTPUT: «0␤»
-my $enriched-str = Str-with-ID.new("3,4");
-say "$enriched-str, {$enriched-str.^name}, {$enriched-str.ID}";
-# OUTPUT: «3,4, Str-with-ID, 1␤»
-=end code
-
```
JJ, line 415: And then, of course, there's this whole thing. Using the same (probably wrong) technique in `BUILD` but, again, consequence of #2077 and serves its purpose. So if you want, you can simply change this `BUILD` to the new version that simply assigns a value to ID.

JJ (changes requested):
Please check out my answers to the changes in Mu.

doc/Type/Mu.pod6
patrickbkr: I have reopened #2077 and changed this PR to close that ticket as you suggested.

doc/Type/Mu.pod6
patrickbkr: I have re-added the example, simplified it a bit and adapted the explanation accordingly.

JJ (changes requested):
A few more comments here...

doc/Language/objects.pod6
```diff
@@ -835,30 +930,36 @@ class Point {
 }
 =end code
 
-However this is considered poor practice, because it makes correct
-initialization of objects from subclasses harder.
-
-Another thing to note is that the name C<new> is not special in Raku. It is
-merely a common convention, one that is followed quite thoroughly in
-L<most Raku classes|/routine/new>. You can call C<bless> from any method at
```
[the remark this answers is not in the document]
patrickbkr: I didn't remove it, but move it to  doc/Language/objects.pod6:913. I put it right before the text that explains why one shouldn't put much logic in `new`. It fits quite nicely there because the fact that `new` is only a regular method and that one shouldn't put much logic into it are related.

doc/Type/Mu.pod6
```diff
@@ -413,16 +367,14 @@ say "$enriched-str, {$enriched-str.^name}, {$enriched-str.ID}";
...
-is going to call C<Str.BUILD>. It will *capture* the value it's looking for, the
-pair C<value => $str> and initialize itself. But we have to initialize also the
-properties of the subclass, which is why within C<BUILD> we use the previously
-explained method to initialize C<$.ID> with the value that is in the C<%args>
-variable. As shown in the output, the objects will be correctly initialized with
-its ID, and will correctly behave as C<Str>, converting themselves in just the
-string in the C<say> statement, and including the C<ID> property as required.
+We are I<enriching> C<Str> with an auto-incrementing ID. We create a custom
+C<new> since we want to be able to initialize it with a bare string. C<bless>
+is going to call C<Str.BUILD>. It will I<capture> the value it's looking for, the
```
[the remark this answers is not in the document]
patrickbkr: `Language/objects.pod6:766` states that `BUILD` will be passed all named arguments that were passed to `bless`. I don't think mentioning `*%` explicitly here is necessary, as that is not specific to `BUILD`. It's documented in detail in [type/Signature/(https://docs.raku.org/type/Signature#index-entry-parameter_*$PERCENT_SIGN>) already.

The order of `BUILD` calls (and the construction process in general) is explained in detail in `Language/objects`.

I think the basic question here is: How much detail should we put in the documentation of `bless` (and `Mu.new`, `BUILD`, `TWEAK`, ...) when there is the exhaustive documentation of the entire construction mechanism in `Language/objects`?

My current approach is to keep the information in the `bless` documentation (and the other methods directly related to object construction) limited and put the context and big-picture documentation into `Language/objects` and link to there from `bless`. This way we have a single place for all the information necessary to fully understand the object construction process. Then one isn't in danger of missing pieces of information that were only mentioned somewhere else.

Following this reasoning I also moved the `Str-with-ID` example to `Language/objects`.

What do you think?

JJ (approved):
Thanks!

JJ:
Thanks for having the patience to work this out. Keep them coming!

Issue #3527, opened 2020-07-21
Checklist for 2020.07


JJ:
Here are the additions in the [new release](https://github.com/rakudo/rakudo/blob/master/docs/announce/2020.07.md)

#  Additions:
 * [x] Set equality operators: `(==)` Unicode `≡` and `≢` which allow to check
      for sameness of the operands using QuantHash semantics
      (for example, `1,2,3 (==) 3,1,2` returns `True`) -[701f4cfb](https://github.com/rakudo/rakudo/commit/701f4cfb)-[dbd69ba7](https://github.com/rakudo/rakudo/commit/dbd69ba7)
 * [x] Add support for Supply.tail(*) and .tail(Inf) -[9ccfc90e](https://github.com/rakudo/rakudo/commit/9ccfc90e)
 * [x] Add support for Supply.tail(*-3) -[eaa72afa](https://github.com/rakudo/rakudo/commit/eaa72afa)
 * [x] Make `Supply.head(*-3)` consistent with `List.head(*-3)` -[5c8c7324](https://github.com/rakudo/rakudo/commit/5c8c7324)-[70e5cfb9](https://github.com/rakudo/rakudo/commit/70e5cfb9)
 * [x] Rakudo now supports the `--rakudo-home` command line option to specify the path of the Rakudo home -[914bcbad](https://github.com/rakudo/rakudo/commit/914bcbad)

lizmat:
Not sure if / where Rakudo command line parameters are documented?

coke:
rakudo-home was already listed when I checked (last unchecked item, lizmat++)

PR #3674, opened 2020-10-25
Add ≡ (equivalent to ==) and ≢ (equivalent to !=)

frithnanth:
## The problem

Two Unicode operators first available in Rakudo v2020.07 don't appear in the table under "Other acceptable single codepoints", as said in #3527 

## Solution provided

This PR updates the table.


coke:
Missing a negation on the second item in the table?

frithnanth:
Fixed

doc/Language/unicode_ascii.pod6
```diff
@@ -112,6 +112,8 @@ X<|«>X<|»>X<|×>X<|÷>X<|≤>X<|≥>X<|≠>X<|−>X<|∘>X<|≅>X<|π>X<|τ>X<
   ⊖      |  U+2296   | (^)        |
   ⊍      |  U+228D   | (.)        |
   ⊎      |  U+228E   | (+)        |
+  ≡      |  U+2261   | ==         | Rakudo v2020.07+
```
JJ, line 115: Wouldn't that be Rakudo *and* Raku?

JJ (approved):
It's basically OK, you might want to consider using Raku instead of Rakudo.

frithnanth:
Changed Rakudo into Raku.

coke:
I think it should be Rakudo / it's showing which version of the compiler the feature was added in

JJ:
> I think it should be Rakudo / it's showing which version of the compiler the feature was added in

The problem is that it's too fine a distinction for the documentation, I'm afraid. As a matter of fact, we often simply eliminate the name altogether.

CIAvash:
`≡` is equivalent to `(==)` set operator, not `==`. Same for `≢`, which is equivalent to `!(==)`.

JJ:
Many thanks. Reverting...

PR #3675, opened 2020-10-25
Revert "Add ≡ (equivalent to ==) and ≢ (equivalent to !=)"

JJ:
Reverts Raku/doc#3674


AlexDaniel:
3 reverts for what? Yes, the op is wrong, but it could've been fixed with just one commit.

JJ:
Well, I use rebase instead of squash usually. No big deal.

AlexDaniel:
@lizmat you're on a mission to downvote all my comments, or what?

AlexDaniel:
So, to clarify, a relatively new contributor submitted a PR that you reviewed poorly. Then, once it was pointed out that there's a little mistake and it talks about a different operator, you reverted the whole thing with three revert commits, instead of just adding missing parens. Not only it unnecessarily clutters the git history, but it also makes it somewhat hard for new contributors to proceed. Why revert it in the first place? You sure that the revert sends the right message?

But, as it seems, this is considered normal around here, and it's not the first time I see this attitude. The downvotes that completely fail to understand the issue, and are there just because I pointed it out, are also very telling.

PR #3676, opened 2020-10-27
Document absolute parameter for IO::Path symlink method

mienaikage:
Flag added in https://github.com/rakudo/rakudo/pull/3980

(Probably a bit early to merge right now)


lizmat (approved):

JJ:
Yep, a bit early. Can you please move it to draft, and un-draft it when it's released? We'll also need some information about the specific release.

PR #3677, opened 2020-10-28
rakudobrew -> rakubrew

patrickbkr:
Replace `rakudobrew` with its successor `rakubrew` where it makes sense. No need to send readers to a page that then forwards them to rakubrew.


JJ (approved):
Thanks!

PR #3678, opened 2020-10-29
Replace non-Spanish phrase

jjatria:
## The problem

The text `în modul` is not Spanish. It was copied by accident from a different language.

## Solution provided

It was replaced by the equivalent "en modo".


JJ (approved):
Good one.

PR #3680, opened 2020-10-30
Remove out of place AUTHORS section

patrickbkr:
If we choose to have an AUTHORS section of the documentation it should be in a central place and not in a random sub page.

This is a draft PR, because I'd like to answer the following questions first:

- Do we want to have an AUTHORS section on the docs page?
- Where would such a section live?
- Who should be credited?


JJ:
As far as I can tell, it's an authors section from the original page. It's just assigning credit where credit is due.

patrickbkr:
Yes, that's what I noticed as well.

PR #3681, opened 2020-10-30
Rename and update Programs/03-environment-variables.pod6

patrickbkr:
`doc/Programs/03-environment-variables.pod6` is a copy of [rakudo/docs/running.pod]<https://github.com/rakudo/rakudo/blob/master/docs/running.pod> but ran out of sync. Merge the two files. Also now include all bits of the rakudo/rakudo version of the file and not only the environment variable paragraph. As a consequence rename the file to `03-running-rakudo.pod6`.

Of note this pulls in the documentation for the new `rakuw.exe` and `rakudow.exe` executables.

### DIFF
--- a/doc/Programs/03-environment-variables.pod6
+++ b/doc/Programs/03-environment-variables.pod6
@@ -1,14 +1,70 @@
 =begin pod :kind("Programs") :subkind("programs") :category("programs")
 
-=TITLE Shell environment variables used by Raku
+=TITLE Running Rakudo
 
-=SUBTITLE The variables you can declare to alter its behavior or gather
-additional information.
+=SUBTITLE How to run Rakudo, a Raku implementation, and modify its behavior
+with environment variables.
 
-The behavior of L<Rakudo|https://rakudo.org>, the Raku interpreter, can be
-tweaked by a (growing) number of environment variables; this section attempts to
-document all those currently in use. They are interpreter specific in all cases,
-except where some use conventional names such as C<PATH>.
+=head1 NAME
+
+raku - Rakudo Raku Compiler
+
+=head1 SYNOPSIS
+
+ raku [switches] [--] [programfile] [arguments]
+
+=head1 DESCRIPTION
+
+With no arguments, enters a REPL. With a C<[programfile]> or the C<-e>
+option, compiles the given program and by default also executes the
+compiled code.
+
+  -c                   check syntax only (runs BEGIN and CHECK blocks)
+  --doc                extract documentation and print it as text
+  -e program           one line of program, strict is enabled by default
+  -h, --help           display this help text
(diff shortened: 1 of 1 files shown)


Altai-man (changes requested):
This will 100% result in broken links.

patrickbkr:
@Altai-man Are you worried about external or internal links? What would you propose?

JJ:
I tend to agree with @Altai-man here. We don't like broken links. We can change the title if needed. Right now it pretty much reflects what's in there, environment variables. Incremental sync would be for me the best way to proceed.

Altai-man:
I am worried about external links.
I would propose waiting until we have redirects, since this is a proper solution, and create a ticket there is such an issue to not forget this.

doc/Programs/03-environment-variables.pod6
```diff
@@ -154,15 +217,45 @@ X<|RAKUDO_SNAPPER>
...
+To receive the output of the program it suffices to redirect it externally:
+
+  rakuw.exe script.raku >stdout.txt 2>stderr.txt
+
+=head1 AUTHORS
 
-The L<initial version of this manual
-page|https://github.com/rakudo/rakudo/blob/master/docs/running.pod> was written
-by Reini Urban, Moritz Lenz and the Rakudo contributors.
+Written by the Rakudo contributors, see the CREDITS file.
```
JJ, line 256: Why not keep the original links?

JJ (approved):
Except for the eliminated links (you probably have a rationale for that), it's OK and I'm surprised we didn't have so far a "running raku" page...

doc/Programs/03-environment-variables.pod6
patrickbkr: If we manage to keep the files in sync, there is no reason to link there. Also the link makes no sense in the rakudo/rakudo version of the file and I'd like to keep the differences we need to maintain minimal.

Altai-man (commented):
Some simple bits, need a more thoughtful review later.

doc/Programs/03-environment-variables.pod6
```diff
@@ -1,29 +1,87 @@
...
+=SUBTITLE How to run Rakudo, a Raku implementation, and modify its behavior
+with environment variables.
 
-The behavior of L<Rakudo|https://rakudo.org>, the Raku interpreter, can be
-tweaked by a (growing) number of environment variables; this section attempts to
-document all those currently in use. They are interpreter specific in all cases,
-except where some use conventional names such as C<PATH>.
+=head1 NAME
+
+raku - Rakudo Raku Compiler
```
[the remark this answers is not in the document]
JJ: This has been already merged, but please keep making your comments, we'll use this as an issue to incorporate them later.

PR #3687, opened 2020-11-08
Fix code highlighting using the correct scope name

stoned:
... when using a recent version of Raku/atom-language-perl6

Recently the scope name for Raku code has changed from 'source.perl6fe'
to 'source.raku' in the package atom-language-perl6.
See Raku/atom-language-perl6#96

We try to detect this change by testing for the presence of grammar file
'raku.cson', which beforehand was named 'perl6fe.cson'

Also, I suspect the script highlights/highlight-folder.coffee is not really used anymore. Perhaps it could be removed ?


JJ (approved):
Will have to test it, but it looks fine.

Issue #3495, opened 2020-06-21
Checklist for 2020.06


JJ:
Here are the changes in the new release:

#  SPECIAL NOTES:
 * [ ] Results of `dir` routine called with dirty base path are presented
      as cleaned up for some paths (like `///` or `a/../a`) unlike
      before. While compatibility with most of other cases is preserved
      and no ecosystem fallout was observed, be cautious of the possible
      observable changes if the code does not clean dirty paths using `cleanup` method.
 * [ ] Since this release, when you start Raku with "-" as a single positional parameter, it
      will now check if STDIN is connected to a terminal.  If it is, then
      the REPL will be started, rather than appearing to freeze waiting
      for the user to enter something and issue an EOF.  If STDIN is *not*
      connected to a terminal, then Raku will read from STDIN and process
      that as the source of a program.
 * [ ] The `run` routine, `Proc.spawn` and `Proc::Async.new` are
      extended with a new argument `:$win-verbatim-args` defaulting to
      `False`. If the argument is left off or set to False, the
      arguments are quoted according to the Microsoft convention. This
      is identical to Rakus previous behavior and thus backwards
      compatible. When passing True the passed arguments are
      concatenated with a single space. No other processing takes
      place. This allows to manually quote the arguments as necessary. 
#  Changes:
 * [ ] Support verbatim argument handling for `Proc` -[709941c6](https://github.com/rakudo/rakudo/commit/709941c6)-[9a75738f](https://github.com/rakudo/rakudo/commit/9a75738f)
 * [ ] Signatures of `slurp` and `spurt` routines not accepting redundant arguments is
      now compile time error rather than run-time one -[22f43444](https://github.com/rakudo/rakudo/commit/22f43444)
 * [ ] The `Distro.desc` method now states codename for MacOS -[b4b2a5ec](https://github.com/rakudo/rakudo/commit/b4b2a5ec)
 * [ ] The `gist` method output for multi-dimensional arrays is truncated now -[cfc3e57f](https://github.com/rakudo/rakudo/commit/cfc3e57f)-[4bfe5bd5](https://github.com/rakudo/rakudo/commit/4bfe5bd5)
 * [ ] Parametrization a variable type in form `my $foo is Type[Foo, Bar]` works for all types now -[de43f193](https://github.com/rakudo/rakudo/commit/de43f193)
[...]

PR #3688, opened 2020-11-10
List.rotate returns a Seq since Rakudo 2020.06

stoned:
ref #3495

While here flesh out a bit the routine description and adopt a more
common example style.


JJ (approved):
Thanks!

PR #3690, opened 2020-11-12
Add a warning paragraph about type of arguments of Signature.new

yguillemot:
## The problem
The runtime creation of a `Signature` object silently fails when arguments of `Signature.new` have a wrong type. This issue currently is not documented.
Moreover the given example precisely show a wrong type for the parameter `count` and would not work.
(See rakudo/rakudo issue #4014)

## Solution provided
1. Modify the example to show the right type
2. Add a small explanatory paragraph to warn the users


JJ (approved):
Thanks!

PR #3692, opened 2020-11-15
Fix IO::Spec::* parent class

stoned:
While here uniformize introduction to these classes

### DIFF
--- a/doc/Type/IO/Spec/Cygwin.pod6
+++ b/doc/Type/IO/Spec/Cygwin.pod6
@@ -4,7 +4,7 @@
 
 =SUBTITLE Platform specific operations on file and directory paths for Cygwin
 
-    class IO::Spec::QNX is IO::Spec::Unix { }
+    class IO::Spec::Cygwin is IO::Spec::Unix { }
 
 An object of this type is available via the variable C<$*SPEC> if the
 Raku interpreter is running on C<Cygwin>.
--- a/doc/Type/IO/Spec/Cygwin.pod6
+++ b/doc/Type/IO/Spec/Cygwin.pod6
@@ -9,14 +9,8 @@
 An object of this type is available via the variable C<$*SPEC> if the
 Raku interpreter is running on C<Cygwin>.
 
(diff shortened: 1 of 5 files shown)


doc/Type/IO/Spec/QNX.pod6
```diff
@@ -4,18 +4,19 @@
...
 
-This sub-class of L<IO::Spec|/type/IO::Spec> specific to QNX will be
-available via C<$*SPEC> if the perl is running on QNX.
+An object of this type is available via the variable C<$*SPEC> if the
+Raku interpreter is running on a C<QNX> platform.
 
-B<NOTE:> the C<IO::Spec::*> classes provide low-level path operations. Unless
-you're creating your own high-level path manipulation routines, you don't
-need to use C<IO::Spec::*>. Use L«C<IO::Path>|/type/IO::Path» instead.
+The C<IO::Spec::*> classes provide low-level path operations. Unless
```
JJ, line 12: Don't like repetition of text in the documentation. As long as you're improving this, maybe you should put all this text in a single place and link it.

JJ (changes requested):
Can you please address the minor concerns indicated in the review? Thanks again for your work.

doc/Type/IO/Spec/Cygwin.pod6
```diff
@@ -4,18 +4,19 @@
 
 =SUBTITLE Platform specific operations on file and directory paths for Cygwin
 
-    class IO::Spec::QNX is IO::Spec { }
+    class IO::Spec::QNX is IO::Spec::Unix { }
```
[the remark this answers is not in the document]
stoned: Yes ! Indeed, and that is what I wanted to fix in the first place! :)

doc/Type/IO/Spec/QNX.pod6
stoned: Sure, what about the update I pushed ? Would that be OK ?

JJ (approved):
:heart:  it. Thanks!

PR #3693, opened 2020-11-15
Introduce a way to mark versioned paragraphs

Altai-man:
## The problem

https://github.com/Raku/doc/issues/302

Right now, there is no standard way to document what things are part of what version. As the documentation is written by a lot of people, eventually people start to write different things in different styles. Given we render this _very important_ text just as the usual text, it is really hard for the user to actually see this important info immediately.

## Solution provided

The discussion in the ticket mentioned suggests Pod6 tags, because "IMO the metadata tags is the way to go: they're trivial to implement, simple to use, and they can always be converted to branches or whatever we need in the future, with a simple porting program.".

Right now, configuration of paragraphs is ignored by most tools, but such metadata creates an opportunity for tools to render
these blocks in a clear way, stating this is a versioned thing.

## Important note

In this PR I _purposely_ went with a simplest way, just "versioned" tag with a direction of the text, stating something is versioned.

This results in e.g.:

2020-11-15-204701_1920x1080_scrot

Let's discuss here and settle the final syntax we have to support && render eventually!

### DIFF
--- a/doc/Language/exceptions.pod6
+++ b/doc/Language/exceptions.pod6
@@ -385,11 +385,13 @@ printing a backtrace along with the message:
 
 =head1 Control exceptions
 
-Control exceptions are raised when throwing an Exception which does the
-L<X::Control|/type/X::Control> role (since Rakudo 2019.03). They are
+=begin para :versioned<down>
+Since Rakudo 2019.03 control exceptions are raised when throwing an Exception which does the
+L<X::Control|/type/X::Control> role. They are
 usually thrown by certain L<keywords|/language/phasers#CONTROL> and are
 handled either automatically or by the appropriate L<phaser|/language/phasers#Loop_phasers>.
 Any unhandled control exception is converted to a normal exception.
+=end para
 
 =begin code
 { return; CATCH { default { $*ERR.say: .^name, ': ', .Str } } }
--- a/doc/Language/faq.pod6
+++ b/doc/Language/faq.pod6
@@ -378,9 +378,13 @@ X<|Data::Dumper (FAQ)>
 Typical options are to use the L<say|/routine/say> routine that uses
 the L<gist|/routine/gist> method which gives the "gist" of the object being
 dumped. More detailed output can be obtained by calling the
-L<perl|/routine/perl> method (soon to be deprecated in favor of C<$obj.raku>,
-available since the Rakudo 2019.11 release) that typically returns an object's
+L<perl|/routine/perl> method that typically returns an object's
 representation in L<EVAL|/routine/EVAL>-able code.
+
+=begin para :versioned<up>
+Since the Rakudo 2019.11 release the C<.raku> method is available, it works just as C<.perl>
+and is advised to use instead of C<.perl> which will be eventually deprecated and removed.
+=end para
 
 If you're using the L<rakudo|https://rakudo.org> implementation, you can use
 the L«rakudo-specific C<dd> routine|/programs/01-debugging#Dumper_function_dd»
--- a/doc/Language/nativecall.pod6
+++ b/doc/Language/nativecall.pod6
@@ -312,15 +312,16 @@ my $ints = CArray[int32].allocate($number_of_ints); # instantiates an array with
 my $n = get_n_ints($ints, $number_of_ints);
 =end code
 
-I<Note>: C<allocate> was introduced in Rakudo 2018.05. Before that, you had to
+=begin para :versioned<down>
+C<allocate> was introduced in Rakudo 2018.05. Before that, you had to
 use this mechanism to extend an array to a number of elements:
+=end para
 
 =begin code :preamble<use NativeCall>
 my $ints = CArray[int32].new;
 my $number_of_ints = 10;
 $ints[$number_of_ints - 1] = 0; # extend the array to 10 items
 =end code
-
 
 The memory management of arrays is important to understand. When you create an
 array yourself, then you can add elements to it as you wish and it will be
--- a/doc/Language/pragmas.pod6
+++ b/doc/Language/pragmas.pod6
@@ -102,8 +102,10 @@ poke;
 # Cannot access '$y' through CALLER, because it is not declared as dynamic
 =end code
 
+=begin para :versioned<up>
 This pragma is not currently part of any Raku specification and was added
 in Rakudo 2019.03.
+=end para
 
 X<|experimental, pragma>
 =head2 experimental
--- a/doc/Language/unicode.pod6
+++ b/doc/Language/unicode.pod6
@@ -108,8 +108,9 @@ Raku supports all Unicode names. X<|\c[] unicode name>
     say "\c[PENGUIN]"; # OUTPUT: «🐧␤»
     say "\c[BELL]";    # OUTPUT: «🔔␤» (U+1F514 BELL)
 
-All Unicode codepoint names/named seq/emoji sequences are now case-insensitive:
-[Starting in Rakudo 2017.02]
+=begin para :versioned<down>
+Starting from Raku 2017.02, all Unicode codepoint names/named seq/emoji sequences are now case-insensitive:
+=end para
 
     say "\c[latin capital letter ae with macron]"; # OUTPUT: «Ǣ␤»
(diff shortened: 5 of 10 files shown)


doc/Language/exceptions.pod6
```diff
@@ -385,11 +385,13 @@ printing a backtrace along with the message:
 
 =head1 Control exceptions
 
-Control exceptions are raised when throwing an Exception which does the
-L<X::Control|/type/X::Control> role (since Rakudo 2019.03). They are
+=begin para :versioned<down>
```
JJ, line 388: You're just making up these values, right? Or is `up` and `down` intended to actually mean something?

JJ (commented):
OK, I see several possible issues here
- Versioning might refer to a whole "unit" (an operator, a method), an argument in a whole unit (a part of the signature or return), or simply a disambiguation of something that does not add any new feature. Versioning paragraphs will help with the later. It's not going to help with the former.
- I don't entirely like the "up" and "down" terminology. We do have versions in Raku, we should probably use versions for that.
- The way I see it, since it's not affecting indexing and it's not going to be exhaustive (although it will probably be required for new things), it's actually an additional piece of style. And, as such, it would probably be much simpler if we added different styles for simple notes, caveat notes, or version notes. So instead of using `:versioned<up>` we should probably have something like `:class<notes>` or `c:lass<versioned`. The thing is, we're not actually using the version, just the fact that it's being versioned. We could as well generalize it.

doc/Language/exceptions.pod6
Altai-man: They do mean something, but probably not what you think they do.

The intent of "up" and "down" is to indicate where this version note belongs to: to previous text or to subsequent text (see an arrow in the right upper corner pointing up in the sample image).

In cases where a version note starts a section or ends it, it is kind of ambiguous where it really belongs, so this is a visual helper and nothing more.

As for Raku versions to put in there: yes, this is what I want to actually settle in this thread. :)

JJ: That would be `above` and `below`, then... But then it's more confusing. Why don't you simply mark the paragraph that's versioned, instead of the versioning note?

Altai-man:
>Versioning might refer to a whole "unit" (an operator, a method), an argument in a whole unit (a part of the signature or return), or simply a disambiguation of something that does not add any new feature. Versioning paragraphs will help with the later. It's not going to help with the former.

A whole unit coverage can be done in the unit's text at the beginning. Same with additional signatures, e.g.

2020-11-15-212109_1920x1080_scrot

>I don't entirely like the "up" and "down" terminology. We do have versions in Raku, we should probably use versions for that

Yes. What would be the format we establish?

>So instead of using :versioned<up> we should probably have something like :class<notes> or c:lass<versioned.

I am totally fine with this and will implement whatever will be decided.

doc/Language/exceptions.pod6
Altai-man: >But then it's more confusing

I'll try to explain in an example.

For example, we have this:

```
method foo is such and such.

[ Oh, by the way, in version 2020.02 you can do this / it behaves this ]

method bar

```

Visually / out of context it is not always obvious to the user if this `Oh, by the way` refers to `foo` or to `bar`. This tiny visual hint tries to help with that.

This is not a replacement for telling in a configurable way "Since Rakudo ...", it is just something simple to make this thing move and we could discuss.

JJ:
@Altai-man I'm not against using actual versioning, but we should really try and find an use case for it; for instance, generating different versions of the documentation based on the version (the way it's done in Python, for instance). That would be cool, but we would have to attend [this issue](https://github.com/Raku/problem-solving/issues/93). If, as I understand it, we're moving from a totally static site to a dynamic site, we might have to do it anyway.
So, baseline is:
- Tags for paragraph styling, including provisions for versioning: minor, incremental change, can be more or less easily accomodated, short-term work. I'm all for it.
- Tags for actual feature versioning, including generation of version-capped documentation: long term, needs some thinking, I'm all for it.
I see it, however, as two different things. And I'm really glad you're starting to do some work in this direction. All mi :heart:

Altai-man:
>I'm not against using actual versioning, but we should really try and find an use case for it

We already have a use case for it, it is ticket #302.

As for other things you mentioned.

When the ticket was active, I suggested generating different versions as e.g. python does, but it was suggested that simple tagging is better. And now I think it is better indeed!

>Tags for paragraph styling, including provisions for versioning: minor, incremental change, can be more or less easily accomodated, short-term work. I'm all for it.

This is what I plan to support just now, quite soon.

>Tags for actual feature versioning, including generation of version-capped documentation: long term, needs some thinking, I'm all for it.

In a dynamic website we will support links in the most backward compatible way possible. This is a nice idea, but not today for sure, maybe somewhere in future. And we still have to establish a format for metadata to rely this upon first, this is what I'd like to do.

JJ:
> > I'm not against using actual versioning, but we should really try and find an use case for it
> 
> We already have a use case for it, it is ticket #302.

You probably mean [this comment within that issue](https://github.com/Raku/doc/issues/302#issuecomment-226827827). The issue does not actually mention an use case. But fair enough.
> 
> As for other things you mentioned.
> 
> When the ticket was active, I suggested generating different versions as e.g. python does, but it was suggested that simple tagging is better. And now I think it is better indeed!
> 
> > Tags for paragraph styling, including provisions for versioning: minor, incremental change, can be more or less easily accomodated, short-term work. I'm all for it.
> 
> This is what I plan to support just now, quite soon.

Great.
> 
> > Tags for actual feature versioning, including generation of version-capped documentation: long term, needs some thinking, I'm all for it.
> 
> In a dynamic website we will support links in the most backward compatible way possible. This is a nice idea, but not today for sure, maybe somewhere in future. And we still have to establish a format for metadata to rely this upon first, this is what I'd like to do.

... and URLs.

Altai-man:
>You probably mean this comment within that issue

Oh, I see now. Well, how I see this whole ticket situation:

1)The issue asks about "a way" to do a thing.
2)It also suggests a number of ways how having such a way would be helpful.
3)I also see a reason, important to me - visual distinction of the versioned info is very important to the reader, that is our ultimate consumer.
4)I want to add "the way" to do the thing requested. Maybe someone will write another renderer with docs capped, this is none of my business now. What I want to do is to provide a way and simultaneously use for my little use case (different styling).

>... and URLs.

Nothing threats URLs in this particular PR. What it adds is a way for documentation sources to say to the renderer (any renderer) "Hey, this section is important, do what you think is worth with it".

JJ:
OK, so please go ahead. My 2c, are that versioning should refer to the paragraph itself, not the previous one or the next one, there should probably be a way of indicating the version itself (which might be useful later on), and that there should be a general spec for highlighting paragraphs additionally to versioning.

coke:
I am happy to see progress here - I agree with JJ, I'd like the versioning note to be for the pod item it's tagged on, and for now we can require it be in a separate paragraph. (If talking about an operator.. add a paragraph description in the operator that has the version note.)

Eventually we could make it smarter to tag something else and generate something more akin to a footnote.

finanalyst:
Just read this thread after focusing on my Collection project. I suggest a different alternative, although it does not preclude this suggestion as well.

For error messages, it would be better if Raku added another field to a Pod Block, namely the line number. This would allow for renderers to pick up the information about where a Pod Block exists in a file, and would therefore make it easier to do versioning (I think).

Altai-man:
I think this suggestion of mine was a bad idea anyway, we need to version whole sets of pages, not just mere paragraphs, former is the future and latter is just a very bad idea, so I kind of abandoned this PR, as https://github.com/Altai-man/docs.raku.org/issues/6 is a far superior way (even though nobody heavy-lifts it, alas).

>For error messages, it would be better if Raku added another field to a Pod Block, namely the line number. This would allow for renderers to pick up the information about where a Pod Block exists in a file, and would therefore make it easier to do versioning (I think).

Can you please provide an example how can it help?

Altai-man:
OTOH I still think that such paragraphs along with UI are awesome to mention not version differences, but maybe compatibility notes, mark important info etc.

PR #3698, opened 2020-11-16
Improve documentation of Map.new

dumarchie:
There are two ways to ensure that a literal pair is not interpreted
as a named argument. Quoting the key is arguably the clearest.


JJ (approved):
Thanks!

PR #3699, opened 2020-11-16
Improve documentation of Hash.push

dumarchie:
* correct signature
* sub `push` throws if it receives an unexpected named argument


JJ (approved):
Thanks!

Issue #3700, opened 2020-11-16
RFC: document array operators as independent routines


dumarchie:
Raku(do) provides a number of subroutines a.k.a. "array operators" such as `pop`, `push`, `shift` and `unshift` that operate on *any* object that implements the corresponding method. Sometimes there's a real advantage to using the subroutine instead of the method. An example is pushing onto a `Hash`: whereas `.push` will silently discard named arguments, the `push` subroutine will throw, informing the programmer they may have accidentally passed a named argument where a `Pair` was intended.

I wonder if it would be a good idea to document these "array operators" as independent routines, instead of documenting them together with the corresponding `Array` methods. If so, should they be grouped under a subheading such as "array operators" or a better name?

I'll happily update the documentation if consensus is reached.

JJ:
It's probably a good idea, yes. And sorry, I missed this issue...

PR #3702, opened 2020-11-21
Update documentation of pop, shift, push, append

dumarchie:
I would like some feedback to find out if I'm on the right track.

### DIFF
--- a/doc/Type/Array.pod6
+++ b/doc/Type/Array.pod6
@@ -22,16 +22,16 @@ called on multi-dimensional arrays.
 Exactly the same as L«C<List.gist>|/type/List#method_gist»,
 except using square brackets for surrounding delimiters.
 
-=head2 routine pop
+=head2 method pop
 
 Defined as:
 
-    multi sub    pop(Array:D )
-    multi method pop(Array:D:)
+    method pop(Array:D:) is nodal
 
-Removes and returns the last item from the array.  Fails for an empty array.
+Removes and returns the last item from the array.  Fails if the array is empty.
 
-Example:
+Like many C<Array> methods, method C<pop> may be called via the corresponding
+L<subroutine|/language/independent-routines#sub_pop>. For example:
 
     my @foo = <a b>; # a b
     @foo.pop;        # b
@@ -40,15 +40,23 @@ Example:
     CATCH { default { put .^name, ': ', .Str } };
     # OUTPUT: «X::Cannot::Empty: Cannot pop from an empty Array␤»
 
-=head2 routine push
+=head2 method push
 
 Defined as:
 
-    multi sub    push(Array:D, **@values --> Array:D)
-    multi method push(Array:D: **@values --> Array:D)
+    multi method push(Array:D: Slip \value --> Array:D)
+    multi method push(Array:D: \value --> Array:D)
+    multi method push(Array:D: **@values is raw --> Array:D)
 
-Adds the C<@values> to the end of the array, and returns the modified array.
-Throws for lazy arrays.
+A subclass that just wants to accept the same arguments as C<Array> can override
+L<all candidates|/routine/push> with a single:
+
+    method push(::?CLASS:D: **@values is raw --> ::?CLASS:D)
+
+The C<Array> class adds the values in the argument list to the end of the array,
+and returns the modified array. If any argument is a C<Slip>, method C<push>
+will add the values produced by the argument's L<iterator|/routine/iterator>.
+It throws if the invocant array or a C<Slip> L<is lazy|/routine/is-lazy>.
 
 Example:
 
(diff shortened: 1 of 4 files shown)


doc/Type/independent-routines.pod6
```diff
@@ -1164,6 +1164,65 @@ assigning L<Empty|/type/Slip#index-entry-Empty-Empty>; for everything else,
...
 
+=head1 Array manipulation
+
+Routines that manipulate arrays and other containers.
+
+=head2 sub pop
+
+Defined as:
+
+    proto sub pop($, *%)
```
JJ, line 1175: I see and appreciate what you are doing here. It's a good idea to move independent subs to where they belong. However, I see some problems, which are not so much in the content itself but in the conventions. First one is here, in the use of `proto`. In general, it's not something we do. There are just 4 proto declarations in the reference section:
```Raku
doc/Type/Range.pod6:    proto method int-bounds(|)
doc/Type/Signature.pod6:proto sub stuff(|) {*}
doc/Type/Signature.pod6:This is often used in C<proto> definitions (like C<proto foo (|) {*}>) to
doc/Type/independent-routines.pod6:proto sub EVAL($code where Blob|Cool|Callable, Str() :$lang = 'Raku',
```
you're adding a bunch (here and in the other files), which would amount as many as there are now. I'm not totally sure about the reason why, but I guess it's simply an implementation detail you don't really need to know to use the feature. So you might consider taking them out.

doc/Type/Array.pod6
```diff
@@ -22,16 +22,16 @@ called on multi-dimensional arrays.
...
 
-Removes and returns the last item from the array.  Fails for an empty array.
+Removes and returns the last item from the array.  Fails if the array is empty.
 
-Example:
+Like many C<Array> methods, method C<pop> may be called via the corresponding
+L<subroutine|/language/independent-routines#sub_pop>. For example:
 
     my @foo = <a b>; # a b
     @foo.pop;        # b
```
JJ, line 37: Right next to this line is the `pop @foo` example statement. You haven't removed it from here... and moved it to the new placement. It might be a good idea to leave it here, but then it would also be convenient to link the new reference for it.

doc/Type/independent-routines.pod6
```diff
@@ -1164,6 +1164,65 @@ assigning L<Empty|/type/Slip#index-entry-Empty-Empty>; for everything else,
...
+
+=head2 sub push
+
+Defined as:
+
+    proto sub push($, |)
+    multi sub push(\a, \b)
+    multi sub push(\a, **@b)
+
+Probably best thought of as a simple:
```
JJ, line 1199: There's a `sub push` example in [`Hash`](https://docs.raku.org/type/Hash#method_push) which is not included in the definition. Maybe a reference to this definition would be needed, as well as an example of how to use it here.

doc/Type/independent-routines.pod6
```diff
@@ -1164,6 +1164,65 @@ assigning L<Empty|/type/Slip#index-entry-Empty-Empty>; for everything else,
...
+
+    sub append(\a, +b)
+
+Calls method L<C<append>|/routine/append> on container C<a> which is supposed to
+add C<b> or the values in C<@b> to the end of a container. In contrast with
+L<sub C<push>|#routine_push>, the container method is supposed to add the values
+produced by the L<iterator|/routine/iterator> if called with a B<single>
+non-slipping C<Iterable> that is not inside a C<Scalar> container.
+
```
JJ, line 1225: There are no examples anywhere of append as a sub, but it might be a good idea to insert them here as long as we're giving it some space.

JJ (commented):
Good job here. Please check out my comments.

doc/Type/Array.pod6
dumarchie: The reference is in line 34 :)

doc/Type/independent-routines.pod6
dumarchie: I noticed that it is not a convention to document prototypes. My reasons for breaking this convention are as follows.

In rakudo/rakudo/issues/1739 it is stated that many core subs have been converted to multis to allow users to add their own candidates. In my opinion this means both the prototype and the core candidates should be documented so users can find out which signatures they can use for their own candidates. The `pop` prototype won't dispatch to a candidate with two positional parameters, for example.

The same reasoning applies to multi methods. Even if the multi implementation was meant as an optimization only (as appears to be the case with `Array`'s `push` and `append`), the user should know about the prototype and the existing candidates so they can make an informed decision about whether to add a candidate or override the method altogether.

Of course a prototype like `(|)` does not need to be documented. I documented the protos of methods `push` and `append` because of the [`is nodal`](https://docs.raku.org/routine/is%20nodal) trait that it applies to all candidates.

Raku is very extensible, and I think it's good if readers of the documentation become aware of it. Of course there's a trade-off. Beginners may just want to get the gist of a sub or method and may be scared of by too many technical details. That's the reason why I added the "Probably best thought of as" sections, but I'm not sure if they are a good idea.

doc/Type/independent-routines.pod6
dumarchie: I agree.

doc/Type/independent-routines.pod6
dumarchie: In [`Hash`](https://docs.raku.org/type/Hash#method_push) I will add a link to the subroutine documentation.

doc/Type/independent-routines.pod6
JJ: It's also a convention that conventions are not changed in a PR :-); they should be discussed first in an issue, and find the best solution for that. I would argue, for instance, that protos should be documented separately from specific instances, since what you want to document for it is how to create a new multi. Totally different use case to calling the routine, which is what's being documented here.

doc/Type/Array.pod6
JJ: Ah, right sorry. Still, what about the example?

doc/Type/independent-routines.pod6
JJ: You might want to move the example too.

doc/Type/Hash.pod6
```diff
@@ -337,13 +337,17 @@ Example:
...
     my %h .= push(e => 6);
-    try push %h, f => 7;
     say %h.raku; # OUTPUT: «{}␤»
 
+Use the corresponding L<subroutine|/language/independent-routines#sub_push> to
+catch this kind of mistake:
+
+    CATCH { default { put .message } };
+    push my %h, f => 7; # OUTPUT: «Unexpected named argument 'f' passed␤»
+
```
JJ, line 350: I'm not totally sure this belongs here now. It should probably be moved to the new reference you have created.

doc/Type/Array.pod6
dumarchie: You wrote it might be a good idea to leave it here, and I agree :-)

doc/Type/Hash.pod6
dumarchie: I thought it would be more useful here. I did it the other way around for 'append'. As both the method and the sub pages contain cross-references between `append` and `push` I thought this would be enough to let users figure out when to use the method (e.g. because they know what they're doing and want maximum efficiency) and when to use the subroutine.

doc/Type/independent-routines.pod6
dumarchie: Ah, I was not aware  of *that* convention and did not intend to violate it :-)

I will remove the protos from this PR and in due time create an issue if no one else does or did.

doc/Type/Array.pod6
JJ: I'm not totally sure that's what I wanted to write then, but point taken.

doc/Type/independent-routines.pod6
JJ: We're surrounded by conventions :-) Thanks a lot.

JJ (approved):
Just tell me when you're ready to merge. Thanks!

dumarchie:
I still need to update the documentation of `unshift`, `prepend` and `splice`. I can add those updates to this PR (and change the title), but you can merge this PR if you prefer a separate PR for those other routines.

dumarchie:
Oh, and of course this PR references #3700

JJ:
Since this is already reviewed and approved, anything you add to it would invalidate that, so it's probably better if you limit it to these functions and create a new one for the rest. Just tell me when this is ready by un-drafting it (add a comment just in case this is not an event sent to watchers)

dumarchie:
You may want to review the changes in my last commit, then :-)

JJ (approved):
Very good. Thanks!

Issue #3706, opened 2020-11-23
OO code example in rb-nutshell doesn't work


raiph:
## Problem or new feature

```
class Person {
    has $.name = 'Jill';
    has $.age  = 42;
    has $!birth_year;
    method BUILD {
        $!birth_year = now.Date.year - $.age;
    }
}
my $p = Person.new( name => 'Jack', age => 23 )
```

is bogus code in several ways, generating warnings, failing to initialize the object, etc.

## Suggestions

Maybe replace with:
```
class Person {
    has $.name = 'Jill';
    has $.age  = 42;
    has $!birth_year = now.Date.year - $!age;
}
my $p = Person.new( name => 'Jack', age => 23 );
```
This compiles and runs without warnings and correctly initializes the object.

I can't vouch for whether it's exactly the same as the Ruby but it's definitely much better than the code that is currently in the doc.

JJ:
Good catch. Thanks!

MorayJ:
"In Raku you get a default constructor that accepts key/value pairs of accessor attributes, and can do further setup in the BUILD method." - For this example, I think we may just need to change BUILD to TWEAK in the code and this line.

lizmat:
Yeah.  A lot of examples in this area predate the existence of the `TWEAK` method.

MorayJ:
Example now changed to TWEAK

PR #3722, opened 2020-12-03
Issue 3706

MorayJ:
## The problem
OO code example in rb-nutshell doesn't work #3706

Code snippet example "is bogus code in several ways, generating warnings, failing to initialize the object, etc."

## Solution provided

The snippet uses BUILD to access properties at instantiation, but they don't yet exist. TWEAK must be used as it runs after the properties are set.

For clarity, have removed private attributes from this example as they are not referred to.


JJ (approved):
Thanks!

PR #3723, opened 2020-12-03
Update haskell-to-p6.pod6 — section Fold

wambash:
## The problem
`reduce` has not named parameter `with`.
In `[+] 0, @numbers`, `@numbers` must be slipped.

## Solution provided
This PR removes the named parameter `with`.
`[+] 0, |@numbers`


JJ (approved):
Thanks!

Issue #3709, opened 2020-11-27
Slip documentation give example result that doesn't mesh with current rakudo result


michael-776:
## Problem

https://docs.raku.org/type/Slip near the top says:

my $l = (1, 2, 3);
...
say (0, slip $l).raku;    # says (0, $(1, 2, 3)), $l does not break apart 

But the current rakudo (Welcome to 𝐑𝐚𝐤𝐮𝐝𝐨™ v2020.11.
Implementing the 𝐑𝐚𝐤𝐮™ programming language v6.d.
Built on MoarVM version 2020.11)
gives this result:
(0, 1, 2, 3)

JJ:
That's correct. We'll need to fix that. Thanks!

PR #3724, opened 2020-12-04
Fix slip example error

MorayJ:
## The problem
Slip documentation give example result that doesn't mesh with current rakudo result #3709

## Solution provided
Fixed example result. Checked with v6.c and the result in the original example was still wrong, so no versioning required.

### DIFF
--- a/doc/Type/Slip.pod6
+++ b/doc/Type/Slip.pod6
@@ -36,7 +36,7 @@ it behaves closer to the C<.Slip> method than the C<slip> subroutine.
 =for code
 my $l = (1, 2, 3);
 say (1, slip 2, 3).raku;  # says (1, 2, 3)      , slips 2, 3 into (1, …)
-say (0, slip $l).raku;    # says (0, $(1, 2, 3)), $l does not break apart
+say (0, slip $l).raku;    # says (0, 1, 2, 3)   , slips from $l into (0, …)
 say (0, $l.Slip).raku;    # says (0, 1, 2, 3)   , slips from $l into (0, …)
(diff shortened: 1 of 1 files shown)


doc/Type/Slip.pod6
```diff
@@ -36,7 +36,7 @@ it behaves closer to the C<.Slip> method than the C<slip> subroutine.
 =for code
 my $l = (1, 2, 3);
 say (1, slip 2, 3).raku;  # says (1, 2, 3)      , slips 2, 3 into (1, …)
-say (0, slip $l).raku;    # says (0, $(1, 2, 3)), $l does not break apart
+say (0, slip $l).raku;    # says (0, 1, 2, 3)   , slips from $l into (0, …)
```
JJ, line 39: Thanks for fixing this, but the problem is that whatever it was trying to illustrate will be left orphan. A more integral approach, looking at the context, would be necessary, illustrating how itemizing will escape slipping (if that's what this particular line was aiming at)

MorayJ: I've added an example that works the way the original writer wanted as well as the fixed example

JJ:
Is it possible that you've done a merge instead of a rebase? I'm afraid all kinds of commits have got mixed, and you might have to force-push to fix it.

MorayJ:
It is possible, though it seemed to me that this had happened in my branch upstream, and I had to pull it down and merge before I could push back to my branch. My concern is that someone may have accidentally pushed to my branch and I'll be blitzing their stuff.

MorayJ:
Thanks. Did a force push and that's all looking a little happier...

JJ (approved):
Great! Thanks!

Issue #3670, opened 2020-10-20
The whole explanation for the difference between block and closure is LTA


JJ:
## Problem

The [example for closures](https://docs.raku.org/language/control#Blocks) includes a block that is "not run", and defines closures as blocks that "do not stand alone as statements".

## Suggestions

I don't really see the difference. They're still blocks. Also, the example does not include an actual use case: it simply drops a block in place of a statement. We should always show intent #1748

MorayJ:
Removing my assignment, in case someone else wants a crack at it as going to do other things first.

PR #3725, opened 2020-12-04
Remove closures discussion from control flow document

MorayJ:
## The problem
The whole explanation for the difference between block and closure is LTA #3670

## Solution provided
Removes reference to closures from this document. Blocks are closures, but it doesn't seem relevant in this context, and is a digression. Document is illustrating blocks that run in place v blocks that are controlled.

Adds some explanation of why you might use a block statement.

Also fixes link to Block type - which, perhaps, should have a reference to closures. Either a link to https://docs.raku.org/language/functions#index-entry-closures or a repeat of the information.

### DIFF
--- a/doc/Language/control.pod6
+++ b/doc/Language/control.pod6
@@ -39,24 +39,38 @@ executed.
     { say 2; say 3 };         # OUTPUT: «2␤3␤»
     say 4;                    # OUTPUT: «4␤»
 
-Unless it stands alone as a statement, a block simply creates a closure. The
-statements inside are not executed immediately. Closures are another topic
-and how they are used is explained
-L<elsewhere|/language/functions#Blocks_and_lambdas>. For now it is just
-important to understand when blocks run and when they do not:
+This might be useful if you are using an editor that offers code folding,
+so you can use these blocks to add visual sections to your code. 
+Blocks also create their own namespace, so can be used for
+scoping.
+ 
+    my $number = 1;
+    print $number ~ ', ';
+    { 
+        my $number = 2;
+        print $number ~ ', ';
+            {
+                my $number = 3;
+                print $number ~ ', ';
+            }
+    }
+    say $number;
+    # OUTPUT: «1, 2, 3, 1␤»
+
+However, if a block is used in an expression, then it doesn't run immediately. 
 
 =for code
-say "We get here";
-{ say "then here." };
-{ say "not here"; 0; } or die;
+say "We get here";             # Statement
+{ say "then here." };          # Block as statement
+{ say "not here"; 0; } or die; # Block used in expression
 
 In the above example, after running the first statement, the first block stands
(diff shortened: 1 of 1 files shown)


MorayJ:
I've made this a draft to prevent it getting accidentally merged without some scrutiny, as my understanding may be flawed.

doc/Language/control.pod6
```diff
@@ -39,24 +39,38 @@ executed.
     { say 2; say 3 };         # OUTPUT: «2␤3␤»
     say 4;                    # OUTPUT: «4␤»
 
-Unless it stands alone as a statement, a block simply creates a closure. The
-statements inside are not executed immediately. Closures are another topic
-and how they are used is explained
-L<elsewhere|/language/functions#Blocks_and_lambdas>. For now it is just
-important to understand when blocks run and when they do not:
+This might be useful if you are using an editor that offers code folding,
+so you can use these blocks to add visual sections to your code.
```
JJ, line 43: Well, this describes intent, only you can simply omit the {} and it will still do the same and look pretty much the same. The example was probably flawed to start with. Blocks are literals for functions, and they should be used as such. They are code.

doc/Language/control.pod6
```diff
@@ -39,24 +39,38 @@ executed.
...
     say 4;                    # OUTPUT: «4␤»
 
-Unless it stands alone as a statement, a block simply creates a closure. The
-statements inside are not executed immediately. Closures are another topic
-and how they are used is explained
-L<elsewhere|/language/functions#Blocks_and_lambdas>. For now it is just
-important to understand when blocks run and when they do not:
+This might be useful if you are using an editor that offers code folding,
+so you can use these blocks to add visual sections to your code. 
+Blocks also create their own namespace, so can be used for
```
JJ, line 44: They don't create a namespace in the Raku sense. It's just a scope.

doc/Language/control.pod6
```diff
@@ -39,24 +39,38 @@ executed.
...
+        print $number ~ ', ';
+            {
+                my $number = 3;
+                print $number ~ ', ';
+            }
+    }
+    say $number;
+    # OUTPUT: «1, 2, 3, 1␤»
+
+However, if a block is used in an expression, then it doesn't run immediately.
```
JJ, line 60: Again, it does not because a block defines a function; it does not run in the same way a function does not run. Maybe you should explain what's the expression context here, and why in the first case it's directly called (as if it had a () behind) and in the second case it's in expression context, so it's not invoked. But the examples weren't too good to start with. Showing intent in this case would mean assigning a block to some other variable, and using it as Callable code.

JJ (commented):
It's a small incremental improvement, but it should maybe be a bit more radical. Thanks!

coke:
Rejecting this - JJ's comments were never addressed - Please feel free to re-submit when these items are addressed.

PR #3726, opened 2020-12-05
Failure.handled() is rw

stoned:


JJ (approved):
Maybe some added explanation in the reference would help, but this is certainly fine.

PR #3729, opened 2020-12-10
Add more CATCH index locations

stoned:


Altai-man:
How would this display during search?

stoned:
> How would this display during search?

This would display
CATCH
CATCH (Catching exceptions)
CATCH (Phasers)

JJ:
I think I know where @Altai-man is going, and it's in the direction of this creating new categories. Commas in the indexing text do, and that's something we really don't want to do now.

stoned:
Ok, understood.

PR #3743, opened 2020-12-20
Update Map.pod6

JustThisGuyAZ:
Clarify what the list method returns.

## The problem
It wasn't clear that the list method returned a list of Pairs.

## Solution provided
Clarified that the list method returns a list of Pairs.


JJ (approved):
Thanks!

PR #3745, opened 2020-12-21
De-trap a trap's description

Altai-man:
Closes https://github.com/Raku/doc/issues/3721

## The problem

The trap example code was incomplete and confused the reader.

## Solution provided

Make it more complete.


JJ (approved):
Thanks!

PR #3746, opened 2020-12-21
Update independent-routines.pod6

Altai-man:
Closes https://github.com/Raku/doc/issues/3730
Closes https://github.com/rakudo/rakudo/issues/4061

## The problem

The EVAL example was outputting expected result due to wrong reasons.

## Solution provided

Added a more complete result demonstrating the code.


JJ (approved):
Thanks!

PR #3748, opened 2020-12-23
Update Path.pod6

JustThisGuyAZ:
Clarify what "it" was referring to in "resolve" docs.

## The problem
It was not clear if "it" was referring to "Windows" or "resolve". I assume "it" is referring to "resolve".

## Solution provided
Clarified what was being referred to.


JJ (approved):
It would probably make sense to check if that "currently" still applies, but thanks!

PR #3751, opened 2020-12-26
Add clarification about private methods

uzluisf:

### DIFF
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -484,12 +484,96 @@ X<|FQN>
 
 Just like attributes, methods can also be private. Private methods are declared
 with a prefixed exclamation mark. They are called with C<self!>
-followed by the method's name. To call a private method of another class the
-calling class has to be trusted by the called class. A trust relationship is
-declared with C<trusts> and the class to be trusted must already be declared.
-Calling a private method of another class requires an instance of that class
-and the fully qualified name (FQN) of the method. Trust also allows access to
-private attributes.
+followed by the method's name. In the following implementation of a C<Region>
+class, methods C<population-info> and C<densely-populated> are private methods,
+while the rest are public ones:
+
+=begin code
+class Region {
+    has $.name;
+    has $.population;
+    has $.area-size;
+    has $.continent;
+    has $.high-population-density;
+
+    method info {
+        say "Name: $!name";
+        say self!population-info;
+    }
+
+    method more-densely-populated(Region:D $r) {
+        my $result = self.population-density > $r.population-density ?? 'more' !! 'less';
+        say "!name is $result densely populated than $r..name.";
+    }
+
+    method population-density {
+        ($!population / $!area-size).round(0.01)
+    }
+
+    method can-be-crowded {
+        self!densely-populated
+        ?? "$!name can be crowdy."
+        !! "There is enough space in $!name."
+    }
+
+    method !population-info(--> Str:D) {
+        "$!population people live here."
+    }
+
+    method !densely-populated(--> Bool:D) {
+        self.population-density > $!high-population-density
+    }
+}
+
+my constant city-hpd = 3_000;
+
+my $wroclaw = Region.new(
+    :name<Wrocław>,
+    :population(638_000),
+    :area-size(293),
+    :continent<Europe>,
+    :high-population-density(city-hpd)
+);
+=end code
+
+Regular methods can be introspected with C<.^methods> and resolved with
+C<.^find-method>.
+
+=begin code :skip-test<compile time error>
+given $wroclaw {
+    say .^methods.first(2);               # OUTPUT: «(info more-densely-populated)␤»
+    say .^find_method('info');            # OUTPUT: «info»
+    say .^find_method('population-info'); # OUTPUT: «(Mu)»
+}
+=end code
+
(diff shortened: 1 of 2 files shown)


doc/Language/classtut.pod6
```diff
@@ -484,12 +484,96 @@ X<|FQN>
...
+    :population(638_000),
+    :area-size(293),
+    :continent<Europe>,
+    :high-population-density(city-hpd)
+);
+=end code
+
+Regular methods can be introspected with C<.^methods> and resolved with
+C<.^find-method>.
+
```
JJ, line 541: You should probably clarify what it actually returns, not show the result of say over them. Also, link to them.

doc/Language/classtut.pod6
```diff
@@ -484,12 +484,96 @@ X<|FQN>
...
+    method !population-info(--> Str:D) {
+        "$!population people live here."
+    }
+
+    method !densely-populated(--> Bool:D) {
+        self.population-density > $!high-population-density
+    }
+}
+
```
JJ, line 527: First, thanks a lot for your contribution. However, there are a couple of things that could be improved here.
1. Methods shouldn't do I/O unless explicitly requested to do so. It's probably better to return a string, and let the client decide what to do with them, than "say" that same string.
2. There are probably better use cases than the one here, which seems to be "these methods are private because I want to illustrate privacy of methods". Both could be private, or not, and it would probably make more sense for them to be public. Documentation should reflect intent #1748, not syntax. So it would be probably better to create an example where it makes sense for them to be private.
3. You should probably also establish some identifier convention, like preceding them with a `_`.
4. Don't particularly like also method returns as strings; you should use the most adequate type for it, or define an enum. `can-be-crowded` should return a `Bool`, for instance, and again let the client decide what to do with that info.

doc/Language/classtut.pod6
```diff
@@ -484,12 +484,96 @@ X<|FQN>
...
+class, methods C<population-info> and C<densely-populated> are private methods,
+while the rest are public ones:
+
+=begin code
+class Region {
+    has $.name;
+    has $.population;
+    has $.area-size;
+    has $.continent;
+    has $.high-population-density;
```
JJ, line 497: Not clear this should be an attribute. Probably a build-time attribute thrown away after the build phase.

doc/Language/classtut.pod6
```diff
@@ -484,12 +484,96 @@ X<|FQN>
...
+
+Since public and private methods don't use the same APIs, we cannot inquire
+about private methods using the same meta-methods. Private methods have their own
+namespace and have a different declaration syntax (C<method !>). Furthermore
+they're not virtual, and thus missing ones are detected at compile time. In
+order to support that in the meta-space, private methods are stored separately,
+added via L<C<.^add_private_method>|/routine/add_private_method>, introspected
+with L<C<.^private_methods>|/routine/private_methods> and resolved by
+L<C<.^find_private_method>|/routine/find_private_method>.
+
```
JJ, line 558: Not totally clear this belongs here, together with the paragraph above. Probably not the place to talk about introspection.

doc/Type/Metamodel/PrivateMethodContainer.pod6
```diff
@@ -42,6 +42,25 @@ Adds a private method C<$code> with name C<$name>.
 
 Returns a hash of C«name => &method_object»
 
+=head2 method private_methods
```
JJ, line 45: This probably belongs in a different PR. It's pretty good, and not worth the while to hold it while we solve the other one.

doc/Type/Metamodel/PrivateMethodContainer.pod6
```diff
@@ -42,6 +42,25 @@ Adds a private method C<$code> with name C<$name>.
...
+=head2 method find_private_method
+
+    method find_private_method($obj, $name)
+
+Locates a private method. Otherwise, returns L<C<Mu>|/type/Mu> if it doesn't
+exist.
+
 =begin comment
 
 TODO: document find_private_method, once we've figured out how to represent
```
JJ, line 66: This TODO should probably be deleted...

JJ (changes requested):
Thanks a lot for your contribution. It would probably be the best to split it in two, if you don't mind.

doc/Type/Metamodel/PrivateMethodContainer.pod6
uzluisf: >You should probably also establish some identifier convention, like preceding them with a _.

Is there any reason for this? I'd daresay that the convention for private methods is already part of the syntax when both declaring them and calling them. Wouldn't doing something like `!_private-method` introduce more syntactic noise?

Thanks for the great feedback, JJ. I'll close this PR and then submit them separately.

JJ: I could say that my sight is not what it used to be, and that a single `!` does not really stand out. On the other hand, mixing `_` with `-` is downright ugly. So I don't know. You're probably right. There's also no convention as long as private attributes go. Still... Well, I really don't know.

tbrowder: As an older Rakoon, I agree with JJ, the '!' really doesn't stand out well, and the leading underscore convention for private methods (and attributes) I believe is used in Perl as well as other languages.

PR #3754, opened 2020-12-28
Update syntax.pod6

alanrocker:
Clarify rules about whitespace and bounding characters of multi-line comments.


JJ (approved):
Thanks a lot! Keep them coming!

PR #3775, opened 2021-01-09
Remove references to Str.sprintf

stoned:
While here minor rewriting of {s,}printf's descriptions, including the
acknowledgment that this document's sprintf section holds the reference
to printf format description :-)


JJ (approved):
Thanks!

PR #3778, opened 2021-01-11
Add clarification about private methods

uzluisf:
Refs: https://github.com/Raku/doc/issues/3521

### DIFF
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -927,6 +927,7 @@ called on C<$mp3>. The C<:local> named argument limits the returned methods to
 those defined in the C<MP3TagData> class and excludes the inherited methods;
 C<MP3TagData> inherits from no class, so providing C<:local> makes no difference.
 
+X<|classes,find_method>
 To check if a type object (or an instance object) implements a certain public
 method, use the L<C<.^find-method>|/routine/find_method> metamethod, which
 returns the method object if it exists. Otherwise, it returns L<C<Mu>|/type/Mu>.
@@ -936,6 +937,8 @@ say $mp3.^find_method('name');   # OUTPUT: «(Mu)␤»
 say $mp3.^find_method('artist'); # OUTPUT: «artist␤»
 =end code
 
+X<|classes,private_methods>
+X<|classes,find_private_methods>
 Type objects can also be introspected for its private methods. However, public
 and private methods don't use the same APIs, and thus different metamethods
 must be used: L<C<.^private_methods>|/routine/private_methods> and
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -483,13 +483,74 @@ X<|FQN>
 =head2 Private methods
 
 Just like attributes, methods can also be private. Private methods are declared
-with a prefixed exclamation mark. They are called with C<self!>
-followed by the method's name. To call a private method of another class the
-calling class has to be trusted by the called class. A trust relationship is
-declared with C<trusts> and the class to be trusted must already be declared.
-Calling a private method of another class requires an instance of that class
-and the fully qualified name (FQN) of the method. Trust also allows access to
-private attributes.
+with a prefixed exclamation mark. They are called with C<self!> followed by the
+method's name. In the following implementation of a C<MP3TagData> class to
+extract L<ID3v1|https://en.wikipedia.org/wiki/ID3> metadata from an mp3 file,
+methods C<parse-data>, C<can-read-format>, and C<trim-nulls> are private
+methods while the remaining ones are public methods:
+
+=begin code
+class MP3TagData {
+    has $.filename where { .IO ~~ :e };
+
+    has Str $.title   is built(False);
+    has Str $.artist  is built(False);
+    has Str $.album   is built(False);
+    has Str $.year    is built(False);
+    has Str $.comment is built(False);
+    has Int $.genre   is built(False);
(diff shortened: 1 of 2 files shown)


doc/Language/classtut.pod6
```diff
@@ -853,6 +910,43 @@ say $o.^parents.map({ $_.^name }).join(', ');
...
+public methods with C<.^methods>:
+
+=begin code :skip-test<compile time error>
+my $mp3 = MP3TagData.new(filename => 'football-head.mp3');
+say $mp3.^methods(:local);
+# OUTPUT: (TWEAK filename title artist album year comment genre track version
+# type Submethod+{is-hidden-from-backtrace}.new)
+=end code
+
+X<|^methods>
```
JJ, line 924: OK, I see it's moved here. Would be interesting then to anchor it to a place that defined it, instead of using it in an example, maybe just adding here a brief explanation

doc/Language/classtut.pod6
```diff
@@ -853,6 +910,43 @@ say $o.^parents.map({ $_.^name }).join(', ');
...
+C<$mp3.^methods(:local)> produces a list of L<Method|/type/Method>s that can be
+called on C<$mp3>. The C<:local> named argument limits the returned methods to
+those defined in the C<MP3TagData> class and excludes the inherited methods;
+C<MP3TagData> inherits from no class, so providing C<:local> makes no difference.
+
+To check if a type object (or an instance object) implements a certain public
```
JJ, line 930: Then ^find_method is not indexed. While you're at it, you could also add an index entry (and explanation, as above)

doc/Language/objects.pod6
```diff
@@ -553,6 +553,13 @@ CATCH { default { put .^name ~ ":\n" ~ .Str } }
...
 =end code
 
+Private methods have their own namespace. Since they're not virtual, missing
+ones are detected at compile time. Unlike in some languages where C<private> is
+an L<accessibility modifier|https://en.wikipedia.org/wiki/Access_modifiers> on a
+method, in Raku "private methods" and "methods" are quite different things -
+that is to say, it's better to read "private method" as a compound noun rather
+than an adjective describing a noun.
+
 Private methods are not inherited by subclasses.
```
JJ, line 563: Then maybe you could explain this as a consequence of the above.

doc/Language/objects.pod6
```diff
@@ -553,6 +553,13 @@ CATCH { default { put .^name ~ ":\n" ~ .Str } }
 # 'FunMath'. Did you mean '!do-subtraction'?␤»
 =end code
 
+Private methods have their own namespace. Since they're not virtual, missing
```
JJ, line 556: What's this namespace? Is it not `GLOBAL`? Is there a different one per class? What do you mean by "virtual"?

JJ (changes requested):
It's fine in general, but maybe a bit of clarification is needed. I've asked a question in the original Rakudo issue, maybe we could ask them to clarify a bit more the terms used here.

doc/Language/classtut.pod6
uzluisf: Do you mean an anchor to the definition of the example (i.e., `MP3TagData`) in particular? Because I used the same format as used [here](https://github.com/Raku/doc/pull/3778/files/a8c3f582daf8d49cf599a67ca0292ea2e9798cfa#diff-a5d9faf4ee78401ed8a8a680beef89e7cd5aa90126d15963f2bfea5584c2170eL802-L803), and linked directly to the section where it's declared.

doc/Language/classtut.pod6
uzluisf: Do you mean here specifically, or within the whole documentation?

doc/Language/objects.pod6
uzluisf: I used the information in this [issue](https://github.com/rakudo/rakudo/issues/3801), so hopefully @jnthn can clarify all these questions for us.

JJ: He clarified a bit further, check it out...

doc/Language/classtut.pod6
JJ: Possibly here, unless you find a better place.

doc/Language/classtut.pod6
JJ: Well, to the first place where you explain what `^methods` are.

uzluisf: >Well, to the first place where you explain what ^methods are.

I'm unclear about what you mean here. I'm myself not explaining what `^methods` does since it was already part of this tutorial and the anchor is already placed there. As for what they're, that's already part of the [documentation](https://github.com/Raku/doc/blob/master/doc/Type/Metamodel/MethodContainer.pod6#L31-L52).

doc/Language/objects.pod6
uzluisf: >Private methods have their own namespace. They're not virtual, i.e.,
private methods cannot be overriden within the inheriting class to provide any
polymorphic behavior, thus missing ones are detected at compile time...

As for the namespace, I don't know.

JJ: OK. Please add it to the document.

doc/Language/classtut.pod6
JJ: OK, just leave it here then. No problem.

JJ (approved):
Looks good now.

PR #3784, opened 2021-01-15
Document new Proc(::Async) arg `arg0`

patrickbkr:
Add documentation for the new Proc::Async :arg0 argument implemented in [rakudo/rakudo#4166](https://github.com/rakudo/rakudo/pull/4166).


JJ:
Thanks, Patrick.

PR #3787, opened 2021-01-20
Update modules.pod6 to include Testing modules

finanalyst:
There is important information in https://stackoverflow.com/questions/55661464/what-is-the-difference-between-i-and-ilib-in-perl6
that is not reflected in the documentation. I have wasted HOURS of time getting a module to install properly because I did not know this.

### DIFF
--- a/doc/Language/modules.pod6
+++ b/doc/Language/modules.pod6
@@ -524,43 +524,85 @@ The directories in the C<RAKULIB> path will be searched for modules when
 Raku C<need>s, C<use>s or C<require>s them. Directories that start with a
 dot are ignored and symlinks are followed.
 
-=head1 Testing modules
+=head1 Testing modules and a distribution
 
-Once a module (for illustration C<MyMainModule>) has been written, 
-the basic test (eg., in t/01-sanity.t) it needs to pass is 
+It is important in this section to repeat the note at the beginning of this document, namely that
+there is a difference between a B<Raku> C<distribution>, which approximates a I<module> in other languages, 
+and a B<Raku> 
+L<module declaration|https://docs.raku.org/language/syntax#index-entry-declarator_unit-declarator_module-declarator_package-Package,_Module,_Class,_Role,_and_Grammar_declaration>.
+The reason for this is that a single file may contain a number of C<class>, C<module>, C<role> etc declarations, or
+these declarations may be spread between different files. In addition, when a C<distribution> is published (see 
+L<Distributing modules>) it may be necessary to provide access to other resources, such as callable utilities.
 
-    use-ok 'MyMainModule';
+B<Raku> also allows for a single distribution to provide multiple C<compunit>s that can be I<use'd> (or I<require'd> etc).
+The information
+about a distribution is contained in the C<META6.json> file in the distribution's root directory. See below
+for more about C<META6.json>. Each entity that the B<distribution> allows to be I<use'd> (or I<require'd> etc), 
+also called a C<compunit>, is 
+placed in the C<provides> section of the C<META6.json> (as specified below). 
+
+It should also be noted that when 
+writing the C<depends> section of the C<META6.json> file, it is the name of the B<distribution> that is listed,
+not the name of the C<compunit> that is desired. Although, for very many entities, the name of the I<distribution>
+and the name of the I<compunit> are the same.
+
+Given this arrangement, the B<rakudo> compiler, or package manager, such as B<zef>, can obtain all the information
+needed about each C<compunit> from a C<META6.json> file in a directory.
+
+This also means that a testing program such as B<Perl>'s C<prove> or the B<Raku> equivalent 
+L<prove6|https://modules.raku.org/dist/App::Prove6:cpan:LEONT> can be run in the root directory of the B<distribution>
+as, eg.,
+
+    prove -e 'raku -I.'
     
-There is an important feature to B<Raku> modules that are distributed, namely
-the presence of a C<META6.json> file in the root of the directory. See below
-for more about C<META6.json>. 
-
-When testing, it is possible to use C<prove6>, the B<Raku> version of C<prove>, eg.
+or 
 
     prove6 -I.
     
-This should be run in the root directory of the Module, and it will run all the tests in 
-the C<t/> sub-directory. The parameter B<-I.>, rather than B<-Ilib>,
-is significant because then the compiler will inspect the C<META6.json> file in the root
-directory of C<MyMainModule>. 
+In both cases, the testing program looks for a directory C<t/> and runs the test programs there (see L<Testing>).'
 
-If you are just beginning to develop the module, and you have not decided what to call it
+The parameter B<-I.>, rather than B<-Ilib>, in the examples above
+is significant because then the compiler will inspect the C<META6.json> file in the root
+directory of the B<distribution>. 
+
+If you are just beginning to develop a module (or series of modules), and you have not decided what to call
 or where to place the code (perhaps splitting it into several C<*.rakumod> files),
 but assuming all the module files are under the C<lib/> directory, then it is possible to use
 
    prove6 -Ilib
 
-This is deceptive. A module may pass all the tests you have set, and you are able to release it,
-but if the Module does not pass all tests
-with C<-I.> then in the future, the module cannot be installed as a dependent module by C<zef>,
+This is deceptive. A distribution may pass all the tests you have set, and you may be able to release it,
+but the distribution (and the modules/classes/roles it contains) cannot be automatically installed. 
+For example, if the distribiton as a dependent module by C<zef>,
 which will test the module using B<-I.> and not C<-Ilib>.
 
-It is also highly recommended to use the C<Test::META> C<meta-ok> test, which verifies the validity of 
-the C<META6.json> file. If you wish to add a module
-to the C<Ecosystem>, then this test will be applied to the Module.
+The most basic test that should be run for each C<compunit> that is defined for the distribution is
 
-Since C<meta-ok> only needs to be tested by the developer/maintainer of a module, it can either
-be made dependent on an Environment variable, eg., in C<t/99-author-test.t> there is the following code
+    use-ok 'MyModule';
+    
+assuming inside C<META6.json> there is a line such as 
+
+    depends: [ "MyModule": "lib/MyModule.rakumod" ],
+
+It is also highly recommended to use the C<Test::META> C<meta-ok> test, which verifies the validity of 
+the C<META6.json> file. If you wish to add a distribution (module)
+to the C<Ecosystem>, then this test will be applied to the distribution.
+
+Since C<meta-ok> only needs to be tested by the developer/maintainer of a module, 
+it can be within a set of extended tests in another directory, eg.
+C</xt>. This is because C<prove6>, for example, only runs the tests in C<t/> by default. Then to 
+run the extended tests:
+
+    prove6 -I. xt/
+
+Indeed it is becoming common practice by B<Raku> community developers to place the extensive
+testing of a distribution in C<xt/> and to put minimal I<sanity> tests in C<t/>. 
+Extensive testing is essential for development and quality control, but
+it can slow down the installation of popular distributions.
+
+An alternative to placing the C<meta-ok> test in an extended test directory, but to ensure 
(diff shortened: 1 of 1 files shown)


doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
 Raku C<need>s, C<use>s or C<require>s them. Directories that start with a
 dot are ignored and symlinks are followed.
 
+=head1 Testing modules
```
JJ, line 527: Maybe testing distributions, as opposed to modules, would be better here? You're talking about testing modules in distributions, not individual modules.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+
+Once a module (for illustration C<MyMainModule>) has been written, 
+the basic test (eg., in t/01-sanity.t) it needs to pass is 
+
+    use-ok 'MyMainModule';
+    
+There is an important feature to B<Raku> modules that are distributed, namely
+the presence of a C<META6.json> file in the root of the directory. See below
+for more about C<META6.json>.
```
JJ, line 536: Maybe link the corresponding resource in the documentation.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+Once a module (for illustration C<MyMainModule>) has been written, 
+the basic test (eg., in t/01-sanity.t) it needs to pass is 
+
+    use-ok 'MyMainModule';
+    
+There is an important feature to B<Raku> modules that are distributed, namely
+the presence of a C<META6.json> file in the root of the directory. See below
+for more about C<META6.json>. 
+
+When testing, it is possible to use C<prove6>, the B<Raku> version of C<prove>, eg.
```
JJ, line 538: Link the module that includes it. Probably drop the mention to prove, too. I wouldn't count on newcomers knowing about it.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+
+    prove6 -I.
+    
+This should be run in the root directory of the Module, and it will run all the tests in 
+the C<t/> sub-directory. The parameter B<-I.>, rather than B<-Ilib>,
+is significant because then the compiler will inspect the C<META6.json> file in the root
+directory of C<MyMainModule>.
```
JJ, line 545: Um. Not totally sure about this. I know `zef` does use it, but do you know `prove6` does that? What for? What does it have to do with the search path to find modules?

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+or where to place the code (perhaps splitting it into several C<*.rakumod> files),
+but assuming all the module files are under the C<lib/> directory, then it is possible to use
+
+   prove6 -Ilib
+
+This is deceptive. A module may pass all the tests you have set, and you are able to release it,
+but if the Module does not pass all tests
+with C<-I.> then in the future, the module cannot be installed as a dependent module by C<zef>,
+which will test the module using B<-I.> and not C<-Ilib>.
+
```
JJ, line 557: The consequence is clear, the cause not so much. Besides, `zef` uses prove6 if it's available, but it can also use prove or its own internal logic. In that sense, it's `zef` the one that uses that prefix for testing, instead of -Ilib, which might cause confusion. Trouble is, it's not totally clear why using this as search path instead of lib, where they are actually stored, causes the confusion.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+It is also highly recommended to use the C<Test::META> C<meta-ok> test, which verifies the validity of 
+the C<META6.json> file. If you wish to add a module
+to the C<Ecosystem>, then this test will be applied to the Module.
+
+Since C<meta-ok> only needs to be tested by the developer/maintainer of a module, it can either
```
JJ, line 562: In this case, I would make it an author test in the xt directory.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+    skip-rest "Skipping author test";
+    exit;
+}
+=end code
+    
+Then when testing the module by the developer, use
+
+    AUTHOR_TEST=1 prove6 -I.
+    
+An alternative is to place the C<meta-ok> test within extended tests and place them in another directory, eg.
```
JJ, line 583: I would reverse the order here; make the alternative the main option and the main option an alternative.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+C</xt>. Since C<prove6> looks at C<t/>, any tests in C</xt> will not be run by default.
+
 =head1 Distributing modules
```
JJ, line 586: Maybe you should add that section *behind* this one, since it's mainly about distributable modules (called *distributions* in our lingo)

JJ (changes requested):
It's basically OK, but I would like to see an explanation about why the search path change might cause problems.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,65 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+
+If you are just beginning to develop the module, and you have not decided what to call it
+or where to place the code (perhaps splitting it into several C<*.rakumod> files),
+but assuming all the module files are under the C<lib/> directory, then it is possible to use
+
+   prove6 -Ilib
+
+This is deceptive. A module may pass all the tests you have set, and you are able to release it,
+but if the Module does not pass all tests
+with C<-I.> then in the future, the module cannot be installed as a dependent module by C<zef>,
```
ugexe, line 555: To be clear it doesn't matter if its a dependency or installing directly by itself. Nor does it matter if it is installed with zef, zef with forced or disabled tests, or a raku one-liner that uses CURI ala

```
# from ./
raku -I. t/my-test.t
raku -e 'my $dist = CompUnit::Repository::FileSystem.new(prefix => $*CWD.absolute).distribution; say CompUnit::RepositoryRegistry.repository-for-name("site").install($dist);'

# from ./lib
raku -Ilib t/my-test.t
raku -e 'my $dist = CompUnit::Repository::FileSystem.new(prefix => $*CWD.add("lib").absolute).distribution; say CompUnit::RepositoryRegistry.repository-for-name("site").install($dist);'
```

doc/Language/modules.pod6
finanalyst: added clarification and 'distribution' to title

doc/Language/modules.pod6
finanalyst: I've tried to make it clearer that it is the compiler that uses META6.json

doc/Language/modules.pod6
```diff
@@ -524,6 +524,106 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+
+B<Raku> also allows for a single distribution to provide multiple C<compunit>s that can be I<use'd> (or I<require'd> etc).
+The information
+about a distribution is contained in the C<META6.json> file in the distribution's root directory. See below
+for more about C<META6.json>. Each entity that the B<distribution> allows to be I<use'd> (or I<require'd> etc), 
+also called a C<compunit>, is 
+placed in the C<provides> section of the C<META6.json> (as specified below). 
+
+It should also be noted that when 
+writing the C<depends> section of the C<META6.json> file, it is the name of the B<distribution> that is listed,
```
JJ, line 545: Well... you know all the trouble about Pod::To::HTML? It was because what you include in depends, or at least how it's interpreted, is the name of the module, not the distribution; actually, it looks up the "provides" section of every distribution.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,106 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+not the name of the C<compunit> that is desired. Although, for very many entities, the name of the I<distribution>
+and the name of the I<compunit> are the same.
+
+Given this arrangement, the B<rakudo> compiler, or package manager, such as B<zef>, can obtain all the information
```
JJ, line 549: Given that now we have some other package managers, maybe say "package manager*s*" and drop the comma.

JJ (changes requested):
Much better, but you might want to clarify the distinction between module and distribution name regarding `depends`

doc/Language/modules.pod6
finanalyst: It was trying to resolve those problems that led me to want to change the documentation. 
However, I think that may be it would be better to commit these changes to the module document, then add more information with another PR.

JJ: You need to change "distribution" to "module" then; depends list modules, not distributions.

doc/Language/modules.pod6
finanalyst: made reference to zef as an example package manager

JJ (approved):

doc/Language/modules.pod6
```diff
@@ -524,6 +524,113 @@ The directories in the C<RAKULIB> path will be searched for modules when
 Raku C<need>s, C<use>s or C<require>s them. Directories that start with a
 dot are ignored and symlinks are followed.
 
+=head1 Testing modules and a distribution
+
+It is important in this section to repeat the note at the beginning of this document, namely that
+there is a difference between a B<Raku> C<distribution>, which approximates a I<module> in other languages,
```
ugexe, line 530: A distribution is a collection of modules and other code, packaged in such a way as to be distributed. I wouldn’t draw a comparison between a distribution and a module; one is simply part of the other.

doc/Language/modules.pod6
```diff
@@ -524,6 +524,113 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+and a B<Raku> 
+L<module declaration|https://docs.raku.org/language/syntax#index-entry-declarator_unit-declarator_module-declarator_package-Package,_Module,_Class,_Role,_and_Grammar_declaration>.
+The reason for this is that a single file may contain a number of C<class>, C<module>, C<role> etc declarations, or
+these declarations may be spread between different files. In addition, when a C<distribution> is published (see 
+L<Distributing modules>) it may be necessary to provide access to other resources, such as callable utilities.
+
+B<Raku> also allows for a single distribution to provide multiple C<compunit>s that can be I<use'd> (or I<require'd> etc).
+The information
+about a distribution is contained in the C<META6.json> file in the distribution's root directory. See below
+for more about C<META6.json>. Each entity that the B<distribution> allows to be I<use'd> (or I<require'd> etc),
```
ugexe, line 540: One can “use” bin scripts, yet they do not belong in the provides section

doc/Language/modules.pod6
```diff
@@ -524,6 +524,113 @@ The directories in the C<RAKULIB> path will be searched for modules when
...
+L<Distributing modules>) it may be necessary to provide access to other resources, such as callable utilities.
+
+B<Raku> also allows for a single distribution to provide multiple C<compunit>s that can be I<use'd> (or I<require'd> etc).
+The information
+about a distribution is contained in the C<META6.json> file in the distribution's root directory. See below
+for more about C<META6.json>. Each entity that the B<distribution> allows to be I<use'd> (or I<require'd> etc), 
+also called a C<compunit>, is 
+placed in the C<provides> section of the C<META6.json> (as specified below). 
+
+It should also be noted that when
```
ugexe, line 544: This is false. Per s22 depends should be by module name (what is in provides). For backwards compatibility reasons though the distribution name is also allowed, although it is not the original design.

ugexe:
I left some more comments after this was rushed to be merged

JJ:
Sorry, I saw all of the issues resolved by @finanalyst. Your comments are
valuable; we'll use them to fix the text. Anyway, I'm not generating a new
set of documentation until this is fixed. Thanks.

ugexe:
To be clear an item being marked as resolved just means the original commenter made changes in the area that someone commented -- it does not in any way ensure correct changes have been made, or that the reviewer has seen the changes to approve them at all.

PR #3793, opened 2021-02-05
Remove unused Pod::Convenience method

coke:
Removed these methods (and their tests), ran ```make xtest``` - no failures. No references to this method elsewhere in the repo; able to do a documentable build. Not all the methods had tests that were removed, but all the tests were for removed methods, so the test file is gone.


coke:
While switching away from this lib entirely has been slow going, removing this function is an easy way to reduce the footprint.

coke:
update coming that will remove more methods...

JJ (approved):
Probably safe if we remove the whole module too...

JJ:
Thanks!

PR #3794, opened 2021-02-05
Largely rework the Object Tutorial

patrickbkr:
With the advent of `is built` all the common use-cases of object
construction are now covered by our attribute syntax. This means that the
need to fall back to using `BUILD` has lessened enormously. Given that
`BUILD` has a number of behaviours that are non-obvious and require
knowing and understanding them, it makes sense to consider `BUILD` the
last resort mechanism to use when all else fails and recommend using
automatic attribute initialization and `TWEAK` for everything else.
Communicating this clearly in our object tutorial is very valuable, it
shows newcomers the sane default first. We do not want to send them off in
the wrong direction as `BUILD` has proven to be a stumbling block for new
users numerous times.

(Probably not entirely complete) List of Changes:

- Try to be more structured by reordering things.
- Add more headings for easier navigation.
- Explain `is built`.
- Rework examples to not use `BUILD` anymore.
- Remove wrong explanation of the `Task` example stating one would need a
  `BUILD` when custom constructors are used.
- Remove in-depth explanation of `BUILD`. Add respective documentation for
  `TWEAK`. `BUILD` is still explained in detail in the Object Reference.
- State that `BUILD` is only intended for advanced use-cases.
- Remove `Hero` example which is made obsolete by `is built`.
- Simplify the `Str-with-ID` example and remove bits promoting making
  `TWEAK` a `method` instead of a `submethod`.
- Remove confusing bit about `my` vs `our` class variables.
- Rename `Static Fields?` heading to `Class variables`
- Replace the `Singleton` example with the `Str-with-ID` example as that
  shows the "variable shared by all instances" aspect better than a
  singleton (of which there usually only is one).

### DIFF
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -5,11 +5,16 @@
 =SUBTITLE A tutorial about creating and using classes in Raku
 
 X<|OOP>
+Raku provides a rich built-in syntax for defining and using classes. It makes writing classes
+expressive and short for most cases, but also provides mechanisms to cover
+the rare corner cases.
 
-Raku has a rich built-in syntax for defining and using classes.
-
-A default constructor allows the setting of attributes for the created
-object:
+Like many other languages the C<class> keyword is used to define a
+class. In general one uses the C<has> keyword to add
+attributes and the C<method> keyword to add behaviors to a class. It is rarely necessary to explicitly write
+a constructor, as the default construction mechanism
+automatically initializes attributes from the default constructors named parameters. The default constructor
+is called C<new>. The following example showcases this:
 
 =begin code
 class Point {
@@ -33,57 +38,80 @@ my $r = Rectangle.new(lower => Point.new(x => 0, y => 0),
 say $r.area(); # OUTPUT: «100␤»
 =end code
 
-In the two classes, the default constructor is being used. This
-constructor will use named parameters in its invocation: C«Point.new(x =>
-0, y => 0)».
+Don't get confused by the C<$.> and C<$!>. They are called twigils and will be explained further down.
 
-You can also provide your own constructor and C<BUILD> implementation.
-You need to do it for cases in which there are private attributes that
-might need to be populated during object construction, as in the
-example below:
+To add custom logic to the construction process, one can use the C<TWEAK> submethod.
 
 =begin code
-# Example taken from
-# https://medium.freecodecamp.org/a-short-overview-of-object-oriented-software-design-c7aa0a622c83
-class Hero {
-    has @!inventory;
-    has Str $.name;
-    submethod BUILD( :$name, :@inventory ) {
-        $!name = $name;
-        @!inventory = @inventory
-    }
+class Point {
+    has Int $.x;
+    has Int $.y;
 
-    method act {
-        return @!inventory.pick;
+    submethod TWEAK(:$log) {
+        say "Constructing Point ($!x|$!y)" if $log;
     }
 }
-my $hero = Hero.new(:name('Þor'),
-                    :inventory(['Mjölnir','Chariot','Bilskirnir']));
-say $hero.act;
+
+my $p = Point.new(x => 1, y => 2, :log);
+# OUTPUT: Constructing Point (1|2)
 =end code
 
-In this case, we I<encapsulate> the private attribute C<@!inventory>; but
-private instance variables cannot be set by the default constructor, which is
-why we add a C<BUILD> submethod that takes care of that.
+Just remember to always make C<TWEAK> a C<submethod> and not a normal C<method> N<If in a class hierarchy a class contains a C<TWEAK> method (declared as a C<method> instead of a C<submethod>) that method is inherited to its subclass and will thus be called twice during construction of the subclass! The effects are even worse with C<BUILD>, but that is best read about in the L<Object Construction Reference|/language/objects#Object_construction>.>.
 
-The following, more elaborate example, shows how a dependency handler might look
-in Raku.  It showcases custom constructors, private and public attributes,
-L<Submethod|/type/Submethod>s, methods, and various aspects of signatures. It's
-not a lot of code, and yet the result is interesting and useful.
+X<|DESTROY>
+X<|submethod DESTROY>
+
+C<DESTROY> is the submethod that gets called when an object is garbage
+collected. That is going to happen only if the runtime needs the memory, so we
+can't rely on when it's going to happen. In particular, it could happen in the
+middle of some running code in a thread, so we must take special care to not
+assume any context during that event. We can use it to close any kind of handles
+or supplies or delete temporary files that are no longer going to be used.
+
+=begin code
+my $in_destructor = 0;
+
+class Foo {
+    submethod DESTROY { $in_destructor++ }
+}
+
+my $foo;
+for 1 .. 6000 {
+    $foo = Foo.new();
+}
+
+say "DESTROY called $in_destructor times";
+=end code
+
+This might print something like C<DESTROY called 5701 times> and possibly only kicks
+in after we have stomped over former instances of C<Foo> a few thousand times. We also can't
+rely, on the order of destruction.
+
+Same as C<TWEAK>, make sure to always declare C<DESTROY> as a C<submethod>.
+
+=head1 The Task example
+
+X<|class>
+X<|classes>
+
+X<|state>
+X<|has>
+X<|classes,has>
+X<|behavior>
+X<|classes,behavior>
+
+The following piece of code implements a dependency handler. It's not a lot of code, and yet the result is interesting and useful.
+It will be used as an example throughout the following sections.
 
 =begin code
 class Task {
-    has      &!callback;
-    has Task @!dependencies;
+    has      &!callback     is built;
+    has Task @!dependencies is built;
     has Bool $.done;
 
-    # Normally doesn't need to be written
     method new(&callback, *@dependencies) {
         return self.bless(:&callback, :@dependencies);
     }
-
-    # BUILD is the equivalent of a constructor in other languages
-    submethod BUILD(:&!callback, :@!dependencies) { }
 
     method add-dependency(Task $dependency) {
         push @!dependencies, $dependency;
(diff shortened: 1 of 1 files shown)


Altai-man:
I wonder what policy do we have about headers (I am sure there is none, though).
Such changes will break local anchors.

While not commenting on the content itself, I suppose other languages have their docs versioned to avoid this trouble completely, right? E.g. `langfoo.com/docs/v1/foo#header` will never change even if `langfoo.co/docs/v1.1/foo` will have no `#header` anchor.

patrickbkr:
Oh. I was totally unaware of that problem. I guess we need a strategy first to tackle that problem. Do you consider the header problem a blocker for this PR?

JJ:
Re: versioning. The problem is that most languages release documentation and code in lockstep. That's not our case, simply because there's a large backlog; tooling does not allow that for the time being. @Altai-man 's proposal would definitely have that advantage, but then there would be other issues regarding the granularity of changes. Anyway, let's get to this specific PR.
In principle, large PRs are always a problem. There are many small issues, every one of which might or might not be a blocker; if there's some problem with one, the whole PR gets blocked. Also, they obviously need time to be reviewed properly.
But the second, and maybe major, issue, is the use of "we prefer" or "is obsolete". Unless blatantly incorrect, there's no good reason to totally hide a behavior, even more so when the text is probably there as the result of solving some issue, and definitely as the work of someone that's taken care of it. So in this case, and in general, I'm more for "build upon" than "rework", and insert something like "From 2021.01, it's considered a better practice..." TIMTOWDI and all that. That takes care of not eliminating headers, for instance, since it's only adding new headers what would be done.
That said, let me take a look at this to check if we can steer it into suggestions for better behavior instead of eliminating old ones.

doc/Language/classtut.pod6
```diff
@@ -192,69 +100,72 @@ X<|classes,has>
 X<|behavior>
 X<|classes,behavior>
 
-Raku, like many other languages, uses the C<class> keyword to define a
```
JJ, line 195: We've already used "class" in the first example, so not a lot of sense in the original inception. Still, the same thing is kept so...

doc/Language/classtut.pod6
```diff
@@ -192,69 +100,72 @@ X<|classes,has>
...
-through the C<has> keyword, and behaviors, introduced through the C<method>
-keyword.
-
-X<|type object>
-X<|DEFINITE>
-X<|.DEFINITE>
+The following piece of code implements a dependency handler. It's not a lot of code, and yet the result is interesting and useful.
+It will be used as an example throughout the following sections.
 
-Declaring a class creates a new I<type object> which, by default, is installed
```
JJ, line 206: Removing index anchors is definitely a no-no. They probably have moved elsewhere but, again, we need to approach stuff piecewise so that we can judge every issue by itself.

doc/Language/classtut.pod6
```diff
@@ -192,69 +100,72 @@ X<|classes,has>
...
 
 =for code
-has &!callback;
+has &!callback is built;
 
 X<|sigils,&>
 X<|twigils>
 X<|twigils,!>
-
+(The C<is built> is explained further down.)
```
JJ, line 168: Then use it in an example further down.

doc/Language/classtut.pod6
```diff
@@ -312,6 +227,35 @@ has Bool $.done is rw;
 The C<is rw> trait causes the generated accessor method to return a container
 so external code can modify the value of the attribute.
 
+=head2 C<is built> trait
+
+=for code
```
JJ, line 232: Again, use case following #1748. Users need to know what to use it for.

doc/Language/classtut.pod6
```diff
@@ -312,6 +227,35 @@ has Bool $.done is rw;
...
+has &!callback is built;
+
+X<|traits,is built>
+By default private attributes are not automatically set by the default constructor. (They are private after
+all.) In the above example we want to allow the user to provide the initial value but keep the attribute
```
JJ, line 237: The thing is that it's not totally clear to me you've explained what private attributes are. There's a bunch of thing that's deleted, I would need to see how this one looks like, but my impression is that it's not covered.

doc/Language/classtut.pod6
```diff
@@ -340,91 +284,40 @@ C<.an-attribute()> syntax. See also
...
 =begin code
 class Str-with-ID is Str {
     my $counter = 0;
-    our $hierarchy-counter = 0;
     has Str $.string;
-    has Int $.ID;
+    has Int $.ID is built(False);
 
-    method TWEAK() {
+    submethod TWEAK() {
```
JJ, line 302: Yep, that's a good one. We should definitely change that. I'll probably create an issue.

doc/Language/classtut.pod6
```diff
@@ -340,91 +284,40 @@ C<.an-attribute()> syntax. See also
...
-considered a better practice by Raku experts. Please check L<the section on
-submethods|/language/classtut#index-entry-OOP> for an alternative example on how
-to do this. Since C<TWEAK> is called in every object instantiation, it's
-incremented twice when creating objects of class C<Str-with-ID-and-tag>; this is
-a I<class hierarchy> variable that is shared by all subclasses of
-C<Str-with-ID>. Additionally, class variables declared with package scope are
-visible via their fully qualified name (FQN), while lexically scoped class
-variables are "private".
-
-=head1 Static fields?
```
JJ, line 389: Again, big elimination here. Probably not a good idea.

doc/Language/classtut.pod6
```diff
@@ -633,6 +521,41 @@ making dinner
 eating dinner. NOM!
 =end code
 
+=head2 A word on types
```
JJ, line 524: OK; I see it's been moved here. It probably makes sense. it creates a new header, though...

doc/Language/classtut.pod6
```diff
@@ -863,10 +786,10 @@ But there are other applications too. For instance, a routine that serializes
 objects to a bunch of bytes needs to know the attributes of that object,
 which it can find out via introspection.
 
-=head2 X<Overriding default gist method>
+=head2 X<Overriding C<gist>>
```
JJ, line 789: Change of headers, also index. Not a good thing.

JJ (changes requested):
In general, I would try and rework this into elimination of only what's an error, and use an incremental approach to add new things without removing old ones. Also, not change anchors or indexing (unless there's an issue that really says you should, in which case do just that)

patrickbkr:
@JJ Thanks for the detailed review! I'll try to work through it!

Independent of this specific PR I think we should talk about how to best approach changes to the documentation, because I think that we currently have different approaches. We should talk about that, so we don't keep annoying each other every time we do something together on the docs.

I think there are two topics:

Small vs large changes
------------------

When working on the documentation I often find it difficult to separate individual changes from one another. I think that's because free-form text is usually not as strictly structured and isolated than source code, so when working on a text it often happens that one change necessitates another (because texts are so interleaved). This has the effect, that when I do changes on a larger scale the text is at no point in the refactoring in a state where it is consistent and suitable for publication. I will try to aim for smaller changes and create more intermediate commits in the future. It probably makes sense to squash them before merging, as changes will probably not be atomic.

Removing content
-----------------

I think there are situations where removing content is appropriate.
- When content does not add any information that was not already explained in a more appropriate place.
- It does not fit into the overall structure or narrative of the document
- It is not easy to understand for the target audience of the document
- Its text to information ratio is too low
- It is factually wrong

In some cases it is possible and makes sense to improve the bit of content in question or move it to some other place where it fits in better. Other times that is not possible.

The topic of removing content is especially controversial with tutorials, as they need not be all-encompassing as a reference needs to be, but must balance multiple interests:
- ease of understanding
- provide enough of information to get the reader started in a topic
- be narrated well
- don't put too high requirements on the things the reader needs to know to understand the tutorial
- "build up" of information (explain things before they are used)

I'd like to hear your opinion on this.

Altai-man:
>Oh. I was totally unaware of that problem. I guess we need a strategy first to tackle that problem. Do you consider the header problem a blocker for this PR?

No, not really. I mean, we did it for literally years and while there were concerns about preserving URLs, the general flow is a bit LTA to stick to it. I am just re-confirming for myself there is an issue in this area.

JJ:
> @JJ Thanks for the detailed review! I'll try to work through it!
> 
> Independent of this specific PR I think we should talk about how to best approach changes to the documentation, because I think that we currently have different approaches. We should talk about that, so we don't keep annoying each other every time we do something together on the docs.
> 
> I think there are two topics:
> ## Small vs large changes
> 
> When working on the documentation I often find it difficult to separate individual changes from one another. I think that's because free-form text is usually not as strictly structured and isolated than source code, so when working on a text it often happens that one change necessitates another (because texts are so interleaved). This has the effect, that when I do changes on a larger scale the text is at no point in the refactoring in a state where it is consistent and suitable for publication. I will try to aim for smaller changes and create more intermediate commits in the future. It probably makes sense to squash them before merging, as changes will probably not be atomic.

That's probably a better approach, and it will allow everyone to pay more attention to them and review them properly.

> ## Removing content
> 
> I think there are situations where removing content is appropriate.
> 
>     * When content does not add any information that was not already explained in a more appropriate place.
> 

In principle, we're no the same page here. Incorrect, repeated, confusing documentation should probably removed. But a PR is not the place for that: first the issue, then a PR for that specific problem. Removing code, moving it around in a PR together with different things means that blocking it for improvements will be throwing away the baby with the bath water. So again atomic, issue-related changes makes everything easier.

>     * It does not fit into the overall structure or narrative of the document
> 
>     * It is not easy to understand for the target audience of the document
> 

Again, these are issues that probably need to be raised independently of a PR, and previous to that. It might include headers, or anchors, or some things that we need to move elsewhere, probably.

>     * Its text to information ratio is too low
> 
>     * It is factually wrong

No doubt is something is wrong it needs to be fixed. Removed, that's a different thing altogether.
> 
> 
> In some cases it is possible and makes sense to improve the bit of content in question or move it to some other place where it fits in better. Other times that is not possible.
> 

Yep, but discussing that in the context of the change itself does not leave a lot of room for maneuvering.

> The topic of removing content is especially controversial with tutorials, as they need not be all-encompassing as a reference needs to be, but must balance multiple interests:
> 
>     * ease of understanding
> 
>     * provide enough of information to get the reader started in a topic
> 
>     * be narrated well

But those are general principles for the documentation we can all agree on. But the problem is that we need to agree how that applies to the original piece of documentation it refers to. That's why I favor an incremental approach. First add the new thing, then check if there's something else it can replace.

patrickbkr:
I am reworking this. Thus I turned it into a draft.

finanalyst:
> > Oh. I was totally unaware of that problem. I guess we need a strategy first to tackle that problem. Do you consider the header problem a blocker for this PR?
> 
> No, not really. I mean, we did it for literally years and while there were concerns about preserving URLs, the general flow is a bit LTA to stick to it. I am just re-confirming for myself there is an issue in this area.

I think that the 'header' issue is a part of a more general issue of what I suggest we call 'target consistency'. Namely the within a collection of content files, there are links to targets in other content files. Since all headers are by default targets (all current pod renderers automatically include an 'id' or 'name' attribute that can be a part of a `<a href="...">` in another file), changing the name of a header will cause links in other files to fail.  But headers are not the only target creators, there are also X<> codes.

It is not cost beneificial to set up an infrastructure to keep track of target/link connections. However, it is possible to create tests that check links across the entire collections.

In this way, when a change in one file (editing a header) causes a link failure, the test will flag a link failure. I have implemented such a test with my Collection tool chain.  The implementation is not yet good because I have too many false positives.

But even tracking that down has led to some undocumented behaviours of Pod::To::HTML and the assumptions of early documentation writers.

Altai-man:
>In this way, when a change in one file (editing a header) causes a link failure, the test will flag a link failure

The issue I am concerned with is that not only one parts of the docs link other parts of the docs (and we control both, so surely can test and fix issues on our side), but that any third-party source like a blog post or a search engine or whatever can (and should) point to the doc pages which will become broken with the flow of time because people rewrite paragraphs, change headers etc. And we cannot test the whole Web and fix people's content.

In the world where Raku has a good documentation web resource, URLs must be set in stone forever and ever. The way to do it is to start real versioning: one has to call the current version, say, 1.0, and all the upward changes go into master/upstream/whatever prefix. Then one day this prefix is frozen into 1.1 and the cycle continues.

It is a schema everyone is using except us (because we lacked tooling back then and lack volunteers to do the heavy-lifting now).

Absence of this mechanism also leads to various ad-hoc solutions like "This and that was working since release 2018.11".

I think https://github.com/Raku/doc/issues/2690 kinda relates to this. The https://github.com/Altai-man/docs.raku.org/issues/6 was intended as a starting point for this, since it is not that hard to implement dynamic serving various versions with Cro, but no volunteers fancy.

patrickbkr:
Superseded by https://github.com/Raku/doc/pull/4047

PR #3801, opened 2021-02-11
Use absolute path to routine's documentation

stoned:
When a type's routine documentation is rendered as a part of
'/routine/RoutineName', a link to the documentation of another routine for
the same type needs to provided with an absolute path ('/type/Type#...').


JJ (approved):
Thanks!

Issue #3800, opened 2021-02-11
Adding a new way of entering UNICODE characters in Linux to the wiki


flo-pe:
Hi everyone,

I just made a little set up on my system to entering unicode characters via copy paste.
To me it is a lot more ergonomical that typing Ctrl-V u + hexadecimals

Basically, I just execute this one line :
echo -e "«\n»\n∈\n∉\n∋\n∌\n≡\n≢\n⊆\n⊈\n⊂\n⊄\n⊇\n⊉\n⊃\n⊅\n∪\n∩\n∖\n∞\n≠\n≤\n≥" | dmenu -i -l 30 -fn "-xos4-terminus-medium-r-*-*-10-*" | tr -d "\n" | { xsel -ib; xsel -ip; }

it makes appear a drop-down menu where you can select the unicode character that you want, and it copies it to the clipboard, after which you can simply paste it in you text editor

you can have dmenu by installing the "suckless-tools" package on Debian and Ubuntu, or "dmenu" or Archlinux
or you can compile it from source from here : 
https://tools.suckless.org/dmenu/

the part -fn "-xos4-terminus-medium-r-*-*-10-*" is optional, it simply set a bigger font, but you may not have this particular font intalled on you system

and to be truly ergonomical, I also use the software sxhkd (Simple X hotkey daemon) also available in the repositories of Debian, Ubuntu and Archlinux.

this program allows you to configure keyboard shortcuts to execute executables, scripts, or direct terminal commands

you configure it by editing the text file at ~/.config/sxhkd/sxhkdrc
im my case, the relevant part I added is this : 

ctrl + bracketright
	echo -e "«\n»\n∈\n∉\n∋\n∌\n≡\n≢\n⊆\n⊈\n⊂\n⊄\n⊇\n⊉\n⊃\n⊅\n∪\n∩\n∖\n∞\n≠\n≤\n≥" | dmenu -i -l 30 -fn "-xos4-terminus-medium-r-*-*-10-*" | tr -d "\n" | { xsel -ib; xsel -ip; }

JJ:
Thanks a lot! We'll try to add it to the documentation.

flo-pe:
I'm glad to help !
Also, there should be a tabuation before echo in the 2 lines of sxhkdrc. The tabulation got deleted when I copy pasted it

flo-pe:
You can also drop the -i flag which stands for ignore case, it has no use here

JJ:
Please feel free to create a PR with this content.

[...]

PR #3804, opened 2021-02-15
Input Unicode characters with two-key combinations

frithnanth:
Please review and advice: I had some doubts about adding a 55 line-long example.

## The problem

Input of Unicode characters can be simplified by using dead keys and XCompose

## Solution provided

This PR shows a way to add several two-key combinations to a user's ~/.XCompose file in order to let them enter easily all the relevant Unicode characters used by Raku.


JJ (approved):
Thanks!

JJ:
Refs: #3800

PR #3805, opened 2021-02-15
Programs/03-Environment... remove F<

finanalyst:
## The problem

`F<...>` is not a standard FormatCode, either in the current Pod6 documentation, or in Damian Conway's original spec. This is the only Raku documentation content file to contain `F<...>`. The current standard HTML renderer just stringifies `F< ... >` to `F< ...>`, which is quite ugly imho. 

## Solution provided

I have replaced `F<...>` with `B<...>`, which is a standard FormatCode.


JJ:
Thanks!

PR #3821, opened 2021-03-18
Clean up unicode operator docs

moon-chilled:


moon-chilled:
I chose ascii as the ‘canonical’ form for the prototypes, somewhat arbitrarily; would unicode be better?

doc/Language/operators.pod6
```diff
@@ -1533,11 +1533,11 @@ element
 say [~] Blob.new([3,4,5]);     # OUTPUT: «Blob:0x<03 04 05>␤»
 say [~] 1|2;                   # OUTPUT: «any(1, 2)␤»
 
-=head2 infix C«∘»
+=head2 infix C<o>, infix C«∘»
```
JJ, line 1536: This might be a problem with indexing...

doc/Language/operators.pod6
```diff
@@ -1619,12 +1618,9 @@ of that element over all arguments).
     say <a a b c a> (&) bag(<a a b c c>); # OUTPUT: «Bag(a(2) b c)␤»
     =end code
 
-C«∩» is equivalent to C«(&)», at codepoint U+2229 (INTERSECTION).
-
 =head2 infix C«(.)», infix C«⊍»
 
     multi sub infix:<(.)>(**@p)
-    multi sub infix:<⊍>(**@p)
```
JJ, line 1627: Don't get this... Is there any reason why you delete it?

doc/Language/operators.pod6
```diff
@@ -2075,7 +2055,7 @@ X<Numeric equality operator>.
 Coerces both arguments to L<Numeric|/type/Numeric> (if necessary); returns C<True>
 if they are equal.
 
-=head2 infix C«!=»
+=head2 infix C«!=», infix C«≠»
```
JJ, line 2058: OK, I see what's being done here, using a coherent header. The problem is that I'm not sure the initial ones were correct. I'll have to check the whole indexing thing to see what we have, but we might have to change the whole thing. =head2 indexes by default if it finds an "infix", so we need to see what happens here actually.

doc/Language/operators.pod6
```diff
@@ -1775,8 +1757,6 @@ the result is a new C<Bag> (or C<Mix>).
     say <a b> ⊖ bag(<b c>); # OUTPUT: «Bag(a c)␤»
     =end code
 
-C«⊖» is equivalent to C«(^)», at codepoint U+2296 (CIRCLED MINUS).
-
```
JJ, line 1779: Again, why? Is that incorrect? Is there any  issue that complains about it?

doc/Language/operators.pod6
```diff
@@ -2131,7 +2099,7 @@ X<Numeric greater than operator>.
 Coerces both arguments to L<Real|/type/Real> (if necessary); returns C<True> if the first argument
 is larger than the second.
 
-=head2 infix C«>=»
+=head2 infix C«>=», infix C«≥»
```
JJ, line 2102: OK, I see what you're doing, but if you add this here, you'll need to add the definition(s) below. Or maybe eliminate them...

JJ (approved):
OK, took a while, but I think I got your intent. It might have helped if you had clarified it in the PR body, or if you had opened an issue indicating the problem here, which I see clearly now.

PR #3832, opened 2021-03-25
Undocumented L<> behaviour

finanalyst:
The `Pod::To::HTML` renderer automatically translates the token `::` in a L<> to `/`

In Perl sibling languages, a Module called `Some::Module` resides in `lib/Some/Module`, so the token `::` is a synonym for `/` or directory marker.

This assumption has been carried unremarked into Raku, even though there is no need for directory markers to be followed. 

In VERY many Pod6 files documentation authors have used the notation `L<some reference|type/X::Inheritance::NotComposed>` to refer to the file `doc/Type/X/Inheritance/NotComposed.pod6`.

Since there tens of such references, changing the behaviour is not worth the effort.

Consequently, the behaviour should simply be documented.

### DIFF
--- a/doc/Language/pod.pod6
+++ b/doc/Language/pod.pod6
@@ -561,6 +561,9 @@ Comments L<#Comments>
 
 L<Comments|#Comments>
 
+Alongside the normal way of describing a path, viz. C< L<Some reference|path/to/filename> >, it is also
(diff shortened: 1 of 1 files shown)


JJ (approved):
Thanks!

PR #3836, opened 2021-03-26
issue #3883 remove 404

finanalyst:
P5rewinddir is referenced, but the module no longer exists at all. I cannot find a substitute.
I see no reason why we don't just state the Raku does not implement the functionality.

### DIFF
--- a/doc/Language/5to6-perlfunc.pod6
+++ b/doc/Language/5to6-perlfunc.pod6
@@ -1619,10 +1619,6 @@ X<| rewinddir - perlfunc>
 
 Not supported in Raku.
 
-The Raku ecosystem has a module L<C<P5rewinddir>|https://modules.raku.org/dist/P5rewinddir>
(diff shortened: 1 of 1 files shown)


doc/Language/5to6-perlfunc.pod6
```diff
@@ -1619,10 +1619,6 @@ X<| rewinddir - perlfunc>
 
 Not supported in Raku.
 
-The Raku ecosystem has a module L<C<P5rewinddir>|https://modules.raku.org/dist/P5rewinddir>
-which exports a C<rewinddir> function that mimics the original Perl
-behavior as much as possible.
-
```
JJ, line 1625: This is now part of [p5opendir](https://modules.raku.org/dist/P5opendir). Can you please change the link to point to that?

finanalyst: @JJ changed to P5opendir as requested.

JJ (approved):

PR #3837, opened 2021-03-26
#3883 remove non-existent host

finanalyst:
The domain `rakudo.party` no longer exists. Since `Undercover` bot only differs from SourceBaby because it points to wtf.rakudo.party,
this link is pointless. Zoffix, perhaps could point to a different domain???


JJ:
Zoffix is, unfortunately, out of the picture, and that site has not been updated. Closest thing is Blin, I guess.

JJ (approved):

PR #3838, opened 2021-03-26
Issue #3883 remove 404 replace with working

finanalyst:
- Replace 'posts' with 'news', which is a working page on rakudo.org
- correct download path reference
- reference 'star/etc/modules.txt' which exists. (Possibly better to eliminate this link altogether, but being minimal here)


JJ (approved):

JJ:
:100:

PR #3847, opened 2021-03-31
Reformat and better define 'daycount' and its relation to MJD

tbrowder:
## The problem

The existing description is a bit incomplete.

## Solution provided

Added more information about the method (a prelude to an upcoming Rakudo PR).


JJ (approved):
Good use of the Oxford comma :-)

PR #3851, opened 2021-04-01
Correct the description of the epoch (posix) DateTime new method.

tbrowder:
## The problem

The description is wrong, the new method for epoch (posix) takes a Numeric:D argument, not an Int.

## Solution provided

Corrected the description


JJ:
But has this already been released?

JJ (approved):
Looks fine, but I want to make sure this is something that is in released Rakudo *now*

jonathanstowe:
> Looks fine, but I want to make sure this is something that is in released Rakudo _now_

It is:

```
[jonathan@menenius katikati]$ raku -e 'say DateTime.new(13456.67)'
1970-01-01T03:44:16.670000Z
```

jonathanstowe (approved):

Issue #3841, opened 2021-03-28
Non-existent links in existing files


finanalyst:
## Links in Primary content files without targets
A primary content file is a pod6 document under the `doc/` directory in this repository.
A link is specified like `L<text|path/to/file#target>`

So the bug is when there is no `target` in `path/to/file` when `path/to/file.html` exists.

For a complete list of these bugs, see [Error Report](https://raku.finanalyst.org/error-report.html).

Some of these bugs seem to have existed from when the Content files were first written, and they are simple typos. The reason they are not picked up is that browsers fail gracefully if the file exists. 

The expected behaviour is for the browser to open the file and move the page so that the target url is at the top (or the middle) of the visible page. Graceful failure leaves the file at the top of the page.

Types of errors: 
- Look at the document `language/io-guide`. You will see that the link texts `.getc` and `slurp` are listed and the underlying urls are shown as `method_getc` and `routine_slurp`. In fact, the references should be `routine_getc` and `method_slurp`.
- In document `language/system` the link text `<code>@*ARGS</code>` (which is browser rendered as `@*ARGS` has an underlying reference to `%24%2AARGS`, which is converted by
a browser as `$*ARGS`. The documentation writer used `%24` instead of `%40`.

There are some *false negatives* as my test regexes are incomplete (the relevant code is in the [link-error-test plugin in the Collection-Raku-Documentation](https://github.com/finanalyst/collection-raku-documentation/tree/master/Website/plugins/link-error-test) distribution). 

For example, there is a reference to `empty set` in `language/math` with underlying url `term_%E2%88%85`. This refers to the target `term_∅`, which exists.  `%E2%88%85` is a combined unicode reference that isn't correctly handled by `link-error-test`. 

However, I think false negatives are now rare.

## Suggestions

These errors can be gradually identified and removed. Since such corrections have no knock-on or side effects, I suggest that they could be assigned to newcomers in Squashathons. 

I shall work on a few and (unless anyone objects) I shall commit corrections without review.

JJ:
This should probably be a test; probably even a normal test, not an xt.

Altai-man:
This is an absolutely awesome suggestion!

I think to effectively approach the "find 404 and fix them" we have to first take a look at how actual anchors are created. Anchor ID generation is tied closely to "metadata" of the `X<>` as well as generated by headers, and so any even sensible changed in this direction will lead to fallout.
See https://github.com/Raku/problem-solving/pull/271/files for further reading.

[...]

PR #3856, opened 2021-04-03
related to #3841

finanalyst:
This is not an obvious mistake. From the link it seems to point at a different file. Since this is a judgment call, I'm flagging my correction for review.


JJ (commented):
Neither link makes a lot of sense. If there's no definition of `new` in that specific class, the first is not linking to what it says, and the second either. So my money is on eliminating the link, or clarifying that there's no specific `new` and that it would call the superclass by default.

finanalyst:
I agree that eliminating the link is the best. Anyone reading this part of the documentation will understand a call to .new, and the link provided would not have helped someone who did not.

@JJ  please merge if you're OK with this.

JJ (approved):

PR #3857, opened 2021-04-03
Related #3841

finanalyst:
Removing B<..> **inside** a X<> for $*COLLATION .  
The problem is that there is no specification for a corresponding `L<>` in another file. As the present, the link author has to know how the target will be rendered. So `B<>` could be rendered as `<b> ... </b>` or `<strong>...</strong>` in HTML. **BUT** if this were to be rendered in some other output format (pdf? epub? latex?), the link would fail. This has already happened in `Language/Operators`, where the link author did not include `<strong>...</strong>` in the link.

So, I have removed B<> inside the X<>, and simplified the X<>, and corrected the link in Operators.
If this has a knock-on effect in other documents, it will be picked up in `Collection` tests when the documentation files are re-rendered.

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -1979,7 +1979,7 @@ aspects of the particular character like capitalization.
 
 The main difference between C<coll> and C<unicmp> is that the behavior of the
 former can be changed by the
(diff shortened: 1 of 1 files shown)


JJ (approved):
Don't see how the body of the PR matches it, but this seems correct.

PR #3859, opened 2021-04-03
see #3841

finanalyst:
removed link altogether. Could not determine the intention of the link author.
Judgement call, so requesting review.

### DIFF
--- a/doc/Language/packages.pod6
+++ b/doc/Language/packages.pod6
@@ -141,7 +141,7 @@ the setting's outermost scope, use C<UNIT::OUTER> instead.
 =head2 Interpolating into names
 X<|interpolating into names>
 
-X<|::()> You may L<interpolate|/language/packages#Interpolating> a string into a
+X<|::()> You may interpolate a string into a
 package or variable name using C<::($expr)> where you'd ordinarily put a package
 or variable name. The string is allowed to contain additional instances of
 C<::>, which will be interpreted as package nesting. You may only interpolate


Altai-man (changes requested):
I think it wants to be a link to the section where we explain the "interpolation" term instead of just being removed.

doc/Language/packages.pod6
```diff
@@ -141,7 +141,7 @@ the setting's outermost scope, use C<UNIT::OUTER> instead.
 =head2 Interpolating into names
 X<|interpolating into names>
 
-X<|::()> You may L<interpolate|/language/packages#Interpolating> a string into a
+X<|::()> You may interpolate a string into a
```
JJ, line 144: Wouldn't this rater be [this link](/language/packages#Interpolating_into_names)? It seems to have changed is all.

Altai-man: No, the link you are suggesting is the exact section where the link in question is situated at, which means in a first paragraph of a section you link to the same section.
It in general talks about interpolating to get packages, but I think just the term "interpolating" in general should be described somewhere.

JJ: Hum, true... But then that's where it pointed at before. So it makes sense to be removed.

JJ (approved):

doc/Language/packages.pod6
Altai-man: No, it makes sense to find where we describe "interpolation" as a general term and update the link to point there.

JJ: There's no such thing right now. There's interpolation in the glossary, but that mainly refers to string. We might open another issue, which, when solved, can be pointed at.

JJ: ... and then where it would make the most sense to define it is around this same link, so no link would be necessary. Really, I think that eliminating it is OK.

Altai-man: https://en.wikipedia.org/wiki/String_interpolation ?

JJ: You mean outside the docs? OK, that might be an alternative. Again, judgement call as Richard says

Altai-man: Yes, something outside the docs as we don't have a section for "interpolation" it seems, only "variable interpolation" and it does not explain the general concept.
https://github.com/Raku/doc/issues/2588 ticket (still opened due to it being too vague, alas) in a confusing, but still a way says that terms like "interpolation" should be explained somewhere and linked to, because not everyone know what does it mean and the docs intent is to serve as an educational material and/or a reference.
Hence just removing the URL will make the situation worse.

finanalyst:
I seem to have corrected this error twice and caused a conflict. I have included @Altai-man wikipedia link. Sorry @JJ if this is n't what you intended.

PR #3863, opened 2021-04-04
Reference error to .^

finanalyst:
Part of  #3841 . But I'm flagging this error because the orginal documentation calls `.^` a postfix op, but in `
language/operators` it is called a metamethod. So I have also change 'postfix' to 'metamethod' to be consistent.


JJ (approved):
Thanks!

PR #3864, opened 2021-04-04
interpolate using wikipedia link (resubmission)

finanalyst:
There was discussion of this and I caused a conflict by editing twice. Here it is again


Altai-man (approved):
This is hopefully better for the reader.

JJ (approved):

PR #3865, opened 2021-04-05
Remove link from List #3841

finanalyst:
This is another false link that probably needs removing.
There is no `routine_Str`or `method_Str` documented in `Any`. I could not find a single reference to `.Str` that I could replace the reference with.
Removing the link does not seem to me to be important, but it probably should be an issue that `Str` is not documented in a single place. Strings and stringification are important enough to need a separate document.

### DIFF
--- a/doc/Type/List.pod6
+++ b/doc/Type/List.pod6
@@ -277,7 +277,7 @@ Defined as:
     method join(List:D: $separator = "")
 
 Treats the elements of the list as strings by calling
-L<C<.Str>|/type/Any#routine_Str> on each of them, interleaves them with
+C<.Str> on each of them, interleaves them with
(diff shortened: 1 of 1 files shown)


doc/Type/List.pod6
```diff
@@ -277,7 +277,7 @@ Defined as:
     method join(List:D: $separator = "")
 
 Treats the elements of the list as strings by calling
-L<C<.Str>|/type/Any#routine_Str> on each of them, interleaves them with
```
JJ, line 280: What about changing it to `/type/Mu#method_Str`? It probably makes more sense. It's also pretty much the correct method, taking into account any object will subclass `Mu`

PR #3866, opened 2021-04-05
Insert missing target into Proc #3841

finanalyst:
`Type/Independent-routines` contains a reference to `Proc#sub_run` but `Type/Proc` does not contain any target for `run`.

So I have inserted a X<> around the first explanation of `run` in Proc. This provides a target for Independent-routines (also opening separate PR with new target)

BUT, for consistency, `run` should be listed as sub along with `new` and `shell`. Also relationship with new should be illustrated. This is not done in this PR.

### DIFF
--- a/doc/Type/Proc.pod6
+++ b/doc/Type/Proc.pod6
@@ -8,7 +8,7 @@
 
 C<Proc> is a representation of an invocation of an external
 process. It provides access to the input, output and error stream as well as
-the exit code. It is typically created through the C<run> subroutine:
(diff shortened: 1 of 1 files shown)


finanalyst:
Change of suggestion: Independent-routines already has much better explanation of `run`. So the reference in `Independent-routines`, in the section for `shell`, should point internally (within Independent-routines). I have edited Independent-routine accordingly.

So this PR can be either approved or rejected.

JJ (approved):

PR #3871, opened 2021-04-21
Update link in Chinese translation

wukgdu:
## The problem

Wrong link in Chinese translation .md

## Solution provided

Use link instead of raw text


JJ (approved):

JJ:
Thanks!

Issue #3744, opened 2020-12-21
Checklist for 2020.12


JJ:
Corresponds to [changes and updates done in the 2020.12 release](https://github.com/rakudo/rakudo/blob/master/docs/announce/2020.12.md)

#  Removals:
 * [ ] Deprecated method candidates (`subbuf(Any:U)` on `Buf`, `chdir(Str(), :$!test)` on `IO::Path`)
     and `indir(IO() $path, &what, :$test!)` subroutine candidate that were
     throwing an exception instead of a deprecation warning for a long time
     were removed from CORE -[9040318c](https://github.com/rakudo/rakudo/commit/9040318c) 
#  Additions:
 * [ ] Add new method `deterministic` to `Iterator` role -[87fc041f](https://github.com/rakudo/rakudo/commit/87fc041f)-[b83b1b34](https://github.com/rakudo/rakudo/commit/b83b1b34)-[b63c0e02](https://github.com/rakudo/rakudo/commit/b63c0e02)-[c37a88e8](https://github.com/rakudo/rakudo/commit/c37a88e8)-[96285afb](https://github.com/rakudo/rakudo/commit/96285afb)
 * [ ] Introduce `%*SUB-MAIN-OPTS<coerce-allomorphs-to>` setting -[bd5eba4b](https://github.com/rakudo/rakudo/commit/bd5eba4b)-[49eecd68](https://github.com/rakudo/rakudo/commit/49eecd68)
 * [ ] Add a new candidate to `spurt` routine and method.
     It does not have an argument taking content to write,
     making it similar to the `touch` utility -[f2ea0a65](https://github.com/rakudo/rakudo/commit/f2ea0a65)
 * [ ] Add `:emit-timed` to `Supply.batch` method -[492651ea](https://github.com/rakudo/rakudo/commit/492651ea)
 * [ ] Make `is DEPRECATED` trait introspectable on `Routine` instances -[0d1c8a8e](https://github.com/rakudo/rakudo/commit/0d1c8a8e) 
#  Changes:
 * [ ] Improve output of `Attribute` documentation when rendered with `Pod::To::Text` -[a0a8a515](https://github.com/rakudo/rakudo/commit/a0a8a515)
 * [ ] Increase sensitivity of `Supply.batch(:seconds)` x 1000 -[aecfc9b3](https://github.com/rakudo/rakudo/commit/aecfc9b3)
[...]

PR #3878, opened 2021-04-27
Note ms granularity for Supply.batch ref #3744

stoned:


doc/Type/Supply.pod6
```diff
@@ -371,10 +371,15 @@ C<&mapper> and emits it to the new supply.
...
-the number of elements in the batch (using C<:elems>) or the maximum number of
-seconds (using the C<:seconds>) or both.  Any remaining values are emitted in
+the number of elements in the batch (using C<:elems>) or a duration
+(using C<:seconds>) or both.  Any remaining values are emitted in
 a final batch when the supply is done.
 
+B<Note>: Since Rakudo version 2020.12, the C<:seconds> parameter has
+a millisecond granularity: for example a 1 millisecond duration could
+be specified as C<:seconds(0.001)>. Before Rakudo version 2020.12,
+the C<:seconds> parameter has a second granularity.
```
coke, line 381: I would say that it "had" the granularity.

coke (changes requested):

doc/Type/Supply.pod6
stoned: Sure. Done !

PR #3879, opened 2021-05-01
Correct a link to Hash

frithnanth:
Hash → /type/Hash

## The problem

A link to the Hash class page generates a 404 error

### DIFF
--- a/doc/Type/Map.pod6
+++ b/doc/Type/Map.pod6
@@ -44,7 +44,7 @@ a C<Map> has been initialized:
    $m{ 'c' } = 'foo'; # WRONG!
                       # Cannot modify an immutable Str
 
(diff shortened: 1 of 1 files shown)


doc/Type/Map.pod6
```diff
@@ -44,7 +44,7 @@ a C<Map> has been initialized:
    $m{ 'c' } = 'foo'; # WRONG!
                       # Cannot modify an immutable Str
 
-Unlike its mutable companion type L<Hash>, Map cannot be parameterized by key or value types.
+Unlike its mutable companion type L<Hash|/type/Hash>, Map cannot be parameterized by key or value types.
```
JJ, line 47: This is fine, but while you're at it, maybe say *a* Map or Map*s*?

frithnanth: Right!

JJ (approved):

PR #3880, opened 2021-05-04
Search/test for the word 'boolean', ref #3517

stoned:
... as it should be written 'Boolean'.

This word variant test is all about case, thus the %variants hash now
contains Regex values with the case-insensitive adverb when applicable.


coke:
I have no opinion on b vs. B here, but this PR does it the right way, approved.

coke (approved):

PR #3884, opened 2021-05-18
minor fix: code quote within code quote

kybr:
## The problem

<...> was code quoted twice

## Solution provided

removed the in-line quoting

### DIFF
--- a/doc/Language/5to6-perlfunc.pod6
+++ b/doc/Language/5to6-perlfunc.pod6
@@ -1441,7 +1441,7 @@ These survive the transition to Raku. Some notes:
 =for code :lang<perl>
 q/.../;  # is still equivalent to using single quotes.
 qq/.../; # is still equivalent to using double quotes.
-qw/.../; # is more commonly rendered as <...> in Raku.
+qw/.../; # is more commonly written as <...> in Raku.
 
 There are some added quoting constructs and equivalents, as explained at
 L<quoting|/language/quoting>.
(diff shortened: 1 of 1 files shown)


coke:
The C<> here is intended to show the text

```
<...>
```

as code, which it's doing. the brackets are part of the text to display, corresponding to the ```qw``` in the Perl 5 code.

coke:
I would recommend changing "rendered" here, actually, because I think that is a source of confusion. "written as" is probably a better choice.

kybr:
Ok. Perhaps I am wrong. But this is what I see on the [Perl to Raku guide - functions](https://docs.raku.org/language/5to6-perlfunc#quoting):

```
q/.../;  # is still equivalent to using single quotes. 
qq/.../; # is still equivalent to using double quotes. 
qw/.../; # is more commonly rendered as C<< <...> >> in Raku. 
```

Is the reader supposed to see the `C<< >>` part? I don't think so. Because the whole block above is itself is under a `=for code :lang<perl>` thing, I think the `C<< <...> >>` renders literally. I checked on `#raku` to see if anyone else thought it might be a mistake and [at least one person thought it was](https://colabti.org/irclogger/irclogger_log/raku?date=2021-05-18#l321).

coke:
Ah, I assumed the pod was rendering and did not verify. Good catch, in that case you’re 100% right!

kybr:
Cool! And, "written as" is better. I'll amend the PR when I get back to a computer.

JJ (approved):
Thanks!

Issue #3883, opened 2021-05-10
Need a complete example of a callable sort routine


tbrowder:
The docs show an example, using infix, that a new user probably would not use.

A better example would be something that encapsulated this:

    sort({$^a.chars cmp $^b.chars}) # sorts by length of word, shortest first

I would provide the complete example, including how to use it for sorting a list of words, but I don't know how, hence this issue.

# UPDATE

Given a list of words, sort by word length:

    my @w = <abcde zero>
    say @w.sort;
    ["abcde", "zero"]
    @w .= sort({$^a.chars cmp $^b.chars});
    say @w.raku
    ["zero", "abcde"]

Util:
I can provide any such example, but I am unclear on exactly what is being requested.

There are four classes whose .sort documentation appears on page https://docs.raku.org/routine/sort :
    Map, Supply, List, Any
and two classes mention `infix:` :
    List, Any
, yet I am still confused as to which class you are targeting, much less exactly which line, or the problem with the line.

I agree that the sort docs need improving, and hope to soon add a Sorting Tutorial page similar to https://docs.raku.org/language/grammar_tutorial .
In the meantime, would you clarify exactly which part of the sort docs this issue seeks to address?

tbrowder:
Start with the Any class, but a List might make a better example. Use my sort requirement as I showed above, make it a "callable," and apply it to sorting a list of words.

Util:
Hi, Tom!

(All written before your code update in the OP.)
I think I understand now; let me know if I am not spot-on.

Note: I changed `cmp` to `<=>`, as a better signal to a reader of the numeric nature of the comparison. I don't know if it is any more efficient, but why signal to a reader or to the compiler that we want Raku to figure out numeric-vs-alpha, when we can just say "numeric" explicitly?

Long version:

    # Sort comparator.
    # Words of the same length will remain in their original order.
    sub by_character_count ( $a, $b ) {
        return $a.chars <=> $b.chars;
    }
    my @play = <Let us impart what we have seen tonight unto young Hamlet>;
    say @play.sort(&by_character_count);
    # OUTPUT: «(us we Let what have seen unto young impart Hamlet tonight)␤»

Short version:

    sub by_character_count { $^a.chars <=> $^b.chars }
    say <Let us impart what we have seen tonight unto young Hamlet>.sort(&by_character_count);
    # OUTPUT: «(us we Let what have seen unto young impart Hamlet tonight)␤»

Are either of those close to what you had in mind?
[...]

PR #3885, opened 2021-05-19
Add complete example of a callable sort routine #3883

Util:
## The problem

The only example of a named sort routine is using the trick of referencing a operator by its "name".
No example is provided of creating a new separate sort routine, and passing it to .sort .

## Solution provided

A minimal code example is added: a short named sub, and usage of that sub on data whose results highlights the sorting *and* the stability,  since ties are included in the data.


tbrowder:
Looks good to me (although I think kebab case is more "Rakui"). Thanks.

JJ (approved):
Cool!

JJ:
> Looks good to me (although I think kebab case is more "Rakui"). Thanks.

Totally right. Probably, as long as we don't draw attention to this fact, TIMTOWDI. I'll wait a bit to merge, just in case @Util wants to follow that advice.

PR #3886, opened 2021-05-19
 Move from freenode to libera.chat

raydiak:
Made a PR instead of committing directly so someone more authoritative can pull the trigger when we decide the time is right, as the Whateverable bots (and probably some other things and people) aren't yet migrated at the time of this writing

### DIFF
--- a/assets/documentable.json
+++ b/assets/documentable.json
@@ -1,7 +1,7 @@
 {
   "title-page": "Raku Documentation",
   "pod-root-path": "https://github.com/Raku/doc/blob/master/doc",
-  "irc-link": "https://webchat.freenode.net/?channels=#raku",
+  "irc-link": "https://kiwiirc.com/client/irc.libera.chat/#raku",
   "kinds": [
     {
       "kind": "language",
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -108,7 +108,7 @@ function setup_debug_mode(){
                 console.info("checking for dead links");
 
                 function report_broken_link(url) {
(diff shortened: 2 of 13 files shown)


JJ (approved):
Thanks a lot for the effort. If anything is revealed by this, is the problems with putting communication channels in so many places. So I understand this change is permanent, right?

raydiak:
Happy to help. Agreed, freenode did seem to be far more prolifically embedded in our docs and website than I expected.

**If** this move happens, yes it would be safe to assume it's permanent. I emphasize "if" because I am currently unaware of an official decision to move the Raku IRC channels. However, every mention I've seen of it in our community seems to point in that direction, and many people including several core developers are already on libre.chat and have established our channels there. From the discussions I've seen both in and outside of the Raku community, the move seemed likely enough to be worth getting this part of the work out of the way in advance. I'll update this PR and Raku/raku.org#164 if we decide to move elsewhere.

lizmat (approved):

lizmat:
Hmmm... isn't a merge a bit premature ?   My approval did not mean: do it now  :-)

raydiak:
@JJ We should turn this live ASAP. The move is official, and freenode's new staff has forcibly seized control of our channels and disposed of them. See https://www.reddit.com/r/rakulang/comments/nlbt0r/freenode_staff_has_nuked_the_raku_channels/ and https://news.ycombinator.com/item?id=27286628

JJ:
Already deployed

raydiak:
My mistake, I apparently still had an old version in my browser cache. Thanks!

PR #3892, opened 2021-05-30
Document Str.samecase

stoned:
... and state that Cool.samecase uses it.


JJ (approved):
Thanks!
(And sorry about the delay)

Issue #3361, opened 2020-05-05
Cool coercers are missing


JJ:
## The problem

All these [coercers in Cool](https://github.com/rakudo/rakudo/blob/5610764374b975516394f5f0be5c602ef71aa138/src/core.c/Cool.pm6#L417-L478) are missing in the documentation. They need to be added.

This also supersedes 
> All Cool types provide .Rat/.FatRat coercers

in #2632, which includes them. 

## Suggestions

What they do should be pretty obvious. Add them.

PR #3894, opened 2021-06-03
Add Cool coercers ref #3361

stoned:


doc/Type/Cool.pod6
```diff
@@ -1494,6 +1501,119 @@ string|/routine/sprintf#Directives>, to format and print the arguments
...
+
+Defined as:
+
+    multi method FatRat()
+
+Coerces the invocant to a L«C<Numeric>|/type/Numeric» and calls its
+L«C<.FatRat>|/routine/FatRat» method. L<Fails|/routine/fail> if the
+coercion to a C<Numeric> cannot be done.
+
+    say 1+0i.Rat;             # OUTPUT: «1␤»
```
JJ, line 1531: Maybe you mean FatRat here?

doc/Type/Cool.pod6
```diff
@@ -1494,6 +1501,119 @@ string|/routine/sprintf#Directives>, to format and print the arguments
...
+
+Coerces the invocant to a L«C<Numeric>|/type/Numeric» and calls its
+L«C<.Num>|/routine/Num» method. L<Fails|/routine/fail> if the
+coercion to a C<Numeric> cannot be done.
+
+    say 1+0i.Num;             # OUTPUT: «1␤»
+    say 2e1.Num;              # OUTPUT: «20␤»
+    say 1.3.Num;              # OUTPUT: «1.3␤»
+    say (-4/3).Num;           # OUTPUT: «-1.3333333333333333␤»
+    say "foo".Num.^name;      # OUTPUT: «Failure␤»
```
JJ, line 1567: Maybe include an allomorph as an example?

doc/Type/Cool.pod6
```diff
@@ -1494,6 +1501,119 @@ string|/routine/sprintf#Directives>, to format and print the arguments
...
+
+=head2 method Real
+
+Defined as:
+
+    multi method Real()
+
+Coerces the invocant to a L«C<Numeric>|/type/Numeric» and calls its
+L«C<.Real>|/routine/Real» method. L<Fails|/routine/fail> if the
+coercion to a C<Numeric> cannot be done.
```
JJ, line 1593: Maybe you mean Real here?

JJ (changes requested):
Can you please check my comments?

doc/Type/Cool.pod6
stoned: Sure ! I'll correct these pastos ! :sweat_smile:

doc/Type/Cool.pod6
stoned: You mean for all these coercers or specificaly for `.Num`  ?

doc/Type/Cool.pod6
JJ: In general, and because of this reason, I don't really like content to be exactly the same but for a small variation, for this reason and others. But it's OK as an initial version.

doc/Type/Cool.pod6
JJ: Well, this comment was about Num, but it would be good to have it anywhere it makes sense.

doc/Type/Cool.pod6
stoned: Well no, current rakudo's code for this method returns a `Failure` when a first  coercion to `Numeric` fails, and if it does not, another coercion, to ` Real` this time, is applied.

All these coercers follow the same pattern, except `.UInt` for which `.Int` coercer is first called and checked for failure.

doc/Type/Cool.pod6
stoned: I understand.

JJ (approved):
Looks good to me. Thanks!

Issue #3782, opened 2021-01-15
Relocate methods from Numeric to Real


coke:
The following methods are documented as part of the Numeric role but aren't present in the Numeric source (found with xt/check-signatures.t)

```
    ok 2 - # SKIP Numeric.Int not found on Numeric
    ok 3 - # SKIP Numeric.Rat not found on Numeric
    ok 4 - # SKIP Numeric.Num not found on Numeric
    ok 8 - # SKIP Numeric.conj not found on Numeric
```

These 4 (there are others not found for Numeric) all appear in Real 's source (which does Numeric), but not in the Real doc page - these method definitions should be moved.

coke:
weirdly, spec seems to only have Str.conj and Complex.conj

arkiuat:
For my first attempt at taking a closer look, I tried using some introspection in the REPL:

```
> .say for (my Complex $c).^methods.map(*.gist).values.grep(* eq 'conj')
conj

> .say for (my Cool $r).^methods.map(*.gist).values.grep(* eq 'conj')
conj

> .say for (my Str $s).^methods.map(*.gist).values.grep(* eq 'conj')
Nil
```
Then I searched in the roast repo with https://github.com/search?q=repo%3ARaku%2Froast%20.conj&type=code

In the results it does look as if `.conj` is tested only against a complex literal and a string literal, but the string test is in a file conspicuously titled cool-num.t and with an opening comment "Basic tests of mathematical functions on Cool"

I don't know how to make a `Cool` literal for that test, I only know how to make `Str` and `Complex` literals. I can make a `ComplexStr` using that type's `.new` method, but if there's a `Cool.new` I don't know how to find out anything about it much less use it.

The upshot of all this for me (and please correct me if I'm wrong) is that although spec is testing the method against a Str literal, the intent is to test `Cool` functionality, which is why the method being tested there is documented under Cool and not Str.

So if that's right, we have complete coverage with the two `.conj` methods documented under Complex and Cool, and the one under Numeric is strictly redundant (as well as slightly misleading) at this point and can just be deleted (and then we can close this issue, if the foregoing is correct).

arkiuat:
> (As I write this I also wonder why `Cool` isn't a `role`. Maybe it was at some point? Or maybe that's a silly idea? If so, what am I missing?)

For what it's worth, I was wondering exactly the same thing when drafting my most recent comment on this issue.

arkiuat:
[...]

PR #3896, opened 2021-06-10
Add Real.{Int,Num,Rat} coercers ref #3782

stoned:
... and remove non-existent Numeric.{Int,Num,Real} ones

### DIFF
--- a/doc/Type/Real.pod6
+++ b/doc/Type/Real.pod6
@@ -73,14 +73,14 @@ its real part and 0e0 as the imaginary part.
     method Int(Real:D:)
 
 Calls the L«C<Bridge> method|/routine/Bridge» on the invocant and then
-calls the L«C<Int> method|/routine/Int» on its return value.
+the L«C<Int> method|/routine/Int» on its return value.
 
 =head2 method Rat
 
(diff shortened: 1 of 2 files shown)


doc/Type/Real.pod6
```diff
@@ -68,11 +68,20 @@ it is an L<Int|/type/Int> when we instantiated C<Temperature> with an L<Int|/typ
 Converts the number to a C<Complex> with the number converted to a C<Num> as
 its real part and 0e0 as the imaginary part.
 
+=head2 method Int
+
+    method Int(Real:D:)
+
+Calls the L«C<Bridge> method|/routine/Bridge» on the invocant and then
+calls the L«C<Int> method|/routine/Int» on its return value.
```
JJ, line 76: The problem with copy/paste is that it propagates (maybe minor) errors like this one. It would have been better to not repeat "call"

JJ (approved):
Thanks a lot!

doc/Type/Real.pod6
stoned: Sure ! I will reformulate these. Thanks !

JJ (approved):
Thanks!

PR #3898, opened 2021-06-14
Update regexes.pod6

rakool:
Change "two vowels together after a C<:>" to "multiple characters after a C<:>".

Both 'vowels'  and  'two'  are too restrictive.


JJ (approved):
Thanks!

PR #3899, opened 2021-06-15
Document Num.Str

stoned:


JJ:
I think this is an infrastructure problem.

JJ:
Let me keep this open, and I'll use to fix the tests.

stoned:
FYI I may be wrong but I opened https://github.com/JJ/perl6-doccer/pull/2

PR #3903, opened 2021-06-19
clarify sentence, fix typo

michaelruigrok:
Previously, 'override' implied that the value of the outer scope may change if a variable of the same name is defined in an inner scope. I have clarified this.


jubilatious1 (commented):
Looks fine. Nice work!

PR #3912, opened 2021-07-10
clarify .& methodop

codesections:
Based on the current docs, I thought that `foo.&{...}` created an unnamed method and called it on `foo`.  However, [Jonathan's recent Stack Overflow answer](https://stackoverflow.com/questions/68308991/what-is-the-intended-value-of-routine/68327774) clarified that `.&` creates a `Block` and calls that `Block` with method syntax.

This commit clarifies the description to avoid similar confusion.


JJ (approved):
Great. Thanks!

PR #3915, opened 2021-07-15
fix Python (dict key string) syntax

mykhal:
was wrong Python syntax, or, dict() definition (with {} literal) in fact tried to use (non-existent) variables instead of string literals


JJ (approved):
Right, thanks!

PR #3919, opened 2021-07-19
add Set operator character explanations

abraxxa:
## The problem

The choice of Set operator characters was awkward to me until @lizmat explained them to me on IRC.

## Solution provided

This is the beginning of a try to add explanations to (all of?) them to clarify why they where chosen to make it easier to remember them.

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -1602,6 +1602,7 @@ details.
 
 X<Intersection operator>.
 
+'&' as in left hand side arguments 'and' right hand side arguments.
 Returns the B<intersection> of all of its arguments. This creates a new
 L<Set|/type/Set> that contains only the elements common to all of the
 arguments if none of the arguments are a L<Bag|/type/Bag>,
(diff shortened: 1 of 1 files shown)


doc/Language/operators.pod6
```diff
@@ -1602,6 +1602,7 @@ details.
 
 X<Intersection operator>.
 
+'&' as in left hand side arguments 'and' right hand side arguments.
```
JJ, line 1605: Please add some subject to this sentence, something like "As a mnemonic... " Or "As an alternative form, this operator uses... "

doc/Language/operators.pod6
```diff
@@ -1662,6 +1663,8 @@ resulting in a positive match. See also L<Junction|/type/Junction> for more deta
 
 X<Union operator>.
 
+'|' as in left hand side arguments 'or' right hand side arguments.
```
JJ, line 1666: Same as above.

JJ (approved):
It's a little grammar nit, but I'd be grateful if you could add something along the lines mentioned.

doc/Language/operators.pod6
abraxxa: As I'm no native English speaker I wasn't able to come up with a better wording. Initially I had appended it to the previous line but decided to keep that separated to make it stand out more is better.

JJ: Just use one of the two alternatives I propose...

coke:
Looks like @JJ approved this; I'll merge it later today with no objection...

JJ:
I'll leave my two nits open to remember and maybe change them afterwards.

PR #3920, opened 2021-07-22
Uniname examples fix/update

mykhal:
This fixes routine examples' excessive nested quoting;
longest uniname probably from older data, current one is 6 chars longer.

### DIFF
--- a/doc/Type/Cool.pod6
+++ b/doc/Type/Cool.pod6
@@ -1035,13 +1035,13 @@ codepoints, and L<uniparse|/routine/uniparse> for the opposite direction.
 
     # Camelia in Unicode
     say ‘»ö«’.uniname;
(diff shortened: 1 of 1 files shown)


mykhal:
.. the longest uniname example is slightly misleading thought, it gets only one, but there are two characters with (different) uninames of that longest length:

```
BOX DRAWINGS LIGHT DIAGONAL UPPER CENTRE TO MIDDLE RIGHT AND MIDDLE LEFT TO LOWER CENTRE
BOX DRAWINGS LIGHT DIAGONAL UPPER CENTRE TO MIDDLE LEFT AND MIDDLE RIGHT TO LOWER CENTRE
```

JJ (approved):
Good catch. Thanks!

PR #3921, opened 2021-07-22
using chaining operators in comparison

mustafaaydn:
hi,
### The Problem
It seems `and` was meant in `hour > 8 or hour < 22` 

### The Solution
 but I think chained comparison is better!


JJ (approved):
Good catch! Thanks

PR #3925, opened 2021-07-27
py-nutshell: Add how to enumerate in iteration

### LINKED ISSUE #3924: Py-nutshell: enumerate()
[Py-nutshell](https://docs.raku.org/language/py-nutshell) doc is lacking known and useful Python `enumerate()`.

Proposal stub:

> ### Enumerate()
> 
> Python
> ```python
> elems = ["aether", "hydrogen", "helium", "lithium"]
> for i, e in enumerate(elems):
>     print "elem no. %d is %s" % (i, e)
> # elem no. 0 is aether
> # elem no. 1 is hydrogen
> # ...
> ```
> 
> Raku
> ```raku
> my $elems = <aether hydrogen helium lithium>; 
> for $elems.kv -> $i, $e { say "elem no. $i is $e" };
> ```

I'm using conservative Python syntax, as current doc examples are in Python 2.  
Also not sure if there's some other more idiomatic Raku way.

treyharris:
Resolves #3924.

Modified from suggestion in issue to include `kv()`’s other use case in maps,
matching Python’s `dict.items()`.

### DIFF
--- a/doc/Language/py-nutshell.pod6
+++ b/doc/Language/py-nutshell.pod6
@@ -357,6 +357,46 @@ I<Raku>
         say $c;
     }
 
+Uses of Python's C<for… in enumerate(…)> and C<.items()> mechanisms for
+iterating lists or dict/maps can both be achieved using the same
+L<kv|/routine/kv> method in Raku (because the "key" of a list is its
+arraylike numeric index):
+
+I<Python>
+
+=begin code :lang<python>
+elems = ["neutronium", "hydrogen", "helium", "lithium"]
+for i, e in enumerate(elems):
+    print "elem no. %d is %s" % (i, e)
+symbols = ["n", "H", "He", "Li"]
+elem4Symbol = {s: e for s, e in zip(symbols, elems)}
+for symbol, elem in elem4Symbol.items():
+    print "Symbol '%s' stands for %s" % (symbol, elem)
+# elem no. 0 is neutronium
+# elem no. 1 is hydrogen
+# elem no. 2 is helium
+# elem no. 3 is lithium
+# Symbol 'H' stands for hydrogen
+# Symbol 'He' stands for helium
+# Symbol 'Li' stands for lithium
+# Symbol 'n' stands for neutronium
+=end code
+
+I<Raku>
+
+    my @elems = <neutronium hydrogen helium lithium>;
+    for @elems.kv -> $i, $e {
+        say "elem no. $i is $e"
+    }
+    
+    my @symbols = <n H He Li>;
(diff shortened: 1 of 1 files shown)


mykhal:
@treyharris i'd prefer classical ''aether'' than Wikipedia-resolved element 0, they are sort of identical

treyharris:
> @treyharris i'd prefer classical _aether_ than Wikipedia-resolved element 0, they are sort of identical

It’s not really important to the code illustration either way. But I’ve never heard of “aether” as element 0 before, while I had heard of neutronium, so I just changed it because I figured you’d made something up so that the array indices matched the atomic numbers. Do you have a reference? What is aether’s chemical symbol in that system?

mykhal:
@treyharris 
Let's have neutronium at least for now, it;s good, I'll dig in my notes later.

It's good. But Raku code has `.kv` twice. Could of of them be converted into `.pairs`, for illustration, if it's approx. equally idiomatic ?

treyharris:
> Raku code has `.kv` twice. Could one of them be converted into `.pairs`, for illustration, if it's approx. equally idiomatic ?

It isn’t—neither equally idiomatic in this use case, nor with an analog commonly used in Python. So, while we _could _change the code from using `kv` to different code using `pairs`, it would then not really qualify for being included in this document at all (since it wouldn’t be an aid to Pythonic idioms in Raku)—so just dropping the hash/dict example entirely would be better.

`kv`’s double utility—where Python uses two different words—was an intentional and useful part of the example as I wrote it.

But if you can think of an example that would both motivate `pairs` and would map to idiomatic Python—by all means, suggest it.

treyharris:
Oops, @JJ — I mistakenly pinged you in a commit message when I meant to do it here in the conversation thread. Repeating:

I noticed that in this py-nutshell doc, `L<>`’s to routine docs are not in `C<>`’s, so appear as regular text instead of monospaced. This seems to be a case where there’s different styles on different pages—but should I submit a separate patch to change all these routine-link `L<>`’s to code font? Thanks

JJ:
> Oops, @JJ — I mistakenly pinged you in a commit message when I meant to do it here in the conversation thread. Repeating:
> 
> I noticed that in this py-nutshell doc, `L<>`’s to routine docs are not in `C<>`’s, so appear as regular text instead of monospaced. This seems to be a case where there’s different styles on different pages—but should I submit a separate patch to change all these routine-link `L<>`’s to code font? Thanks

If you have the time and will, please do it. We generally do it when we're doing some other changes on the file, because it's a lot of work. Thanks for your contribution anyway.

Issue #3906, opened 2021-06-19
Add overview comment to util/*


coke:
The files in util/ are not commonly used, having the comments will help with maintenance of the repo.

- [x] util/clean-spell
- [x] util/create-brackets-table.raku
- [x] util/github-action-test.sh
- [x] util/ignored-methods.txt
- [x] util/list-missing-methods.raku
- [x] util/missing-types.raku
- [x] util/new-type.raku
- [x] util/perl-nbsp.raku
- [x] util/sort-words.raku
- [x] util/test-modified.sh
- [x] util/unskip.raku
- [x] util/update-and-test

coke:
Related: we have both Dockerfile and util/Dockerfile - They appear to be basically duplicative, with some formatting differences, and direct invocation of commands vs. invoking a script. If we can eliminate one as part of this effort great, otherwise let's include documents on the top level Dockerfile as well so we know why they're different.

JJ:
You mean inline comments here?
Also, re: Dockerfile. I need to find out why these were done. I think that one is used in Circle, and the other elsewhere. I'm going to check it out.

JJ:
Right, one is a `no-docs` Dockerfile used to create the docs (and thus probably useless now, since building is no longer done here). Refers to #2525

PR #3936, opened 2021-08-21
Fix file truncate and extrapolation errors in stub doc generation

k3ut0i:
## Stub Documentation generation is buggy
```
util/new-type.raku --kind=role Some::Role
```
This generated the following stub
```

=head1 Methods

=head2 method flurb

    method flurb(Some::Role:D: *@args --> Str)

method description here

=end pod

# vim: expandtab shiftwidth=4 ft=perl6
```
It should have been
```
=begin pod

=TITLE role Some::Role

=SUBTITLE ...

    role Some::Role is SuperClass { ... }

Synopsis goes here


=head1 Methods

=head2 method flurb

    method flurb(Some::Role:D: *@args --> Str)

method description here

=end pod

# vim: expandtab shiftwidth=4 ft=perl6
```

It happened due to file truncation when writing to it twice.
- fixed the truncation issue.
- a variable interpolation issue.
- check to only create the file if it does not exist.
- a comment to help with  #3906.

### DIFF
--- a/util/new-type.raku
+++ b/util/new-type.raku
@@ -17,8 +17,8 @@ sub MAIN($typename, :$kind='class') {
 
     $path ~= "/$filename";
     if $path.IO ~~ :e {
-	say "The file $path already exists.";
-	exit 1;
+        say "The file $path already exists.";
+        exit 1;
     }
     my $fh = open $path, :x;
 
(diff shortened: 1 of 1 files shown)


util/new-type.raku
```diff
@@ -12,8 +16,13 @@ sub MAIN($typename, :$kind='class') {
     }
 
     $path ~= "/$filename";
+    if $path.IO ~~ :e {
+	say "The file $path already exists.";
```
JJ, line 20: Please mind the indentation.

JJ (approved):
Thanks for this! Except for the nit, it looks fine.

Issue #2673, opened 2019-03-14
Checklist for 2019.03


JJ:
Check the announcement [here](https://github.com/rakudo/rakudo/blob/master/docs/announce/2019.03.md)

## Additions:
 * [x] Added a candidates method to `CompUnit::Repository::FileSystem`
        and `CompUnit::Repository::Installation`, which acts as the API
        for module lookup. Internals were refactored to use this API for
        module lookup, which fixes some issues as a side effect (i.e.
        old bin scripts being invoked instead of a newer version)
        -[2d007bf4](https://github.com/rakudo/rakudo/commit/2d007bf4)-[07f62c60](https://github.com/rakudo/rakudo/commit/07f62c60)-[bed3bbd0](https://github.com/rakudo/rakudo/commit/bed3bbd0)-[a25d5fa4](https://github.com/rakudo/rakudo/commit/a25d5fa4)-[29d74709](https://github.com/rakudo/rakudo/commit/29d74709)-[085fa19c](https://github.com/rakudo/rakudo/commit/085fa19c)
        -[78de05eb](https://github.com/rakudo/rakudo/commit/78de05eb)-[b378a157](https://github.com/rakudo/rakudo/commit/b378a157)-[d0a3fec9](https://github.com/rakudo/rakudo/commit/d0a3fec9)-[d533d450](https://github.com/rakudo/rakudo/commit/d533d450)-[877e7dc0](https://github.com/rakudo/rakudo/commit/877e7dc0)-[270b355e](https://github.com/rakudo/rakudo/commit/270b355e)
        -[683df8bf](https://github.com/rakudo/rakudo/commit/683df8bf)-[e1ef571e](https://github.com/rakudo/rakudo/commit/e1ef571e)-[077f88a5](https://github.com/rakudo/rakudo/commit/077f88a5)-[22021fdd](https://github.com/rakudo/rakudo/commit/22021fdd)-[603010be](https://github.com/rakudo/rakudo/commit/603010be)-[27e141cc](https://github.com/rakudo/rakudo/commit/27e141cc)
[...]

PR #3939, opened 2021-08-22
Caret parameterize

JJ:
## The problem

Metaclass methods were added in 2015; so far they hadn't been working, or actively used. During release 2019.03 #2673 they were used, which prompted us to request docs for them.

## Solution provided

Looking at the implementation, and through a few examples, some use cases have been written and included in the documentation. They can then be pointed at when referring to the corresponding changes in Rakudo, implemented in that release.


JJ:
@lizmat can you please take a look at this? I can merge it alright, but would like a second opinion.

Issue #3940, opened 2021-08-24
Add info on qq quotes for behavior of left angle brackets


tbrowder:
## Problem or new feature

Given the following code:

    my $a = 'foo';
    say qq:to/HERE/;
        <a>$a</a>
    HERE
    
one gets this result:

    (to be entered shortly)

The docs should expand a bit more on such behavior and its mitigation.

jonathanstowe:
I get

```
Type Str does not support associative indexing.
  in block <unit> at -e line 3
```

That can be fixed by doing either:

```
my $a = "foo";
say qq:to/HERE/;
    <a>{$a}</a>
HERE
```

or

```
my $a = "foo";
say qq:to/HERE/;
    <a>$a\</a>
HERE
```

It's not so much the left angle brackets, rather the constructs that look like indexing.

jonathanstowe:
You'd get similar if the characters after the variable were `{..}` or `[..]` for instance.

tbrowder:
You are correct, and that's how I fixed my problem (with {$a}, my goto solution for qq problems). What was confusing me in my case was where the problem was since the code in the HERE block is quite lengthy with javascript, html, and other embedded strings I had already put in curlies (as well as escaped curlies in the js code). 

I finally found the offending line by starting with the first line only and gradually adding one line at a time. I had failed to spot the problem because it was so innocent looking!

tbrowder:
I will offer a PR later.

tbrowder:
See PR #3941

PR #3941, opened 2021-08-24
Expand info on Heredocs

tbrowder:
Expands on heredoc information to satisfy doc issue #3940

### DIFF
--- a/doc/Language/quoting.pod6
+++ b/doc/Language/quoting.pod6
@@ -472,7 +472,7 @@ string
 END
 =end code
 
-The contents of the heredoc always begin on the next line, so you can (and
+The contents of the I<heredoc> always begin on the next line, so you can (and
 should) finish the line.
 
 =begin code :preamble<sub my-escaping-function { $^x }>
@@ -483,7 +483,7 @@ TERMINATOR
 =end code
 
 If the terminator is indented, that amount of indention is removed from the
-string literals. Therefore this heredoc
+string literals. Therefore this I<heredoc>
 
 =begin code
 say q:to/END/;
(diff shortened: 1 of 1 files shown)


doc/Language/quoting.pod6
```diff
@@ -525,6 +525,38 @@ option {
...
+say qq:to/HERE/;
+<a href='https://usafa-1965.org'>$title</a>
+HERE
+# Output:
+Type Str does not support associative indexing.
+  in block <unit> at here.raku line 2
+=end code
+
+The right angle bracket makes '$title' look like a hash call
+to Raku when it is actually a Str variable, hence the error message.
```
JJ, line 543: Nit: `C<Str>` Maybe also clarify you refer to `$title</a>`. And maybe something instead of hash "call". Hash index?

doc/Language/quoting.pod6
```diff
@@ -525,6 +525,38 @@ option {
...
+The cure is to enclose the scalar with curly braces:
+
```
JJ, line 545: Maybe clarify this is simply the way used in quoting constructs for expressions?

JJ (approved):
To be honest, I am of two minds about this PR. On the face of it, it's perfectly fine, and addresses the issue. But a couple of things bug me
1. This is clearly a trap. We have a specific page for traps. That does not mean we don't document traps elsewhere. At the end of the day, it's a judgment call.
2. This is probably an example of a wider problem. Probably the same will happen with curly braces, and most probably with square brackets. It's probably not limited to heredocs, I'm afraid. Any quoting construct is bound to have the same problem, there's nothing special about heredocs. However, that's outside the scope of this PR and it would probably be reasonable to open another PR for that

PR #3945, opened 2021-08-27
default base type for subset declaration

k3ut0i:
Clarify that the default base type in the declaration of subset is `Any`[1].

  [1]: https://github.com/rakudo/rakudo/blob/c0a6823b0c0ef3382c281202adcc6ce079afc141/src/Perl6/Actions.nqp#L5218


k3ut0i:
@JJ  Yes, I accepted the invitation. I am just getting familiar with the repo, so I am hesitating to commit directly.

JJ:
No problem. You're very welcome to do it when you're ready.

PR #3948, opened 2021-08-30
Explain behavior of non-alphanumerics in <[ ]>

codesections:
The docs previously mentioned that you can  use `\` to escape characters inside `<[  ]>` ; this PR adds the fact that you don't need to escape most non-alphanumeric characters the way you do in the portion of a regex outside `<[ ]>` and that you must escape (rather than quote) characters inside `<[ ]>` for them to have their literal meaning.


coke:
PR fails 'make xtest', has trailing whitespace. Otherwise looks good.

codesections:
I guess that's what I get for using the web editor.  Fixed, thanks!

PR #3951, opened 2021-09-02
Clarify where rx declarator allows adverbs

codesections:
A regex declared with `rx` allows adverbs both before the delimiter (`rx:s/pattern/`) and after (`rx/:s pattern/`), but the docs only mentioned the first option.  This fixes that, and brings the rx section of the page into alignment with the adverb section.


coke:
Passes 'make xtest'

PR #3954, opened 2021-09-06
Provide a practical example of interpolating package names

tbrowder:
## The problem

No example shown of interpolating and using a subroutine.

## Solution provided

This PR aims to provide that example.


JJ (approved):
Looking fine. Thanks!

PR #3957, opened 2021-09-09
Note that &?ROUTINE refers to current multi

codesections:
This PR adds a note that `&?ROUTINE` refers to the current `multi` candidate and compares it to `samewith`, which allows self-recursion to the entire proto.


JJ:
Thanks again!

Issue #1355, opened 2017-06-01
Don't use `.WHAT` in examples to print the type info


zoffixznet:
It's [not quite the right tool for the job](https://www.reddit.com/r/perl/comments/6dfj2g/perl_6_interactive_shell_values_in_variables/di2g70s/)

Currently there's a ton of uses many of which shouldn't be using `.WHAT`:

```
zoffix@leliana/tmp/tmp.jxnsvMuh21 (master)$ grep -FR '.WHAT'
doc/Language/functions.pod6:multi mistake-proto($str, $number) { say $str.WHAT }
doc/Language/functions.pod6:   say $bar.WHAT; # OUTPUT: «(Bar)␤»
doc/Language/typesystem.pod6:    say $a.WHAT;
doc/Language/typesystem.pod6:identity between the object and its C<.WHAT> pseudo-method. Note that the
doc/Language/typesystem.pod6:    say so $a // $a === $a.WHAT;
doc/Language/typesystem.pod6:    say A.new.WHAT;    # OUTPUT: «(A)␤»
doc/Language/typesystem.pod6:    say $a.WHAT;     # OUTPUT: «(Names)␤»
doc/Language/typesystem.pod6:    say one.WHAT; # OUTPUT: «()␤»
doc/Language/typesystem.pod6:    say $e.WHAT;  # OUTPUT: «(Map)␤»
doc/Language/variables.pod6:    say @bar.WHAT;          # OUTPUT: «(Array)␤»
doc/Language/variables.pod6:    say $baz.WHAT;          # OUTPUT: «(List)␤»
doc/Language/faq.pod6:    say @a.WHAT;            # OUTPUT: «(Array)␤»
doc/Language/faq.pod6:    say $scalar.WHAT;       # OUTPUT: «(Array)␤»
doc/Language/exceptions.pod6:            say .WHAT.perl, do given .backtrace[0] { .file, .line, .subname }
doc/Language/classtut.pod6:    say $o.WHAT;
doc/Language/classtut.pod6:The C<.WHAT> method returns the type object associated with the object
doc/Language/classtut.pod6:know about, finding its type with C<.WHAT>, seeing a construction recipe for
doc/Language/containers.pod6:    say @a.WHAT;                # OUTPUT: «(List)␤»
doc/Language/5to6-nutshell.pod6:        say $query.WHAT;       # (Block)
doc/Language/objects.pod6:    my $int-type-obj = 1.WHAT;
[...]

PR #3962, opened 2021-09-14
Explain "list" vs "destructuring" assignment

### LINKED ISSUE #3961: Explain/correct the distinction between "list assignment" and "destructuring assignment"
The docs currently use "destructuring assignment" as a synonym for "list assignment".  For example, in the [Item and List assignment](https://docs.raku.org/language/variables#Item_and_list_assignment) section, we provide these examples

```raku
my $f;
($f,) = 7,8,9;                 # list assignment to List with one element 
say $f;                        # OUTPUT:  «7␤» 
say ( ($f,) ).VAR.^name;       # OUTPUT:  «List␤» 

# ATTENTION: special declaration syntax! 
my ($g) = 7,8,9;               # list assignment to List with one element 
say $g;                        # OUTPUT:  «7␤» 
say ( ($g) ).VAR.^name         # OUTPUT:  «List␤»
``` 
and then say that
> The last two examples above are simple destructuring assignments that select the first item of the right-hand side list. 

Conflating these two terms is understandable – as far as I can tell, most languages don't make this distinction (e.g., JavaScript calls both features [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment); PHP calls both features [list assignment](https://www.php.net/list).  

However, in Raku there _is_ a distinction.  As @jnthn pointed out in rakudo/rakudo/issues/4522:

> At the heart of the issue is that the grammar parses a signature here; in the case we really do destructuring (`my (:$numerator, :$denominator) := 4.2`) then the signature is exactly what we want. While "destructuring" is often used loosely to describe things like `my ($x, $y) = 1, 2`, strictly this is an example of list assignment, not destructuring.
>
> The case given in the issue here is not destructuring either; there's no binding operator, and I think for attributes there cannot be. Rather, the situation here is that we are declaring a list of attributes, just as `my ($x, $y);` is declaring a list of lexical variables. In this case, the signature should "decay" into just declarations (and in the case of `my ($x, $y) = 1, 2`, we expect it to also produce a list of the declared variables; what happens from there is also list assignment rather than destructuring).

Apparently this issue has been confusing people for a while – it came up in the 2017 [Review of Perl 6](https://www.evanmiller.org/a-review-of-perl-6.html) post:

> I don’t understand the logic of making a syntactic distinction between list assignment and full destructuring assignment — and silently flattening the lefthand list structure in the first case — but there you have it. 

And I'll admit that it's not a distinction I'd really focused on until it came up in the context of that Rakudo issue.  I think I now understand the distinction and will prepare a PR – but I'll be especially looking for feedback on it to be sure I don't misstate anything.

codesections:
This PR is an attempt to resolve #3961 by significantly reworking the `list assignment` section and adding a new `destructuring assignment` section to clarify that these two methods of assignment are different.

I would appreciate a second pair of eyes on this PR to make sure that I didn't misstate any of the differences between the types of assignment or document any behavior that is a bug rather than intentional/in accord with the spec.  

(I'd also be interested to know whether anyone has views on how the docs should refer to the concept I called "destructuring assignment": I went with that term because it was already in the docs, it's the term other languages seem to use, and it's the one that was used in /rakudo/rakudo/issues/4522 (which is what set me off on this in the first place).  However, roast doesn't seem to use the term "destructuring assignment" – the closest it comes is in S06-signature, which talks about "destructuring".  On the other hand, Roast doesn't really seem to have another term for this form on assignment. (Two tests in S04-declarations use the term "signature bind", but that isn't widespread and seem significantly more confusing).

I sometimes feel that it's hard to get a sense of how the docs look just from the pod source; in case anyone else feels the same, I uploaded a [copy of the changed sections](https://codesections.com/DRAFT-destructuring-assignment.html) so that you can review the rendered HTML if that's easier.

### DIFF
--- a/doc/Language/variables.pod6
+++ b/doc/Language/variables.pod6
@@ -84,105 +84,6 @@ CATCH { default { put .^name, ': ', .Str } }
 For information on variables without sigils, see
 L<sigilless variables|#Sigilless_variables>.
 
-=head2 Item and list assignment
-
-There are two types of variable assignment, I<item assignment> and I<list
-assignment>.
-
-An item assignment copies a single value from the right-hand side into a Scalar
-variable on the left. An assignment to anything other than a simple Scalar
-variable is parsed as a list assignment. A list assignment leaves the choice of
-what the assignment operation entails to the variable on the left. For example,
-L<Array|/type/Array> variables (C<@> sigil) empty themselves on list assignment,
-and then iteratively copy all values from the right-hand side into themselves as
-elements.
-
-The two types of assignment both use the equal sign C<=> as their operator and
-are both right associative, but differ in operator precedence: item assignment
-has a higher precedence level (level: Item assignment) than list assignment
-(level: List prefix). In situations in which a comma-separated list of elements
-is assigned, these precedences should in particular be contrasted with that of
-the comma operator C<,> which sits in between. So without any list-delimiting
-parentheses (or other construct to hold the list's elements together), item
-assignment will only assign the first element of the specified list, and not the
-full list.
-
-In an assignment expression the context of the left-hand side determines whether
-an C<=> means item or list assignment. As mentioned, item assignment is
-restricted to simple Scalar variables. Accordingly, assignment to a Scalar
-container (scalar-context) triggers item assignment, unless the Scalar is
-explicitly put in list-context by surrounding parentheses C<( )>:
-
-    my $a;
-    $a = 1,2,3;        # item assignment to Scalar
-    say $a;            # OUTPUT: «1␤» ( '=' has higher precedence than ',' )
-
-    my $b = 1,2,3;     # item assignment to Scalar (same as preceding example)
-    say $b;            # OUTPUT: «1␤»
-
-    my $c;
-    ($c) = 4,5,6;      # list assignment to Scalar; '( )' is list-contextualizer
-    say $c;            # OUTPUT:  «(4,5,6)␤»
-
-    (my $d) = 4,5,6;   # list assignment to Scalar (same as preceding example)
-    say $d;            # OUTPUT:  «(4,5,6)␤»
-
-Assignment to a List container (list-context) always triggers list assignment:
-
-    my @e;
-    @e = 7,8,9;                    # list assignment to Array
-    say @e;                        # OUTPUT:  «[7,8,9]␤»
-
-    my $f;
-    ($f,) = 7,8,9;                 # list assignment to List with one element
-    say $f;                        # OUTPUT:  «7␤»
-    say ( ($f,) ).VAR.^name;       # OUTPUT:  «List␤»
-
-    # ATTENTION: special declaration syntax!
-    my ($g) = 7,8,9;               # list assignment to List with one element
-    say $g;                        # OUTPUT:  «7␤»
-    say ( ($g) ).VAR.^name         # OUTPUT:  «List␤»
-
-The last two examples above are simple I<destructuring assignments> that select
-the first item of the right-hand side list. See for a more elaborate discussion
-of destructuring assignments in the context of variable declarations the section
-on L<declaring a list of variables with lexical or package
-scope|/language/variables#index-entry-declaring_a_list_of_variables>.
-
-Chained assignments are parsed having regard to the precedence of the assignment
-operators and, where applicable, their right associativity. For instance, in the
-example below there is one chained assignment statement comprising two
-assignment operators. The assignment to C<@array> is a list assignment having a
-lower precedence than the item assignment to the Scalar variable C<$num>. The
-assignment expression involving the item assignment to the Scalar variable
-C<$num> is thus evaluated first. It returns the assigned value C<42>, which in
-turn forms part of the List C<(42, "str")> constructed by the comma operator
-that also has a higher precedence than the list assignment. Finally, the
-List C<(42, "str")> is list-assigned to C<@array>:
-
-    my @array;
-    @array = my $num = 42, "str";   # parsed as @array = ( (my $num = 42), "str )
-    say @array.raku;                # OUTPUT: «[42, "str"]␤» (an Array)
-    say $num.raku;                  # OUTPUT: «42␤» (a Num)
-
-Here's a variant:
-
-    my ( @foo, $bar );
-    @foo = ($bar) = 42, "str";       # parsed as @foo = ( $bar = (42, "str") )
-    say $bar.raku;                   # OUTPUT: «$(42, "str")␤» (a List)#
-    say @foo.raku;                   # OUTPUT: «[(42, "str"),]␤» (an Array)
-
-In this case, the list contextualizer C<( )> puts C<$bar> in a list context, and
-thus triggers a list assignment to the Scalar variable C<$bar>. This means that
-there are two chained list assignments, both having a lower precedence than the
-comma operator C<,> that constructs the List C<(42, "str")>. Due to their right
-associativity, the list assignment expression that is evaluated first is the
-assignment to C<$bar>, which returns the assigned value C<$(42, "str")>, i.e. a
-Scalar containing a two-element List. This value is in turn list-assigned to
-C<@array>, such that it becomes a Array with a single element, namely a List.
-
-See L<operators|/language/operators> for more details on precedence and
-associativity.
 
 =head2 Sigilless variables
 X<|\ (sigilless variables)>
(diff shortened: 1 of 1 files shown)


JJ:
Thanks for this new PR. I need to check it out carefully; it might take some time.

tbrowder:
Just a comment from a Raku fan: VERY helpful for someone who can't always remember such details and needs a good example. Excellent! Thanks, Daniel!

codesections:
@p6steve sent very helpful comments via email, which I'll quote from and respond to below.  (I've made a few changes to paragraph breaks, punctuation, and capitalization in the quotes below)

> from your preview, it looks like this is now the top section of the Variables page on docs.raku.org … I would rather keep the Sigils section above the Assignments section

I agree that the Sigil's section should stay at the top.  I wasn't suggesting changing that – sorry if my excerpted changes made that unclear.  Here are the first few sections and subsections in the Variables page before my PR:

* Sigils
  - Item and list assignment
  - Sigilless variables
* Twigils
  - [8 twigil subsections]
* Variable declarators and scope
  - The `my` declarator
  -  The `our` declarator
  - Declaring a list of variables with lexical (`my`) or package (`our`) scope
  - [more]

And here's the order I suggested in my PR:

* Sigils
  - Sigilless variables
 * Twigils
  - [8 twigil subsections]
* Item assignment, list assignment, and destructuring assignment
  - destructuring assignment
* Variable declarators and scope
  - The `my` declarator
  -  The `our` declarator
  - Declaring a list of variables with lexical (`my`) or package (`our`) scope
  - [more]

That is, I'm not suggesting that we move Sigils _down_, I'm suggesting that we move Twigils up, and make the Assignment section its own top-level heading instead of a subheading under Sigils (since it isn't really a sub-point you need to understand to grasp the concept of Sigils).  Does that make sense, or do you still have concerns about the re-ordering?

> Now, my main homily is that for newbies (both to raku and to coding in general), assignment is a simple concept – everyone knows from junior school that 2+2=4 

I generally agree (though, I'll also note that many early programmers have trouble with assignment precisely _because_ they're familiar with it from math – but the assignment our elementary school teachers taught would never allow expressions like `$x = $x + 5`)

> Since [assignment] is such a basic building block of common currency (as opposed to eg. binding), then I think it should stay as section 2.  [It's] important to me is that we (the documentation authors) end up with an audience that is not confused and disoriented […] So – please don’t shoot the messenger, but I am a bit worried that your version, while more precise and detailed than the original, is too quickly deep and thus a confusing a way to convey the raku implementation of such a simple concept as assignment.
>
> My guess that the original author of the docs was well aware of the details (of destructuring), but specifically chose to describe only item and scalar assignment to avoid diving too quickly into the detail for the average reader.  In my role as “meat and potatoes” coder, the title here is very scary … 3 types of assignment!!! Another worry is that this section now carries too much information and has grown from <3/2> pages to about 4 pages … so my feedback is to pare it back to the original’s length.

I think those are all very good points, and I agree 100% that we should aim to avoid diving too quickly into details, as you say.  And I also agree that the increased length of this version is a warning sign – all else being equal, we should aim to be as concise as possible.

After considering the points you raised above, I wonder if we should take an even larger step to make the Variables page shorter: Extract the Assignment section into its own page, and then cut the entire discussion of assignment on the Variables down to a short "Binding and assignment" paragraph that mostly just links to that page.  This would let us both keep thing simple here, and to provide more details for users who are interested.  Both before and after my PR, the Variables page is the 6th largest `/Language` page (out of ~70), so splitting off Assignment would still leave plenty of content here.  And, at least to me, Assignment seems like a distinct enough topic to be worth its own page (especially if we pull in some of the info from the [Binding](https://docs.raku.org/language/containers#Binding) section as well).

How would you feel about that change?

so my hope is that you will apply a little presentational concision, closer to the old version, to subtly hide the complexity of destructuring assignment as one of the features of (list) assignment … my proposal is to move Chained Assignment and Destructuring Assignment to their own headings

codesections:
(As in ^^^, the quotes are from emailed comments from @p6steve.)

> my proposal is to move Chained Assignment and Destructuring Assignment to their own headings

I like that suggestion.  I'm not sure exactly what level the Chained Assignment heading should be at – maybe a third-level heading, since it's not really it's own _form_ of assignment (the way Item, List, and Destructuring assignment are), but splitting it out definitely makes sense. 
>
```raku
say $a.VAR.^name;     # OUTPUT: «1␤Scalar␤» 
```
> I think .WHAT is less line noisy for this

The problem is that `$a.WHAT` prints `Int`, not `Scalar`, so it'd need to be `$a.VAR.WHAT`.  That's actually what I wrote at first, but one of the doc xtests informed me that, per #1355, we're encouraged to use `.^name` instead of `.WHAT`.  That's probably not the call I would have made (to avoid confusing users who haven't encountered the `.^` operator), but I get that consistency is important, so ¯\_ (ツ)_/¯

> [several helpful suggestions for simplifying examples]

Thanks, I generally agree with these.

> so my hope is that you will apply a little presentational concision, closer to the old version, to subtly hide the complexity of destructuring assignment as one of the features of (list) assignment …

I think this is the only place where I actually disagree with anything you said.  While I _do_ want to keep this as simple and approachable as possible, I also feel that clearly distinguishing between list assignment and destructuring assignment is extremely important.

If you don't mind, I'd like to provide slightly longer quote from the 2017 post I mentioned in the original issue. In [A Review of Perl6 (Raku)](https://www.evanmiller.org/a-review-of-perl-6.html), Even Miller wrote:

> In this way, list assignment follows somewhat naturally:
> ```raku
>  my $value1 = 1;
>  my $value2 = 2;
>    
>  ($value1, $value2) = ($value2, $value1);
> ```
> Don’t get too excited though, as this assignment won’t behave as you’d expect:
> ```raku
> my $value1 = 1;
> my $value2 = 2;
> my $value3 = 3;
>
> (($value1, $value2), $value3) = ((100, 200), 300);
> ```
> For reasons that aren’t clear to me, the lefthand side is flattened before assignment, and is equivalent to this:
> ```raku
> ($value1, $value2, $value3) = ((100, 200), 300);
> ```
> Thus in both cases `$value1` gets assigned (100, 200), `$value2` gets 300, and `$value3` gets nothing. There’s a separate colon prefix for doing a destructuring assignment (note that bind-to `:=` is required in this usage):
> ```raku
> :(($value1, $value2), $value3) := ((100, 200), 300);
> ```
> In the above line, `$value1` gets 100, `$value2` is 200, and `$value3` becomes 300. I don’t understand the logic of making a syntactic distinction between list assignment and full destructuring assignment — and silently flattening the lefthand list structure in the first case — but there you have it. 

I point out this passage for two reasons: first, to show that this list assignment/destructuring assignment issue has been confusing people for _at least_ four years (and, given the length and depth of that post, Evan was a pretty serious student of Raku).

Second, I like that blog post and have read it at least three times (once before I'd otherwise heard of Raku, and twice since) – and yet the distinction between list assignment and destructuring assignment **still** didn't click for me until Jonathan made the distinction in the linked Rakudo issue.  But, having now realized that what piece I was missing, I can look back on several times when that misunderstanding was what stopped me from writing the Raku code I wanted to: I was trying to get list assignment to do something that needed to be done with destructuring assignment.

Now, I'm 100% open to the possibility that I'm just not all that bright and that other users haven't had the same issue.  But, on the chance that others have been tripped up by this distinction, I'd like to make it crystal clear in our docs.

> btw – I learned a lot by reading your draft – my feedback is done with the best will and honest intention as feedback requested on reddit … it is given freely and sincerely … but you are the boss!

Thanks for the kind words and your very helpful comments.  But I'm definitely **not** the boss of this or anything else – both Raku and docs.raku.org are entirely community projects, and I'm just one more user hoping to do my small part.

raiph:
> I'd also be interested to know whether anyone has views on how the docs should refer to the concept I called "destructuring assignment": I went with that term because it was already in the docs

Imo "destructuring assignment" is the wrong term, the doc is wrong if it uses that term, and it ought be "destructuring binding".

> it's the term other languages seem to use

The word "assignment" is often used in PL contexts to mean what's called "binding" in Raku. As a couple examples, the term SSA -- Static Single Assignment -- refers to what in Raku is technically Static Single Binding, and what Python calls "assignment" is what in Raku is called "binding".

> it's the one that was used in /rakudo/rakudo/issues/4522

Yes, but only by you and the doc, and very notably not jnthn. I think both you and the doc are wrong to use the term "destructuring assignment" for features that are actually list assignment (per jnthn) or destructuring binding (per @Larry, the historical design docs, jnthn, and Rakudo).

> reworking the list assignment section and adding a new destructuring assignment section to clarify that these two methods of assignment are different.

I currently agree with your idea (reacting to @p6steve's notes) to shrink the bits about assignment and binding to a bare minimum on the variables page, and instead move the bulk of discussion of assignment and binding to their own page(s).

> Just as list assignment uses the same operator as item assignment, destructuring assignment uses the same characters as the Binding operator, `:=`.

Right. Which obviously strongly suggests calling it a binding operation. But...

> Nevertheless, destructuring assignment is *assignment*

Because? I suspect that's because you're thinking that it:

> still creates `Scalar`s rather than binding directly.

Aiui neither assignment nor binding creates `Scalar`s:
```
sub foo ($,$) {} # Creates two `Scalar`s in a signature object.
foo 1,2;         # Bind list (1,2) to `foo`'s signature object.
my ($,$);        # Create two `Scalar`s in a signature object.
my ($,$) = 1,2;  # Create signature object, decay to list, list assign (1,2) to it.
my ($,$) := 1,2; # Create signature object, bind (1,2) to it.
```

(I've used the term "decay" from jnthn's comment in https://github.com/rakudo/rakudo/issues/4522)

> This is exactly the same behavior you get by default with a function's parameters

Exactly.

The doc has entries like [binder](https://docs.raku.org/language/glossary#index-entry-binder):

> When you pass an argument list to a function (or any other callable, like a method or a block), the argument list gets *bound* to the parameters in the signature. The code that does this is called the *binder*.

Binder code is being run to process `my ($,$) := ...`.

> An item assignment copies a single value from the right-hand side into a `Scalar` variable on the left. For example:

Maybe s/`Scalar` variable/`Scalar` variable (revealed by [`.VAR`](link to `.VAR` doc))/? That'll mean you'll have already given the reader a link to click on if they're not sufficiently familiar with `.VAR` as they encounter it in the code block that follows.

```
my $a;
$a = 1; # Item assignment of an Int to a Scalar in $a;
say $a; say $a.VAR.^name; # OUTPUT: «1␤Scalar␤»

my $b = (1, 2, 3); # Item assignment of a List to a Scalar in $a;
say $b; say $b.VAR.^name; # OUTPUT: «(1 2 3)␤Scalar␤»
```
Maybe s:/to/into/?

I suggest consistently using "into" when discussing *assignment* (copying) *into* a `Scalar` (*container*), and "to" when discussing *binding* (whether that's to the `.value` of a symbol `Pair` -- `$foo` or `@bar` or `%baz` etc. -- or to an individual item/element within a `Positional` or individual `Pair`'s `.value`).

(In both prose and code comments.)

> Specifically, a variable with a `$` sigil or without a sigil takes one value from the right hand side and stores it in a `Scalar`

Maybe s/or without a sigil //?

```
my \a = 1, 2;
say a; # (1 2)
```

Sigilless variable "assignment" is actually (static single) binding, despite it using `=` and despite it sometimes being called SSA. (Recall that it's really SSB using Raku's terminology.)

```
a := 42; # Cannot use bind operator with this left-hand side
```
This is because `a` was subject to *static single* binding, not because `my \a = 1,2;` wasn't binding -- it was.

While we're on this topic, note also:
```
my \b = $;
b = 42, 99;
say b; # (42 99)
```
Also, neither `a` nor `b` are `Scalar`s.

Fwiw I discussed a bunch about this stuff in serious detail in [Differences between sigilless and sigilled variables that are bound to](https://www.reddit.com/r/rakulang/comments/kdq9uk/differences_between_sigilless_and_sigilled/).

----

> a `@`-sigiled variable takes all remaining variables from the right hand side and copies them into a `Positional` (usually an `Array`);

Maybe s/usually /by default /?

> and a `%`-sigiled variable takes all remaining values on the right hand side and copies them into an `Associative` (usually a `Hash`).

Ditto?

> If those values are `Pair`s, 

You've written a three sentence paragraph:

* "The effect of a list assignment is ... 

* "Specifically ..."

* "If those values are `Pair`s ..."

The "those" referent is the last semi-colon part of the middle sentence, but someone who doesn't know Raku well enough to realize that might be quite confused.

Maybe break the paragraph up to be the first sentence as a sentence, and then the three parts of the second sentence become a bullet list or mini table, with one entry each for `$`, `@`, and `%`?

```
my ($a) = (1, 2, 3);      # list assignment to a Scalar;            # s/Scalar;/Scalar/

my @b = (1, 2, 3);        # list assignment to a Positional         # s:g?/to/into/

# If there are fewer values than LHS variables, some will get Any   # s/Any/Nil, ending up as Any/
```

The next block hurts my brain:
```
my $a = 1, 2, 3; # item assignment; higher precedence than ,
say $a;          # OUTPUT: «1␤»

my ($c) = 1, 2, 3; # list assignment; lower precedence than , but using ( )
say $c;            # OUTPUT:  «1␤»
```
I'm looking at that as an inner item assignment as part of an outer list assignment. In concrete terms, item `$c` within the LHS list is assigned item `1` from within the RHS list.

----

I'm hoping this helps rather than hinders; it's great to see you take this on!

p6steve:
hi ... my input is (deliberately) presentational ... I DO think that there ought to be one "first stop" page in the docs that succinctly covers BOTH the concepts of variables, sigils and assignment (and maybe twigils)

jnthn:
> I think both you and the doc are wrong to use the term "destructuring assignment" for features that are actually list assignment (per jnthn) in some cases, or destructuring binding (per @larry, the historical design docs, jnthn, and Rakudo) in the others.

I'd agree "destructuring assignment" feels wrong; you can achieve assignment semantics with use of `is copy` in the signature, but binding is certainly the default. Given both are options, I'd probably just title the topic "destructuring".

I remember early on in my involvement with the language realizing that destructuring had been expressed in terms of signatures, and found this really neat: you only need to learn the signature language once, and then get to use it for all of "boring" signatures, signatures that want to unpack data structures (by recursive application of the same syntax), and for doing destructuring at the time of a variable declaration.

raiph:
@jnthn

> I'd agree "destructuring assignment" feels wrong; you can achieve assignment semantics with use of `is copy` in the signature, but binding is certainly the default. Given both are options, I'd probably just title the topic "destructuring".

Ah. So I think you're saying it's not "destructuring assignment", but it's not really "destructuring binding" either. It's just plain "destructuring", and the latter only happens as part of signature binding, which is where I picked up the notion of using "binding".

TITIL that signature binding can involve list assignment of individual parameters/variables!

Am I right in saying that (coercion) types need some clean up relative to this?:
```
my Int() @a;
:(@a is copy) := \([42, 99, 'a']);
say @a; # [42 99 a]
```
I wrote that thinking it would throw a typecheck error on the `'a'`.

Am I right in calling what's happening with the initialization of `@a` in `@a is copy` "list assignment"?

codesections:
> Am I right in saying that (coercion) types need some clean up relative to this?:
>
```raku
my Int() @a;
:(@a is copy) := \([42, 99, 'a']);
say @a; # [42 99 a]
```

@raiph note that the code example you provided can be golfed to a version that doesn't involve coercion types or `is copy`:

```raku
my Int @a;
:(@a) := \([42, 99, 'a']); #  or golfed further to     :(@a) := \['a']
say @a; # [42 99 a]        #  or, for the further golfed version:  [a] 
```

I believe that this is another example of the re-binding issue that you and I have been [discussing on Stack Overflow](https://stackoverflow.com/questions/69231506/what-are-the-rules-for-re-binding) – currently, Raku is _very_ permissive in what values are allowed to be re-bound to existing symbols.  As I mentioned, I am not sure at all whether this behavior is intentional or a bug.

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
 }
 =end code
 
+
+=head1 Item assignment, list assignment, and destructuring assignment
+
+There are three types of variable assignment, I<item assignment>, I<list
```
JJ, line 405: Not totally clear why this is not side by side with the other. That implies it's going to take longer to see what was taken out, and what was left. In general, we're not very fond of totally eliminating old documentation, for basically the same reason we're not fond of eliminating old code unless there's a very good reason for doing it.

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
 }
 =end code
 
+
+=head1 Item assignment, list assignment, and destructuring assignment
```
JJ, line 403: Nit: don't repeat "assignment"
Also: headers generate links I would appreciate if you looked up links to the previous `=head2` and redirected them here.

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+
+There are three types of variable assignment, I<item assignment>, I<list
+assignment>, and I<destructuring assignment>.
+
+An item assignment copies a single value from the right-hand side into a Scalar
+variable on the left.  For example:
+
+    my $a;
```
JJ, line 411: OK, this is a good addition of an example

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+    $a = 1; # Item assignment of an Int to a Scalar in $a;
+    say $a; say $a.VAR.^name; # OUTPUT: «1␤Scalar␤»
+
+    my $b = (1, 2, 3); # Item assignment of a List to a Scalar in $a;
+    say $b; say $b.VAR.^name; # OUTPUT: «(1 2 3)␤Scalar␤»
+
+Any assignment with anything other than a simple Scalar variable to the left
```
JJ, line 418: Pleass wrap Scalar with `C<>`

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+from the right hand side and copies them into a L<Positional|/type/Positional>
+(usually an C<Array>); and a C<%>-sigiled variable takes all remaining values on
+the right hand side and copies them into an L<Associative|/type/Associative>
+(usually a C<Hash>).  If those values are C<Pairs>, then the C<Pair>'s key and
+value will become the C<Hash>'s key and value. If the right-hand-side values are
+anything other than C<Pair>s, then the C<%>-sigiled variable will take
+alternating values as keys and values.  Here are some examples:
+
+
+    my ($a) = (1, 2, 3);      # list assignment to a Scalar;
```
JJ, line 435: Good amount of examples here. Thanks.

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+    say $m; say %n; say $o;      # OUTPUT:  «1␤{2 => 3}␤(Any)␤»
+
+    # if a %-sigiled variable can't construct a Hash, you get an error
+    try my ($q, %r) = (1, 2);
+    say $!;
+    # OUTPUT: «Odd number of elements found where hash initializer expected:␤
+    #          Only saw: 2»
+
+
+Both item assignment and list assignment use the equal sign (C<=>) as their
```
JJ, line 463: Nit: repetition of both (actually carried over from original text, but as long as you're changing stuff...)

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+assignments. For instance, in the example below there is one chained assignment
+statement comprising two assignment operators. The assignment to C<@array> is a
+list assignment having a lower precedence than the item assignment to the Scalar
+variable C<$num>. The assignment expression involving the item assignment to the
+Scalar variable C<$num> is thus evaluated first. It returns the assigned value
+C<42>, which in turn forms part of the List C<(42, "str")> constructed by the
+comma operator that also has a higher precedence than the list
+assignment. Finally, the List C<(42, "str")> is list-assigned to C<@array>:
+
+    #`[this statement:] my @array = my $num = 42, "str";
```
JJ, line 502: This should probably be text instead of a comment

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+    # is parsed as      my @array = ( (my $num = 42), "str" )
+    say @array.raku;    # OUTPUT: «[42, "str"]␤» (an Array)
+    say $num.raku;      # OUTPUT: «42␤» (an Int)
+
+Here's a variant:
+
+    #`[this:]      my @foo = (my $bar) = 42, "str";
```
JJ, line 509: OK, literary programming with embedded comments used here... This might be confusing for newcomers. Just put it in different lines.

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+associativity.  For a more elaborate discussion of list assignment in the
+context of variable declarations, see the section on L<declaring a list of
+variables with lexical or package
+scope|/language/variables#index-entry-declaring_a_list_of_variables>.
+
+With list assignment, you can have multiple groups of variables on the left hand
+side and multiple groups of values on the right hand side – but the groupings on
+the left hand side do not impact the assignments.
+
+    #`[this line is] (my ($a, $b), my $c) = (1, (2, 3), 4);
```
JJ, line 533: Same as above. I don't think we've used this elsewhere in the documentation.

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+
+Destructuring assignment allows you to destructure a (potentially nested) group
+of values on the right hand side into a (potentially nested) group of variables
+on the right hand side. Like list assignment, you trigger destructuring
+assignment by having anything other than a Scalar on the left hand side, but
+then follow the LHS by C<:=> rather than C<=>.  Using C<:=> causes the right
+hand side values to be destructured into the LHS variables using the same syntax
+that is used for L<destructuring
+arguments|/type/Signature#Destructuring_arguments> in function signatures.
+
```
JJ, line 551: OK, but `:=` is a binding assignment. So maybe point to its definition, and indicate this is essentially another way of binding?

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+Some examples:
+
+    # Positionals destructure into their elements
```
JJ, line 554: Why use comments when you can directly write text above them?

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+Just as list assignment uses the same operator as item assignment, destructuring
+assignment uses the same characters as the L<Binding|/language/containers#Binding>
+operator, C<:=>.  Nevertheless, destructuring assignment is I<assignment>, and
+still creates Scalars rather than binding directly.
+
+    my ($a, ($b, $c)) := (1, (2, 3));
+    say $a.VAR.^name;                 # OUTPUT: «Scalar␤»
+
+However, unlike list and item assignment, the variables created through
+destructuring assignment are, by default, read only:
```
JJ, line 576: Because what you're doing is binding. See above.

doc/Language/variables.pod6
```diff
@@ -498,6 +399,212 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
...
+
+    my ($a, ($b, $c)) := (1, (2, 3));
+    try $a = 42;
+    say $!;      # OUTPUT: «Cannot assign to a readonly variable or a value␤»
+
+This is exactly the same behavior you get by default with a function's
+parameters and, as in that case, you can opt out of that default by using the
```
JJ, line 583: Because that's also binding.

JJ (approved):
It's really good overall, and I really appreciate the work you've put into this. You might want to address some nits, anyway. And sorry for the delay, and thanks for the low pressure.

codesections:
I'm closing this for now.  #3961 does need a solution, but I no longer think that focusing on "'list' vs 'destructuring'" is quite the right approach.  I'll open a new PR (which may pull in some of the material from this PR) when I have something new.

PR #3963, opened 2021-09-15
Document attribute aliases

codesections:
Part of the spec in Roast [S12-class/attributes.t](https://github.com/Raku/roast/blob/master/S12-class/attributes.t#L23-L33).


JJ (approved):
Who'd have known. Thanks!

PR #3965, opened 2021-09-21
Describe use of non-$ sigils with ^ and : twigils

### LINKED ISSUE #3953: callable placeholder vars are not documented nor indexed
Callable placeholder vars are well specced in S02-types/anon_block.t, S06-signature/mixed-placeholders.t, S06-signature/positional-placeholders.t, S14-traits/routines.t but not mentioned in /language/variables . As a result `&^` is missing from the search index.

Placeholder vars do create a `Signature`, which is not touched in /type/Signature nor does /language/variables link there.

Named callable placeholders are not documented at all and as such not indexed by `&:`. Those are specced ... a lot.

codesections:
This PR describes the use of non-`$` sigils (that is, `@`, `%`, and `&`) with the `^` and `:` twigils.  It also adds `@^`, `%^`, `&^`, `@:`, `%:`, and `&:` to the index.

Closes #3953

### DIFF
--- a/doc/Language/variables.pod6
+++ b/doc/Language/variables.pod6
@@ -241,9 +241,9 @@ variable, that is, where that variable is defined and can be changed.
 
 =end table
 
-X<|$*>
 X<|Dynamically scoped variables>
 =head2 The C<*> twigil
+X<|twigil,*>X<|twigil,$*>X<|twigil,@*>X<|twigil,%*>X<|twigil,&*>
 
 This twigil is used for dynamic variables which are looked up through the
 caller's, not through the outer, scope. Look at the example below.N<The example
@@ -307,7 +307,7 @@ symbol tables introduced with C<our> are two orthogonal issues.
 
 
 =head2 The C<?> twigil
-X<|$?>
+X<|twigil,?>X<|twigil,$?>X<|twigil,@?>X<|twigil,%?>X<|twigil,&?>
 
 Compile-time variables may be addressed via the C<?> twigil. They are known
 to the compiler and may not be modified after being compiled in. A popular
@@ -321,7 +321,7 @@ For a list of these special variables, see
 L<compile-time variables|/language/variables#Compile-time_variables>.
 
 =head2 The C<!> twigil
-X<|$!>
+X<|twigil,!>X<|twigil,$!>X<|twigil,@!>X<|twigil,%!>X<|twigil,&!>
 
 L<Attributes|/language/objects#Attributes> are variables that exist per instance
 of a class. They may be directly accessed from within the class via C<!>:
@@ -344,7 +344,7 @@ L<object orientation|/language/objects>.
 
 
 =head2 The C<.> twigil
-X<|$.>
+X<|twigil,.>X<|twigil,$.>X<|twigil,@.>X<|twigil,%.>X<|twigil,&.>
 
 The C<.> twigil isn't really for variables at all. In fact, something along
 the lines of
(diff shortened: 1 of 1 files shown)


jubilatious1:
Quote: "twigil can be combined with any sigil to create a placeholder variable"

Clarifying terminology: does a secondary non-alnum such as a `^` create the twigil? My seat-of-the-pants understanding of the etymology of "twigil" is that it has it's origin in the word "twin" (...with the "s" in sigil maybe signifying `singular`?). 

So my guesstimate is that you need two non-alnums to create a twigil, most commonly with the `^` caret. If i'm way off base here, please ignore. 

@TimToady

codesections:
In the code `my $^variable`, the docs currently describe the `^` – that is, the second character in `$^variable` – as the `twigil`.  For example, the [table of Twigils](https://docs.raku.org/language/variables#index-entry-Twigil) lists each twigil as a single character that can be used after a sigil.

I'm not opposed to describing the two-character cluster as the twigil (so that we'd say that `$^variable` has a `$^` twigil, instead of saying that it has a `$` sigil and a `^` twigil) but that's not how the docs currently do it.  I'm not sure I see a whole lot to be gained by making that change but, either way, it seems like it'd be separate from this PR.

jubilatious1:
Hi Daniel! Honestly I don't know either way, whether the single-leading-character signs are "Sigils" and the double-aka-twin-leading-character signs are "Twigils", or not.

I'd like it to be that way.

Regardless, if not we maybe should start thinking more about how to describe this accurately. If it is as you describe, so be it. Otherwise, there is a bit of a middle ground here if we start saying "`^`-class twigils" and "`:`-class twigils", etc. Again, just my 2-cents.

@TimToady

JJ:
TBH, I understood the same as @jubilatious1. Obviously, we have to stick to the published definition.

doc/Language/variables.pod6
```diff
@@ -411,8 +414,16 @@ Placeholder variables cannot have type constraints or a variable name with a
 single upper-case letter (this is disallowed to enable catching some
 Perl-isms).
 
+The C<^> twigil can be combined with any sigil to create a placeholder variable
+with the appropriate type constraint.  Thus C<@^positional>,C<%^associative>,and
```
JJ, line 418: Wouldn't you describe "positional" and "associative" as "roles" more than type constraints? After all, it can be any type that does that role.

doc/Language/variables.pod6
```diff
@@ -448,6 +459,11 @@ only one (but refers to it twice).  This can be especially significant with cons
 such as C<with>, C<for>, and C<if> that are often used without much consideration of
 the fact that they create blocks.
 
+Just like the C<^>twigil, the C<:> twigil can can be combined with any sigil; using
+C<:> with a sigil will create a formal named parameter of the appropriate type.  Thus
```
JJ, line 463: Again, wouldn't you say "with the proper|corresponding role" would be more precise here than "the appropriate type"?

JJ (approved):
Thanks a lot for this. It looks good, however you might want to consider the two comments I've made.

codesections:
@jj thanks.  I'm not 100% sure whether it's better to describe "positional" and "associative" as "roles" or "type constraints" – they constrain the type to something that `does` that role.  I edited the PR to just link to the [Sigils table](https://docs.raku.org/language/variables#Sigils), which avoids the issue and is probably clearer anyway.

I also added a few more `X<index>` entries, which should help with the issue @jubilatious1 brought up – it doesn't resolve it, but at least if someone is thinking of `@^` as a twigil (rather than a sigil followed by a twigil), they'll now get something in the search box.

doc/Language/variables.pod6
raiph: @JJ

The `$` sigil's type constraint isn't a role.

"type constraint" seems to me to be the appropriate generic for this reason alone.

I would also think a newbie is more likely to correctly  guess what a "type constraint" is (a constraint on a variable's type) than guess what a "role" is (a type) *and* guess why it's relevant (it's a type constraint for the associated variable) *and* recognize that `$`'s default type constraint isn't a role.

PR #3968, opened 2021-09-28
Add output to "multiple heredoc" example

codesections:
The docs previously stated that you can start multiple heredocs on a single line, but neither said what effect that had nor showed the output in the accompanying example.

This PR adds both an explanation and example output.


tbrowder:
Looks good to me. Thanks!

PR #3969, opened 2021-09-28
Clarify that enum iteration is unordered

codesections:
Because enums are `Map`s, they do not iterate in any defined order.  But the docs didn't mention this, and provided an example that suggested ordered iteration.  This PR adds a note to that effect and updates the example.

### DIFF
--- a/doc/Language/typesystem.pod6
+++ b/doc/Language/typesystem.pod6
@@ -788,7 +788,12 @@ the keys.
     enum E <one two>;
     my @keys = E::.values;
     say @keys.map: *.raku;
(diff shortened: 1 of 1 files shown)


vrurg:
Enums are not `Map`s. Depending on enum values they can be either numeric or stringy. The reason for undefined order of keys is method `Any.keys` candidate for enums uses method `enums` which, in turn, returns a `Map`.

codesections:
@vrurg, thanks for the correction.  I've correspondingly updated the PR.

JJ (approved):
Thanks!

PR #3973, opened 2021-10-17
Document pop, shift, push and append more concisely

dumarchie:
## The problem

In #3702 I included various signatures in "Defined as" routine introductions that were exact copies of those found in the Rakudo implementation. I now think that was not a good idea, because

* some of the parameter names in these signatures are quite meaningless, and
* some of the multi implementations exist only for performance reasons.

Consequently, these introductions did not provide a good gist of these routines.

## Solution provided

I used more meaningful parameter names and removed multi signatures that are functionally irrelevant. Additionally, I edited the documentation of these routines where it referred to parameters or I just thought it could be improved.

### DIFF
--- a/doc/Type/independent-routines.pod6
+++ b/doc/Type/independent-routines.pod6
@@ -1224,6 +1224,8 @@ container is empty.
 
 Example:
 
+    say shift [1,2]; # OUTPUT: «1␤»
+
 =for code
 my @a of Int = [1];
 say shift @a; # OUTPUT: «1␤»


lizmat:
But you can not push anything that's not an native int to a native int array.  An itemized native int array is not a native int.  So either that should throw, or DWIM.  I guess we're arguing on whether DWIM isn't more of a WAT ?

dumarchie:
B.t.w. there's also an issue with `.push` on native arrays:
```
> my int @a = 1, 2;
[1 2]
> my int @b = 3, 4;
[3 4]
> @a.push: @b.Slip, @b.Slip;
[1 2 3 4 3 4]
> @a.push: @b.Slip;
Type check failed in push to int array; expected int but got Slip (slip(3, 4))
  in block <unit> at <unknown file> line 1
```

PR #3975, opened 2021-10-30
Improve documentation of the list routine

dumarchie:
The documentation of the `list` subroutine is a good place to explain the
single argument rule yet again.

Most core types implementing a `list` method only accept an object instance
as the invocant. If the core type implements `list` as a multi method,
listing the type object is handled by

    multi method list(Any:U: --> List)

### DIFF
--- a/doc/Type/List.pod6
+++ b/doc/Type/List.pod6
@@ -207,7 +207,7 @@ list contextualizer|/type/Any#index-entry-@_list_contextualizer>:
     put list($tuple<>).raku;    # OUTPUT: «(1, 2)␤»
     put list(@$tuple).raku;     # OUTPUT: «(1, 2)␤»
 
-Note that listing the elements of a type object may not do what you expect:
+Note that converting a type object to a list may not do what you expect:
 
     put List.list.raku;         # OUTPUT: «(List,)␤»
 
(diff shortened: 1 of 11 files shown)


doc/Type/List.pod6
```diff
@@ -175,6 +175,69 @@ If C<$topic> is I<not> an L<Iterable|/type/Iterable>, returns the invocant if
...
+    put list(1, 2).raku;        # OUTPUT: «(1, 2)␤»
+
+There are other ways to list the elements of an itemized single argument. For
+example, you can L«decontainerize|/routine/<>» the argument or use the L<C<@>
+list contextualizer|/type/Any#index-entry-@_list_contextualizer>:
+
+    put list($tuple<>).raku;    # OUTPUT: «(1, 2)␤»
+    put list(@$tuple).raku;     # OUTPUT: «(1, 2)␤»
+
+Note that listing the elements of a type object may not do what you expect:
```
JJ, line 210: Maybe just use "casting to a list" here?

doc/Type/List.pod6
```diff
@@ -175,6 +175,69 @@ If C<$topic> is I<not> an L<Iterable|/type/Iterable>, returns the invocant if
...
+is provided by L<C<Any>|/routine/list#(Any)_method_list>. That candidate returns
+a list with one element: the type object self. If you're developing a collection
+type whose type object should be a valid representation of an empty collection,
+you may want to provide your own candidate for undefined invocants or override
+the C<Any:> candidates with an "only" method. For example:
+
+    my class LinkedList {
+        has $.value;            # the value stored in this node
+        has LinkedList $.next;  # undefined if there is no next node
+
```
JJ, line 224: not totally sure about this... If you have blank lines in the middle of a code block, it will create a new one. Pod does not understand indentation. To be on the safe side, just enclose it in a code block.

JJ (approved):
Great addition to the documentation. You might want to address the nits.

doc/Type/List.pod6
dumarchie: Good point. I'll change this.

doc/Type/List.pod6
dumarchie: In Raku we "coerce" rather than "cast", don't we? But coercion typically results in a specific type, so how about "converting a type object to a list"?

JJ: Whatever that's not "listing" is probably OK. Your call.

PR #3980, opened 2021-11-13
Talk about collections rather than containers

dumarchie:
In Raku parlance, a container generally refers to a scalar container. 
Arrays and hashes are mutable _collections_ that store values in scalar 
containers. To avoid confusion it's better to refer to these composite
types as  collections.

### DIFF
--- a/doc/Type/independent-routines.pod6
+++ b/doc/Type/independent-routines.pod6
@@ -1200,9 +1200,9 @@ Routines that manipulate arrays and other mutable collections.
 
 Defined as:
 
-    multi sub pop(@collection) is raw
+    multi sub pop(@a) is raw
 
-Calls method C<pop> on the C<Positional> collection. That method is supposed to
+Calls method C<pop> on the C<Positional> argument. That method is supposed to
 remove and return the last element, or return a L<C<Failure>|/type/Failure> if
 the collection is empty.
 
@@ -1213,9 +1213,9 @@ for an example.
 
 Defined as:
 
-    multi sub shift(@collection) is raw
+    multi sub shift(@a) is raw
 
-Calls method C<shift> on the C<Positional> collection. That method, on
+Calls method C<shift> on the C<Positional> argument. That method, on
 a mutable collection that actually implements it (such as an
 L<C<Array>|/routine/shift#(Array)_method_shift> or a
 L<C<Buf>|/routine/shift#(Buf)_method_shift>), is supposed to remove
@@ -1235,40 +1235,41 @@ say shift @a; # ERROR: «Cannot shift from an empty Array[Int]␤»
 
 Defined as:
 
-    multi sub push(\collection, **@values is raw)
-    multi sub push(\collection, \arg)
+    multi sub push(\a, **@b is raw)
+    multi sub push(\a, \b)
 
-Calls method C<push> on the collection, which is supposed to add the provided
-values to the end of the collection or parts thereof. See the documentation of
-the L<C<Hash> method|/routine/push#(Hash)_method_push> for an example where
+Calls method C<push> on the first argument, passing the remaining arguments.
+Method C<push> is supposed to add the provided values to the end of the
+collection or parts thereof. See the documentation of the
+L<C<Hash> method|/routine/push#(Hash)_method_push> for an example where
 indirection via this subroutine can be helpful.
 
-The C<push> method of the collection is supposed to flatten all arguments of
-type C<Slip>. Therefore, if you want to implement a conforming method for a new
-collection type, it should behave as if its signature was just:
+The C<push> method is supposed to flatten all arguments of type C<Slip>.
+Therefore, if you want to implement a conforming method for a new collection
+type, it should behave as if its signature was just:
 
-    multi method push(::?CLASS:D: **@values --> ::?CLASS:D)
+    multi method push(::?CLASS:D: **@values is raw --> ::?CLASS:D)
 
 Autovivification to an instance of the new type is L<provided by the default
 base class|/routine/append#(Any)_method_append> if the new type implements the
 C<Positional> role. If the new type is not C<Positional>, autovivification can
-be implemented by an additional multi method with a signature like
+be supported by an adding a multi method with a signature like
 
-    multi method push(::?CLASS:U: **@values --> ::?CLASS:D)
+    multi method push(::?CLASS:U: **@values is raw --> ::?CLASS:D)
 
 =head2 sub append
 
 Defined as:
 
-    multi sub append(\collection, **@values is raw)
-    multi sub append(\collection, \arg)
+    multi sub append(\a, **@b is raw)
+    multi sub append(\a, \b)
 
-Calls method C<append> on the collection, which is supposed to add the provided
-values to the end of the collection or parts thereof. Unlike method C<push>,
-method C<append> should follow the L<single argument
-rule|/language/functions#Slurpy_conventions>. So if you want to implement a
-conforming method C<append> for a new collection type, it should behave as if
-its signature was just:
+Calls method C<append> on the first argument, passing the remaining arguments.
+Method C<append> is supposed to add the provided values to the end of the
+collection or parts thereof. Unlike method C<push>, method C<append> should
+follow the L<single argument rule|/language/functions#Slurpy_conventions>. So
+if you want to implement a conforming method C<append> for a new collection
+type, it should behave as if its signature was just:
 
     multi method append(::?CLASS:D: +values --> ::?CLASS:D)
 
--- a/doc/Type/independent-routines.pod6
+++ b/doc/Type/independent-routines.pod6
@@ -1194,17 +1194,17 @@ hashes, whose use is advised.
 
 =head1 Array manipulation
 
-Routines that manipulate arrays and other containers.
+Routines that manipulate arrays and other mutable collections.
 
 =head2 sub pop
 
 Defined as:
 
-    multi sub pop(@container) is raw
+    multi sub pop(@collection) is raw
 
-Calls method C<pop> on the C<Positional> container. That method is supposed to
+Calls method C<pop> on the C<Positional> collection. That method is supposed to
 remove and return the last element, or return a L<C<Failure>|/type/Failure> if
-the container is empty.
+the collection is empty.
 
 See the documentation of the L<C<Array> method|/routine/pop#(Array)_method_pop>
 for an example.
@@ -1213,14 +1213,14 @@ for an example.
 
 Defined as:
 
-    multi sub shift(@container) is raw
+    multi sub shift(@collection) is raw
 
-Calls method C<shift> on the C<Positional> container. That method, on
-a mutable container that actually implements it (such as an
+Calls method C<shift> on the C<Positional> collection. That method, on
+a mutable collection that actually implements it (such as an
 L<C<Array>|/routine/shift#(Array)_method_shift> or a
 L<C<Buf>|/routine/shift#(Buf)_method_shift>), is supposed to remove
 and return the first element, or return a L<C<Failure>|/type/Failure> if the
-container is empty.
+collection is empty.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


JJ:
> In Raku parlance, a container generally refers to a scalar container. Arrays and hashes are mutable _collections_ that store values in scalar containers. To avoid confusion it's better to refer to these composite types as collections.

Not really. A container in Raku is generally called "variable" somewhere else; it tries to distinguish the content, a value, from what contains it, which is also an object by and of itself. In general, I would appreciate a previous issue to arrive to a consensus on this kind of things. However, I'll take a look at the PR because it's probably more straightforward than that.

doc/Type/independent-routines.pod6
```diff
@@ -1194,17 +1194,17 @@ hashes, whose use is advised.
 
 =head1 Array manipulation
 
-Routines that manipulate arrays and other containers.
+Routines that manipulate arrays and other mutable collections.
```
JJ, line 1197: OK, I see what you mean now.

doc/Type/independent-routines.pod6
```diff
@@ -1194,17 +1194,17 @@ hashes, whose use is advised.
...
 
 =head2 sub pop
 
 Defined as:
 
-    multi sub pop(@container) is raw
+    multi sub pop(@collection) is raw
 
-Calls method C<pop> on the C<Positional> container. That method is supposed to
+Calls method C<pop> on the C<Positional> collection. That method is supposed to
```
JJ, line 1205: alt: positional "argument" or "collection passed as a positional argument"

doc/Type/independent-routines.pod6
```diff
@@ -1213,14 +1213,14 @@ for an example.
 
 Defined as:
 
-    multi sub shift(@container) is raw
+    multi sub shift(@collection) is raw
```
JJ, line 1216: This is simply defined as 

https://github.com/rakudo/rakudo/blob/db074a8bc49a0dbcdfc4ecd392c7eaadb0c078ed/src/core.c/array_operators.pm6#L26

As long as we're changing it, the best is to just use the original, as is the default policy.

doc/Type/independent-routines.pod6
```diff
@@ -1235,17 +1235,17 @@ say shift @a; # ERROR: «Cannot shift from an empty Array[Int]␤»
 
 Defined as:
 
-    multi sub push(\container, **@values is raw)
-    multi sub push(\container, \arg)
+    multi sub push(\collection, **@values is raw)
```
JJ, line 1238: Ditto here. It uses \a as an argument.

JJ (approved):
Great clarificaton. You might want to address the nits before merging, anyhow.

doc/Type/independent-routines.pod6
dumarchie: In your review of PR #3973 you still agreed that it was better to use more meaningful parameter names.

I don't mind changing the parameter name if you now think that `@a` is better after all, because the `@` sigil probably adds enough meaning. But as an aside, I think that Rakudo should not restrict the argument to `Positional` as non-`Positional` collection types may implement a `shift` method too. And I don't think that `\a` would be meaningful enough. 

So should I change the parameter to `@a` or not?

doc/Type/independent-routines.pod6
dumarchie: Ditto here. An example of a non-`Positional` collection type implementing `.pop` would be a heap.

dumarchie: In this case using `\a` would really make the signature less meaningful.

doc/Type/independent-routines.pod6
dumarchie: Hmm, this is a bit tricky. We have a positional argument that should do the `Positional` role.  How about "the `@collection` passed as a positional argument"?

This b.t.w. illustrates why I prefer parameter name `@collection` instead of `@a`, because I don't think there's as much meaning in the phrase "the `@a` passed as a positional argument".

doc/Type/independent-routines.pod6
JJ: The problem, again, is unambiguously making something that is really better. "Collection" is implicit in the sigil, for instance. And that's also correct, it's not positional, it's "thing-with-shift". I seem to remember I opened an issue somewhere about that, to create more atomic roles. Anyway, to the point, it's really up to you. I really appreciate the effort for improving the documentation you've put in.

doc/Type/independent-routines.pod6
JJ: Well, it's what it is. If we want to get really picky about it, it's `\object_can_push`. Which is why in the absence of a consensus, it's really the best to stick to the source. But, again, your call. We at least agree that the original `\container` was not a good idea.

doc/Type/independent-routines.pod6
JJ: as a `Positional` argument. It's a `Positional` argument that's also a positional argument. There's a tongue-twister for you :smile_cat:

raiph:
Fwiw I'd love to see introduction of the word "collections" as a generic term covering lists and hashes. I've seen it used to mean this in a lot of other PLs.

> A container in Raku is generally called "variable" somewhere else; it tries to distinguish the content, a value, from what contains it, which is also an object by and of itself.

That may be where we've ended up, and I may have contributed to that, but I think it's unfortunate and that it's wise to move toward using the word container strictly to mean `Scalar` (and perhaps `Proxy` too).

> In general, I would appreciate a previous issue to arrive to a consensus on this kind of things.

I'm with you on the need for consensus but would say a PR like this one is a classic healthy way to get to that. By "like this" I mean one that's simple, small, done well, and with a humble attitude, all of which I think apply here.

> However, I'll take a look at the PR because it's probably more straightforward than that.

Indeed. Your follow ups are notably sweet, and I thank you for your continued careful stewardship of the doc.

I love this shift. :)

doc/Type/independent-routines.pod6
dumarchie: I like the idea of more atomic roles. Then, for example, we could document that the "Shiftable" role has a stub

    method shift(Shiftable:D:)

and change the subroutine form to something like

    multi sub shift(Shiftable:D \object)

which will probably give us a more awesome error message when attempting to `shift` something that is not shiftable.

Maybe we could even combine such roles into more complex ones, like:

    role Stacky does Pushable does Poppable {}
    role Queuey does Pushable does Shiftable {}

so we can have a

    class Array does Stacky does Queuey ...

showing what an `Array` can be used for.

JJ:
> Fwiw I'd love to see introduction of the word "collections" as a generic term covering lists and hashes. I've seen it used to mean this in a lot of other PLs.
> 

Well... on the pure terminological side, a collection might include anything that's not "atomic". Would Junctions be collections? Hashmaps? On the low-level side, however, it breaks down to "thing-with-a-push", "thint-with-a-shift" and so on an so forth. Is an `Iterable` a collection? Anyway, that's something that would be long to discuss. In this PR, it's quite clear what we mean by that.

> > A container in Raku is generally called "variable" somewhere else; it tries to distinguish the content, a value, from what contains it, which is also an object by and of itself.
> 
> That may be where we've ended up, and I may have contributed to that, but I think it's unfortunate and that it's wise to move toward using the word container strictly to mean `Scalar` (and perhaps `Proxy` too).
> 

Well, if I had to steer it towards something, I'd go for something "meta". A container is a meta-object, because it adds structure and metadata (roles, HOWs) to single values.

> > In general, I would appreciate a previous issue to arrive to a consensus on this kind of things.
> 
> I'm with you on the need for consensus but would say a PR like this one is a classic healthy way to get to that. By "like this" I mean one that's simple, small, done well, and with a humble attitude, all of which I think apply here.
> 

As I have commented, the PR is great work. The underlying issue, however, would probably need an, er, different issue. Think that what we're discussing here will be lost when merged.

> > However, I'll take a look at the PR because it's probably more straightforward than that.
> 
> Indeed. Your follow ups are notably sweet, and I thank you for your continued careful stewardship of the doc.
> 
> I love this shift. :)

:+1:

dumarchie:
> Fwiw I'd love to see introduction of the word "collections" as a generic term covering lists and hashes. I've seen it used to mean this in a lot of other PLs.

Then we should probably rename Rakudo's  **array_operators.pm6** to **collection_operators.pm6** (or **collection_operators.rakumod**). I can change the heading in the independent routines documentation from "Array manipulation" to "Collection operators" already, if we agree that is a better name.

JJ:
> > Fwiw I'd love to see introduction of the word "collections" as a generic term covering lists and hashes. I've seen it used to mean this in a lot of other PLs.
> 
> Then we should probably rename Rakudo's **array_operators.pm6** to **collection_operators.pm6** (or **collection_operators.rakumod**). I can change the heading in the independent routines documentation from "Array manipulation" to "Collection operators" already, if we agree that is a better name.

Er... No renaming, please. Again, implies loss of incoming links. Creating a new one that would point to all included concepts would be OK.

dumarchie:
> > > Fwiw I'd love to see introduction of the word "collections" as a generic term covering lists and hashes. I've seen it used to mean this in a lot of other PLs.
> > 
> > 
> > Then we should probably rename Rakudo's **array_operators.pm6** to **collection_operators.pm6** (or **collection_operators.rakumod**). I can change the heading in the independent routines documentation from "Array manipulation" to "Collection operators" already, if we agree that is a better name.
> 
> Er... No renaming, please. Again, implies loss of incoming links. Creating a new one that would point to all included concepts would be OK.

I searched for "Array_manipulation" in the whole project and found nothing, so I don't think there are any links inside the project. I created the heading not so long ago so I doubt there are incoming links from other resources.

JJ:
> > > > Fwiw I'd love to see introduction of the word "collections" as a generic term covering lists and hashes. I've seen it used to mean this in a lot of other PLs.
> > > 
> > > 
> > > Then we should probably rename Rakudo's **array_operators.pm6** to **collection_operators.pm6** (or **collection_operators.rakumod**). I can change the heading in the independent routines documentation from "Array manipulation" to "Collection operators" already, if we agree that is a better name.
> > 
> > 
> > Er... No renaming, please. Again, implies loss of incoming links. Creating a new one that would point to all included concepts would be OK.
> 
> I searched for "Array_manipulation" in the whole project and found nothing, so I don't think there are any links inside the project. I created the heading not so long ago so I doubt there are incoming links from other resources.

The principle is still the same. Pages disappear from search engines, lose their ranking. We aren't even fond of losing fragments in URLs. There's an alternative, which I'm not fond of, because it implies ops, that's returning a redirect. Also, let me remind you we're not agreed either in what's a collection, or if we should call it that way. Please open an issue and we'll discuss terminology, and what to do about it either.

dumarchie:
> > > > > Fwiw I'd love to see introduction of the word "collections" as a generic term covering lists and hashes. I've seen it used to mean this in a lot of other PLs.
> > > > 
> > > > 
> > > > Then we should probably rename Rakudo's **array_operators.pm6** to **collection_operators.pm6** (or **collection_operators.rakumod**). I can change the heading in the independent routines documentation from "Array manipulation" to "Collection operators" already, if we agree that is a better name.
> > > 
> > > 
> > > Er... No renaming, please. Again, implies loss of incoming links. Creating a new one that would point to all included concepts would be OK.
> > 
> > 
> > I searched for "Array_manipulation" in the whole project and found nothing, so I don't think there are any links inside the project. I created the heading not so long ago so I doubt there are incoming links from other resources.
> 
> The principle is still the same. Pages disappear from search engines, lose their ranking. We aren't even fond of losing fragments in URLs. There's an alternative, which I'm not fond of, because it implies ops, that's returning a redirect. Also, let me remind you we're not agreed either in what's a collection, or if we should call it that way. Please open an issue and we'll discuss terminology, and what to do about it either.

Okay, I'll keep that change out of this PR. Please let me know if you agree with the changes in my second commit.

JJ:
> Okay, I'll keep that change out of this PR. Please let me know if you agree with the changes in my second commit.

Like the proverbial credit card, it was already pre-approved. I really appreciate the effort you have put into accommodating my nits.

raiph:
@JJ

Yeah, I got carried away with excitement at @dumarchie having a go, and have already inappropriately over-simplified.

Some things I think we can agree on are the goodness of the desire to improve things, and of @dumarchie's intent and writing skills, and of the need for patient discussion and nuanced thinking about the conceptual, terminological, and practical complications (don't break links!) related to what are really a fundamentally thorny and intertwined mess of conundrums.

> Would Junctions be collections?

I'll mostly postpone my further commentary about the term "collections" for now. The issue is, as you've said, complicated.

> A container is a meta-object, because it adds structure and metadata (roles, HOWs) to single values.

I would be quite unhappy to see us use what is already a precisely defined and scary term ("meta-object") to mean anything other than what it currently does in Raku. It's just a scarier name for a `HOW` object, which is exactly the same as what the term "meta-object" means in PL theory, and as used in related terms like "meta-object protocol". I get that we aren't scared to buck norms and trends among PL languages, but I don't see good things coming of making "meta-object" more ambiguous and mind-bending in Raku than it already is in Raku and other PLs.

Perhaps consistently using the specific term "`Variable`-value" would work? `Variable` is the name of the Raku type that directly corresponds to a variable to which a value of the sort we're talking about can be bound. A `Variable`-value doesn't *have* to be so bound, but the name is suggestive of what one is dealing with. Anyhoo, just an idea.

> As I have commented, the PR is great work. The underlying issue, however, would probably need an, er, different issue. Think that what we're discussing here will be lost when merged.

Yeah. I've written this comment fully aware that it's plausible that, at most, only you and/or @dumarchie will read it.

Also, no need to reply. I'm really just thinking out loud...

PR #3983, opened 2021-11-20
101-basics: correct secondary/primary designation

flwyd:
The previous paragraph points out that matches is the primary sort criterion.
This line takes advantage of stable sort by sorting by the _secondary_ criterion (sets) _first_ and then sorts by the _primary_ (matches).


JJ (commented):
Good catch! Thanks.

PR #3986, opened 2021-12-01
Add $#array equivalent

Smylers:
## The problem

I was converting Perl 5 code to Raku and didn't know what the equivalent of `$#array` was, to find the max element of the array. I looked in the Perl-to-Raku guides, but couldn't find the string `$#` in any of them.

## Solution provided

The `.end` method appears to be what I was looking for. That was hard to find. Add a mention in the Perl-to-Raku overview, so others can find it.

I'm not completely sure this is definitely the best place to put it, but it's with other array-index things, and having it somewhere is better than not having it at all.

### DIFF
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -404,7 +404,6 @@ An array's biggest index is now available with the C<.end> method:
 say $#item;    # Perl
 =for code :preamble<no strict;>
 say @item.end; # Raku
-=end item
 
 =head2 {} Hash indexing/slicing
 
--- a/doc/Language/5to6-nutshell.pod6
+++ b/doc/Language/5to6-nutshell.pod6
@@ -398,6 +398,14 @@ operator rather than a special syntactic form, and thus L<checking for
 existence of elements|#exists> and L<unsetting elements|#delete> is done
 with adverbs.
 
+An array's biggest index is now available with the C<.end> method:
+
+=for code :lang<perl>
+say $#item;    # Perl
+=for code :preamble<no strict;>
+say @item.end; # Raku
+=end item
+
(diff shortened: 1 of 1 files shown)


lizmat (approved):

doc/Language/5to6-nutshell.pod6
```diff
@@ -398,6 +398,14 @@ operator rather than a special syntactic form, and thus L<checking for
 existence of elements|#exists> and L<unsetting elements|#delete> is done
 with adverbs.
 
+An array's biggest index is now available with the C<.end> method:
+
+=for code :lang<perl>
+say $#item;    # Perl
+=for code :preamble<no strict;>
```
JJ, line 405: I would probably have defined `@item`, but OK.

JJ (approved):

doc/Language/5to6-nutshell.pod6
```diff
@@ -398,6 +398,14 @@ operator rather than a special syntactic form, and thus L<checking for
...
+say @item.end; # Raku
+=end item
```
JJ, line 407: You need to delete this; `=for` does not need closing.

JJ (changes requested):

doc/Language/5to6-nutshell.pod6
Smylers: Ah, sorry. It was a copy-and-paste from the block above. I just changed the code and hoped the Pod would still be valid.

Smylers:
Is there a way in the GitHub interface I can squash these into 1 commit? I'm not very familiar with pull requests or the GitHub website interface.

lizmat:
The person merging this PR has the option to squash it into a single commit

PR #3988, opened 2021-12-02
Correct equivalent for split without a pattern

Smylers:
## The problem

In Perl, split without a pattern is different from split "" — telling users of the former to switch to the latter changes from splitting by words to splitting by characters.

## Solution provided

Suggest using words instead, which is probably the most useful Raku alternative, while also mentioning the correct explicit pattern.

### DIFF
--- a/doc/Language/5to6-perlfunc.pod6
+++ b/doc/Language/5to6-perlfunc.pod6
@@ -1935,8 +1935,10 @@ also now works as a method: C<"a;b;c".split(';')>
 
 =item split
 
-The zero argument version must now be called with an explicit empty string, as
-described above.
+C<split> now requires a pattern. For the equivalent of Perl's behaviour
+of splitting on whitespace when no pattern is specified, use
+L<words|/routine/words> (or specify the pattern explicitly with
+C<split /\s+/>).
 
 X<| sprintf - perlfunc>
 =head2 sprintf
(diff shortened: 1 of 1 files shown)


lizmat (approved):

raiph (changes requested):
Use of `words` is correct, but `split /\s+/` will not strip leading/trailing whitespace.

Smylers:
> `split /\s+/` will not strip leading/trailing whitespace.

Ah, OK. Is there an exact equivalent pattern to Perl's behaviour?

Obviously there are several patterns which might exhibit the same behaviour in a user's specific scenario (such as there not being any leading or trailing spaces, or all the whitespace being actual space characters, or all being horizontal whitespace, or all being a single character), so I've gone for not mentioning any pattern in particular.

raiph (commented):
[With apologies. Perhaps there's a way I could have just directly patched your patch to save you having to apply one of my two alternative fixes, but if there is, I don't yet know how it.]

Tiny nit, with two possible fixes:

* s/appropiate/appropriate/

or 

> (or specify a C<split> with C</\s+/> as the pattern and C<:skip-empty> as a named argument).

instead of:

> (or specify the appropiate space-matching pattern explicitly).

raiph (approved):

PR #3990, opened 2021-12-04
Fix error re: delegation and simplify some text.

rakool:
The doc says that  an object delegates all methods to one delegatee.

 The quoting of delegator /delegatee/delegation is inconsistent.

I have simplified the language for finding 'current' and 'original' unhelpful.


JJ (approved):
Looks fine. Thanks!

PR #3991, opened 2021-12-04
\d is for decimal

codesections:
As we correctly note [further down the page](https://docs.raku.org/language/regexes#Predefined_character_classes), `\d` matches _decimal digits_ (Unicode `Nd`) not all numbers (Unicode `N`).  This PR fixes the description here, adds an example of where that makes a difference, and provides a way to match against Numbers if that's what the user meant.

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -472,16 +472,16 @@ character that is not whitespace.
 
 =head3 X<C<\d> and C<\D>|regex,\d;regex,\D>
 
-C<\d> matches a single digit (Unicode property C<N>) and C<\D> matches a
-single character that is not a digit.
+C<\d> matches a single decimal number (Unicode property C<Nd>) and C<\D> matches a
+single character that is not a decimal number.
 
     'ab42' ~~ /\d/ and say ~$/;     # OUTPUT: «4␤»
     'ab42' ~~ /\D/ and say ~$/;     # OUTPUT: «a␤»
 
 Note that not only the Arabic digits (commonly used in the Latin
-alphabet) match C<\d>, but also digits from other scripts.
+alphabet) match C<\d>, but also decimal numbers from other scripts.
 
-Examples of digits are:
+Examples of decimal numbers are:
 
 =begin code :lang<text>
 U+0035 5 DIGIT FIVE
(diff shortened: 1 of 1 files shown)


doc/Language/regexes.pod6
```diff
@@ -472,16 +472,16 @@ character that is not whitespace.
 
 =head3 X<C<\d> and C<\D>|regex,\d;regex,\D>
 
-C<\d> matches a single digit (Unicode property C<N>) and C<\D> matches a
-single character that is not a digit.
+C<\d> matches a single decimal number (Unicode property C<Nd>) and C<\D> matches a
```
JJ, line 475: I'd prefer the "digit" term to appear, if only as mnemonic. Full name in the unicode standard is comething like digit with decimal radix, but I'd very much prefer "digit" to "number" here.
Also, what it said initially is not really false. `N` includes `Nd`, so it might be worth the while to mention it.

doc/Language/regexes.pod6
```diff
@@ -490,6 +490,21 @@ U+0E53 ๓ THAI DIGIT THREE
 U+17E5 ៥ KHMER DIGIT FIVE
 =end code
 
+Also note that "decimal number" is a narrower category than "Number" because (Unicode) numbers include
+not only decimal numbers (C<Nd>) but also leter numbers (C<Nl>) and other numbers (C<No>)
```
lizmat, line 494: s/leter/letter/ ?

doc/Language/regexes.pod6
```diff
@@ -490,6 +490,21 @@ U+0E53 ๓ THAI DIGIT THREE
...
+Examples of Unicode numbers that are not decimal numbers include:
+
+=begin code :lang<text>
+U+2464 ⑤ CIRCILED DIGIT FIVE
```
JJ, line 498: typo: `CIRCLED`

JJ (changes requested):
Mainly the typo, but please consider also my other nits.

doc/Language/regexes.pod6
codesections: > I'd prefer the "digit" term to appear, if only as mnemonic. Full name in the unicode standard is something like digit with decimal radix, but I'd very much prefer "digit" to "number" here.

It looks like the Unicode name is "Number, decimal digit" (according to [tbl. 4-4](https://www.unicode.org/versions/Unicode14.0.0/ch04.pdf#G124206), anyway) and that's the name that we use on the [syntax](https://docs.raku.org/language/syntax#Ordinary_identifiers) page.  Since that still has the mnemonic "digit" in it, I changed to that here.

> Also, what it said initially is not really false. `N` includes `Nd`

Yeah, but it `N` is broader: `/<:N>/` matches things like `' ⑤'`, which aren't matched by `/\d/` (and are matched by `/\D/').

JJ (approved):
:+1:

tbrowder:
Late to the dance, but wouldn't an example of finding only Arabic digits be useful just for completeness?

JJ:
> Late to the dance, but wouldn't an example of finding only Arabic digits be useful just for completeness?

Would care to create another PR? Or to commit it directly?

PR #3994, opened 2021-12-05
Remove &eager sub form

codesections:
This removes the `&eager` Sub, which does not seem to exist.

(There's an `.eager` method and an `eager` statement prefix, but no Sub)


JJ (approved):
Thanks!

PR #3996, opened 2021-12-08
Document sequences like 'aa'…'cc'

codesections:
The sequence `'aa'…'cc'` does not use a `*.succ` default generator  (and this is correct [per Roast](https://github.com/Raku/roast/blob/master/S03-sequence/nonnumeric.t#L58)).

See also [Can Raku range operator on strings mimic Perl's behaviour?](https://stackoverflow.com/questions/70239228/can-raku-range-operator-on-strings-mimic-perls-behaviour/70239585) and the links cited in the answers.

### DIFF
--- a/doc/Language/operators.pod6
+++ b/doc/Language/operators.pod6
@@ -3113,7 +3113,7 @@ L<Junctions|/type/Junction>.
 
 The sequence operator invokes the generator with as many arguments as necessary.
 The arguments are taken from the initial elements and the already generated
-elements. 
+elements.
 
 An endpoint of C<*> (L<Whatever|/type/Whatever>), C<Inf> or C<∞>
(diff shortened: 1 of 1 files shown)


JJ (approved):
Nice!

coke:
is the output right? don't see the "ba" in yours.

```
$ raku -e "say 'aa'...'cc'"
(aa ab ac ba bb bc ca cb cc)
```

coke:
fails xt/trailing-whitespace.t

codesections:
Thanks, @coke, both fixed.

coke:
Thanks for the update: found a typo: "finial" should probably be "final".

coke:
Thanks for the content and quick updates, merged.

PR #3997, opened 2021-12-14
State that isa-ok matches Roles

codesections:
`Mu.isa` explicitly does *not* match roles, so it seems worth calling out that `&isa-ok` behaves differently.


JJ (approved):
Great. Thanks!

PR #4005, opened 2021-12-31
Add filename extensions page

finanalyst:
## The problem

No handy reference for best current filenaming practices

Email thread on perl6-users with approval  for idea.

## Solution provided

Page written
Propose to put in at the beginning section


doc/Language/filename-extensions.rakudoc
```diff
@@ -0,0 +1,49 @@
...
+Filename extensions can be used to associate behaviour with the file content, such as opening
+a specialised reader or editor, running a program with the file content as input, etc. On some
+operating systems, this is almost the required behaviour. Even when it is not required by the
+operating system, filename extensions provide a useful mnenomic, or reminder, about the contents
+of the file.
+
+Early filesystems, the part of the operating system handling filenames, constrained filename extensions
+to three characters after the '.'. This is no longer a constraint.
+
+=head1 History and Extensions
```
JJ, line 35: Maybe simply history. But if you decide to keep "and extensions", just put it in lowercase.

doc/Language/filename-extensions.rakudoc
```diff
@@ -0,0 +1,49 @@
...
+The Raku language was for many years known as Perl 6, and filename extensions
+were adopted accordingly. Consequently, there is a best current practice and a previous best
+practice. The following table is an attempt to list all the filename extensions, starting with the
+earliest on the left.
+
+The most recent best practice is shown in the last column.
+
+Some modules will become widely used, and will have filename extensions associated with them.
+
+A module developer / maintainer may add to the table and create a PR with the change in the
```
JJ, line 46: This is probably redundant.

doc/Language/filename-extensions.rakudoc
```diff
@@ -0,0 +1,49 @@
...
+
+=SUBTITLE The extensions recommended for files with Raku content.
+
+=begin table
+File contents      | Extensions recommended or used at one time | Current recommendation
+===================================================================================
+B<Language files>  |                                            |
+Raku script        | .pl, .p6                                   | .raku
+Raku module        | .pm, .pm6                                  | .rakumod
+Raku documentation | .pm, pm6, pod, pod6                        | .rakudoc
```
JJ, line 13: Did it ever use `.pm`?

JJ (approved):
Thanks a lot for this. I am not totally sure `documentable` will work with the new extensions, or for that matter that we want different extensions. But let's go ahead and then decide what's what.

doc/Language/filename-extensions.rakudoc
finanalyst: Redundant yes, but I tend to restate things for beginners

doc/Language/filename-extensions.rakudoc
finanalyst: Not sure .pm was ever used for documentation, but Pod::To::HTML allowed for it. So on this basis pm/pm6 were once considered acceptable.

PR #4006, opened 2022-01-02
change run on config explanation to match actual behaviour

finanalyst:
## The problem
The documentation about run-on config data supplied to Pod::Blocks is not entirely correct.

The intent of the Pod6 design was for config data to be allowed after a `= ` in the following line, so long at the `=` of the Pod6 block declaration and the subsequent `=` align vertically.
But there is *no* constraint about the whitespace after the `=`, except that it should exist (that is there should be at least one `\h`.

In fact the actual behaviour of Rakudo requires the subsequent config data to also be vertically aligned so that the data starts after the Pod::Block identifier. 

There is a test for this in the Raku test suite [here](https://github.com/Raku/roast/blob/f151ad5c9338ddf678531499f29b804525c5cead/S26-documentation/09-configuration.t#L44).

I have looked at the Rakudo grammar, [which I think is here](https://github.com/rakudo/rakudo/blob/4960802167dd63eda69b9c33899edd0503e2d3e2/src/Perl6/Grammar.nqp#L5093), and it seems to me to provide the documented behaviour. I do not understand why it does not.

## Solution provided

Rewritten the Pod6 section to match the actual behaviour, which requires two alignments, one with `=` in line, the second with the config data after the Block identifier.

The secondary alignment looks OK, and it is not inconsistent with intended design of Pod6. 

Also, the secondary alignment would be a source of error because an incorrect secondary alignment (too few spaces on the subsequent lines) would not be picked up as an error by the compiler.

Alternatively, the secondary alignment information could be put in a footnote as being a workaround.


JJ (approved):
It looks essentially OK, but @tbrowder is the expert here so he should probably take a look if possible.

tbrowder:
Richard, I appreciate your look at this, but it will take me a while to refresh my memory about the guts of the issue before I can make any kind of reasoned judgement. If I haven't responded in a week, please ping me.

finanalyst:
After looking at this some more, I have found that the 'secondary' alignment is NOT needed.
The actual behaviour of the Rakudo compiler matches the expected documented behaviour.

So I have removed all mention of the 'secondary' unexpected behaviour.

The final version of this PR 
1) removes the footnote saying that extra Config is in the future, because it is available now.
2) provides an example
3) also explains in more detail what is meant by a 'virtual margin'.

JJ (approved):
Looks fine. Thanks!

PR #4009, opened 2022-01-08
remove unneeded assignment to nameless state variable

sdondley:
## The problem

In a code example to demonstrate slurping, a signature object get assigned to the nameless state variable. This assignment does not appear to be needed and may add confusion for readers.

## Solution provided

The code example removes the assignment.

### DIFF
--- a/doc/Type/Signature.pod6
+++ b/doc/Type/Signature.pod6
@@ -668,9 +668,9 @@ These are called "slurpy" because they slurp up any remaining arguments
 to a function, like someone slurping up noodles.
 
 =begin code
-$ = :($a, @b);  # exactly two arguments, where the second one must be Positional
-$ = :($a, *@b); # at least one argument, @b slurps up any beyond that
-$ = :(*%h);     # no positional arguments, but any number of named arguments
+:($a, @b);  # exactly two arguments, where the second one must be Positional
+:($a, *@b); # at least one argument, @b slurps up any beyond that
+:(*%h);     # no positional arguments, but any number of named arguments
 
 sub one-arg (@)  { }
 sub slurpy  (*@) { }
--- a/doc/Type/Signature.pod6
+++ b/doc/Type/Signature.pod6
@@ -668,9 +668,10 @@ These are called "slurpy" because they slurp up any remaining arguments
 to a function, like someone slurping up noodles.
 
 =begin code
-:($a, @b);  # exactly two arguments, where the second one must be Positional
-:($a, *@b); # at least one argument, @b slurps up any beyond that
-:(*%h);     # no positional arguments, but any number of named arguments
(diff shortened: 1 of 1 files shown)


JJ:
Did you test-compile the examples? The assignment might be there to avoid failure, probably.

JJ:
And that's correct, that example was introduced by @gfldex 6 years ago in this commit f001e174cb16cabc0d7ca421f7a72e61e81b2ada that literally says "make examples compile"

I really appreciate your effort to improve the documentation, but there are enough outstanding issues to start working. I can suggest one if you want.

jonathanstowe:
> literally says "make examples compile"

I dunno, it looks like it isn't a problem any more:

```
jonathan@souffle:~/devel/RPi-PCA6985 $ raku -e ':($a, @b);:($a, *@b); '
jonathan@souffle:~/devel/RPi-PCA6985 $ 
```

It may have been some "_useless of signature in sink context_" or so before but it doesn't do it now.

coke:
You can verify the change works with

```
TEST_FILES=doc/Type/Signature.pod6 make xtest
```

doc/Type/Signature.pod6
```diff
@@ -668,9 +668,9 @@ These are called "slurpy" because they slurp up any remaining arguments
 to a function, like someone slurping up noodles.
 
 =begin code
-$ = :($a, @b);  # exactly two arguments, where the second one must be Positional
-$ = :($a, *@b); # at least one argument, @b slurps up any beyond that
-$ = :(*%h);     # no positional arguments, but any number of named arguments
+:($a, @b);  # exactly two arguments, where the second one must be Positional
```
JJ, line 671: I'm going to assume this works, but in that case it might be better to show these examples in a more real context, or simply eliminate the `;`, which gives the (possibly confusing) impression that this would be a valid statement.

BTW, there are similar assignments further down in the document, you might wan to change that in the same PR.

JJ (changes requested):
Looks good, but I'd be grateful if you addressed my nits.

coke:
To remove the semicolon, you'd have to split the code into a section per line;

for example:

```
:(Str $id)
:($base = 10)
```
isn't valid.

JJ:
> To remove the semicolon, you'd have to split the code into a section per line;
> 
> for example:
> 
> ```
> :(Str $id)
> :($base = 10)
> ```
> 
> isn't valid, even though each line by itself is.

In general, #1748 should be the rule to follow when adding new docs or making any clarification. Once we agree this can be improved, it's better if it's used in a real(ish) use case, and not as simple bare signatures in a statement (which does not make a lot of sense)

JJ (approved):
Thanks a lot!

PR #4010, opened 2022-01-08
Fix foreign key reference in profile SQL query

flwyd:
## The problem

The example query would produce results with a few rows with high `inclusive_time` and then lots of small `inclusive_time`, each of which appeared to only happen once.

## Solution provided

Use the foreign key when joining, rather than joining two tables by their own primary keys.


jonathanstowe (approved):

JJ (approved):
Seems to work OK, so many thanks!

PR #4011, opened 2022-01-08
add test for version of raku

coke:
Avoid issues where an older version might fail an example/signature test.


JJ (approved):
Thanks for this. It's really convenient that we use the last, since many features will be only available in the last one.

coke:
Verified test failed with 2021.10 and passes with 2021.12

PR #4008, opened 2022-01-04
update information about methods for distributing modules

sdondley:
## The problem

information about how to distribute modules is out of date

## Solution provided

updates info about how to distribute a module

PR #4013, opened 2022-01-12
Update information about methods for distributing modules

sdondley:
## The problem

information about how to distribute modules is out of date

## Solution provided

updates info about how to distribute a module

This is the same PR as #4008 which was accidentally obliterated by me.


lizmat (approved):

JJ (approved):
👍

PR #4014, opened 2022-01-13
Mention adverbial pair in 5to6

0rir:
Being slightly competent with Perl, I took the fat arrow description in 5to6-perlop as a disincentive to read about Pairs.  That was an efficient choice except for my later ignorance of adverbially constructed pair syntax.

Here, I have added a link to the type Pair and to the glossary entry for adverbial pair.


JJ (approved):
Thanks!

PR #4015, opened 2022-01-14
Expand qq docs re esc sequences

codesections:
This expands the description of interpolation with qq, especially with regards to backslash escape sequences, which it now lists in a table.  It also separates the discussion of variable, closure, and backslash-escape interpolation and emphasizes the connection to the equivalent quoting adverb(s). 

Finally, this also fixes a link on the syntax page that incorrectly implied that backslash escape sequences work in `q` rather than `qq`.

### DIFF
--- a/doc/Language/quoting.pod6
+++ b/doc/Language/quoting.pod6
@@ -174,19 +174,70 @@ My favorite color is blue!
 =end code
 
 X<|\ (quoting)>
-The C<qq> form – usually written using double quotes – allows for
-interpolation of backslash sequences and variables, i.e., variables can be
-written within the string so that the content of the variable is inserted into
-the string. It is also possible to escape variables within a C<qq>-quoted
-string:
 
-=begin code :allow<B> :preamble<my $color = 'blue'>
-say B<">The B<\>$color variable contains the value '$color'B<">;
+The C<qq> form – usually written using double quotes – allows for
+interpolation of backslash escape sequences (like C<q:backslash>), all sigiled
+variables (like C<q:scalar:array:hash:function>), and any code inside C<{...}>
+(like C<q:closure>).
+
+=head3 Interpolating variables
+
+Inside a C<qq>-quoted string, you can use variables with a sigil to trigger
+interpolation of the variable's value.  Variables with the C<$> sigil are
+interpolated whenever the occur (unless escaped); that's why, in the example
+above, C<"$color"> became C<blue>.
+
+Variables with other sigils, however, only trigger interpolation when you follow
+the variable with the appropriate postfix (C<[]> for Arrays, C«<>», for Hashes,
+C<&> for Subs). This allows you to write expressions like
+C<"documentation@raku.org"> without interpolating the C<@raku> variable.
+
+To interpolate an Array (or other L<Positional|/type/Positional> variable),
+append a C<[]> to the variable name:
+
+=begin code :allow<B>
+my @neighbors = "Felix", "Danielle", "Lucinda";
+say "@neighborsB<[]> and I try our best to coexist peacefully."
 =end code
 
 =begin code :lang<text>
-The $color variable contains the value 'blue'
+Felix Danielle Lucinda and I try our best to coexist peacefully.
 =end code
+
+Alternatively, rather than using C<[]>, you can interpolate the Array by
+following it with a method call with parentheses after the method name. Thus the
+following code will work:
+
+=begin code :allow<B L> :preamble<my @neighbors = "Felix", "Danielle", "Lucinda">
+say "@neighborsB<.L<join|/routine/join>(', ')> and I try our best to coexist peacefully."
+=end code
+
+=begin code :lang<text>
+Felix, Danielle, Lucinda and I try our best to coexist peacefully.
+=end code
+
+However, C<"@example.com"> produces C<@example.com>.
+
+To call a subroutine, use the C<&>-sigil and follow the subroutine name with parentheses.
+X<|& (interpolation)>
+
+    say "uc  'word'";  # OUTPUT: «uc  'word'»␤
+    say "&uc 'word'";  # OUTPUT: «&uc 'word'»␤
+    say "&uc('word')"; # OUTPUT: «WORD»␤
+    # OUTPUT: «abcDEFghi␤»
+
+To interpolate a Hash (or other L<Associative|/type/Associative> variable), use
+the C«<>» postcircumfix operator.
+
+    my %h = :1st; say "abc%h<st>ghi";
+    # OUTPUT: «abc1ghi␤»
+
+The way C<qq> interpolates variables is the same as
+C<q:scalar:array:hash:function>.  You can use these adverbs (or their short
+forms, C<q:s:a:h:f>) to interpolate variables without enabling other C<qq>
+interpolations.
+
+=head3 Interpolating closures
 
 Another feature of C<qq> is the ability to interpolate Raku code from
 within the string, using curly braces:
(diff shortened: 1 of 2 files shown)


doc/Language/quoting.pod6
```diff
@@ -174,38 +174,26 @@ My favorite color is blue!
...
-the string. It is also possible to escape variables within a C<qq>-quoted
-string:
+interpolation of backslash escape sequences (like C<q:backslash>), all sigiled
+variables (like C<q:scalar:array:hash:function>), and any code inside C<{...}>
+(like C<q:closure>).
 
-=begin code :allow<B> :preamble<my $color = 'blue'>
-say B<">The B<\>$color variable contains the value '$color'B<">;
-=end code
+=head3 Interpolating variables
```
JJ, line 183: I'm not usually in favor of deletion, unless it's patently wrong and it's addressing some issue. That text is usually there for a reason, and it might be a subtle reason, but reasonable nonetheless.

JJ (commented):
While the PR subject says "expand", you're actually changing many paragraphs, eliminating examples (or moving them around). Expansion is good, substitution I'd appreciate first an issue that says why it fails. Rewriting for the sake of rewriting might eliminate some stuff that's there for a good reason.

doc/Language/quoting.pod6
codesections: This code wasn't deleted, just moved.  In this commit, it's at lines 303–315.

codesections:
That's a fair point and I probably should have opened an issue first.  (I started this PR intending to just add a table of allowed backslash escape sequences, but then couldn't find a good place to put it and ended up moving more text around than I originally meant to.)

That said, I don't believe I deleted anything – I just moved a few things around to create some subsections within the `Interpolation: qq` section.

Let me know if you'd like me to open an issue, though; I'd be happy to do so.

JJ:
> That's a fair point and I probably should have opened an issue first. (I started this PR intending to just add a table of allowed backslash escape sequences, but then couldn't find a good place to put it and ended up moving more text around than I originally meant to.)
> 
> That said, I don't believe I deleted anything – I just moved a few things around to create some subsections within the `Interpolation: qq` section.
> 
> Let me know if you'd like me to open an issue, though; I'd be happy to do so.

Fair enough. Let me check out what you've done. I'd also appreciate, if possible, that you made this a bit clearer in the PR subject.

doc/Language/quoting.pod6
```diff
@@ -174,38 +174,26 @@ My favorite color is blue!
 =end code
 
 X<|\ (quoting)>
+
 The C<qq> form – usually written using double quotes – allows for
-interpolation of backslash sequences and variables, i.e., variables can be
-written within the string so that the content of the variable is inserted into
-the string. It is also possible to escape variables within a C<qq>-quoted
-string:
+interpolation of backslash escape sequences (like C<q:backslash>), all sigiled
```
JJ, line 179: nit: just scratch "all"

doc/Language/quoting.pod6
```diff
@@ -174,38 +174,26 @@ My favorite color is blue!
...
+variables (like C<q:scalar:array:hash:function>), and any code inside C<{...}>
+(like C<q:closure>).
```
JJ, line 181: I was a bit confused here, but then I understood that what you want to say is that `qq` works in the same way as q with a series of adverbs. It might be clear withing the whole context of the page, but either a link or a fleshed out clarification might be clearer.

doc/Language/quoting.pod6
JJ: Fair enough. This is a subsection of a section that's called `Interpolation: qq`. Subsections create index entries, so you might want to flesh this out to something that more clearly shows the context or can be found more easily.

doc/Language/quoting.pod6
```diff
@@ -174,38 +174,26 @@ My favorite color is blue!
...
 
-=begin code :allow<B> :preamble<my $color = 'blue'>
-say B<">The B<\>$color variable contains the value '$color'B<">;
-=end code
+=head3 Interpolating variables
 
-=begin code :lang<text>
-The $color variable contains the value 'blue'
-=end code
+Inside a C<qq>-quoted string, you can use variables with a sigil to trigger
```
JJ, line 185: This is the second time you mention variables `with a sigil`; however, if they don't have a sigil, how can they be identified as variables? Also, arguably you're talking about identifiers, not variables. Something with a sigil is clearly an identifier; no sigil, well, it could be, but not always.

doc/Language/quoting.pod6
```diff
@@ -174,38 +174,26 @@ My favorite color is blue!
...
+interpolation of the variable's value.  Variables with the C<$> sigil are
+interpolated whenever the occur (unless escaped); that's why, in the example
```
JJ, line 187: typo: they

doc/Language/quoting.pod6
```diff
@@ -174,38 +174,26 @@ My favorite color is blue!
...
-=begin code :lang<text>
-This room is 4 m by 3.5 m by 3 m.
-Therefore its volume should be 42 m³!
-=end code
-
-By default, only variables with the C<$> sigil are interpolated normally.
-This way, when you write C<"documentation@raku.org">, you aren't
-interpolating the C<@raku> variable. If that's what you want to do, append
-a C<[]> to the variable name:
+To interpolate an Array (or other L<Positional|/type/Positional> variable),
```
JJ, line 195: Probably just talk about anything with the `@` sigil instead. I really have no idea why this happens, BTW

doc/Language/quoting.pod6
```diff
@@ -174,38 +174,26 @@ My favorite color is blue!
...
+append a C<[]> to the variable name:
```
JJ, line 196: Apparently, same goes for `%`

```
> my %h = :a, :b, :c
{a => True, b => True, c => True}
> say "%h"
%h
> say "%h[]"
a	True
b	True
c	True
> 
```

doc/Language/quoting.pod6
```diff
@@ -216,9 +204,9 @@ say "@neighborsB<[]> and I try our best to coexist peacefully."
 Felix Danielle Lucinda and I try our best to coexist peacefully.
 =end code
 
-Often a method call is more appropriate. These are allowed within C<qq>
-quotes as long as they have parentheses after the call. Thus the following
-code will work:
+Alternatively, rather than using C<[]>, you can interpolate the Array by
+following it with a method call with parentheses after the method name. Thus the
+following code will work:
```
JJ, line 209: All this seems strangely ad hoc and does not seem to match the "strangely consistent" nature of Raku. There must be some underlying simple rule we're not getting

doc/Language/quoting.pod6
```diff
@@ -230,20 +218,63 @@ Felix, Danielle, Lucinda and I try our best to coexist peacefully.
...
 
-    say "abc&uc("def")ghi";
+    say "uc  'word'";  # OUTPUT: «uc  'word'»␤
+    say "&uc 'word'";  # OUTPUT: «&uc 'word'»␤
+    say "&uc('word')"; # OUTPUT: «WORD»␤
     # OUTPUT: «abcDEFghi␤»
 
-Postcircumfix operators and therefore L<subscripts|/language/subscripts> are
-interpolated as well.
+To interpolate a Hash (or other L<Associative|/type/Associative> variable), use
```
JJ, line 229: As shown above, it also works with []

doc/Language/quoting.pod6
```diff
@@ -230,20 +218,63 @@ Felix, Danielle, Lucinda and I try our best to coexist peacefully.
...
+This provides the same functionality as the C<q:closure>/C<q:c> quoting form.
+
+=head3 Interpolating escape codes
+
+The C<qq> quoting form also interpolates backslash escape sequences.  Several of
+these print invisible/whitespace ASCII control codes or whitespace characters:
+
+=begin table :allow<L>
+Sequence    Hex Value       Character
+\0          \x0000          L<Nul|https://util.unicode.org/UnicodeJsps/character.jsp?a=0000&B1=Show>
```
JJ, line 265: I've usually seen this in ALL CAPS, right?

doc/Language/syntax.pod6
```diff
@@ -568,8 +568,8 @@ String literals are surrounded by quotes:
     say "a string literal\nthat interprets escape sequences";
 
 See L<quoting|/language/quoting> for many more options, including
-L<the escaping quoting C<q>|/language/quoting#Escaping:_q>. Raku uses
-the standard escape characters in literals: C<\a \b \t \n \f \r \e>,
+L<interpolation quoting C<q>|/language/quoting#Interpolation:_qql>. Raku uses
```
JJ, line 571: Thanks for taking care of links.

JJ (approved):
Except for typos, good job. Still baffled at the interpolation rules, though. Will postcircumfix operators trigger it?

JJ:
I have been testing with defined postcircumfix operators and it's not their presence what triggers the interpolation; it's apparently either (), [] or <>. So, no idea.

TimToady:
Look at the use and definition of the bracket_ending rule in Grammar.nqp.

Larry

PR #4019, opened 2022-01-19
Document `handles(**)`

codesections:
This PR explains that you can pass a `HyperWhatever` to the `handles` method and describes how that's different from passing a `Whatever` (specifically, the `HyperWhatever` delegates method calls that will end up in the attribute's `FALLBACK` method; the `Whatever` only delegates if the attribute `.^can` do a method of that name).


JJ (approved):
Thanks!

raiph:
Hi @codesections,

> the HyperWhatever delegates method calls that will end up in the attribute's FALLBACK method; the Whatever only delegates if the attribute .^can do a method of that name

Is the latter, and more generally the doc / your doc change consistent with what I wrote here? https://www.reddit.com/r/perl6/comments/c7l3d9/something_is_wrong_with_handles/

PR #4024, opened 2022-02-01
Remove unnecessary files and links

sumanstats:
I have removed circleci as we have have more robust github actions and circleci link in README is dead. Also removed shippable yml file.


JJ:
CircleCI is working without a glitch:
Captura de pantalla de 2022-02-01 20-36-22
Let me check Shippable.

JJ:
Right, it's not been working for 3 years. Let's get rid of Shippable.

sumanstats:
Ok I see. But since we have more robust github actions, is circleci necessary?

Capture

Also the circleci badge does not render in README.

JJ:
We can fix the badge then. In general, I would appreciate if things were discussed in an issue before creating a PR.
CircleCI also does a different thing; it tries and build the documentation, while GitHub actions do pass tests over existing documentation. Shippable also did a different thing; tests are *never* replicated in different platforms, why would anyone want to do that?

JJ (approved):
Thanks!

Issue #4029, opened 2022-02-10
Operator precedence table is needlessly confusing


codesections:
The [operator precedence table](https://docs.raku.org/language/operators#Operator_precedence) currently has a column labeled `A` with single-letter abbreviations for various types of associativity.  To interpret these letters, the reader needs to cross-reference another table.  These letters aren't used in the source code; as far as I know, this is the only place they're used.

Unless anyone objects, I'll simplify the table by directly listing the associativity and remove the letter-to-meaning table.

JJ:
By all means. Please go ahead.

codesections:
Fixed by #4031

PR #4031, opened 2022-02-11
Clarify operator precedence description and table

codesections:
This PR clarifies the operator precedence table (as discussed in #4029) and clarifies/expands the accompanying discussion of operator precedence and associativity.

This PR makes the following changes:
 * In the operators table, replaced the column that contained L, R, N, C, and X with a column listing associativites of 'left', 'right',
   'non', 'chain', and 'list'.  These single-letter abbreviations were  used in S03, but are not currently used in elsewhere in  Raku/Rakudo/the docs other than in a table in `language/functions`.
 * Made corresponding changes to the `language/functions` table.
 * Split the discussion of precedence and associativity into separate subsections for clarity.
 * Marked precedence levels that do not currently have any operator subroutines (these mostly do have *operators*, but those operators are special-cased by the compiler and aren't accessible to users).
 * Slightly expanded the associativity table and changed the example infix operator from `!` to `§` to avoid confusion with `prefix:<!>`.
 * Added a brief discussion of how associativity applies to non-infix operators and a note that setting the associativity for non-infix operators is NYI (see https://github.com/rakudo/rakudo/issues/4779, https://github.com/rakudo/rakudo/issues/4780).
 * Removed the table depicting associativity between unary operators, which misleadingly implied that creating right-associative unary operators is possible. (I commented out the table rather than deleting it so that we can add it back if/when custom `assoc` is implemented for unary ops.

So, as I guess I should expect by now, a seemingly simple doc update grew into a more significant change than I expected, led me to realize I didn't understand something as well as I thought, and led me into an edge case that triggers a Raku bug!  But, on the other hand, it also put me on the path to solving that bug – and it's one that's been around since 2014, so at least there's that :grin:

### DIFF
--- a/doc/Language/functions.pod6
+++ b/doc/Language/functions.pod6
@@ -944,19 +944,23 @@ Raku has the following possible associativity configurations:
 
 =begin table
 
-    A   Assoc     Meaning of $a ! $b ! $c
-    =   =====     =======================
-    L   left      ($a ! $b) ! $c
-    R   right     $a ! ($b ! $c)
-    N   non       ILLEGAL
-    C   chain     ($a ! $b) and ($b ! $c)
-    X   list      infix:<!>($a; $b; $c)
+    Associativity | Meaning of $a ! $b ! $c
+    ==============+========================
+    left          | ($a ! $b) ! $c
+    right         | $a ! ($b ! $c)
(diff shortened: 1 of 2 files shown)


JJ (approved):
Very valuable. Thanks!

PR #4032, opened 2022-02-12
Methods in a class that `is hidden` don't get *%_

codesections:
This PR adds a note to the `is hidden` docs to explain that methods in a class declared with `is hidden` don't get the `*%_` parameter and explain why. See [this recent Stack Overflow answer](https://stackoverflow.com/a/71088429/) and the design docs [A12](https://raku.org/archive/doc/design/apo/A12.html#Interface%20Consistency) and [S12](https://design.raku.org/S12.html#Interface_Consistency).


JJ (approved):
Thanks!

PR #4035, opened 2022-02-13
Remove incorrect text & non-compiling example

### LINKED ISSUE #4020: xt/examples-compilation.t doc/Language/containers.pod6
from

```$ xt/examples-compilation.t doc/Language/containers.pod6```

```
   # my Int \z = 42;
    # z := 100; # OK
    # z := "x"; # Typecheck failure
    # Cannot bind to 'z' because it is a term and terms cannot be rebound
    not ok 4 - doc/Language/containers.pod6 chunk 4 starts with “my Int \z” compiles
```

This example is not currently correct  and dies on the first bind:

```
23:43 < [Coke]> m: my Int \z = 42; z :=100;
23:43 < camelia> rakudo-moar 284b56232: OUTPUT: «===SORRY!=== Error while compiling <tmp>␤Cannot bind to 'z' because it
                 is a term and terms cannot be rebound␤at <tmp>:1␤------> my Int \z = 42; z :=100⏏;␤»
```

Bisectable6 output: https://gist.github.com/Whateverable/6f770a2858b5477de41b1581e4e067e6

Shows a change in behavior, but this may have never worked as intended. The example needs to be improved (and if it isn't intended to compile, should skip the compilation test.

Also check the other code that has the comment "Typecheck failure" as it is not failing - the example there is compiling and doesn't look like it intended to.

This code hasn't change in some time in the file.

### PR DESCRIPTION
**Pull request by codesections**
This PR removes some text describing binding semantics that, after discussion, turned out to be the result of a bug rather than
intended behavior (the intended behavior is still under-documented, see Raku/doc#3961).  It also removes the accompanying test, which
failed to compile once the bug was fixed.

Finally, it adds a doc-comment clarifying that a nearby test fails at runtime even though it *looks* like it would throw a compile-time
error (and thus would need to be marked for xt/example-compilation to skip).

Resolves #4020

### DIFF
--- a/doc/Language/containers.pod6
+++ b/doc/Language/containers.pod6
@@ -351,18 +351,6 @@ The constraint is a property of the variable, not the container.
 In this case, the type constraint is the (compile-type defined) subset
 C<Three-letter>.
 
-Variables may have no container in them, yet still offer the ability to
-re-bind and typecheck that rebind. The reason for that is in such cases the
-binding operator L<:=|/language/operators#infix_:=> performs the typecheck:
-
-    my Int \z = 42;
-    z := 100; # OK
-    z := "x"; # Typecheck failure
-
-The same isn't the case when, say, binding to a L<Hash|/type/Hash> key, as
-the binding is then handled by a method call (even though the syntax remains
-the same, using C<:=> operator).
-
 The default type constraint of a C<Scalar> container is L<Mu|/type/Mu>.
 Introspection of type constraints on containers is provided by C<.VAR.of>
 method, which for C<@> and C<%> sigiled variables gives the constraint for
@@ -383,6 +371,7 @@ declaration:
     my Int:D $def = 3;
     say $def;   # OUTPUT: «3␤»
     $def = Int; # Typecheck failure
+=comment ^^^ fails at runtime, so xt/example-compilation still passes
 
 You'll also need to initialize the variable in the declaration, it can't be
 left undefined after all.

### CODE REVIEW

File: doc/Language/containers.pod6 [hunk 1]
```diff
@@ -383,6 +371,7 @@ declaration:
     my Int:D $def = 3;
     say $def;   # OUTPUT: «3␤»
     $def = Int; # Typecheck failure
+=comment ^^^ fails at runtime, so xt/example-compilation still passes
```
JJ, line 374: Not terribly fond of these comments. I would really prefer to open an issue, to be closed by whoever.

doc/Language/containers.pod6
```diff
@@ -351,18 +351,6 @@ The constraint is a property of the variable, not the container.
 In this case, the type constraint is the (compile-type defined) subset
 C<Three-letter>.
 
-Variables may have no container in them, yet still offer the ability to
-re-bind and typecheck that rebind. The reason for that is in such cases the
-binding operator L<:=|/language/operators#infix_:=> performs the typecheck:
-
-    my Int \z = 42;
-    z := 100; # OK
-    z := "x"; # Typecheck failure
```
JJ, line 360: This returns now 

```
Cannot bind to 'z' because it is a term and terms cannot be rebound
```

So I guess we're good deleting this. But I'm not so fond myself of deleting stuff.  This was introduced here 94c2270cb3c35ff12b91e111b901a50b7e5ef9a6 Apparently that's no longer the case, but apparently also that definition is now creating a Term. So maybe a reference to it being a term (and thus, not rebindable) might be better than total elimination.

JJ (changes requested):
Can you please consider my objections? Thanks-

JJ:
OK, not so important now, and it solves an issue, so merging this.

PR #4039, opened 2022-02-23
Replace "Perl Grammar Guide" with "Raku Grammar Guide"

mreitinger:
Fix grammar tutorial still referring to 'Perl' instead of 'Raku'.


lizmat:
Good catch!

JJ:
Thanks!

PR #4045, opened 2022-03-29
Document MacOS quirk wrt DYLD_LIBRARY_PATH and SIP

patrickbkr:
MacOS has some annoying behavior when dynamically loading C libs. Document those.
See https://github.com/Raku/App-Rakubrew/issues/9 for the very long story.

### DIFF
--- a/doc/Language/nativecall.pod6
+++ b/doc/Language/nativecall.pod6
@@ -1246,6 +1246,28 @@ AF_INET6 SOCK_RAW
 2607:f8b0:4006:800::200e
 =end code
 
+=head1 Platform Specific Notes
+
+=head2 MacOS - DYLD_LIBRARY_PATH is ignored
+
+On MacOSX El Capitan and later the I<System Integrity Protection>, short SIP,
+prevents protected processes from passing through several environment
+variables, among them C<DYLD_LIBRARY_PATH>. The C<env> program, often used in
+I<shebang> lines, is one of those programs. This means that whenever C<env> is
+involved in calling a program the C<DYLD_LIBRARY_PATH> variable will be
+cleared. To work around this effect one has to either make sure no protected
+process is involved (this can be difficult) or disable SIP.
+
+This is usually not a problem when using popular installation methods such as
+I<Homebrew> or I<MacPorts>, tends to be more problematic when installing into
+non standard locations such as ones home directory and most definitely
+problematic when explicitly utilizing the C<DYLD_LIBRARY_PATH> variable.
+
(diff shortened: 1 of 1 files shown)


zaucker:
I wonder who the target audience for this description is. While it helps me (as a Rakudo user) understand that there is an issue with SIP, it doesn't really tell me what I have to do (if anything) when installing and using it.

vrurg:
> it doesn't really tell me what I have to do (if anything) when installing and using it.

Because there is no single solution for all. For example, installing `rakudo` into standard MacPorts or Homebrew paths would not be affected by the problem due to how SIP is implemented.

doc/Language/nativecall.pod6
```diff
@@ -1246,6 +1246,20 @@ AF_INET6 SOCK_RAW
...
+
+=head2 MacOS - DYLD_LIBRARY_PATH is ignored
+
+On MacOSX El Capitan and later the I<System Integrity Protection>, short SIP,
+prevents protected processes from passing through several environment
+variables, among them C<DYLD_LIBRARY_PATH>. The C<env> program, often used in
+I<shebang> lines, is one of those programs. This means that whenever C<env> is
+involved in calling a program the C<DYLD_LIBRARY_PATH> variable will be
+cleared. To work around this effect one has to either make sure no protected
+process is involved (this can be difficult) or disable SIP.
```
vrurg, line 1259: Sorry for being late. I missed your request for review.

I think we need an addition here stating that custom installations of Rakduo into non-standard paths, like under one's home directory, are the most affected ones. But more details in the link below.

Otherwise looks good. Thanks!

zaucker:
> > it doesn't really tell me what I have to do (if anything) when installing and using it.
> 
> Because there is no single solution for all. For example, installing `rakudo` into standard MacPorts or Homebrew paths would not be affected by the problem due to how SIP is implemented.

I currently use the binary releases from https://rakudo.org/downloads unpacking the tar-file into `$HOME/opt/` and symlink that to `$HOME/opt/rakudo`. My PATH then includes `$HOME/opt/rakudo/bin:$HOME/opt/rakudo/share/perl6/site/bin`.
This seems to work fine so far, but of course I might miss some usages that might fail.

This is **not** meant to criticise the suggested documentation, just trying to let you know what some ordinary user might (not) get from it.

patrickbkr:
@vrurg @zaucker I added a paragraph listing situations that tend to trigger the problem and added another link to a good overview of the problem (that google only showed me *after* I found out about the offender and added "SIP" to the search terms. Grrr.)

vrurg (approved):

vrurg:
Looks good. Thank you, Patrick!

PR #4048, opened 2022-04-05
Correct META6.json `provides` snippet

patrickbkr:
To use a module of the own distribution, it must be listed
in the `provides` section, not the `depends` section.


JJ (approved):
Good catch!

PR #4050, opened 2022-04-07
Some tweaks on CATCH

lizmat:
In response to https://github.com/Raku/doc/issues/4049

### DIFF
--- a/doc/Language/exceptions.pod6
+++ b/doc/Language/exceptions.pod6
@@ -100,7 +100,7 @@ in the same namespace will look alike but won't match in a C<CATCH> block.
 Note that the C<CATCH> block semantics apply to the B<entire> lexical scope
 in which it is defined, B<regardless> of where it is defined inside that
 lexical scope.  It is therefore advised to put any C<CATCH> block at the
-starte of the lexical scope to which they apply so that the casual reader
+start of the lexical scope to which they apply so that the casual reader
 of the code can immediately see that there is something special going on.
 
 =head2 Exception handlers and enclosing blocks
--- a/doc/Language/exceptions.pod6
+++ b/doc/Language/exceptions.pod6
@@ -36,15 +36,19 @@ For example, if while executing C<.frobnicate> on an object, a needed path
 C<foo/bar> becomes unavailable, then an
 L<X::IO::DoesNotExist|/type/X::IO::DoesNotExist> exception can be thrown:
 
-    method frobnicate($path) {
-        X::IO::DoesNotExist.new(:$path, :trying("frobnicate")).throw
-          unless $path.IO.e;
-        # do the actual frobnication
-    }
+=begin code
+method frobnicate($path) {
+    X::IO::DoesNotExist.new(:$path, :trying("frobnicate")).throw
+      unless $path.IO.e;
+    # do the actual frobnication
+}
+=end code
 
-    frobnicate("foo/bar");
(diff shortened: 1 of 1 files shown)


coke:
The updated examples don't compile cleanly, you can see the output with

```
TEST_FILES=doc/Language/exceptions.pod6 make xtest
```

I am happy to make them work in the branch or post-merge. (Need to wrap in an explicit pod block and predeclare some things, I think)

doc/Language/exceptions.pod6
```diff
@@ -84,9 +93,15 @@ methods apply to C<$_>, which holds the C<Exception> within the C<CATCH> block.
...
     }
 
-Note that the match target is a role. To allow user defined exceptions to match
-in the same manner, they must implement the given role. Just existing in the
-same namespace will look alike but won't match in a C<CATCH> block.
+Note that the match target is a role. To allow user defined exceptions to
+match in the same manner, they must implement the given role. Just existing
+in the same namespace will look alike but won't match in a C<CATCH> block.
+
+Note that the C<CATCH> block semantics apply to the B<entire> lexical scope
```
JJ, line 100: Starts with the same word as above. Consider alternatives "Please also" or "Note also"

doc/Language/exceptions.pod6
```diff
@@ -84,9 +93,15 @@ methods apply to C<$_>, which holds the C<Exception> within the C<CATCH> block.
          }
     }
 
-Note that the match target is a role. To allow user defined exceptions to match
-in the same manner, they must implement the given role. Just existing in the
-same namespace will look alike but won't match in a C<CATCH> block.
+Note that the match target is a role. To allow user defined exceptions to
+match in the same manner, they must implement the given role. Just existing
+in the same namespace will look alike but won't match in a C<CATCH> block.
```
JJ, line 98: I don't quite understand this sentence. It was there prior to this PR, but as long as you're editing it, some nits
1. Not clear what's that precise role you're mentioning.
2. The second sentence is difficult to parse. It refers to the existence of what? Will look like what?

doc/Language/exceptions.pod6
```diff
@@ -84,9 +93,15 @@ methods apply to C<$_>, which holds the C<Exception> within the C<CATCH> block.
...
+
+Note that the C<CATCH> block semantics apply to the B<entire> lexical scope
+in which it is defined, B<regardless> of where it is defined inside that
+lexical scope.  It is therefore advised to put any C<CATCH> block at the
+starte of the lexical scope to which they apply so that the casual reader
+of the code can immediately see that there is something special going on.
```
JJ, line 104: This clarification is on point, but it might be better right after the example, or alternatively providing another example.

JJ (changes requested):
Since @japhb and @coke have pointed a couple of things that could be changed, I'll go for requesting changes; however, none of my comments are actually blockers. Thanks a lot!

coke:
at @lizmat's request, I fixed the examples and the typos in this branch. Might have to manually merge it to get those fixes?

coke:
I did push before I posted. was asked if I wanted to make *another* PR, which I didn't.

JJ:
> I did push before I posted. was asked if I wanted to make _another_ PR, which I didn't.

I'm confused. There's only one commit in this branch, and it's by @lizmat

coke:
branch differs by case only. fixing.

JJ (approved):
Thanks!

PR #4051, opened 2022-04-10
Search categories update

Altai-man:
## The problem

See https://github.com/Raku/problem-solving/issues/250#
See https://github.com/Raku/doc/issues/1410

## Solution provided

A curated list of categories was created, along with a sensible rules (see https://github.com/Raku/Documentable/pull/159) on how to set one, plus warnings for duplicate refs we had plenty, and all the index references were updated to match the list.


coke:
Current `make xtest` failures on this branch:

```
xt/aspell.t              (Wstat: 256 Tests: 844 Failed: 1)
  Failed test:  19
  Non-zero exit status: 1
xt/search-categories.t   (Wstat: 768 Tests: 412 Failed: 3)
  Failed tests:  10, 19, 76
  Non-zero exit status: 3
xt/links-not-links.t     (Wstat: 65280 Tests: 51 Failed: 0)
  Non-zero exit status: 255
  Parse errors: Bad plan.  You planned 406 tests but ran 51.
xt/space-after-comma.t   (Wstat: 512 Tests: 422 Failed: 2)
  Failed tests:  53, 59
  Non-zero exit status: 2
```

Please wait to merge until these are resolved.

doc/Language/about.pod6
```diff
@@ -153,12 +153,12 @@ limitations, special steps have to be taken to use B<V< X<> >> with other
...
 
 =begin code :lang<pod6>
-=item X<B<foo>|foo> a fancy subroutine
+=item X<B<foo>|Subroutines,foo> a fancy subroutine
 =end code
 
 renders like this
 
-=item X<B<foo>|foo> a fancy subroutine
+=item X<B<foo>|Subroutines,foo> a fancy subroutine
```
JJ, line 161: Again, just an example.

doc/Language/classtut.pod6
```diff
@@ -276,10 +264,9 @@ The third attribute represents the state of completion of a task:
 =for code
 has Bool $.done;
 
-X<|twigils,.>
-X<|twigils,accessors>
-X<|accessor methods>
-X<|classes,accessors>
+X<|Tutorial,.>
+X<|Tutorial,accessors>
```
JJ, line 268: Hum. I really don't know here. It's not about accessors, it's about accessors that use twigils. You're losing that information here.

doc/Language/5to6-perlvar.pod6
```diff
@@ -19,13 +19,11 @@ please see the Raku documentation for each of them.
 
 =head1 SPECIAL VARIABLES
 
-X<|Special Variables (Perl)>
+X<|Foreign,Special Variables (Perl)>
```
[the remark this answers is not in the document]
Altai-man: The list of categories I used was standardized at https://github.com/Raku/problem-solving/pull/271 and is now "set in stone" I believe.

>Why not "Other languages"?

It's a subjective solution I thought was the most fitting. Nobody can win here, as it's subjective. Can be reversed: "Other languages"? Why not "Foreign"?

doc/Language/about.pod6
```diff
@@ -153,12 +153,12 @@ limitations, special steps have to be taken to use B<V< X<> >> with other
 formatting codes; for example:
 
 =begin code :lang<pod6>
-=item X<B<foo>|foo> a fancy subroutine
+=item X<B<foo>|Subroutines,foo> a fancy subroutine
```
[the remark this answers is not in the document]
Altai-man: Yes, but rules are rules, why not show them in examples?

doc/Language/classtut.pod6
Altai-man: No, there is no "twigil accessors" category.

Categorization is complex, I'm sure you're aware of that better than anyone. You cannot be too "high-level" and at the same time you cannot categorize every tiniest detail creating hundreds of categories to be lost.

This page is a tutorial for everything related to OOP and in this paragraph accessors are explained, so "accessors in category Tutorial" looked fitting to me.

doc/Language/classtut.pod6
```diff
@@ -300,7 +287,7 @@ without having to write the method by hand. You are free instead to write
 your own accessor method, if at some future point you need to do something
 more complex than returning the value.
 
-X<|traits,is rw>
+X<|Tutorial,is rw>
```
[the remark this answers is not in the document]
Altai-man: I'm not sure about that.

`is rw` as a trait is indexed in another place, in fact in three different places:

```
doc/Type/Attribute.pod6
104:=head2 X<trait is rw|Traits,is rw (Attribute)>

doc/Language/typesystem.pod6
424:=head3 trait X«C<is rw>|Traits,is rw (class)»

doc/Type/Signature.pod6
1084:X<|Traits,is rw>
```

So we have explanation of `is rw trait` for attributes, `is rw trait` for classes and `is rw trait` when used in signatures.

And this `is rw` is on the tutorial page, so it makes sense to put it under the tutorial term, as it's a quick introduction.

If it would be `Traits,is rw`, it would conflict with the anchor elsewhere, and when you have a search giving ambiguous results it's not that awesome. Am I wrong?

Altai-man:
>I have seen several cases where categories are substituted and simply dropped. For instance, classes, methods has become Tutorial, methods.

Well, we have no `classes` category. A method is a _concept_ within a language, so it either goes into Language or Tutorial. As this is a tutorial page, I placed it under Tutorial category.

>In many other cases the category was redundant or implicit, but I would have wanted to have a more calmed discussion on some of them, including the inclusion of some new categories

On the search page we can look up things filtering by a category, so if something is indexed in search, how can it be redundant? If we categorize the data to query it we cannot just leave some of it without a category.

>implicit

Yes, that's the real horror. For years rules of applying a category were obscure and I am sure almost nobody of our tech writers contributing to docs really knew them properly.

People had items indexed with categories like "<sym>" and "->" and nobody paid attention to it being bogus, as people were using syntax not really described anywhere, processed by obscure code with 6 different ways to do something wrong.

I believe moving from implicit to a clear list of how people should index things, both the syntax and the semantics, and giving them tooling (to detect items with wrong syntax or to detect duplicated terms) is the right direction.

>I'll open an issue, or a series of issues to discuss and propose solutions on specific changes

Sure, thanks.

doc/Language/5to6-perlvar.pod6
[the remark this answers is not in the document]
JJ: "Foreign" does not usually apply to programming languages. A language is not a country, other languages are simply other languages, not "foreign" as in "citizens of a different country"

doc/Language/about.pod6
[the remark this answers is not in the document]
JJ: Because examples should only illustrate what they do illustrate, not something else. This is simply an example of formatting codes within indexing marks. If you want to illustrate categories, that's *very* fine. not in this case. And if you want to illustrate there are no first-level terms, explain as much below.

doc/Language/classtut.pod6
JJ: Again, this is not explaining accessors. It's explaining a specific type of accessors that use twigils. By dropping twigils you're negating the possibility of searching that, as opposed to simple accessors.

doc/Language/classtut.pod6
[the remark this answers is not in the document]
JJ: Well, it was originally traits, is rw. You would be simply capitalizing it. My point is that you're substituting some terms by more general ones, thus losing search information.

JJ:
The problem is not the fact that we use categories. That PR was accepted, we're beyond that point. I would have appreciated, however, when there's some ambiguity on how to index something, to have opened a little discussion. We can do that later, of course. You've said that categorization is complex. Let's try and agree on how to categorize things properly in a series of follow-up issues and/or PRs.

doc/Language/5to6-perlvar.pod6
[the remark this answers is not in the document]
Altai-man: >"Foreign" does not usually apply to programming languages. A language is not a country, other languages are simply other languages, not "foreign" as in "citizens of a different country"

You may be correct about the "usually" part, but I dare to disagree it doesn't mean "always".

For example,
Python has ctypes (https://docs.python.org/3/library/ctypes.html) and it's "A foreign function library".
Haskell has a similar thing called FFI, which stands for "Foreign Function Interface", meaning another language (C).

So apparently the term is used to refer to other languages. I believe "Other languages" as a category is a bit confusing, not only it's longer, but it's meaning also blurred with the natural language the documentation is written in, while "Foreign" when it comes to things like FFI has a single meaning.

doc/Language/about.pod6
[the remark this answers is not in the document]
Altai-man: You are right. We need a ticket then, to teach Documentable to skip `X<>` elements if they are placed within `code`, then this can be removed.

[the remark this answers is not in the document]
JJ: Sorry, what's Documentable got to do with this? Did this generate an indexed term?

doc/Language/5to6-perlvar.pod6
```diff
@@ -19,13 +19,11 @@ please see the Raku documentation for each of them.
...
 
 =head2 General variables
 
-=head3 $ARG, $_
-
-X<|$_ (Perl)>
+=head3 X<$ARG, $_|Foreign,$_ (Perl)>
```
JJ, line 26: You're dropping `$ARG` as an indexing term.

doc/Language/5to6-perlvar.pod6
```diff
@@ -48,21 +44,17 @@ you can call methods on it. Unlike C<$_> you cannot assume C<@_> as the
 default variable for those methods to operate on (i. e. C<@_.shift> works,
 C<.shift> does not).
 
-=head3 $LIST_SEPARATOR, $"
+=head3 X<$LIST_SEPARATOR, $"|Foreign,$" (Perl)>
```
JJ, line 47: In all these, you're dropping the first term in the list, or the alias. As a matter of fact, these were not indexed before, but if you index them, you should use both terms.

doc/Language/5to6-perlvar.pod6
```diff
@@ -48,21 +44,17 @@ you can call methods on it. Unlike C<$_> you cannot assume C<@_> as the
...
 
 Currently, there is no equivalent of the List Separator variable in Raku.
 Design document L<S28|https://design.raku.org/S28.html> says there isn't one,
 so you probably don't want to hold your breath.
 
-=head3 $PROCESS_ID, $PID, $$
-
-X<|$$ (Perl)>
+=head3 X<$PROCESS_ID, $PID, $$|Foreign,$$ (Perl)>
```
JJ, line 53: Again, three terms to index here.

JJ: Agreed, they were not indexed before, just the one you're indexing. And badly, at that (below the title instead of above). But as long as you're re-indexing it, let's go all the way.

doc/Language/classtut.pod6
Altai-man: >Again, this is not explaining accessors. It's explaining a specific type of accessors that use twigils

Ah, I see what you mean! Now I believe we have different views on the concept/purpose of index categorization.

Please, bear with me for a bit then.

In my view, we have an index of things to search. Things go from particular functions like `run` or `dir` to language-wide concepts like classes or twigils.

The idea is to categorize those things into certain buckets, so you can, say, search for all the methods starting with `is-` or all the `$` variables (but you don't need explanation of what $ twigil is, etc).

So buckets/categories are meant to be some sort of an umbrella to a meaningful number of entities. Say "subroutines" is a bucket for subroutines or "tutorial" is a bucket for places in the documentation when some term is explained in a tutorial-like manner for a quick dive.

You claim we need a "twigil accessors" bucket/category, but what belongs to this category? Only this anchor where the doc talks about twigil accessors? Even if there are other places like that (we can arguably sweep things like `$.` syntax there), there are but a few of them. Trying to store additional information on the labels of the buckets (we cannot have a bucket "accessors", but only "twigil accessors" and maybe some other types of accessors) creates not a lot less categories than we have items.

I agree with you the information the category gives is great (speaking of the other example, we have `is rw` in Tutorial section, `is rw` trait explanation for classes and in signatures, so you can decide if you want a tutorial or not), but maybe we don't agree on the minimal number of items necessary to start a new "bucket". IMO anything with less than say 15-20 items doesn't need a bucket by itself, but maybe the line is different for you?

So, in my view, categories are used to narrow the search query down as their primary purpose, and the additional information is the secondary purpose. The anchor itself should give the information when it's necessary.

Does it match your view on the issue in any way?

doc/Programs/03-environment-variables.pod6
```diff
@@ -44,70 +44,70 @@ version 6.e.  Early implementation has been available in Rakudo compiler
...
-X<|RAKUDO_BACKTRACE_SETTING>
+X<|Programs,RAKUDO_BACKTRACE_SETTING>
 =item C<RAKUDO_BACKTRACE_SETTING> (I<Bool>; B<src/core.c/Backtrace.pm6>)
 
 Controls whether C<.setting> files are included in backtraces.
 
 =head2 Affecting precompilation
 
-X<|RAKUDO_PREFIX>
+X<|Programs,RAKUDO_PREFIX>
```
JJ, line 72: Not a program. Neither of these is a program. It's an environment variable.

doc/Type/Bool.pod6
```diff
@@ -6,9 +6,9 @@
 
     enum Bool <False True>
 
-X<|True>
-X<|False>
-X<|Boolean>
+X<|Syntax,True>
+X<|Syntax,False>
+X<|Language,Boolean>
```
JJ, line 11: It behaves like a type. Why not Type?

doc/Type/IO/Path.pod6
```diff
@@ -976,8 +976,6 @@ Defined as:
 Coerces L«C<.basename>|/routine/basename» to L<Int|/type/Int>. L<Fails|/routine/fail>
 with C<X::Str::Numeric> if base name is not numerical.
 
-X<|symlink (sub)>
-X<|symlink (method)>
 =head2 routine symlink
```
JJ, line 979: These have simply disappeared?

doc/Language/5to6-perlvar.pod6
Altai-man: Yes, those pages are a bit troublesome in general. I had to go through hundreds and hundreds of anchors when doing this, so some things certainly slipped, and I also decided to work on them when the main pack is done.

doc/Programs/03-environment-variables.pod6
Altai-man: We have types of pages like `type`, `language` and also `programs`.

Not sure if it was intended, but `programs` ended up to relate to everything "near" the programming process, like how to debug Raku programs or env variables for the compiler.

I am not sure what'd be a better term for this.

doc/Language/5to6-perlvar.pod6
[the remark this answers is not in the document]
JJ: This is not FFI. You're simply referring to indexing terms in the reference manual that don't belong to raku, but, well, other languages. Can't see how having $_ in the category "Other languages" would be more confusing than in the category Foreign.

doc/Type/X/Bind/Slice.pod6
```diff
@@ -18,7 +18,7 @@ and
     CATCH { default { put .^name, ': ', .Str } };
     # OUTPUT: «X::Bind::Slice: Cannot bind to Hash slice␤»
 
-you get an exception of type X<::Bind::Slice>
+you get an exception of type C<::Bind::Slice>.
```
JJ, line 21: This is out of scope in this PR. Not a bad call, but would have been better in a separate PR.

doc/Type/IO/Path.pod6
Altai-man: I believe they are bogus. The symlink routine is already indexed as subroutine because of the `=head2 routine symlink` syntax, there is no need to double/triple index it. Others don't have it, just look at the https://github.com/Raku/doc/blob/bab14801aa96c36e357645bf9212f74a3acab59b/doc/Type/IO/Path.pod6 <- or others, no other method/subroutine is indexed this way.

So I believe it's just the author wanting to make an anchor but being ignorant (and those who reviewed it) to how the anchors are indexed, resulting in those two - just another consequence of indexing being obscure thing for years.

doc/Language/classtut.pod6
JJ: No, I'm not saying that should be a category. It's simply an indexing term.

doc/Language/5to6-perlvar.pod6
JJ: Actually, not dropping, but not including where it wasn't included before, but anyway.

doc/Type/X/Bind/Slice.pod6
JJ: Scratch that, not a PR, but simply commit to main. We should maybe add a test for these.

doc/Programs/03-environment-variables.pod6
JJ: As long as we're introducing or changing categories, why not go the whole nine yards? I'd say "Environment variables" would be simply it.

doc/Language/about.pod6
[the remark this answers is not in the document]
Altai-man: Documentable is the current tool used to create search indexes out of Pod. There might be other tools too, I used Documentable as it 1)exists; 2)works well after some improvements.

So right now we have a test that walks through all the anchors and if it sees some that is bogus, it causes the test to fail, so to make it work we need to change the test and how this case (anchor in a code) is processed in Documentable.

coke:
Comment says "Add test for these" - For what? There is a test for enforcing valid categories that was added with the PR - if something else, please let me know.

JJ:
> Comment says "Add test for these" - For what? There is a test for enforcing valid categories that was added with the PR - if something else, please let me know.

Trying to add indexing to X while in X, which is indexed automatically. A page describing Foo and including X<Foo>

coke:
@JJ, @Altai-man  - there are discussions going on in comments that are very hard to follow. If there are action items as a result of that discussion, can someone summarize that in one of these comments with checkboxes so we can tell when we're done with followup?

doc/Language/5to6-perlvar.pod6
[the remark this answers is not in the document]
Altai-man: Alright, let's change it to "Other languages" then.

JJ:
> Comment says "Add test for these" - For what? There is a test for enforcing valid categories that was added with the PR - if something else, please let me know.

Here you go: https://github.com/Raku/doc/issues/4057

JJ:
> @JJ, @Altai-man - there are discussions going on in comments that are very hard to follow. If there are action items as a result of that discussion, can someone summarize that in one of **these** comments with checkboxes so we can tell when we're done with followup? (Emphasis, summarize in the comments in the conversation tab, not comments on commits)

I'm creating items for what I consider actionable. Disagreement with specific indexing terms can possibly be fixed on the fly, or discussed in a follow up PR.

doc/Language/5to6-perlvar.pod6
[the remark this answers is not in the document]
JJ: Please let's do that at the same time we address #4056

doc/Programs/03-environment-variables.pod6
Altai-man: I was using the list we agreed upon in https://github.com/Raku/problem-solving/pull/271/files with your approval, but yes, can be changed in theory.

JJ: Totally different to seeing it applied to specific terms.

coke:
For any indexing changes, be sure to re-run `make xtest`. Thanks.

Altai-man:
@coke 

On the weekend I hope to dedicate some time to squash at least some valid bugs JJ pointed out, so no need to create a ticket, and summarize rest of the "questionable" things.

I'd really like to avoid bike-shedding over things like perfect naming as much as possible as we have mountains of work besides this and not enough time.

doc/Type/IO/Path.pod6
JJ: The issue is that you're mixing several different issues in a single PR. One thing is adding categories to terms that didn't have one. That's one thing. Another is fixing issues with indexing. That's another thing. Changing categories is yet another. And finally, dropping indexing for whatever you think does not need to be indexed is yet another. Every one of them should have been given some time and, if possible, feedback from the community. In general, I'm against deleting things from the documentation (and have said it in so many words in many PRs). In particular, indexing terms that can be either subs or methods *in the same place* under two different terms, is possibly not awesome, so I might agree with you. However, that's a different discussion, would need a separate issue. Since this PR has 90 different files and lots of changes, it's simply impossible to give it the attention it deserves (even more so if you merge it during someone's holidays)

coke:
Regarding the holiday merge, that’s on me, FYI.

doc/Language/5to6-perlvar.pod6
Altai-man: >You're dropping $ARG as an indexing term.

But it never was indexed, so it's equivalent, the text is `$ARG, $_` and `$_ (Perl)` is indexed, the same in the updated version.

doc/Type/IO/Path.pod6
Altai-man: I agree things always could've been done better, but I am not sure what is your point.

Yes, you had a vacation and the PR was merged, but this work was here for over a year and nobody paid attention or wanted to instigate a discussion to revive it, nothing. So, in the end, you either get something working and then polish it until it's good or you wait forever waiting for some miracle perfect solution.

Is there something particular you ask from me?

doc/Language/5to6-perlvar.pod6
JJ: Fair enough. Still, choosing one over the other is arbitrary and might not have been the best call initially.

doc/Type/IO/Path.pod6
JJ: What I'm not sure is that even if I make the point it's going to make any difference. So, again, at this point what I ask from you is to have patience with the issues I initiate trying to follow up on this PR.

PR #4069, opened 2022-04-29
Document the .are method

lizmat:


doc/Type/Any.pod6
```diff
@@ -781,6 +781,24 @@ Instead of throwing away the first C<$n> elements, it throws away everything
 I<but> the elements indicated by the WhateverCode, in this case all but the last
 one.
 
+=head2 method are
+
+Defined as:
+
+    multi method are()
+
+Return the strictest type / role to which all elements of the list will
```
Altai-man, line 790: "type or role" or "type (or role)".

doc/Type/Any.pod6
```diff
@@ -781,6 +781,24 @@ Instead of throwing away the first C<$n> elements, it throws away everything
 I<but> the elements indicated by the WhateverCode, in this case all but the last
 one.
 
+=head2 method are
+
+Defined as:
+
+    multi method are()
+
+Return the strictest type (or role) to which B<all> elements of the list
```
Altai-man, line 790: Ugh, I'm sorry to notice it late. I think most of the method descriptions use the "does" style of describing rather than "do", so it should be "returns" to be more consistent.
Otherwise looks good.

dwarring:
Jusr a thought. Does it make sense to have a 1 argument form of `are` that returns a Bool?
`say (2, 5, -3).are(UInt) # False`

lizmat:
> say (2, 5, -3).are(UInt)

Why not `say (2, 5, -3).are ~~ UInt` ?  Except that apparently gives the wrong answer as `Int ~~ UInt` returns `True` atm.

dwarring:
`(1,2,-3).all ~~ UInt` seems reasonable.

coke:
Looks good. Minor nit: output should be of the form

```
    say $range;         # OUTPUT: «1..5␤»
```

coke:
Updated output, passes `make xtest`

PR #4070, opened 2022-04-30
Document X::Cannot::Empty

dumarchie:
Documenting `X::Cannot::Empty` allows the ecosystem to reuse the class, rather than copy it.

### DIFF
--- a/doc/Type/X/Cannot/Empty.pod6
+++ b/doc/Type/X/Cannot/Empty.pod6
@@ -1,0 +1,26 @@
+=begin pod :kind("Type") :subkind("class") :category("exception")
+
+=TITLE class X::Cannot::Empty
+
+=SUBTITLE Error due to inappropriate usage of an empty collection
+
+    class X::Cannot::Empty is Exception { }
+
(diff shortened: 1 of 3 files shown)


coke:
passes `make xtest`

doc/Type/X/Cannot/Empty.pod6
```diff
@@ -0,0 +1,26 @@
+=begin pod :kind("Type") :subkind("class") :category("exception")
+
+=TITLE class X::Cannot::Empty
+
+=SUBTITLE Error due to inappropriate usage of an empty collection
+
+    class X::Cannot::Empty is Exception { }
+
+Error thrown (or wrapped in a C<Failure>) when inappropriately using an empty
+collection.
```
Altai-man, line 10: Can you add an example or two please?

dumarchie: Good point. I'll add an example where a routine fails with an `X::Cannot::Empty` if the collection is empty.

Altai-man (approved):

PR #4081, opened 2022-06-06
Simple typos, misphrasings, and punc

0rir:
Simple typos, misphrasings, and punctuation errors.

With problems little problems like these,  I think thirty changes would be a reasonably size pull request.   Absent feedback, I'll roughly double the count over the next two PRs to twenty.

Thanks.


coke (changes requested):
I don't think the addition of 'the' is needed/correct here. Otherwise looks fine.

0rir:
@coke, thanks for the catch.

Altai-man (approved):

coke:
Sorry for the delay, didn't see that the change had been made.

PR #4086, opened 2022-06-18
Fix precedence table ordering

m-dango:
Table header names changed in https://github.com/Raku/doc/pull/4031. Closes https://github.com/Raku/doc/issues/4085.


m-dango:
FYI, I have not tested this change yet as I'm in the middle of writing something else. Went to the table to check something and saw things didn't look right.

Altai-man (approved):

PR #4094, opened 2022-06-28
Set max AssetPack version in cpanfile

m-dango:
\>= 2.12 requires Mojolicious 9.0.


coke:
Thank you - but this will likely be going away in the next few weeks as we migrate the website build to https://github.com/Altai-man/docs.raku.org, leaving this repository for only the documentation itself.

coke:
... that said, merging it should reduce harm if anyone tries to build in the meantime. Thanks.

PR #4100, opened 2022-08-03
Add examples of obtaining path parts

tbrowder:
## The problem

No good example of getting path parts exists.

## Solution provided

I added a new head2 section with some examples. (Note I ran "make xtest before this commit and saw no problems with my code.)

# UPDATE

Well, I obviously didn't look very carefully!


coke:
The failure here is my fault: 74508673dd5015bef1daee78c7c2481a5b80729

tbrowder:
I'll fix and resubmit.

Issue #4089, opened 2022-06-19
deprecated code in lib/


coke:
```
Saw 1 occurrence of deprecated code.
================================================================================
Use of the 'CompUnit::PrecompilationStore::File' class seen at:
  /Volumes/BUKKIT/Users/coke/sandbox/raku-doc/lib/Pod/Convenience.rakumod (Pod::Convenience), line 11
Please use the 'CompUnit::PrecompilationStore::FileSystem' class instead.
--------------------------------------------------------------------------------
Please contact the author to have these occurrences of deprecated code
adapted, so that this message will disappear!
```

coke:
May have to run with `v2022.06` to see it.

coke:
Fix was literally just renaming the invoked class.

coke:
@finanalyst : FYI

coke:
Your fix may be slightly more complicated, e.g. picking this one if it exists, and falling back to the original name if not.

finanalyst:
@coke I solved this for Pod::From::Cache, which I wrote to replace Pod::Cached.
I asked about this on Stackoverflow. Liz provided a nice answer that checks to see which  class spelling is OK at Compile time.

coke:
We do need the smarter fix, because our CI is using an older raku.

lizmat:
Run with `RAKUDO_NO_DEPRECATIONS=1` ?

coke:
That doesn't fix the problem, it delays it, by trading the deprecation warning now for a hard failure later when the feature is finally removed.

PR #4101, opened 2022-08-08
Fix for #4089

coke:
Properly handle the deprecation notice for earlier rakudos. Original fix ONLY worked on
new compilers, died on old ones. This will work on both and avoid the deprecation warning on newer


coke:
cherry picked the one commit to master - only did PR with the doc change to verify CI now works.

PR #4115, opened 2022-10-06
Including :x adverb

massa:
## The problem

`Language/regexes` omitted the `:x` adverb, so this patch is just to add that.

### DIFF
--- a/doc/Language/regexes.pod6
+++ b/doc/Language/regexes.pod6
@@ -2523,6 +2523,13 @@ In this case, one of them exists (1), so it returns True. Observe that we have
 used C<:st>. As said above, it's functionally equivalent, although obviously
 less legible than using C<:nth>, so this last form is advised.
 
+=head3 X<Counting|Adverbs,:x>
+
+The C<:x> counting adverb makes the expression match many times, like the C<:g>
+adverb, but only up to the limit given by the adverb expression, stopping once the
+specified number of matches has been reached. The value must be a
+L<Numeric|/type/Numeric> or a L<Range|/type/Range>.
+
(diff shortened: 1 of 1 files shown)


lizmat (approved):

coke:
looks good - can we add an example as well to highlight the syntax?

from my REPL as an example

```
[1] >  "aaaaa" ~~ m:x/ 'a' /
(｢a｣)
[2] >  "aaaaa" ~~ m:x3/ 'a' /
===SORRY!=== Error while compiling:
Adverb x3 not allowed on m
------> "aaaaa" ~~ m:x3/ 'a' /⏏<EOL>
[2] >  "aaaaa" ~~ m:x(3)/ 'a' /
(｢a｣ ｢a｣ ｢a｣)
```

coke:
this is an improvement by itself (merged!), but if you can add an example in a future PR, that'd be awesome.

Thanks!

massa:
> looks good - can we add an example as well to highlight the syntax?
> 
> from my REPL as an example
> 
> ```
> [1] >  "aaaaa" ~~ m:x/ 'a' /
> (｢a｣)
> [2] >  "aaaaa" ~~ m:x3/ 'a' /
> ===SORRY!=== Error while compiling:
> Adverb x3 not allowed on m
> ------> "aaaaa" ~~ m:x3/ 'a' /⏏<EOL>
> [2] >  "aaaaa" ~~ m:x(3)/ 'a' /
> (｢a｣ ｢a｣ ｢a｣)
> ```

I think the right thing here would be `"aaaaa" ~~ m:3x/a/`

coke:
Sure, :3x is the same as :x(3) and probably more readable. Thanks.

PR #4121, opened 2022-11-08
add link; lose my; change phrasing

0rir:
In Type/Scalar.pod6,

- Add a link to anonymous scalar,
- Delete a `my` from example,
- Refer to `$` as 'anonymous scalar variable`


coke:
Thanks

PR #4126, opened 2022-11-14
sub-signatures demand a space before parenthesis

gfldex:
This makes the example compile.


coke:
@gfldex - as we discussed on IRC, this looks to me like it's not an issue in the latest version of rakudo - I'll close this PR, but please feel free to re-open if needed.

PR #4132, opened 2022-11-19
Remove bot examples

2colours:
Relates to https://github.com/Raku/doc/issues/711.

These were the only mentions to IRC bots I found. It seems that their entries had been removed earlier. Now the description links to the irc page on raku.org.


coke:
Thanks

PR #4146, opened 2022-12-02
Document Metamodel::Primitives' parameterization cache API

Kaiepi:
## The problem

The parameterization cache is *very* handy in MOP-heavy code, but is rather unintuitive and undocumented.

## Solution provided

Document the `set_parameterizer`, `parameterize_type`, `type_parameters`, and `type_parameter_at` methods of `Metamodel::Primitives`.

Adds "parameterizer" and "noop" as words.


coke:
checking `set_parameterizer`, I don't see it specced - are these rakudo specific? (not a blocker to merge)

Kaiepi:
I'd say these are in a similar limbo to the rest of `Metamodel`. Sometimes `Metamodel` behaviour gets specced in one way or another, but not with these. At the same time, they're not marked `implementation-detail`, and are depended on occasionally in the ecosystem.

coke (approved):

Issue #4117, opened 2022-10-18
language/control and language/statement-prefixes cross contaminate


0rir:
1. `Statement Prefixes` is a misleading name since all of the prefixes apply to Blocks.
2. There are three categories of controllers on these pages:  a) execution flow, b) message flow, and c) execution modifier.
3. The docs for `quietly` should merged into `Prefixes`.
4. The docs for `try`, `do`, `sink`, `once`, `start`, `gather`, `react`, and `supply` should be consolidated to `Control Flow`.

Altai-man:
That's a nice one! Thank you very much.

>Statement Prefixes is a misleading name since all of the prefixes apply to Blocks.

Should it be removed / renamed?

0rir:
The distinction of taking a block and/or statement is not worth splitting the documentation of any one of these words onto two pages.

So I suggest renaming it  'Pragmatic Prefixes' and leaving only `lazy`, `eager`, `hyper`, `race`, and `quietly` on that page.  The effect of these keywords is practical and interfering which matches the English meaning; and that effect also echos well that of 'pragmas'.

0rir:
I suggest "Prefixes", the only other Raku use seems to be in creating operators.  I don't see that as confusing, though document index entries might need a extra word.

An amendment because I just suffered for believing the "statement prefix" tag.   I liked "Code Prefixes" but worried that the `Code`-type implication is not correct (my ignorance about that class is great).

`lazy my @seq = 0 ... 1_000_000_000_000;`  is probably  unwanted enough to warrant a compile time warning.  It is clear that 'my @seq = lazy 0...$x;` is using `lazy` **in** a statement.
Lacking discouragement, I will submit an issue for this.

And, as a starting point, I volunteer to merge the two pages while  conserving their current  meanings.

Altai-man:
>And, as a starting point, I volunteer to merge the two pages while conserving their current meanings.

If you can tweak the search categories in this area, welcome! This will be a great help.

raiph:
> `lazy ...;` is probably unwanted enough to warrant a compile time warning. ... Lacking discouragement, I will submit an issue for this.

I presume that should always have raised a compile time sink warning. But, as I once wrote in [an SO answer](https://stackoverflow.com/a/54469963/1077672):

[...]

PR #4151, opened 2022-12-02
Consolidate quietly into statement-prefix.pod6 from control.pod6.

0rir:
Put SPoT for `quietly` into `language/statement_prefixes` following from  #4117.

### DIFF
--- a/doc/Language/statement-prefixes.pod6
+++ b/doc/Language/statement-prefixes.pod6
@@ -80,8 +80,8 @@ don't care.
 
 =head2 X<C<quietly>|Syntax,quietly (statement prefix)>
 
-As a prefix, C<quietly> suppresses all warnings produced by the
-block or statement it precedes (currently only run-time warnings).
+As a prefix, C<quietly> suppresses all run-time warnings produced by the
+block or statement it precedes.
 
 =for code
 sub marine() {};
@@ -91,7 +91,7 @@ quietly { told-you; warn 'kaput!' };
 warn 'Telling you now!';  # OUTPUT: «Telling you now!␤ [...] ␤»
 
 Calling L<C<.Str> on C<code> produces a warning|/type/Code#method_Str>.
-Preceding the code with C<quietly > will just produce the output without warning.
+Preceding the code with C<quietly> will just produce the output without warning.
 
 =head2 X<C<try>|Syntax,try (statement prefix)>
 
--- a/doc/Language/statement-prefixes.pod6
+++ b/doc/Language/statement-prefixes.pod6
@@ -80,16 +80,18 @@ don't care.
 
 =head2 X<C<quietly>|Syntax,quietly (statement prefix)>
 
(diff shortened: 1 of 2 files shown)


doc/Language/statement-prefixes.pod6
```diff
@@ -80,16 +80,18 @@ don't care.
 
 =head2 X<C<quietly>|Syntax,quietly (statement prefix)>
 
-As a statement prefix, C<quietly> suppresses all warnings produced by the
-statement it precedes.
+As a prefix, C<quietly> suppresses all warnings produced by the
+block or statement it precedes (currently only run-time warnings).
```
coke, line 85: Suggestion:

``
As a prefix, C<quietly> suppresses all (run-time) warnings produced by the
block or statement it precedes.
``

doc/Language/statement-prefixes.pod6
```diff
@@ -80,16 +80,18 @@ don't care.
...
 sub marine() {};
 quietly say ~&marine; # OUTPUT: «marine␤»
+sub told-you { warn 'hey...' }; 
+quietly { told-you; warn 'kaput!' };
+warn 'Telling you now!';  # OUTPUT: «Telling you now!␤ [...] ␤»
 
 Calling L<C<.Str> on C<code> produces a warning|/type/Code#method_Str>.
-Preceding the statement with C<quietly > will just produce the output, the name
-of the routine.
+Preceding the code with C<quietly > will just produce the output without warning.
```
coke, line 95: Can you remove the extra space in the C<>? I know it was there before, thanks.

doc/Language/statement-prefixes.pod6
0rir: Done.

doc/Language/statement-prefixes.pod6
0rir: Done.  @coke, thanks twice.

coke:
Having these as 10 different PRs makes it more challenging to review. About how many more are there coming?

0rir:
Sorry about that, @coke.  I was asking in  #4117 if this should be procedurally  different than a small "one and done" PR.  If it will help,  I can stack the commits on a branch and submit the whole ball of wax.  As they stand, the PRs are independent by each touching only the one topic in each file.

The original problem that I saw was duplication between control.pod6 and statement-prefixes.pod6.  I feel that `do`, `lazy`, `eager`, `hyper/race`, and `sink` can all stay in statement-prefixes.pod6.  So other than tracking down links and such, we can say this is done, or not.

Either way, I intend to address the following:
- moving `try` to the exceptions topic,
- moving `react` to the currency topic,
- renaming statement-prefixes to prefixes, code-prefixes, or whatever.

The above glosses over likely disagreement about the destination location of some of the above.  That would efficient to resolve before hand, at least for me.

coke (approved):

PR #4152, opened 2022-12-02
Altering Online communities section of the community page

2colours:
Includes removal of IRC bot list - it's also present on the raku.org site, apart from the undersightable bot that doesn't seem to work.


coke (approved):

coke:
I'm happy with these; any reason not to take it out of draft?

2colours:
I felt something was still missing, for example I was hoping for any feedback regarding the undersightable bot, at least. I think if we can settle that undersightable is gone for good, we can call it a day.

raiph:
FWIW I've never heard of undersightable. But I see [this](https://gist.github.com/Whateverable/6abf1cbf1f77a6ec617f3303a34ea410), which looks like a legit generated bot result from 5 years ago.

> I think if we can settle that undersightable is gone for good, we can call it a day.

Can't we just remove things for now anyway? Forgiveness > permission, and if someone wants to get broken things working again, and they do, and the newly working thing is considered useful, a reference to it can always be added back where appropriate, right?

coke:
I'm happy if this gets merged.

2colours:
If nobody has objections - apparently - then sure, go ahead.

By the way, is there no such principle as "communication > ivory tower"? "forgiveness > permission" is pretty good but ultimately, the general indifference. often excused by volunteer work, is causing a lot of damage to the community - emotional damage just as much as quality problems.

PR #4158, opened 2022-12-07
fix cross-reference formats

tbrowder:
## The problem

Several instances of cross-references are in the wrong format or have invalid categories.

## Solution provided

Fix the format, and use a valid (but possibly not the most appropriate) category.


coke:
This resets your `similarly` change?

coke:
Applied this fix by hand in 7e91a0018

PR #4164, opened 2022-12-10
Add a practical example of class introspection

tbrowder:
## The problem

The current docs do not show a practical, simple, and safe example of the use and implementation of class introspection. 

## Solution provided

This PR attempts to rectify the problem by including such an example at the end of the *Classes and objects* tutorial.

### DIFF
--- a/doc/Language/classtut.pod6
+++ b/doc/Language/classtut.pod6
@@ -930,4 +930,79 @@ class instances; in this case, the class method uses the alembic symbol, and the
 instance method, defined below it, aggregates the data we have on the cook to show
 it in a narrative way.
 
+=head2 A practical introspection example
+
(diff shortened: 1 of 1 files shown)


doc/Language/classtut.pod6
```diff
@@ -930,4 +930,79 @@ class instances; in this case, the class method uses the alembic symbol, and the
 instance method, defined below it, aggregates the data we have on the cook to show
 it in a narrative way.
 
+=head2 A practical introspection example
+
+When one creates a new class, it is somtimes useful to have
```
coke, line 935: `sometimes`

coke (approved):

tbrowder:
Fixed type with 9c8feb4

Issue #2138, opened 2018-07-03
Fix left sidebar in the "Language" page


uzluisf:
## The problem
Whenever I'm reading the docs and reach a certain point, I need to do some scrolling if
I want to get back to the table of contents and jump to a certain section.

## Suggestions
Make the sidebar (table of contents) fixed so that it stays in place when scrolling the doc.

zoffixznet:
> I need to do some scrolling

FWIW, you can also press "Home" button.

> Make the sidebar (table of contents) fixed

Is there a way to make it behave decently when the entire table of contents doesn't fit on the screen? How to scroll it if it stays fixed?

cfa:
Related: pages without a table of contents (e.g. [this](https://docs.perl6.org/language/temporal)) currently occupy far more horizontal space than pages with toc entries (e.g. [that](https://docs.perl6.org/language/create-cli)).

We should probably render an empty toc div to ensure consistency of presentation?

sdondley:
This is a must. Docs are a real pain to use. Even just a simple css "sticky" tag for the TOC would be a huge improvement.

raiph:
[What is a css "sticky" property?](https://www.google.com/search?q=What+is+a+css+%22sticky%22+property%3F)

Altai-man:
The whole site needs a re-design at this point, but the last attempt to do so was rejected by the community.

Since then a year has passed and nobody volunteered to contribute more time on this to make it happen. The original contributors don't have much time (and experience in some areas) to put into it due to life circumstances, and the young contributors were forcefully shooed away. Due to this, the docs project is maintained to some minimum degree, but I would not expect much from it.

sdondley:
It makes an html element "stick" to a specific location on the screen when you scroll down the page. See demo here: https://philipwalton.github.io/polyfill/demos/position-sticky/

I use it on my site as well to more dramatic effect: https://climatechangechat.com/setting_up_lsp_nvim-lspconfig_and_perl_in_neovim.html

I took a quick look at the Raku doc site. The TOC is done in a float so it's not going to be quite so simple. Some changes to the structure of the html would need to happen. But it's not a huge task.

patrickbkr:
@Altai-man Speaking about a year having passed. Would it be possible to set up the redesigned docs page on a more permanent URL to pass around / look at?

Altai-man:
>Would it be possible to set up the redesigned docs page on a more permanent URL to pass around / look at?

[...]

PR #4177, opened 2022-12-15
Add scrollbar for navbar

### LINKED ISSUE #4176: On Firefox, difficult to scroll through the table of contents on the left of each page
## Problem or new feature

See this screenshot: 
2022-12-15-175315_1105x731_scrot
Note that I've scrolled halfway down the page, but the table of contents is still at the top.
In order to see the end of the TOC, it seems I need to scroll all the way to the bottom of the page.
Related: the behaviour when I search through the page in Firefox is confusing normally Firefox scrolls to the current search result, but it doesn't if the search result is in the TOC, so the search result often ends up being off-screen.

I tried it in Chromium and found the TOC scrolls with the rest of the page:
2022-12-15-175517_1103x768_scrot
This seems much better to me, because it is much easier to scroll through the TOC (with the minor disadvantage that the TOC scrolls completely off-screen at some point).

codesections:
The navbar was made sticky in b2e1f85 to avoid situations where the user needs to scroll the main content up to use the ToC navbar pane (as described in #2138). However, this resulted in some portions of the ToC being unreachable when the ToC was long enough that it didn't fit on a single screen.

This PR fixes that by adding a scrollbar to the navbar.  Fixes #4176.

### DIFF
--- a/assets/sass/style.scss
+++ b/assets/sass/style.scss
@@ -395,6 +395,9 @@ td, th {
         float: left;
         position: sticky;
         top: 0;
+        height: 100vh;
+        overflow-y: scroll;
+        overflow-x: hidden;
     }
 
     table#TOC {


2colours:
I think it would be much better to reflect the state of [the new site](https://github.com/Altai-man/docs.raku.org) because this site probably won't be updated ever again anyway.

coke:
Rejecting - we can't currently deploy the existing site, and @Altai-man 's new site is a different code base for the UI.

PR #4180, opened 2023-01-13
Simplify links and fix syntax

Prince213:
## The problem

Some links are prefixed by https://docs.raku.org/, which is not needed. Also, one link is using wrong syntax for links.

## Solution provided

Links are simplified, and correct syntax are used.


coke:
Good catch, thanks!

PR #4193, opened 2023-02-08
NFG link modifications

2colours:
I added another 6guts blog link that I found particularly useful about the topic - probably more useful than its prequel. Also "deseriousalized" the chat discussion: it's clearly cute and funny if you already kind of grasp the topic but perhaps a newcomer shouldn't take it too seriously, that wouldn't seem good PR. :v


coke:
Thanks.

PR #4198, opened 2023-02-11
Making "our declaration" more accurately described

2colours:
## The problem

As @niner pointed out, the docs had the "what is an alias of what" backwards.

## Solution provided

 This change brings the description closer to S04-control, which also seems more factual for this matter.


lizmat (approved):

coke:
Thanks!

Issue #119, opened 2015-08-29
Match.chars missing


MARTIMM:
Hi,

According to synopsis of regex (S05) the Match class has a ```chars()``` method which returns the difference of the values of methods ```to()``` and ```from()```. On the doc page about Match there is a ```chars()``` method mentioned under the Cool class. But that one has a different meaning.

Greets

moritz:
It's not a different meaning. Match stringifies to the matched string, and its length is the same as `.to - .from`. And `Cool.chars` coerces to Str, so the two are the same.

PR #4205, opened 2023-02-20
remove TM char for consistency

finanalyst:
This addresses issue #119 in doc-website, was in Raku/doc Either all references to Raku should have tm or none. My preference is for none as it is obvious

### DIFF
--- a/doc/Language/101-basics.pod6
+++ b/doc/Language/101-basics.pod6
@@ -1,6 +1,6 @@
 =begin pod :kind("Language") :subkind("Language") :category("beginning")
 
-=TITLE  Raku™ by example 101
+=TITLE  Raku by example 101
 
 =SUBTITLE A basic introductory example of a Raku program
 


coke:
Thanks for this - team, hold off on applying this until we have a consensus on the plan for the trademark symbol.

lizmat:
Trademarks only work if you mention and enforce them.  Having been dragged to court about that in a previous lifetime, I'm all too aware of this.

FWIW, I think that specifying the trademark on the homepage, and perhaps a mention that all references to "Raku" should be interpreted as having the trademark mentioned, in small lettering at the bottom, should be enough  :-)

finanalyst:
I absolutely agree with having a TM mark with Raku on one or more of the index pages (homepage, language), or in a footer on all pages.

Adding a TM to Raku in 101-Basic  because it was the first content page to appear is a hack and should be removed.

coke:
no response from @codesections - so we'll go with this for now, and add it in a few key places on the website itself

PR #4207, opened 2023-02-21
Replace the 'Opener graphemes' table with 'Brackets'

tbrowder:
The original table was incomplete in that it
didn't show the closing bracket for each opener.

The new table shows the opener/closer pairs
and provides some helpful introductory
information. In addition, the new table
has a header row to guide the user
on its contents.


tbrowder:
Yes, I will add that.

tbrowder:
So done.

PR #4210, opened 2023-02-23
Change supported Unicode version

thundergnat:
It's kind-of dependent on the VM it's running on. Moar is up to Version 15.0 at least.

Interesting enough, the repository said 13.1, the [live site ](https://docs.raku.org/language/unicode) says 12.1. I would suspect that unicode.pod6 hasn't been regenned for a while.


coke:
Looks good; we should probably have a table of supported versions by Rakudo compiler range at some point.

PR #4215, opened 2023-02-26
Modify the 'Brackets' table generator to read the data source file

tbrowder:
The original table generator still has the embedded data, but its current default is to read the source file and extract the data so it can be updated if the source should change.

For now the script allows for other options that can be removed when the default mode satisfies the current requirements.

Possible changes:

+ modify the behavior for a failed download of the source HLL::Grammar.nqp
+ change the table name
+ change the pod6 file name

### DIFF
--- a/util/create-brackets-table.raku
+++ b/util/create-brackets-table.raku
@@ -1,74 +1,102 @@
 #!/usr/bin/env raku
 
-# global array defined in the BEGIN block at the end:
-my @bracket-chars;
+# TODO
+#   + get desired exit behavior from Coke
+#   + settle on table name and output file name
 
-my $f = "../doc/Language/brackets.pod6";
+# Global array initially defined in the BEGIN block at the end.
+# The array is then defined dynamically by reading the source
+# file at https://github.com/Raku/nqp/src/HLL/Grammar.nqp.
+our @bracket-chars;
+
 if !@*ARGS {
     say qq:to/HERE/;
-    Usage: {$*PROGRAM.IO.basename} go | reorder
+    Usage: {$*PROGRAM.IO.basename} go | Reorder [refresh][original][debug]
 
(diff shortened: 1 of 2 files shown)


util/create-brackets-table.raku
```diff
@@ -1,74 +1,102 @@
...
+# Global array initially defined in the BEGIN block at the end.
+# The array is then defined dynamically by reading the source
+# file at https://github.com/Raku/nqp/src/HLL/Grammar.nqp.
+our @bracket-chars;
 
-my $f = "../doc/Language/brackets.pod6";
 if !@*ARGS {
     say qq:to/HERE/;
-    Usage: {$*PROGRAM.IO.basename} go | reorder
+    Usage: {$*PROGRAM.IO.basename} go | Reorder [refresh][original][debug]
```
coke, line 14: Let's pick an order and stick with it. Supporting 3 different order types when we'll only use one is adding maintenance for this script for no value - this would remove the need for this option entirely. (If we keep the option: please use lowercase "reorder" making it easier to type.

util/create-brackets-table.raku
```diff
@@ -1,74 +1,102 @@
...
+    when /^ r/ { ++$refresh }
+    when /^ R/ { ++$reorder }
+    when /^ o/ { ++$want-original }
+    when /^ d/ { ++$debug }
+}
 
+# The output table file:
+my $repopath = $*PROGRAM.IO.absolute.IO.parent(2);
+my $fdir1 = $repopath ~ "/doc/Language";
+my $f1    = "$fdir1/brackets.pod6";
```
coke, line 46: You have based this against an old version - I've already made similar changes to the file that are committed in the repository.

You'll have to rebase on your end and fix the PR or submit a new one.

coke (changes requested):
Need to rebase and resubmit.

tbrowder:
Will do.

PR #4219, opened 2023-02-28
Rewrite the "make your own operator" example

2colours:
Relates to https://github.com/rakudo/rakudo/issues/5221

This example got attention because of a related regression, however it was a severely problematic example:
- the implementation was incorrect
- hence the described output wasn't matched
- `||=` exists natively anyway so this example didn't illustrate what it claimed to

Here is a proposed alternative that does illustrate the point and the output is also correct.

### DIFF
--- a/doc/Language/js-nutshell.pod6
+++ b/doc/Language/js-nutshell.pod6
@@ -363,16 +363,16 @@ operator overloading natively! Since all operators are subroutines, you can
 define your own like so:
 
 =begin code
-multi sub infix:<||=>($a, $b) is equiv(&infix:<+=>) { $a || $b }
+# "distance operator": the distance of two numbers is the absolute value
+of their difference
+multi sub infix:<|-|>($a, $b) is equiv(&infix:<->) { abs $a - $b }
 
-my $foo = 0;
-$foo ||= 1;
-say $foo; # OUTPUT: 1
+say -1 |-| 3; # OUTPUT: 4
 =end code
 
 Operators can be defined as C<prefix>, C<infix>, or C<postfix>. The
 C<is tighter>, C<is equiv>, and C<is looser> traits optionally define the
-operator's precedence. In this case, C<||=> has the same precedence as C<+=>.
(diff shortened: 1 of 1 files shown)


lizmat:
What makes you think that `||=` exists natively?

lizmat (approved):

2colours:
This does: https://glot.io/snippets/ginhwcf4f0

lizmat:
Ok, I guess we have a misunderstanding on what is meant with "natively".

In my book, it means that the operator exists in a native form in that way.  This means you can ask for its identity:

````raku
dd &infix:<||>;  # Sub infix:<||> = proto sub infix:<||> (|) {*}
````

You cannot do that for `||=`, because that's a meta-operator.  The `=` on **any** existing infix operator, will change `foo op= bar` into `foo = foo op bar`.  And that includes `||`.

Yes, you don't have to do anything explicitely to make `||=` work.  But it really doesn't exists until you use it.

2colours:
I think, in the context of a "from JS to Raku" introduction, it's better if "your own operator" means something that isn't syntactically already understood. What's the point of "making your own operator" if it worked for the purpose of the example without making it (and fun fact, doesn't work afterwards)?

Out of curiosity: is it language-specified that something like `||=` would *not* have an operator identity, or is that just Rakudo in action? Because so far I could imagine it either way.

lizmat:
> I think, in the context of a "from JS to Raku" introduction, it's better if "your own operator" means something that isn't syntactically already understood.

I agree, that's why I approved this PR

> Out of curiosity: is it language-specified that something like ||= would not have an operator identity, or is that just Rakudo in action? Because so far I could imagine it either way.

Oddly enough, I found out that things like `&infix:<+=>`  actually **do** return something akin to an identity, although it's generated:

````raku
say &infix:<+=>;  # -> \a, \b { #`(Block|4078171635712) ... }
````

Except for `||=` and `&&=`, as these bomb, probably because `||` and `&&` are really macros in Raku because of their short-circuiting characteristics.

PR #4224, opened 2023-03-01
Correct the intended zero-width index entries

tbrowder:


coke (changes requested):
You added a user visible TODO here.

Is there an open issue and/or docs about what you're trying to fix here? (Why is "Classes" wrong?)

tbrowder:
No, I forgot to remove it after my edits.

tbrowder:
Requested changes sbmitted.

tbrowder:
> Why is "Classes" wrong?

My whole intent when I submitted this doc chunk was to (1) ensure the programmatic technique was exemplified and (2) its location could be found with a search for "programmtic." This secion is specifically on class and instance methods, not classes themselves. As I understand the X<> use, using the "Class" before the pipe would have been misleading.

The description of X<> could probably be improved with an explicit example of this situation. I'll see about that.

PR #4227, opened 2023-03-01
Update faq.pod6

finanalyst:
edit the zef url so that it is recognised by modules.raku.org remove reference to p6doc, which is not being found there,


coke (changes requested):
p6doc is now called rakudoc

finanalyst:
reverting and re-editing with new name

PR #4229, opened 2023-03-01
Update modules-extra.pod6

finanalyst:
## The problem
p6doc no longer available on modules.raku.org

## Solution provided
Remove reference to it and link


coke (changes requested):
New name is

https://modules.raku.org/dist/rakudoc

finanalyst:
reverting and replacing with new name

PR #4230, opened 2023-03-01
Update DateTime.pod6

finanalyst:
## The problem
Link attached to 'propleptic' is tagged as a security hazard using Firefox,

## Solution provided

Replace link with similar content in Wikipedia.


coke (changes requested):
Lost the https

2colours:
Also - pretty sure it was just a typo from wik**t**ionary

cfa:
As @2colours pointed out, the intended link was probably https://en.wiktionary.org/wiki/proleptic.

finanalyst:
reverting to original, then correcting typo

PR #4234, opened 2023-03-02
Document Date.daycount|Int|Numeric

lizmat:


coke:
Passes xtest, looks good, thanks.

PR #4239, opened 2023-03-03
Update faq.pod6

finanalyst:
change malformed links and change to rakudoc


coke (commented):
Changed the zef to an incorrect one. I'll apply this p6doc change manually.

PR #4242, opened 2023-03-04
Partially revert changes to `Distribution::Resource`

cfa:
Reverts two lines changed in 4118f0c: the directory name has yet to be changed from `perl6` to `raku`.


coke:
good catch

Issue #2954, opened 2019-08-15
Styling of types


uzluisf:
## The problem

While proofreading some documentation pages (e.g., [types](https://docs.perl6.org/type)), I've noticed that when a particular type is mentioned, the type is either linked to (e.g., `L<Hash/type/Hash>`), formatted as code (e.g., `C<Hash>`) or just plain text (e.g., `Hash`). In some instances, the same type is linked to multiple times over the length of the page. Other times, it's never linked to but just formatted as code. In other, it's just plain text w/o any formatting that blends with the surrounding text.

## Suggestions

Most type pages have two top-level sections: the introduction to the type (e.g., `class Hash`) which follows the class's title and subtitle and the Methods (and Type Graph) sections. In the following paragraphs, I'll focus only on the introduction section which itself might have sections. The case for routines documentation under the Methods section seems to be [already covered here](https://github.com/perl6/doc/commit/0d57bfa1d37e6948ab941f661c2474475dd4bf8f#diff-da89ae830582df8fe02e0957d2ee00f0R35-R38). Thus depending on how these pages are meant to be read, we could've some sort of convention when styling the types in the introduction:

* **Top-to-bottom**. If each type page is meant to be read from top-to-bottom, then:
    
     - The first time a type is mentioned, we should link to it (`L<>`). Subsequent mentions (regardless if it's in the same section, same section's subsection, new section, etc.) should be formatted as `C<>` accordingly.
     - References to types and other constructs in a specific code example should be formatted with `C<>`. However, some exceptions may apply.

* **Granular**. If each type page is meant to be read granular-ly, then:

     - In a single h1-level section, the first time a type is mentioned we should link to it (`L<>`) and subsequent mentions of it within the same section (and its subsections) should be formatted as `C<>`. New mentions of the same type in other h1-level sections follow the same procedure.
     - References to types and other constructs in a specific code example still should be formatted with `C<>`. However, some exceptions may apply.

JJ:
I see merit in your argument. The problem is to actually enforce it. Either we get expert proofreaders, or we have somehow to develop an xt test to check that. If that's the case, I'd happily adopt that policy (if everyone else is happy with it)

coke:
Adding onto the new [xt/rakudoc-c.rakutest] in origin/type-links.

Lots of e.g.`C<Hash>` - the WIP test is recommending these appear as C<L<Hash|/type/Hash>>.

coke:
[...]

PR #4256, opened 2023-03-06
Type links

coke:
## The problem

Inconsistent formatting when referring to types in the docs that have their own primary page. (#2954)

## Solution provided

```xt/rakudoc-types.rakutest```

### DIFF
--- a/xt/rakudoc-types.rakutest
+++ b/xt/rakudoc-types.rakutest
@@ -1,0 +1,142 @@
+#!/usr/bin/env raku
+
+=begin overview
+
+When referring to items that are types, the required format is:
+
+   L<C<Thing>|/type/Thing>
+
+Any other formatting code that refers to a type will fail the test; any C<>
+that isn't inside of an L<> will fail, and any L<> that doesn't have a C<>
+will fail. Links may end with an optional #id.
+
+Exceptions:
+
+=item Referring to a type on its own page should only use C<>.
+=item It's OK to refer to a routine page with the same name instead.
+
+=end overview
+
+use Test;
+use lib $*PROGRAM.parent(2).child('lib');
+
+use Test-Files;
+use Pod::Convenience;
+
+my @files = Test-Files.pods;
+
+if @files {
+    plan +@files;
+} else {
+    plan :skip-all<No rakudoc files specified>
+}
+
+# TODO: Renders #tags oddly.
+sub render-node($node) {
+    my $type = $node.contents.join('');
+    my $result = $node.type ~ '<' ~ $type;
+    if $node.type eq 'L' {
+        $result ~= '|' ~ $node.meta.join('');
+    }
+    $result ~= '>';
+
+    $result;
+}
+
+# given a slashy type, see if that file exists on disk
+# To work on case-insensitive file systems, we grep the dir listing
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by coke** [comment]
need to allow #tags on the links, probably.

**Comment by cfa** [comment]
This is also erroneously treating links like the following at bad,

```
L<C<Proc/Async|/type/Proc::Async>> should be L<C<Proc/Async>|/type/Proc/Async>
```

**Comment by coke** [comment]
We should standardize one way or the other, let's pick.

**Comment by cfa** [comment]
A few more false positive cases:

- Index entries
  - e.g. `Failed test 'X<Scalar> should be L<C<Scalar>|/type/Scalar>'`
    - This is from the heading `=head2 X<Scalar|Tutorial,scalar (Basics)>` in [Language/101-basics.rakudoc](https://github.com/Raku/doc/blob/main/doc/Language/101-basics.rakudoc#L128).
  - Suggestion: exclude these from consideration altogether.
- Explicitly formatted terms
  - e.g.
    - `'I<method> should be L<C<method>|/type/method>'`, 
    - `'I<block> should be L<C<block>|/type/block>'`
  - Two subissues here: 
    1. "method" (like "block", "map", "match", "raku" etc.) should probably be excluded from consideration as FPs are very likely.  (More on this below.)
    2. I'm not sure that we should assume that explicitly formatted terms are missing _additional_ formatting codes. It's probably reasonable to omit `I<...>`, `B<...>` etc. from consideration altogether (or to have the test check specific codes). This could cover the `X<...>` case too.
- Case insensitivity
  - e.g. `'L<exception|/type/Exception> should be L<C<exception>|/type/exception>'`
  - Two issues here:
    1. `/type/exception` was derived from "exception" and located on disk (a byproduct of running this on a case insensitive filesystem?)… but the casing is incorrect (this should be `/type/Exception`, as it was originally).
    2. When word and type case don't match, perhaps skip? "exception" here—like "method" above—is likely to be a false positive. Nevertheless, requiring consistent case between word and type name might omit a few more FPs?

**Comment by coke** [comment]
* I only have case insensitive filesystems for easy access testing, agree we have to address that.
* Will definitely exclude `X<>`
* my thought about `B<>` and `I<>` was that they were likely intended to be type refs but were non standard, so wanted to call them out. if we want `B<stuff>`, we could have `B<L<C>|..>>`?

**Comment by cfa** [comment]
Sounds good.

(The false positives for `B<>` and `I<>` might be addressed by a combination of word exclusion and case sensitivity.)

File: xt/rakudoc-types.rakutest [hunk 1]
```diff
@@ -10,7 +10,10 @@ Any other formatting code that refers to a type will fail the test; any C<>
 that isn't inside of an L<> will fail, and any L<> that doesn't have a C<>
 will fail. Links may end with an optional #id.
 
-One exception: Referring to a type on its own page should only use C<>.
+Eexceptions:
```
patrickbkr, line 13: Ee

coke:
> Rebased accidentally against master and then (hopefully correctly) against main.

Just did the same thing again!

coke:
Most of the issues from cfa's review are covered. One doc file is now passing in the branch - will merge back to main and start addressing the formatting everywhere.

PR #4257, opened 2023-03-06
Add test for <> (missing formatcode)

coke:
## The problem

Authors occasionally writing `<foo>` instead of, e.g.: `C<foo>`

## Solution provided

xt/rakudoc-brackets.rakutest


coke:
All tests pass in branch, minor updates to docs.

PR #4263, opened 2023-03-11
Add utility to test doc-website templates

coke:
## The problem

templates in doc-website exist but are not verified to pass raku/doc's basic pod structure.

## Solution provided

util/test-website

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ util/Grammar.nqp
 
 # IDE
 *.iml
+
+# website clone for testing
+doc-website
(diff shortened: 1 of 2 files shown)


coke:
If we keep the tests in raku/doc and the templates in doc-website, what's the advantage of running the tests from either area? Either way we have to cover the other repo, yes?

cfa:
Just that this requires cloning the other repo in order to run tests on its content.

coke:
And in the other case, we clone this repo to get the tests to run.

coke:
So, since either way we have to do a clone to get the tests and the files in the same place, let's go with this one for now.

finanalyst:
see https://github.com/Raku/doc-website/issues/103

PR #4265, opened 2023-03-12
Cleanup examples and some text

### LINKED ISSUE #4264: fix examples on doc/Type/Buf.rakudoc
## Problem or new feature

The examples in type/Buf require the entire document for a reader to understand; as per writing-docs/EXAMPLES.md, on a Type page, the examples should be self contained.

coke:
Closes #4264

### DIFF
--- a/doc/Type/Buf.rakudoc
+++ b/doc/Type/Buf.rakudoc
@@ -190,7 +190,7 @@ with the removed elements.
 
     my $bú = Buf.new( 1, 1, 2, 3, 5 );
     say $bú.splice:  0, 3, <3 2 1>;  # OUTPUT: «Buf:0x<01 01 02>␤»
-    say $bú.raku;                    # OUTPUT: «Buf.new(3,2,1,2,3,5)␤»
+    say $bú.raku;                    # OUTPUT: «Buf.new(3,2,1,3,5)␤»
 
 =head1 Methods on buf8 only (6.d, 2018.12 and later)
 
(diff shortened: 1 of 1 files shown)


coke:
@2colours - can you double check my work here?

doc/Type/Buf.rakudoc
```diff
@@ -120,72 +118,79 @@ Returns a C<List> of integers.
...
-Substitutes elements of the buffer by other elements
+Substitutes elements of the buffer by other elements, returning a buffer
+with the removed elements.
 
-=for code :preamble<my $bú>
-$bú.splice:  0, 3, <3 2 1>;
-say $bú.raku; # OUTPUT: «Buf.new(3,2,1,2,3,5,8,13,21,34,55,89)␤»
+    my $bú = Buf.new( 1, 1, 2, 3, 5 );
+    say $bú.splice:  0, 3, <3 2 1>;  # OUTPUT: «Buf:0x<01 01 02>␤»
+    say $bú.raku;                    # OUTPUT: «Buf.new(3,2,1,2,3,5)␤»
```
2colours, line 193: Seems to me that this output is actually wrong, and it was wrong previously as well. It will be `3, 2, 1, 3, 5`, without the two.

2colours:
Thank you - I think apart from the inherited output error spotted, it resolves the problem of out-of-context results involving `Buf`.

doc/Type/Buf.rakudoc
coke: good catch, fixed.

PR #4272, opened 2023-03-18
Update modules.rakudoc

finanalyst:
Link url has a typo (should be `/language/filename-extensions`, not  `/Language/...`

## The problem
Link for filename-extensions leads to a 404


coke:
This should have been caught by xt/rakudoc-l.rakutest, except the test was run on a case-insensitive filesystem.

coke:
Inspired updated test 02a2b1342 - which wouldn't have caught this because of missing leading / (which may inspire another test)

PR #4275, opened 2023-03-21
Fix example output

kjpye:
## The problem

The length of a vector (2,3) is most definitely not 6.

## Solution provided

Change the numbers to give nice integral values, with the correct output.


coke:
Good catch- shouldn't the second updated example print 13?

kjpye:
No. The point of the second example is that it doesn't update the length.

PR #4281, opened 2023-03-29
mention dist tools

tony-o:
## The problem

The how to section for distributing your module could also mention that dist management tools can also make the boilerplate faster

## Solution provided

Added `fez` to the list of ways you can initialize a skeleton dist repo


coke:
Thanks!

PR #4284, opened 2023-04-04
Add Buffers section to the JS/NodeJS to Raku migration guide

uzluisf:
## The problem

Missing section about buffers on the NodeJs - Raku migration guide.

## Solution provided

Add a section about buffers  on the NodeJs - Raku migration guide.

### DIFF
--- a/doc/Language/js-nutshell.rakudoc
+++ b/doc/Language/js-nutshell.rakudoc
@@ -830,7 +830,7 @@ constructs in NodeJS and Raku:
        Class/Role  | NodeJS      | Raku
     ===============+=============+===========
     `Buffer`/`Buf` | Fixed-length sequence of bytes (No methods such as `push`, `pop`, etc.) | Sequence of bytes that can grow or shrink dynamically. You can use methods such as `push`, `pop`, etc.
-                   | Iterable using the `for..of` syntax                                     | It cannot be iterated over using a looping construct. Use the method `list` to get hold of an iterator.
+                   | Iterable using the `for..of` syntax                                     | It can be iterated over using a looping construct.
                    | Each byte can be updated using array indexing, e.g., `buf[i]++`.        | Same as NodeJS.
     `Blob`         | Fixed-length sequence of bytes (No methods such as `push`, `pop`, etc.) | Same as NodeJS.
                    | It's not iterable.                                                      | Same as NodeJS.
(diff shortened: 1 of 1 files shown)


uzluisf:
>Nice and elaborate work, I'd like to also check it rendered. I have two slight issues with the content.

Thanks! If I were to render it locally, would I need to add a screenshot or something?

>The more important is that Bufs (and Blobs alike!) can be iterated using the `for`structure, even though they do not mix Iterable in.

Understood! I think I've a wrong notion about iterables then. I was using the definition given in this [page](https://docs.raku.org/language/iterating.html#The_Iterator_and_Iterable_roles) (and the pages for `Iterator` and `Iterable`). For example, in Raku v2022.02 a `for` loop doesn't iterate over each item within a buffer:

```
my $buf = Buf.new(98, 97, 99);
.say for $buf; # Buf:0x<62 61 63>
```

Again, this is probably a misunderstanding on my part and only using it to show my thought process when I was writing this doc section :)

>The other thing is that I find toJSON confusing, even in JS. It returns a JS object (probably one that can round-trip from JSON text), not actual JSON text. In the Raku version, you are using a Hash which makes perfect sense but then I wouldn't say it "returns a similar JSON object", nor would I name the function toJSON.

I agree with your assessment. Overall I don't think it adds much value, and thus can be removed.

2colours:
> Thanks! If I were to render it locally, would I need to add a screenshot or something?

That can work but I thought I could do that on my own (haven't gotten to it yet). It's not that difficult to generate the files with a local instance of the website; all one needs to do is to checkout your branch.

> my $buf = Buf.new(98, 97, 99);
> .say for $buf; # Buf:0x<62 61 63>

Oh this is a tricky one... `$buf` is created with a `Scalar` container and the concept of that is mixed with itemization. You could even have a `List`, it would still iterate as one item. 

However, if you remove the extra layer using [zen slicing](https://docs.raku.org/language/subscripts.html#Zen_slices) - a fairly common way to achieve decontainerisation - you will see that it can be iterated on a per-element basis.

```raku
my $buf = Buf.new(98, 97, 99);
.say for $buf[]; # 98\n97\n99
```

You could also just bind(!) it to a `@variable` and then use it as something that naturally looks like a list. The advantage is that you can just use the assignment from that point on and it will set the content of the Buffer the way it makes sense. (The disadvantage is that you can't downright replace the object without binding.)

2colours:
@uzluisf oh, another thing. Please rebase the branch to the up-to-date main. There was a breakage that has been fixed in the meantime so this branch can't be built, despite the fact that it wouldn't break anything for the current state of things.

coke:
The rebase would be nice but isn't strictly necessary, we can do it when accepting the PR.

uzluisf:
@2colours Thanks for the great explanation.

>$buf is created with a Scalar container and the concept of that is mixed with itemization. You could even have a List, it would still iterate as one item.

I totally forgot about `Scalar`'s itemization, despite being one of its key features. I figured I can iterate directly over the object (e.g., `.say for Buf.new(98, 97, 99);`) or bypass containerization by binding it as you stated.

2colours:
What still holds up for me is that `Blob`s can be iterated on the same terms as `Buf`s.

uzluisf:
>What still holds up for me is that Blobs can be iterated on the same terms as Bufs.

Fixed! A blunder on my part :)

2colours (approved):
Thank you very much. I'm not an expert of this field but if it matters, it looks good to me now. :)

uzluisf:
>Thank you very much. 

You're welcome!

>I'm not an expert of this field but if it matters, it looks good to me now. :)

Me neither :)! You're probably more experienced and knowledgeable than me so you can gauge where I stand haha. My intention is to add a raw diamond if you will (assuming we can ignore the presumptuousness), in the hope that more experienced Rakoons can polish it and fill in the gaps.

PR #4286, opened 2023-04-14
Describe additional word splitters for quote protection

2colours:
## The problem
It's not obvious that interpolating word quotes break on any interpolation, even without a space.

## Solution provided
I added this remark and the analoguous remark for "sub-strings" in general. Also, I thought an example of being able to interpolate within qww "sub-strings" can give the reader a bit of sanity boost.

### DIFF
--- a/doc/Language/quoting.rakudoc
+++ b/doc/Language/quoting.rakudoc
@@ -386,10 +386,16 @@ Thus, if you wish to preserve quoted sub-strings as single items in the resultin
 you need to use the C<qww> variant:
 
     say qww{"a b" c}.raku; # OUTPUT: «("a b", "c")␤»
+    
+The delimiters of sub-strings are always considered word splitters:
+
+    say qww{'alpha'beta'gamma' 'delta'"epsilon"}.raku; # OUTPUT: «("alpha", "beta", "gamma", "delta", "epsilon")␤»
 
 Other kind of quotes are also supported
 
-    say qww{ ’this and that’ “here or there” ｢infinity and beyond｣ }.raku;
+    my $one = 'here';
+    my $other = 'there';
+    say qww{ ’this and that’ “$one or $other” ｢infinity and beyond｣ }.raku;
     # OUTPUT: «("this and that", "here or there", "infinity and beyond")␤»
 
 =head2 X<Word quoting with interpolation: qqw|Syntax,qqw>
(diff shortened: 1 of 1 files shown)


coke:
I prefer "substring" over "sub-string" to the point that I'm going to add this to xt/word-variants. Can you make that update?

2colours:
To be honest, the reason I don't like using "substring" here is that its de facto meaning is "any continuous slice of characters within a string" while here it's about a full string that happens to be embedded in this quoting structure.

This is not so disturbing when talking about "preserving [quoted] sub(-)strings as items" or "the delimiter of sub(-)strings" but it seems quite disturbing at "[interpolation happens] inside sub(-)strings". Like where? Between two arbitrary characters? The point is meant to be that it can happen in quoted strings, or embedded strings.

A somewhat broader objection is that I feel in `<<'foo bar'baz qux'xd lol'>>`, `'baz qux'` is kind of a "quoted substring", and that an embedded string could be arbitrarily deep regarding embedding. It would be good to avoid these misinterpretations when it's quite possible that someone unsure will eventually say: "well, what do the docs say about it"?

I get the point that distinguishing "substring" and "sub-string" can also lead to confusion. What do you think about a rephrasing that embraces the aspect that "quote protection" _allows you to embed_ certain quoting structures, and then talk about embedded strings?

jubilatious1:
"Quote Protection" has always seemed excessively jargonistic to me.  OTOH, I prefer substring (no-hyphen). Period. 

1. "String with single identical prepended/postpended character" == quoting (single or double quoting).

2. "String with a matched pair of prepended/postpended characters" == parens, square brackets, curlie braces (not applicable to this conversation).


PS. Please, whatever we do here, let's document that we're miles ahead of the mess below:
 
https://www.austingroupbugs.net/bug_view_page.php?bug_id=249

coke:
> rephrasing

sure, that works

2colours:
Sorry, I missed the messages somehow.

For what it's worth, I also don't like "quote protection"; actually it rather makes me think of the opposite thing i.e you are allowed to include quotes in your strings... but it's kind of coined. Maybe it even appears in links. I don't think this is the right time to change it, if there even is a right time.

I raised my objections about "substring" (it makes sense but means something else) and I maintain them with regards to "interpolation happens inside [quoted?] substrings". I maintain my position and hope that talking about "string embedding" is not too heavy on the jargon side.

Hopefully I can show a new version today.

2colours:
Updated now.

coke (commented):
Should be "other kinds" - sorry I missed this earlier.

2colours:
Fair enough... I was also thinking about that but English is not my first language, after all. Changed now.

coke (approved):

PR #4287, opened 2023-04-14
Inconsistent commafication in precedence table

0rir:
Commas got into the last two lines of some nice work.


coke (approved):

coke:
Good catch, thanks.

coke:
Those commas have been there since the original version of the table, 11 years!

PR #4297, opened 2023-04-27
Name the right type of bracket in :100[99, 2, 3]

falsifian:
## The problem

I think angle brackets are `<>`; this text is referring to `[]`.


coke (approved):

coke:
Good catch, thanks.

PR #4307, opened 2023-05-21
Fix typo in structures documentation

roosemberth:
The sentence is explaining the difference between `deepmap` and `duckmap` as it continues to exemplify `duckmap`, but it was erroneously typed `deepmap`.


lizmat:
Thanks for the spot!

Issue #4294, opened 2023-04-23
Parameter.modifier is ENODOC


gfldex:
In https://docs.raku.org/type/Parameter.html `method modifier` is missing. This returns type-smilies with `Str` values of <:D :U :_>.

PR #4312, opened 2023-05-28
Document Parameter.modifier

uzluisf:
## The problem

[`Parameter.modifier`](https://github.com/rakudo/rakudo/blob/c2449ef601a12eec8014b8893f90ab5151d7c869/src/core.c/Parameter.pm6#L307C5-L313) was undocumented. See #4294.

## Solution provided

This PR documents `Paramete.modifier`.

image

### DIFF
--- a/doc/Type/Parameter.rakudoc
+++ b/doc/Type/Parameter.rakudoc
@@ -373,6 +373,20 @@ returns the empty string.
     say $named-sig.params[0].suffix; # OUTPUT: «!␤»
     say $named-sig.params[1].suffix; # OUTPUT: «␤»
 
+=head2 method modifier
+
+    method modifier(Parameter:D: --> Str:D)
+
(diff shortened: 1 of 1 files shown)


coke:
Looks good, but I would prefer that the text of the link not be type smiley, here. it's a nickname, and isn't mentioned until several paragraphs in on the linked page.

uzluisf:
What's a better term? I could link to the term in the [glossary](https://docs.raku.org/language/glossary#Type_smiley).

coke:
If you add type smiley in the glossary, link to that and then have a glossary entry link to the original page that’s fine

uzluisf:
I linked straight to the term in the glossary. In the glossary, there was already a link to the [page](https://docs.raku.org/language/signatures.html#Constraining_argument_definiteness) this was linking to originally.

Issue #4188, opened 2023-01-30
Remove new() method showing positional arguments from Mu.new() documentation?


sdondley:
## Problem or new feature

Looking at https://docs.raku.org/type/Mu#method_new

It shows a multi method which accepts a positional argument. However, that method just throws an error: https://github.com/rakudo/rakudo/blob/2022.12/src/core.c/Mu.pm6#L121

Not sure it makes sense to leave this multi method in the docs when its only purpose is to throw an error.  

## Suggestions

Remove the `multi method new($, *@)` line from the documentation or explicitly explain that this version of the method will only throw an error.

coke:
Thanks, removed.

PR #4313, opened 2023-05-28
Remove multi new candidate

uzluisf:
## The problem

Remove the `multi method new($, *@)` candidate from the docs. It throws an error when used, and going by the description of `Mu.new` it wasn't meant to be documented.

See #4188.

## Solution provided

Remove `multi method new($, *@)` from the docs.


coke:
Thanks

Issue #4299, opened 2023-05-03
uniname(s) & uniparse should refer to each other.


coke:
Wherever `uniname`(s) and `uniparse` are defined, a reference to the inverse routine should be present as well (this is already in some, but not all places)

PR #4315, opened 2023-05-31
Cross reference uniname(s) and uniparse routines

uzluisf:
## The problem

The routines uniname(s) & uniparse should refer to each other. See #4299.

## Solution provided

The routines uniname(s) & uniparse now reference each other when they're defined. Also surround some types with `C<>`.


coke:
Thanks!

PR #4327, opened 2023-07-17
fix broken shell link

2colours:
See https://github.com/Raku/doc-website/issues/270


lizmat:
Thanks!

2colours:
@lizmat I just discovered that unfortunately that won't just cut it: the whole site used to be under language/ and now it's under type/, there are even tests for the old path and about a dozen similar links lurking around. :( So at this point I don't know if this fix really fixes anything or just does more harm.

PR #4329, opened 2023-07-21
Improve examples in ff() and related docs

sjn:
## The problem

When quickly looking at a specific example for ff() (or related functions), the examples
provided don't show the core functionality of the function – namely that the function ignores
elements until the first match, and after the second match.

## Solution provided

Add an element in the example array before and after the objects matched, so the
main functionality of the functions is show, in addition to the nuance offered by the
specific variation.

### DIFF
--- a/doc/Language/operators.rakudoc
+++ b/doc/Language/operators.rakudoc
@@ -2859,7 +2859,7 @@ condition).
 
 A comparison:
 
-    my @list = <A B C>;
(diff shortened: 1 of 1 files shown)


coke:
Please remove the temp files - these should be set in your local .gitignore, not the project one (we should consider removing the IDE one that was in there also)

coke (approved):

PR #4331, opened 2023-07-24
Polish the "list prefix" entry

2colours:
Addressing https://github.com/Raku/doc/issues/4073

Moreover removing fake "operators" that are mere
functions from the list.


coke:
Thanks

PR #4334, opened 2023-07-30
Add a reference to Whitespace

tbrowder:
+ Grammar tweaks
+ Add more examples
+ Align examples of Unicode naming corrections for clarity
  - Use comments to distinguish correct versus incorrect naming versions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,7 +77,7 @@ channel and/or the [issues for this repository](https://github.com/Raku/doc/issu
 before you proceed further. After you get consensus on a title, subtitle,
 section, and filename, you can add the document by following these steps:
 
-+ create a **filename.pod6** file in the **doc/Language** directory and
++ create a **filename.rakudoc** file in the **doc/Language** directory and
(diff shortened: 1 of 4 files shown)


coke (approved):

coke:
Thanks - note the ```X<...>``` needed to be ```X<|...>``` which I had forgotten since the last time I looked at xt/search-categories.

tbrowder:
I forgot about the proper format. I did look at another example and thought I had copied it correctly, but, alas. I'll check the docs again. Maybe that's why "make test was failing!"

tbrowder:
Coke, I did look at the docs before I used the reference link, but I failed to finish it as I intended. My initial motivation for the PR was because an existing search for "whitespace" turned up very little, and nothing about that section. I searched all the source docs and found no reference links. 

What I think that line needs to look like is more like this (as I keep forgetting, the part before the vertical bar is not case-sensitive, but the part following it is):

    =head1 X<whitespace|Language, whitespace> Whitespace...

coke:
OK - this one is merged, feel free to change it directly if you have a small update.

PR #4339, opened 2023-08-11
Remove numeric comparison operators from DateTime

2colours:
The numeric comparison operators were previously documented here because they were specifically implemented for this type.

However, this changed with https://github.com/rakudo/rakudo/commit/72856df91b2e1cafb439bcd8bcc7ae46c378d01e .

Now, the fact that this type can be used with these operators is simply a natural consequence of the provided coercion.


coke (approved):
~~Can we include a note~~

coke:
Thanks for the cleanup

PR #4342, opened 2023-08-18
Rakuast links

coke:
Updates the links test to use the new RakuAST API instead of the old parser.

Allows URLS in the same positions as the original test.

add utility to dump AST of doc files.

### DIFF
--- a/util/dumpast.raku
+++ b/util/dumpast.raku
@@ -4,5 +4,5 @@
 
 unit sub MAIN($file);
 
-dd $file.IO.slurp.AST.rakudoc
+.say for $file.IO.slurp.AST.rakudoc
 
--- a/util/dumpast.raku
+++ b/util/dumpast.raku
@@ -1,0 +1,8 @@
+#!/usr/bin/env raku
+
+# Generate the rakudoc AST for a given file
+
+unit sub MAIN($file);
+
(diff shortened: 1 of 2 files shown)


util/dumpast.raku
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env raku
+
+# Generate the rakudoc AST for a given file
+
+unit sub MAIN($file);
+
+dd $file.IO.slurp.AST.rakudoc
```
lizmat, line 7: `.say for ` instead of `dd` might give better readable results

lizmat:
If I get this correctly, you want a list of all the URLs in `L<>` markup, right?

I don't think `.rakudoc` is the best solution then.  You probably want to use `.map` on the RakuAST tree directly:

```raku
.say for $ast.map: { .meta.Slip if $_ ~~ RakuAST::Doc::Markup && .letter eq "L" }
```

coke:
I want all the urls that are NOT in an ell to throw an error. I’ll investigate the map

coke:
I don't think the map buys me anything, it still looks nested so I still have to walk the children, and I have a bunch of specific types of rakudoc I'm skipping, which I feel better about in the recursive walk sub than fitting into a map. Thanks.

PR #4346, opened 2023-08-23
fix typo in link to :kv adverb

jakebman:
## The problem

The link generated for :kv near https://docs.raku.org/language/subscripts#:k:~:text=in%20which%20case%20the%20return%20value%20will%20be%20determined%20by%20those%20adverbs is malformed.

## Solution provided

The pipe character appears to have been omitted. It is now provided.


coke:
Nice catch!

PR #4353, opened 2023-08-25
Using chained comparator

habere-et-dispertire:
Does this read better ?

~~~diff
-$/.Int <= 255 && $/.Int >= 0
+0 <= $/.Int <= 255
~~~

### DIFF
--- a/doc/Language/regexes.rakudoc
+++ b/doc/Language/regexes.rakudoc
@@ -2195,7 +2195,7 @@ octet:
 
 =begin code
 my $localhost = '127.0.0.1';
-my regex ipv4-octet { \d ** 1..3 <?{ $/.Int <= 255 && $/.Int >= 0 }> }
+my regex ipv4-octet { \d ** 1..3 <?{ 0 <= $/.Int <= 255 }> }
 $localhost ~~ / ^ <ipv4-octet> ** 4 % "." $ /;
 say $/<ipv4-octet>;   # OUTPUT: «[｢127｣ ｢0｣ ｢0｣ ｢1｣]␤»
 =end code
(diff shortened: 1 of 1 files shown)


coke:
I think so, yes.

raiph:
> `0 <= $/.Int <= 255`

I don't know the context, intent of the example, etc, but why not `0 <= $/ <= 255`?

Is there some numeric coercion edge case that I'm not taking into account but would occur in the intended context?

habere-et-dispertire:
> my regex ipv4-octet { \d ** 1..3 <?{ $/.Int <= 255 && $/.Int >= 0 }> }

It looks to me like the match could only be a number comprising of 1 to 3 digits and so an implicit cast is safe ?

FWIW if I run the code snippet, then I get the same result.

raiph:
Ah. OK. Yeah. But be careful. The regex probably needs to anchor the token start/end:
```
my regex ipv4-octet { ^ \d ** 1..3 <?{ $/ ~~ 0..255 }> $ } }
```
(otherwise you'll get things like `say '301' ~~ &regex ipv4-octet # ｢30｣`.)

And perhaps it ought be a `token` rather than a `regex`? (Not that it'll make any difference for this pattern in this case, but Rakoons, and thus examples, ought use `token` as their goto default rule kind, not `regex`.)

raiph:
Ah. OK. Yeah. But a few things to note:

* I was assuming the issue here was some code showing an example of chained comparison. If not, then fair enough.

* Regardless of anything else, at least for parsing an ip octet (or ip address) it should probably be a `token` rather than a `regex`. (Rakoons, and thus examples, ought generally use `token` as the rule kind they default to using, not `regex`. And if I understand what's intended with this example, a token would fail faster for many invalid inputs, and thus have better performance.)

* The `token` (or `regex`) needs to be carefully anchored. Consider, for example:

```
my token ipv4-octet { \d ** 1..3 <?{ $/ ~~ 0..255 }> }
say '301' ~~ &ipv4-octet; # ｢30｣  <-- Uhoh
```

* In practice one would likely write a `token` (or `regex`) that deals with a whole ip address in one go. Perhaps any example ought follow suit. Then again, it's definitely getting *very* complicated for a simple example:

```
my token ipv4 { ^ (\d ** 1..3) ** 4 % '.' <?{ $0.all ~~ 0..255 }> $ }

say '192.168.0.0.' ~~ &ipv4
```

habere-et-dispertire:
The document is about regexes, so not sure if going into `token` land is on topic. It feels like non-trivial changes needs liaison with the primary author. I was just making a small suggestion while reading it.

coke:
I agree just the small change is fine. Are you happy with the PR as is? @raiph can submit a separate PR for the change they are suggesting.

habere-et-dispertire:
Yes, thank you.

coke:
Done, thanks for the cleanup.

PR #4354, opened 2023-08-31
Show how to access the %?RESOURCES variable in run-time code

tbrowder:

### DIFF
--- a/doc/Language/variables.rakudoc
+++ b/doc/Language/variables.rakudoc
@@ -1340,7 +1340,7 @@ Module-Foo/
 =end code
 
 Additionally, a relative path (starting from the root resources directory of a
-distribution) to a file must be specified under C<"resources"> field in the
+distribution) to a file I<may> be specified under the C<"resources"> field in the
 C<META6.json> file:
 
 =begin code :lang<json>
@@ -1370,9 +1370,10 @@ all available resource files in a distribution by printing or by using other
 ways to inspect its value. Instead, use the API described above to
 access particular files.
 
-Since the C<%?RESOURCES> variable is only accessible as a compiled object,
-it requires a special effort to access it during testing or other run-time 
-use. One way to do that is to create a routine in the C<lib> directory to
+The C<%?RESOURCES> variable is only accessible inside of modules.
+If you want to access C<%?RESOURCES> outside of a module, you'll need
+to expose that API yourself.  One way to do that is to create a 
+routine in the C<lib> directory to
 return its value:
 
 =begin code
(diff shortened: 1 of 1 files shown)


doc/Language/variables.rakudoc
```diff
@@ -1340,7 +1340,7 @@ Module-Foo/
 =end code
 
 Additionally, a relative path (starting from the root resources directory of a
-distribution) to a file is specified under C<"resources"> field in the
+distribution) to a file must be specified under C<"resources"> field in the
```
ugexe, line 1343: I think the original text reads better, because nothing `MUST` be under the `resources/` field (its optional to list anything)

doc/Language/variables.rakudoc
```diff
@@ -1370,6 +1370,30 @@ all available resource files in a distribution by printing or by using other
 ways to inspect its value. Instead, use the API described above to
 access particular files.
 
+Since the C<%?RESOURCES> variable is only accessible as a compiled object,
```
ugexe, line 1373: I'm not sure this first sentence is entirely accurate, as its a little more nuanced than that. For instance there are other compile time variables that can be used during runtime (such as `$?FILE`) inside of an e.g. test.

Maybe it would be better to say something like `%?RESOURCES is only accessible inside of modules. If you want to access %?RESOURCES outside of a module you'll need to expose that API yourself. One way to do that is to create a routine in the C<lib> directory to
return its value: ...`

tbrowder: I'll work on changes tomorrow. Thanks.

doc/Language/variables.rakudoc
tbrowder: So done. 

BUT, should there be a note as to why it's even used? Is there a possibility it might be used someday? 

Also, is any mismatch caught by anything other than possibly App::Mi6?

doc/Language/variables.rakudoc
```diff
@@ -1370,6 +1370,31 @@ all available resource files in a distribution by printing or by using other
 ways to inspect its value. Instead, use the API described above to
 access particular files.
 
+The C<%?RESOURCES> variable is only accessible inside of modules.
+If you want to access C<%?RESOURCES> outside of a module, you'll need
+to expose that API yourself.  One way to do that is to create a 
+routine in the C<lib> directory to
+return its value:
```
ugexe, line 1377: Looks like this could go on the previous line

ugexe:
LGTM. There is an extra newline in one of the changes though

doc/Language/variables.rakudoc
tbrowder: So corrected.

tbrowder:
I think it's ready now.

ugexe (approved):

PR #4360, opened 2023-09-18
Mention ordering from dir is arbitrary

0racle:
Question came up in raku-beginner Discord about why `dir` was returning files in a "strange order", and it was noted the docs do not mention this.

### DIFF
--- a/doc/Type/IO/Path.rakudoc
+++ b/doc/Type/IO/Path.rakudoc
@@ -466,7 +466,7 @@ Returns the contents of a directory as a lazy list of C<IO::Path> objects
 representing relative paths, filtered by
 L<smartmatching|/language/operators#infix_~~> their names (as strings) against
 the C<:test> parameter. The path of returned files will be absolute or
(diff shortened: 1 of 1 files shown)


raiph (commented):
It looks like a tweak of the whole paragraph is appropriate (it currently starts "representing relative paths"), and I think it would help to make it explicit where the "arbitrariness" comes from (and maybe mention the lack of `.` and `..`), and slightly emphasize the "as strings" aspect of the filter (that's discussed in the next paragraph). So something like:

> Returns the contents of a directory as a lazy list of C<IO::Path> objects, optionally filtered by
L<smartmatching|/language/operators#infix_~~> their names I<as strings> per the C<:test> parameter. The file system determines the order of entries. Entries C<.> and C<..> are not included. C<$path> determines whether the path of entries are absolute or relative.

coke:
The change from @0racle is acceptable as is: @raiph, if you'd like to submit another PR on top, I'd be happy to take that one as well.

PR #4362, opened 2023-09-22
Update `rakudoc -h` output to match current rakudoc

codesections:
The [programs/02-reading-docs](https://docs.raku.org/programs/02-reading-docs) page showed usage info for the old version of `rakudoc`; this PR updates the info to the current version.


coke (approved):
I feel weird about having zef:coke here, but I guess it's just a reality of the current distribution.

coke:
Thanks.

PR #4365, opened 2023-09-22
Fix codos in `pick` signatures in Bool.rakudoc

raiph:
Dunno if I should just be committing changes like this...


lizmat:
Please do fix factual errors directly if you have the commit bit  :-)   Should anyone disagree, there's always the revert!

PR #4366, opened 2023-09-22
Rewrite `pick` description of Bool.rakudoc

raiph:
I found the existing prose description way too complicated to understand, and when rewriting it decided to separate the description and example that corresponds to one signature from the description and examples that corresponds to the other signature.


lizmat (approved):

coke:
Great, thanks!

PR #4367, opened 2023-09-24
Change `contains` snippet to a non-trap

2colours:
The earlier example reassured the misconception that it's a good idea to use `contains` to check for elements in a list. I changed it to a more idiomatic use of `contains`.


lizmat (approved):

coke:
Good catch, thanks.

Issue #4369, opened 2023-09-29
Second backtracking example is comprehensivly wrong?


codesections:
The second example in the [Backtracking](https://docs.raku.org/language/regexes#Backtracking) section of the Regex page appears to be entirely wrong. I'm referring to the example that starts:

> It is worth noting that disabling backtracking will not prevent the engine to try several ways to match the regular expression.

This example provides code that, according to the text, will result in only two iterations (without `:ratchet`).  Running that code, however, results in 30 iterations.  Additionally, the text incorrectly describes how the match proceeds by saying that it "will match against the rightmost word SQL and go forward from there" – which it won't.  I cannot tell what the example is trying to communicate.  It appears to be ~6 years old; maybe it was documenting old/buggy behavior?  I am inclined to favor deleting this second example entirely rather than trying to salvage it.

However, since it's fairly long and well-written, I thought I'd check if anyone else knows what it's trying to do and/or how to get it to do so.

pmichaud:
I can't speak for the rest of the example (yet), but the author misreads the result of the second example entirely and perhaps that confuses the entire answer.

The reason the example cited fails to match is not because there's not a second "SQL" to be matched, but rather because the first SQL that got matched is now in $0 instead of $1.   The author went from these prior examples:

```
my $string = 'PostgreSQL is an SQL database!';
say $string ~~ /(.+)(SQL) (.+) $1/;      # OUTPUT: «｢PostgreSQL is an SQL｣␤» 
say $string ~~ / :r (.+)(SQL) (.+) $1/;  # OUTPUT: «Nil␤» 
```

to this one

```
my $string = 'PostgreSQL is an SQL database!';
say $string ~~ / (SQL) (.+) $1 /; # OUTPUT: «Nil␤» 
```

and assumes that the failed match in the last one is because of backtracking, or that the first capture grabbed the second SQL, and other incorrect reasoning.  

The reason the match fails in this last example is because the ```(SQL)``` capture now appears in ```$0``` instead of ```$1``` as it did in the previous examples.  If you correct the backreference everything works as expected, including backtracking:

```
[0] > my $string = 'PostgreSQL is an SQL database!';
PostgreSQL is an SQL database!
[1] > say $string ~~ / (SQL) (.+) $0 /;
｢SQL is an SQL｣
 0 => ｢SQL｣
 1 => ｢ is an ｣
[1] > 
```

[...]

PR #4371, opened 2023-09-29
Add backtracking control metacharecters :, :!, and :?

### LINKED ISSUE #4370: Backtracking control characters are un-/under documented
Roast specs 5 terms to control backtracking in regexes (including tokens/etc).  Of these, 2 are NYI (`::` and `:::`).  But 3 are implemented in Rakudo and should be documented:
 - `:` — turns off backtracking for the preceding atom
 - `:!` — turns on greedy backtracking for the preceding atom
 - `:?` — turns on frugal backtracking for the preceding atom
 
Of these, only `:` is documented at all, and it's hiding under the [Quantifiers section](https://docs.raku.org/language/regexes#Preventing_backtracking:_:), which probably isn't the best place for it.  All three should be documented in the [Backtracking](https://docs.raku.org/language/regexes#Backtracking) section.

See #4356 for related discussion.

codesections:
This PR is an attempt to resolve #4370 by extending the language/regex#Backtracking section to include a explanations of the `:`, `:!`, and `:?` metacharecters.  This PR includes only new material; it's possible that some of the information from language/regexes#Preventing_backtracking:_: should be moved to this section (or deleted, see #4369), but this PR hasn't done so.

### DIFF
--- a/doc/Language/regexes.rakudoc
+++ b/doc/Language/regexes.rakudoc
@@ -2693,6 +2693,8 @@ eat up as much as possible in the first match and then
 adjust going backwards in order to ensure all regular expression parts
 have a chance to match.
 
+=head2 Understanding backtracking
+
 In order to better understand backtracking, consider the following example:
 
 =begin code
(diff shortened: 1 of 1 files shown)


coke:
tagging @pmichaud

pmichaud:
Overall the update looks good to me.  My only comment is that ```:?``` and ```?``` aren't merely "closely related" when used on quantifiers -- the ```?``` after a quantifier is in fact considered a shorthand for ```:?```.     Thus    ```.*?``` is really the shorthand of ```.*:?```.   Same for ```!``` and ```:!```.

It might not need mentioning, but this syntax enables ```.??``` (short for ```.?:?```) to be the frugal form of greedy ```.?```.   The frugal version first tries zero instances and then one instance, while the greedy form first tries one instance and then zero.

But all of these are relatively minor details, perhaps too much in the weeds for this section of documentation.

Pm

pmichaud:
Also, perhaps worth mentioning (not certain that this is the correct thread for it):

In the rakudo docs and in recent discussions I've seen several instances where people write a construct like ```[:ratchet \s*]```  or ```[:!ratchet \s*]``` to enable/disable ratcheting of a single quantifier.  This is exactly what the ```:``` and ```:!``` modifiers are for, to enable/disable ratcheting at an atomic level.   The ```:ratchet``` and ```:!ratchet``` adverbs were really intended to be used when you had several quantifiers or subexpressions where ratcheting should be enabled/disabled.

So, it might be worth going through the docs and seeing if any existing instances of ```:ratchet``` can be replaced with ```:``` to increase clarity.  Then more people would be likely to know that ```:``` even exists.

In my code I would have never written ```[:ratchet \d*]``` to turn off backtracking in a regex, I would have always written ```\d*:```.

Pm

codesections:
Thanks for the feedback.  I'll go ahead and merge this as is for now.  I'll put some thought into how to clarify the relationship between `:?`/`:!` and quantifiers – that might belong in the Quantifiers section.

Issue #4356, opened 2023-09-09
Error describing <ws>; it is not the same as <!ww>\s*


deoac:
On [Regexes > Predefined Regexes](https://docs.raku.org/language/regexes#Predefined_Regexes) the page says

> \<ws> | no | Whitespace, same as: <!ww>\s*

This is not quite true.  As nemokosch [explains](https://irclogs.raku.org/raku-beginner/2023-09-09.html), `\s` can backtrack, but `<ws>` cannot.  Here's an example to demonstrate the difference:
```
say "a\n" ~~ / 'a' <!ww>\s* \n /   #Output ｢a
                                   #       ｣   
say "a\n" ~~ / 'a' <.ws> \n /      #Output Nil
```

raiph:
I think the ws rule should be described as a token. So it's the same as:

    :r <!ww> \s* :!r

raiph:
Hi @deoac,

I'm wondering why you haven't responded. I worry that I wasn't clear enough so will be even more explicit below. Plus I have another thing to mention that's technically a red herring relative to this issue but might be confusing some folk so I'm going to write another comment about that too.

> "`<ws>` is not the same as `<!ww> \s*`.

That's incorrect. It is exactly the same.

raiph:
As I tried to demonstrate/explain in my first comment, the key doc improvement that would technically fully address this issue per its stated substance would be to explicitly classify `<ws>` as a `token`, though I recommend broadening it to cover all the "rules" in the table in the doc.

raiph:
The red herring that I chose not to mention before (precisely because it's a red herring relative to the substance of this issue), but may cause some folk confusion (because it *is* confusing -- as far as I can tell it directly contradicts statements in the design docs, the user docs, and @Larry discussions) is that `:sigspace` / `rule`s enable backtracking for the atom preceding any inserted `<ws>`. This is my current best formulation of the surprise discussed in Rakudo issue https://github.com/rakudo/rakudo/issues/3726.

deoac:
Hi raiph!

Your explanation that `<ws>` is a token, and hence doesn't backtrack, was very clear and helpful.

Thanks,
  
Shimon

2colours:
Well, "rules" also don't backtrack, if we can trust the documentation with this: https://docs.raku.org/language/grammars#Rules

[...]

PR #4372, opened 2023-09-30
Clarify built-in regexes

codesections:
This PR is my attempt to solve (most of) the issues discussed in #4356.  I've split the discussion of Raku's built-in regexes into two sections: one discussing the anchors/zero-width assertions and one discussing the two built-in tokens (`<ws>` and `<ident>`) – and calling them tokens.  I've also edited the table to show a "rexex equivalent" that uses the `:` backtracking control character.  Hopefully this will help clarify the ratcheting behavior of `<ws>`.

However, this PR does not document the fact that `<ws>` enables backtracking for the _previous_ atom.  That's largely because there's no consensus about whether that behavior is intentional (see raku/problem-solving#390; it seems likely that `<ws>`'s behavior is a bug).

### DIFF
--- a/doc/Language/regexes.rakudoc
+++ b/doc/Language/regexes.rakudoc
@@ -601,16 +601,22 @@ includes a leading dot: C«<.name>».
 
 =head2 Predefined Regexes
 
-Besides the built-in character classes, the following other rules are built into Raku:
+Besides the built-in character classes, Raku provides built-in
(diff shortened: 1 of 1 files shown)


coke:
as long as @pmichaud is active on tickets, tagging him here!

pmichaud:
This looks like a very good clarification to me.  Well done!

Pm

coke:
Thanks!

PR #4376, opened 2023-10-02
Clarify use of the '$?DISTRIBUTION.ver' (and '.content') methods

tbrowder:
+ Add examples to accompany the descriptions

### DIFF
--- a/doc/Language/variables.rakudoc
+++ b/doc/Language/variables.rakudoc
@@ -1487,12 +1487,32 @@ unit module MyFoo;
 
 sub module-version {
     say "MyFoo is version:";
-    say $?DISTRIBUTION.meta<ver>;
+    say $?DISTRIBUTION.meta<ver>; # OUTPUT: «0.0.1␤»
 }
+=end code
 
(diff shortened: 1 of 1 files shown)


raiph (approved):
I've never clicked an approve in a review before but I presume it's the PR equivalent of a heart and thumbs up emoji in an issue, and better than me just silently reading PRs and thinking "cool", which I do often enough that I'd like to confirm that me doing this is better. (And if instead it's noise, I know to go back to silent nodding of my head.)

coke:
@raiph  - you have a lot of feedback on PRs, this is a perfect way to say you approve of the PR. Helps me to get through them faster and get them merged.

coke:
My one question for @tbrowder : is the `:solo` necessary because of the variable being tested?

tbrowder:
Coke, I think I left it because it was on the original single block of code. I don't pretend to know how that works. I tried to do all the tests you recommend but it always fails with errors from Doc::TypeGraph.

coke:
OK. I'll clean it up in post if needed.

Issue #4379, opened 2023-10-08
Topic as implicit method invocant under-documented  


codesections:
When a method call is used without an explicit invocant, it's called on the topic (e.g., `for 1..3 { .say }`.  This is mentioned in a few places (and is used in multiple examples) but isn't prominently documented.  It's currently documented in passing at https://docs.raku.org/language/classtut#Methods

But it should probably be mentioned in [Language/syntax](https://docs.raku.org/language/syntax), [Language/operators](https://docs.raku.org/language/operators) and/or [type/Method](https://docs.raku.org/type/Method).

2colours:
I'm not a great fan of adding tutorial-ish stuff to `type/` pages; I think those pages are meant to be references of very specific things, like here the `Method` type, not "methods" as a broader concept.

With `language/syntax`, I'm not sure what you'd like to see. It points to the class tutorial at an appropriate place, in my opinion. If you'd like to see more, please give further hints...  
By the way, there is a problem there:

> When declared within a class, a subroutine is named "method"

This is utterly wrong. One can declare subroutines inside of classes, moreover I think one can even declare methods outside of classes. This is the reason I don't like arguments like "nobody ever complained so it must be good". For now, I may just open an issue for it because it's not immediately obvious to me how to fix that.

For `language/operators`, I agree. Perhaps you would be interested in [my rework ideas](https://github.com/Raku/doc/discussions/4337) for the operator? In this particular case, I'm not sure what to do. Perhaps the whole concept of "methodop" should go. Some of them can be used as prefix, implying the topic, but not all of them, e.g `>>.` notoriously cannot be called that way; it will be a syntax error. This could be yet another aspect of operators that could be put into a table, I mean whether they can be used as "prefix", implying the topic variable as the self-object, or not.

PR #4381, opened 2023-10-11
Expand code declaration/invocation syntax

### LINKED ISSUE #4380: methods are not "subroutines declared within a class"
## Problem or new feature

https://docs.raku.org/language/syntax#Subroutine_calls mentions methods the following way:

> When declared within a class, a subroutine is named "method": methods are subroutines invoked against an object (i.e., a class instance).

This is confusing at least but I think it's much rather outright wrong. Whether something is declared in a class or outside of it has little to do whether it's "declared within a class" or not.

## Suggestions

I don't know if it's a good idea in the first place to show methods under an entry called "subroutine calls" but in any case, being a method is much rather about argument handling and dispatch than the place of declaration. That property should be highlighted. Maybe it's described better at some place already.

codesections:
This PR replaces the "Subroutine declaration" section with a "Callable declarations" section, which now contains sub-sections on "Subroutine declaration", "Multiple dispatch subroutine declaration", "Block declaration", "Method declaration", and "Multiple-dispatch method declaration".  Similarly, it replaces the "Subroutine calls" section with an "Invoking code objects" section, which contains "Invoking subroutines or blocks", "Invoking methods", and "Method-like function calls & function-like method calls".

These sections document previously un- or under-documented syntax such as `method $invocant:` and `$obj.&subroutine`.

Closes #4380 and #4379

### DIFF
--- a/doc/Language/syntax.rakudoc
+++ b/doc/Language/syntax.rakudoc
@@ -857,18 +857,136 @@ See L<Variable Declarators and
 Scope|/language/variables#Variable_declarators_and_scope>
 for more details on other scopes (C<our>, C<has>).
 
-=head3 Subroutine declaration
+=head3 Callable declarations
 
-    # The signature is optional
-    sub foo { say "Hello!" }
+Raku provides syntax for multiple L<C<Callable>|type/Callable> code objects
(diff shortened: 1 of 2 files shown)


coke (approved):

coke:
Thanks!

Issue #3796, opened 2021-02-08
Curry is not defined


tbrowder:
The verb "curry" is not defined in the context of programming.

JJ:
Right. We have [whatever-currying](https://docs.raku.org/type/Whatever#index-entry-Whatever-currying), but not what's understood, in functional programming, as currying. Maybe add to the glossary?

tbrowder:
Yes, that would help. I notice Zoffix also created an issue on the whatever-currying discussion.

raiph:
There's no built in in Raku that does currying, there's no obvious affinity of the notion with anything in Raku, mentioning it is like mentioning COBOL picture clauses outside a COBOL to Raku guide. Why?

PR #4397, opened 2023-11-09
Replace incorrect use of "currying" with "priming"

codesections:
The documentation currently uses the term "currying" to refer to what is more broadly known as "partial function application" (but frequently incorrectly called "currying").  However, Raku generally doesn't use the phrase "partial function application", probably because it's a real mouthful and could be off-putting to people not familiar with functional programming jargon.  Instead, Raku prefers the term "priming", see
[S06](https://design.raku.org/S06.html#Priming).

Accordingly, this PR changes (nearly) all uses of "currying" or related words to the "priming" equivalent; it also adds a "priming" entry to the glossary.

The only two (intentional) uses of "currying" left are in a TODO section of the Haskell-to-Raku guide and in the documentation for the (Rakudo-specific, unspecified) type `CurriedRoleHOW`.

See related discussion with @raiph in the comments to [this StackOverflow answer](https://stackoverflow.com/a/77207543).

### DIFF
--- a/doc/Type/HyperWhatever.rakudoc
+++ b/doc/Type/HyperWhatever.rakudoc
@@ -13,7 +13,7 @@ than a single one.
 =head1 Standalone term
 
 Just like with L<Whatever|/type/Whatever>, if a
-L<HyperWhatever|/type/HyperWhatever> is used as a term on its own, no currying
+L<HyperWhatever|/type/HyperWhatever> is used as a term on its own, no priming
 is done and the L<HyperWhatever|/type/HyperWhatever> object will be used as-is:
 
     sub foo ($arg) { say $arg.^name }
@@ -33,7 +33,7 @@ being smartmatched.
 
 =head1 Currying
 
-When it comes to currying, the L<HyperWhatever|/type/HyperWhatever> follows the
+When it comes to priming, the L<HyperWhatever|/type/HyperWhatever> follows the
 same rules as L<Whatever|/type/Whatever>. The only difference is
 L<HyperWhatever|/type/HyperWhatever> produces a L<Callable|/type/Callable> with
 a L«C<*@> slurpy|/language/signatures#Flattened_slurpy» as a signature:
@@ -47,7 +47,7 @@ that simply maps each element in the arguments over:
     my &hyper-whatever = sub (*@args) { map *², @args }
     say hyper-whatever(1, 2, 3, 4, 5); # OUTPUT: «(1 4 9 16 25)␤»
 
-When currying, mixing L<HyperWhatever|/type/HyperWhatever> with
+When priming, mixing L<HyperWhatever|/type/HyperWhatever> with
 L<Whatever|/type/Whatever> is not permitted.
 
 =end pod
--- a/doc/Language/glossary.rakudoc
+++ b/doc/Language/glossary.rakudoc
@@ -817,6 +817,16 @@ easily applied using the GitHub web user interface. It means you request
 someone to do a git pull from your L<repository|#Repository> to hers. PR
 is its usual acronym.
 
+=head1 X<priming|Reference,priming>
+
+Priming (sometimes called L<partial function
+application|https://en.wikipedia.org/wiki/Partial_application> in other
+programming languages) is the act of creating a new function by providing some
+but not all of the arguments to an existing function.  For example, to create a
+function that takes one argument and adds 42 to it, you could prime the addition
+operator with C<42> as a single argument: C<* + 42>.  See
+L<Whatever-priming|/type/Whatever> and L<assuming|/type/Code#method_assuming>.
+
 =head1 X<property|Reference,property>
 
 In this context, it either refers to an
--- a/doc/Language/subscripts.rakudoc
+++ b/doc/Language/subscripts.rakudoc
@@ -183,7 +183,7 @@ important. Passing a bare negative integer (e.g. C<@alphabet[-1]>) like you
 would do in many other programming languages, throws an error in Raku.
 
 What actually happens here, is that an expression like C<*-1> declares a code
-object via L<Whatever|/type/Whatever>-currying - and the C<[ ]> subscript reacts
+object via L<Whatever|/type/Whatever>-priming - and the C<[ ]> subscript reacts
 to being given a code object as an index, by calling it with the length of the
 collection as argument and using the result value as the actual index. In other
 words, C<@alphabet[*-1]> becomes C<@alphabet[@alphabet.elems - 1]>.
--- a/doc/Type/Code.rakudoc
+++ b/doc/Type/Code.rakudoc
@@ -42,9 +42,10 @@ code object's C<Signature> do not contribute, nor do named parameters.
 
     method assuming(Callable:D $self: |primers)
 
-Returns a C<Callable> that implements the same behavior as the
-original, but has the values passed to C<.assuming> already bound to the
-corresponding parameters.
+Returns a new C<Callable> that has been L<primed|/language/glossary#priming>
+with the arguments passed to C<assuming>.  In other words, the new function
+implements the same behavior as the original, but has the values passed to
+C<.assuming> already bound to the corresponding parameters.
 
     my sub slow($n){ my $i = 0; $i++ while $i < $n; $i };
 
--- a/doc/Type/Whatever.rakudoc
+++ b/doc/Type/Whatever.rakudoc
@@ -12,8 +12,8 @@ gets its semantics from other routines that accept C<Whatever>-objects
 as markers to do something special. Using the C<*> literal as an operand
 creates a C<Whatever> object.
 
-X<|Syntax,Whatever-currying>
-Much of C<*>'s charm comes from I<Whatever-currying>. When C<*> is used
+X<|Syntax,Whatever-priming>
+Much of C<*>'s charm comes from I<Whatever-priming>. When C<*> is used
 in term position, that is, as an operand, in combination with most
 operators, the compiler will transform the expression into a closure of
 type L<WhateverCode|/type/WhateverCode>, which is actually a
@@ -37,7 +37,7 @@ Calling a method on C<*> also creates a closure:
 =for code
 <a b c>.map: *.uc;      # same as    <a b c>.map: -> $char { $char.uc }
 
-As mentioned before, not all operators and syntactic constructs curry
+As mentioned before, not all operators and syntactic constructs prime
 C<*> (or C<Whatever>-stars) to C<WhateverCode>. In the following cases,
 C<*> will remain a C<Whatever> object.
 
@@ -54,8 +54,8 @@ C<*> will remain a C<Whatever> object.
 
 =end table
 
-The range operators are handled specially. They do not curry with
-C<Whatever>-stars, but they do curry with C<WhateverCode>
+The range operators are handled specially. They do not prime with
+C<Whatever>-stars, but they do prime with C<WhateverCode>
 
     say (1..*).^name;       # OUTPUT: «Range␤»
     say ((1..*-1)).^name;   # OUTPUT: «WhateverCode␤»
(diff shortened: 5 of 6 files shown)


2colours:
I think the aim is noble but do you think this is enough? I mean, we have been using the term "whatever-currying" somewhat regularly for years now. How many involved people know that we are practically breaking away from that terminology?

lizmat:
I don't think this applies to whatever-currying, but more to what `.assuming` does?

duncand:
I support changing the documentation from "currying" to "priming".  The change makes the documentation more correct, and just because some people were accustomed to using words in the wrong way before, that doesn't mean they need to keep doing so going forward.  Using more standard terms will also make it easier for the wider world to come to Raku.  And a key thing is that it isn't changing actual language keywords so it isn't going to break code.

I will also say I'm surprised this wasn't already done almost a decade ago, as I distinctly recall being in a discussion about this very matter then.  I have my own other language heavily influenced by Raku, and I used the term "currying" originally to refer to the partial function application, following Raku's example, and then the discussion convinced me to refer to it as "priming" instead.  I had thought Raku was making the same change at that time.

codesections:
@lizmat wrote:
> I don't think this applies to whatever-currying, but more to what `.assuming` does?

But the point is that whatever-priming/currying does exactly the same thing as `.assuming`.  Here are functionally equivalent lines of code:

```raku
my &add1-v1 = 1 + *;
my &add1-v2 = &[+].assuming(1);

my &find-primes-v1 = *.grep(&is-prime);
my &find-primes-v2 = List.^lookup('grep').assuming(*, &is-prime);
```

And since using `*` achieves the same result as using `.assuming`, it makes sense to call them both "priming".

I didn't see this until now, but [S06 makes this point explicitly](https://design.raku.org/S02.html#Autopriming_of_Unary_and_Binary_Operators_with_Whatever):

> Perl 6 has several ways of performing partial function application. Since this is an unwieldy term, we've settled on calling it priming. (Many folks call this "currying", but that's not really a correct technical usage of the term.) Most generally, priming is performed on a Callable object by calling its .assuming method, described elsewhere. This section is about a convenient syntactic sugar for that.
> 
> For any unary or binary operator (specifically, any prefix, postfix, and infix operator), if the operator has not specifically requested (via signature matching) to handle * itself, the compiler is required to translate directly to an appropriately primed closure at compile time. We call this autopriming. 

Maybe ^^^^ is what @duncand is remembering being settled, and the docs just never got updated?

lizmat:
> Here are functionally equivalent lines of code:
>    my &add1-v1 = 1 + *;
>    my &add1-v2 = &[+].assuming(1);

Except that the first creates a `Callable` **out of thin air**, and the second one creates a `Callable` by wrapping an existing `Callable`.  I can see the second one be called "priming".  But calling the first one "priming" as well, feels wrong to me.

For instance, what would be the "assuming" version of `*.frobnicate` ?

codesections:
> For instance, what would be the "assuming" version of *.frobnicate ?

That's just my second example but without assuming any args.  So it'd be:

```raku
my &frobnicate = WhateverTypeHasFrobnicateMethod.^lookup('frobnicate');
# or, if you prefer to _explicitly_ assume nothing:
my &frobnicate2 = WhateverTypeHasFrobnicateMethod.^lookup('frobnicate').assuming();
```

(But all that does is let you call `frobnicate` with subroutine syntax, so you might as well have just used `frobnicate $arg:;`)

lizmat:
Except that method calls are late bound, so that won't fly.

In any case, to *me* there is a difference between creating a `Callable` out of thin air (which I think the syntax `*.frobnicate` is doing), and calling `.assuming` on an existing `Callable`.

codesections:
> Except that method calls are late bound, so that won't fly.

I see what you mean; `Str.^lookup('tc').assuming()` always calls the `Str` method, whereas `*.tc` will call the first `tc` method in the `mro` of whatever it's called on.

So here's how I'd describe that difference: `&assuming` is a function that takes a function and 0 or more arguments; it returns a function that calls the original function with the both the arguments passed to `&assuming` and the arguments passed to the inner function.

Or, in code a simplified version (e.g., no named args) of `&assuming` would be:

```raku
sub assuming(&code, *@args) {
    -> *@rest { code(|@args, |@rest) }
}
```

Whereas `*`, when used before a method call, is a function (er, syntax that is equivalent to a function) that takes a method-name and zero or more arguments.  It returns a function that takes 1 or more arguments; this inner function calls the method named in the outer function with the arguments passed to `*` and the arguments passed to the inner function.

Or in code (again, simplified and this time also ignoring `*` when used as an operand).:

```raku
multi whatever(Str $method-name, *@args) {
    -> $invocant,  *@rest { $invocant."$method-name"(|@args, |@rest)}
}
```

So, I now agree with @lizmat that these are two (subtly) different things.  And that `&assuming` meets the strict definition of "partial function application" whereas `*` does not – even though `*` _does_ perform partial application, it does something more.

But even if what `*` does isn't quite partial application, it's certainly not currying.  And I agree with S06 that it makes sense to call what `*` does "priming" – which, after all, doesn't have to be 100% the same as a strict definition of partial function application.  For example, when we have the expression `*.grep(&so)` it makes sense to say that we've primed the `grep` method with `&so`.  More generally, I'd say that – though your point about late-binding is important – what `&assuming` and `*` do is similar enough that we ought to use the same terminology when discussing them.

duncand:
@codesections wrote:
> I didn't see this until now, but [S06 makes this point explicitly](https://design.raku.org/S02.html#Autopriming_of_Unary_and_Binary_Operators_with_Whatever):
> 
> > Perl 6 has several ways of performing partial function application. Since this is an unwieldy term, we've settled on calling it priming. (Many folks call this "currying", but that's not really a correct technical usage of the term.) Most generally, priming is performed on a Callable object by calling its .assuming method, described elsewhere. This section is about a convenient syntactic sugar for that.
> > For any unary or binary operator (specifically, any prefix, postfix, and infix operator), if the operator has not specifically requested (via signature matching) to handle * itself, the compiler is required to translate directly to an appropriately primed closure at compile time. We call this autopriming.
> 
> Maybe ^^^^ is what @duncand is remembering being settled, and the docs just never got updated?

Yes, thank you!

I should also clarify that I don't necessarily support the actual changes of this specific pull request, but rather the principle of using "priming" rather than "currying" where it is the more clearly correct meaning.  Before it seemed like this was recognized but not completely followed through on years ago, like you found some vestigial uses of the wrong term.  But now I recognize it is possible that what you found is some cases where "currying" is referring to something else.  But I don't know enough to judge so I'm neutral on the current PR itself.

On a different subject matter I noticed it seems GitHub allows me to edit other people's comments, which is odd, I would have thought only the author of a comment would be allowed to edit it.

ab5tract:
Just a heads up that in RakuAST there is `CurryThunk`. This could be changed to `PrimeThunk` without any issue.

At first I was against this change but the more I think about it the more I think that "priming" something is a lot more evocative to me than "currying" something is, even though the latter still feels more familiar.

2colours:
My point is specifically that it might cause confusion if this change happens without people really knowing about it, and that in general it almost feels more important to inform people about it than what the doc says. If the doc just "contradicts" what people say and nothing reflects on that, I think that's almost automatically bad PR ("they are indecisive", "look how many of them use plain wrong terminology" etc.). Just think of the "path to Raku" topic; it would be great to do better than that, regarding clear communication and expressing some sort of expectations to people who want to help anyway.

raiph:
@ab5tract 

> At first I was against this change

Perhaps it would help to understand what you were initially against:

1. One issue is `.assuming`. In particular, which of the two aspects of the `.assuming` related verbiage change were you against? Was it the idea of getting rid of (arguably problematic use of) "currying" (by switching to some other suitable term)? Or was it the idea of picking "priming" instead of the term that's generally used for what `.assuming` does, namely "partial application" (fixing some arguments of a function), as per the Wikipedia page for Partial application?

2. Another issue is transforming `Whatever`s into `WhateverCode`s. The same two questions can be applied to this issue too.

> I think that "priming" something is a lot more evocative to me than "currying"

I imagine that's because "currying" isn't really an evocative word in terms of any programming feature unless you make the connection with Haskell Curry's nearly hundred year old technique of transforming a function of N parameters to N functions of 1 parameter and/or the word "currying", used since the 1960s to name use of that technique in a programming language context.
 
And if that's what it evokes it's the wrong concept for what it's doing for either `.assuming` or transforming `Whatever`s into `WhateverCode`s. For `.assuming` the correct concept is clearly "partial application", unless you think that "currying" means partial application, which leads to the arguably problematic aspect of thinking that. As the Wikipedia page on "Partial application" says:

> Partial application is sometimes incorrectly called currying, which is a related, but distinct concept

Larry held the same position as Wikipedia's, but thought "priming" worked as a layperson's alias for "partial application".

"Priming" has the advantages of:

1. Not being confusable with "partial evaluation";

2. Being shorter;

3. Not being constrained to the accepted definition of "partial application", which would arguably not work for what is done when `Whatever`s are turned into `WhateverCode`s;

4. Being evocative of the term/concept "partial application". I'm wondering if you can nail down what "priming" evokes for you. I'm thinking things like priming a surface before applying other coats of paint. I have a vague sense that it more generally connotes doing a preliminary part of some task that consists of doing several parts to complete the task, but I wonder if I'm inventing that.

I half wonder if we might be able to do better. But I know I routinely discover Larry's decisions reflect admirable wisdom, knowledge, and thinking about complications or opportunities to simplify things for most folk that I just don't see until I figure things out months or years after working with his design decisions.

> "currying" ... still feels more familiar.

Do you have a sense why? Is it just because that's the word that's been used in the Raku user doc since it was started? Or is it to do with use of the word outside of the Raku world?

The accepted answer to the 15 year old cs.stackexchange.com question ["What is the difference between currying and partial application"](https://stackoverflow.com/questions/218025/what-is-the-difference-between-currying-and-partial-application) rejects conflating "currying" and "partial application", as do the two current Wikipedia pages for the two concepts.

But perhaps we need to yield to misuse of the term? If such misuse is increasingly prevalent, and sufficiently so that it's clear that the majority of devs will one day think of it as a generic term that means something like "futz with a function so you get one or more variants of it that prebind some of its parameters", then maybe the better decision is to accept the will of the people that is to manifest at some future point. (I'm pretty sure that currently the opinion of the majority of devs who have an opinion on what "currying" means are aligned with the Wikipedia and cs.stackexchange.com view, and it may actually be the vast majority. Perhaps we need to assess that.)

If we do conclude misuse is or will be rampant, and conclude that an if you can't beat 'em, join 'em response is appropriate, then we presumably close this issue, and forget the line of my thinking and the exchange between @codesections and myself that led to him opening this issue.

----

(This all began with me reading @codesections's SO answer. It originally linked a mention of "currying" to the Wikipedia page of that name. That page describes currying as it's supposed to be. This made his link highly confusing, given that it was supposed to be explaining what `.assuming` did. I pointed out something to the effect that it would make more sense if his answer linked to the "Partial application" Wikipedia page. That describes _exactly_ what `.assuming` does. He responded to that by linking instead to the "Priming" section of the old design doc S06. I then responded to _that_ by suggesting we might want to consider switching to that. And the next thing you know he had filed this issue.)

codesections:
> My point is specifically that it might cause confusion if this change happens without people really knowing about it, and that in general it almost feels more important to inform people about it than what the doc says. If the doc just "contradicts" what people say and nothing reflects on that,

Those all seem like fair points.  Do you have a suggestion for what we could/should do to inform people/reflect on the change?  The only thing that's coming to mind for me is one of us writing a personal blog post, but I'm not sure if that would address your concerns or not?

2colours:
I think as far as the existing patterns go, writing, if only a short article about the rationale, and "circulating" it in the weekly/mastodon/reddit/IRC is mostly good. I can't immediately think of a better way now.

doc/Type/HyperWhatever.rakudoc
```diff
@@ -33,7 +33,7 @@ being smartmatched.
 
 =head1 Currying
```
silentTeee, line 34: Was leaving this section header as "Currying"  intentional, or was it just forgotten? Wasn't sure if it made sense to maintain this header text if we're trying to be consistent.

silentTeee (commented):
Think I may have found a missed case, but not sure.

coke:
See ticket #3796

coke:
I am in general in favor of this change; can we address existing comments and can someone summarize what's left to decide?

coke:
@codesections, @lizmat - is there anything left to address in this PR or can I apply it?

raiph:
@coke,

I'm with codesections (and Larry) that the word "Currying" is a poor choice (because what it refers to is not currying), and the word "Priming" is a good choice (because it's a good choice of English word that describes what it's doing), and thus the "Currying" heading @silentTeee spotted above should be changed to "Priming".

coke:
Did silentTree's as a separate followup commit. Thanks everyone.

PR #4409, opened 2023-12-22
Correct for bad links: classtut.rakudoc

finanalyst:
## The problem
Bad links in file `has` and `Attributes` are not source files.

## Solution provided

- `L<>` markup is used for `has`, but probably should be `I`/`B`. Chose `B`
- Missing `#` in attributes link, to refer to heading later


finanalyst:
closing as not correct. Missing link exists, but correction is wrong

PR #4410, opened 2023-12-22
Correct missing links faq.rakudoc

finanalyst:
## The problem
3 missing links

## Solution provided

- '/' should be '/index' as otherwise there are other errors
- 'language' will be automatically redirected to 'introduction', so this is a soft error
- 'with orwith with' does not have a composite file, so change to the main source 'language/control', which does


finanalyst:
The change to `language/control#` from `syntax/` can be excluded. Not sure how to remove this change from the patch.

finanalyst:
closing as incorrect.

PR #4411, opened 2023-12-22
Missing links classtut.rakudoc

finanalyst:
## The problem
2 Missing links 

## Solution provided

- Link to Attributes is to the heading in same file, but missing a `#`
- Link text `L<has>` would reference a primary source file `has`, which does not exist. But there is a `/syntax/has` file


finanalyst:
@cfa @coke I have largely fixed my link-error plugin. In doing so, I have eliminated most of the `index-entry` and other types of systemic errors.
The remaining link errors seem to be author errors. I am now working through them and suggesting changes. 

Please let me know if there remain systemic errors in the page generation software.

PR #4422, opened 2023-12-23
missing links - Update Hash.rakudoc

finanalyst:
## The problem

3 missing links all different

## Solution provided

- `L<Mu>` needs to point at a source, provided /type/Mu
- `L<subroutine|/language/...>` should be type/independent...
- `is default` points to an incorrect Composite. Found the most probable one.


finanalyst:
@coke The 'is default' missing link points at a deeper issue in the doc sources. If you type 'is default' into search, there are several, eg. is default, is default (Attribute)l etc. 
These should be combined into a single Composite, but they will not be because of the way the Source files X<> markup is created.

PR #4427, opened 2024-01-28
Update quoting.rakudoc

0rir:
Fixing  misdirection for regex interpolation rules.  I guess the /reference page will be rebuilt with the heading changed here.


dontlaugh (approved):
Looks good to me. I like moving this to the top of the page. This will remove a search result (a match on the heading "Regexes" for this page), but that seems like a good thing.

PR #4428, opened 2024-01-29
haskell partial application example in raku

k3ut0i:
a simple example of partial application in raku.
Couldn't figure out what `method chaining` has to do with currying, so I just left it as it is.


coke (approved):
can you remove the todo note?

looks good.

PR #4437, opened 2024-03-16
Update contexts.rakudoc

0rir:
My link on line 9 could use review.

Line 18 was poorly constructed.   Beyond that this is mostly about readability by use of the active voice, present tense, and by moving or deleting parenthetcal phrases.

My thanks to all the previous doc authors.


coke (approved):

coke:
Thanks!

PR #4439, opened 2024-03-21
Update typesystem.rakudoc "the ... metamethod .^."

0rir:
<.^.> reads like a typo.  I don't like my explicit parentheses as breaking the usual style of  links, but the link is unusual.


coke (approved):

coke:
Thanks

PR #4447, opened 2024-04-18
Add return type to Real .round

0rir:
I have no doubt that `.round` must return a definite value, but my belief is uninformed.

### DIFF
--- a/doc/Type/Real.rakudoc
+++ b/doc/Type/Real.rakudoc
@@ -125,7 +125,7 @@ otherwise.
 
 =head2 method round
 
-    method round(Real:D: $scale = 1)
+    method round(Real:D: $scale = 1 --> Num:D)
 
 Rounds the number to scale C<$scale>. If C<$scale> is 1, rounds to an
 integer. If scale is C<0.1>, rounds to one digit after the radix point (period or comma), etc.


coke:
So this isn't part of the source, and I don't think it's part of the spec. Given the implementation in Rakudo it seems reasonable:

``
    multi method round(Real:D:) {
        (self + 1/2).floor; # Rat NYI here, so no .5
    }
    multi method round(Real:D: Real() $scale) {
        (self / $scale + 1/2).floor * $scale;
    }
```

But I'd prefer to have the docs match the spec (and failing that, match Rakudo). Comments welcome

**Comment by Util** [comment]
KEY POINT: patch is incorrect about `Num`.

I agree that the compiler does not guarantee any return type, and that that is an argument against documenting a return type. 
(Do we have an official docs policy on that meta-point? I know we started with a scraping of the API from the compiler, and I see that we don't have anything to check for mis-matches over time, like `.round` being defined in the compiler's core.c/Int.rakumod, but missing completely from our doc/Type/Int.rakudoc . Maybe worth raising as a separate issue)

https://github.com/Raku/roast/blob/master/S32-num/rounders.t has lots of tests that could be interpreted as collectively implying "result must be defined", but I do not see an explicit test for such. 

More importantly, the patch is incorrect about *type*. Where the patch asserts `--> Num:D`, other types besides `Num` are currently delivered:

    raku -e 'my Real @rs = 15, 1.5, 1.5e0, pi; for @rs X @rs -> (Real $a, Real $b) { say $a.round($b).WHAT; }'|sort|uniq -c
       4 (Int)
       8 (Num)
       4 (Rat)

A reminder/primer for readers who are as confused as I was when I started in Raku: `Num` is not `Numeric`, and the similarity is confusing, and often when you think `Numeric` you mean `Real`:

    role Numeric : Number or object that can act as a number
    role Real : Non-complex number : `role Real does Numeric`
    class Num : Floating-point number : `class Num is Cool does Real`

Complex numbers, quaternions, octonions, etc are all mathematical objects that are Numeric because you want to do regular math with them, but are composed of many Reals. `Num` is IEEE floating-point, which in Raku we usually must force with exponent notation; `1.5e0` is a `Num`, while `1.5` is the `Rat` of `15` over `10`. Both are forms of `Real` and `Numeric`.


If this PR moves forward, I think a change to `--> Real:D` is required.

**Comment by raiph** [comment]
@Util,

I agree with the substance of your comment. I think the following may be obvious to all and/or trivial and/or nonsense but I'd like to try nail the following down:

> I agree that the compiler does not guarantee any return type, and that that is an argument against documenting a return type.

Is that accurate, or at least not problematic, for any definition of the words "guarantee any return type" in a computer science sense? I'm thinking one of us should post an SO question about this to try elicit a response by jnthn, after we first nail down the verbiage here.

My thinking is to make the question have two parts, one about the language, and the other about Rakudo. As an example of the kind of thing I'm unsure about, I *think* it's fair to say that both Raku and Rakudo guarantee that any value can be viewed as being of type `Mu`, or a subclass thereof, including any returned value, and that, presuming this is true, while it may be viewed as trivial and/or obvious, it may not be, or it may even be nonsense, and the truth be that either or both of Raku or Rakudo literally do not guarantee anything, for a computer science definition of those words. (I am of course ignoring bugs in Rakudo; while that feels like a related topic, it also feels fundamentally different.)

**Comment by raiph** [comment]
@Util,

> `...rounders.t` has lots of tests that could be interpreted as collectively implying "result must be defined"

I'd say no operation in Raku should be considered guaranteed to return a "defined" result (nor even a "definite" one which is what 0rir mentioned, and is not the same thing).

My reasoning is that *any* operation is allowed to return either a `Failure` instance (which is not `.defined`), or `Nil` which is a type object, so isn't even `.DEFINITE`:
```
sub foo (--> Int:D) { fail 'foo' }
sub bar (--> Int:D) { Nil }
say (.defined, .DEFINITE) given foo; # (False True)
say (.defined, .DEFINITE) given bar; # (False False)
```

0rir:
I submit that the return of a `Failure` is a red herring in the context of documentation of signatures (using the term loosely); because any `Callable` could fail.

Closing this and opening an issue.

PR #4450, opened 2024-05-11
Fix broken link in `uniprop`

### LINKED ISSUE #4449: bad unicode icu link
https://docs.raku.org/type/Cool#routine_uniprop has a link to

    http://userguide.icu-project.org/strings/properties

but this ends up on

    https://unicode-org.github.io/icu/userguide/

Looks like updated URL might be:

    https://unicode-org.github.io/icu/userguide/strings/properties.html

mustafaaydn:
Attempts to fix #4449 with a Wikipedia link. There is a more formal [link](https://www.unicode.org/versions/Unicode15.0.0/ch04.pdf) from the Unicode standard itself but i) the next sentence in the documentation refers to the Wikipedia page (a subsection therein) already and ii) formal link is harder to digest as it is a reference document and it's also a PDF.

### DIFF
--- a/doc/Type/Cool.rakudoc
+++ b/doc/Type/Cool.rakudoc
@@ -901,7 +901,7 @@ L«C<Str.samecase>|/type/Str#method_samecase» on it.
     multi sub uniprop(Int:D $code, Stringy:D $propname)
     multi method uniprop(|c)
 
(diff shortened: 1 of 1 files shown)


coke:
Great, thanks!

PR #4451, opened 2024-05-12
add announcements.rakudoc

finanalyst:
## The problem
new announcement / alert functionality created, needs content

## Solution provided
announcement.rakudoc


finanalyst:
@coke when this file merges to main, it will be picked by doc-website and the popups will start working
I thought after a bit that the first announcement should be about the announcements, then on Wednesday-ish  I'll add an announcement about the EBook. I thought  it would be more logical and less startling for experienced users.

coke:
Made new file pass 'make xtest', fixed minor typo (which wasn't caught by the `aspell` test!)

coke (approved):

coke:
Looks good to me; If someone else thinks my changes on top are OK, go ahead and merge.

Issue #4452, opened 2024-05-13
Add $¢ to list of lexical vars


coke:
https://docs.raku.org/language/variables#Pre-defined_lexical_variables is missing $¢ as a special lexical variable.

Noticed missing here: https://stackoverflow.com/questions/78466312/what-does-the-%c2%a2-variable-do-in-raku

PR #4459, opened 2024-05-13
Add $¢ to the list of always available lexical var

coke:
## The problem

#4452

### DIFF
--- a/doc/Language/variables.rakudoc
+++ b/doc/Language/variables.rakudoc
@@ -1112,15 +1112,16 @@ short.
 
 =head2 Pre-defined lexical variables
 
-There are three special variables that are always available:
+There are four special variables that are always available:
 
 =begin table
 
     Variable    Meaning            Scope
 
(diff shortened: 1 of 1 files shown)


doc/Language/variables.rakudoc
```diff
@@ -1112,15 +1112,16 @@ short.
 
 =head2 Pre-defined lexical variables
 
-There are three special variables that are always available:
+There are four special variables that are always available:
```
dontlaugh, line 1115: How should we word this? The special vars are not _always_ available.

Maybe something like

> There are four special variables that are available in certain scopes:

coke: I considered this, but you can access the variable anywhere, it's just Nil if we're not in the middle of a regex.

coke: But your point is taken. I'll update the text.

dontlaugh (approved):

Issue #4270, opened 2023-03-15
Add language/REPL


coke:
create a single page to discuss REPL and link disparate references to it.

coke:
consider moving the separate bits into this page, also.

PR #4462, opened 2024-05-16
First pass at Languages/REPL

coke:
For  #4270

### DIFF
--- a/doc/Language/REPL.rakudoc
+++ b/doc/Language/REPL.rakudoc
@@ -58,7 +58,9 @@ shortcuts for editing. There are currently four options:
 =item 3
 
 =for code :lang<shell>
-    zef install Readline # requires a readline install
+    zef install Readline
+
+This requires an installation of the Readline development library and a working C<C> toolchain.
 
 =item 4
 
--- a/doc/Language/REPL.rakudoc
+++ b/doc/Language/REPL.rakudoc
@@ -1,0 +1,71 @@
+=begin pod :kind("Language") :subkind("Language") :category("tutorial")
+
+=TITLE REPL
+
+=SUBTITLE Read-eval-print loop
+
+=head1 Overview
+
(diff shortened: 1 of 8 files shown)


doc/Language/REPL.rakudoc
```diff
@@ -0,0 +1,71 @@
...
+
+=item 1
+
+=for code :lang<shell>
+    zef install Terminal::LineEditor
+
+=item 2
+
+=for code :lang<shell>
+    zef install Linenoise
```
patrickbkr, line 56: Mention that this requires a working C build toolchain.

doc/Language/REPL.rakudoc
```diff
@@ -0,0 +1,71 @@
...
+
+In a fresh install, there is no command history. Running the C<REPL> in this mode will
+prompt you to install one of the various modules that provide this support. If enabled,
+you can use the arrow keys to scroll through previous commands and use standard terminal
+shortcuts for editing. There are currently four options:
+
+=item 1
+
+=for code :lang<shell>
+    zef install Terminal::LineEditor
```
patrickbkr, line 51: Technically this also requires a working C build toolchain at the moment, but I wouldn't mention that. It's a temporary thing. It depends on `Term::termios`, which has C code in it, but none of the `Terminal::*` modules require the functionality of that module that uses the C code. I plan to change this (possibly by creating a `Term::termios::PureRaku` or so), if no one beats me to it.

coke: Internalized. :)

doc/Language/REPL.rakudoc
patrickbkr: This is not right.

Readline does *not* require a C compiler, but *does* require the readline library to be installed.

Linenoise *does* require a C compiler as it contains and builds a full copy of linenoise (which is a C library).

coke: Resolved

PR #4473, opened 2024-05-24
Corrected CX::Oops to CX::Vaya (which matches the rest of the page)

wayland:
## The problem

For some reason, the documentation referred to CX::Oops, but the code used CX::Vaya.  

## Solution provided

Made documentation match code


coke:
Great catch, thanks!

PR #4480, opened 2024-06-01
Update contexts.rakudoc

voldenet:
Fix `Attempt to return outside of any Routine` error

## The problem
The subroutine would've not worked if it was properly called


dontlaugh (approved):
ty

PR #4495, opened 2024-06-28
Avoid pod6 markup in table cells

librasteve:
Sadly the table in my last commit - when rendered did not produce desired output. My POD6 skills are low, afaict it is not possible to embed markup in table cells. So I have rewritten accordingly and tested locally.

Feel free to edit if you can improve.

https://github.com/Raku/doc/issues/4493

### DIFF
--- a/doc/Language/signatures.rakudoc
+++ b/doc/Language/signatures.rakudoc
@@ -991,12 +991,14 @@ $sig2 = :(:$excludes = ['.', '..']);        # a new Array for every call
 Table showing checks of whether an argument was passed for a given parameter:
 
 =begin table
- Parameter kind | Example | Comment | Check for no arg passed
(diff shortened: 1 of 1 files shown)


lizmat:
@finanalyst I forgot what we decided for new style tables in rakudoc

finanalyst:
@librasteve @lizmat Markup in old tables remains difficult.

FWIW The new procedural table specification in RakuDoc v2 does not have this limitation, and in fact it is possible to have  table inside a table cell.

RakuDoc v2 can now be rendered into HTML, but it will be some months, at least, before the whole website is rendered with the new RakuDoc::To::HTML.

coke:
We'll take the PR as is for now and review formatting in tables when it's better supported.

PR #4497, opened 2024-07-04
Add simple `use` example

librasteve:
## The problem

Recently on the Discord beginner channel, a user was confused by the detailed Package / Module features and documentation

## Solution provided

Create a minimal example to declare and call a class in an external file in the same dir.

### DIFF
--- a/doc/Language/101-basics.rakudoc
+++ b/doc/Language/101-basics.rakudoc
@@ -461,6 +461,34 @@ say "we have @flavors.sort.join(', ')";
                                   # OUTPUT: «we have peach, vanilla␤»
 =end code
 
+=head1 Calling a raku method from another file
+
+To call a raku subroutine from another file, the convention is to have a minimal
+raku module or class file.
+
+=begin code
+# MyClass.rakumod  
+class C {
+    has $.x;
+}
+=end code
+
+Then your script can use this file.
+
+=begin code
+# myscript.raku
+use MyClass;
+ 
+my $instance = C.new(x=>42);
+say $instance.x;
+=end code
+
+The C<-I.> directive tells the raku interpreter to look in directory C<.>.
+
+=begin code
+> raku -I. myscript.raku
+=end code
+
 =head1 Exercises
 
 B<1.> The input format of the example program is redundant: the first line


lizmat:
Feels like at this point, it should also be mentioned (or at least referred to) on setting up a module for proper installation, and tools such as `App::Mi6`

librasteve:
@lizmat - I certainly get the point of Mi6 ... but I feel that this is a tool for those that want to make a module and that are reading the relevant docs page already - which may be one of these:

for example in the `Introductory` section we have:
https://docs.raku.org/language/modules                     <=== here be the gold
https://docs.raku.org/language/module-packages
https://docs.raku.org/language/modules-extra

and then in the `Reference` section we have:
https://docs.raku.org/language/module-packages

don't know if you watch raku-beginner - I am just taking a whiney person at face value - and yet agreeing that the minimal variant of `use` should be set out somewhere

actually calling subs from another file is a bit odd - but I tried to be nice and not push back on that since I know raku is very lightweight for this kind of thing

librasteve:
@lizmat, I see that I have been unable to convince you that this PR is a standalone minimal example for beginners who care about splitting code over multiple files --- my comment about is to try and separate this PR from the wider wiork that needs doing for package/module docs in general and I believe that referencing how to build a module (with Mi^ or any other way) doesn't belong here

I hope that you will come to see that adding in references to "how to build a module" pages reduces the simplicity, brevity and purity of raku in "minimal multi file mode" and urge you to allow this merge forward as written since - or at least on the basis that the perfect is the enemy of the good

or, feel free to make changes if you prefer

either way, please comment here and I will close this PR either way

doc/Language/101-basics.rakudoc
```diff
@@ -461,6 +461,34 @@ say "we have @flavors.sort.join(', ')";
...
+# myscript.raku
+use MyClass;
+ 
+my $instance = C.new(x=>42);
+say $instance.x;
+=end code
+
+The C<-I.> directive tells the raku interpreter to look in directory C<.>.
+
+=begin code
```
coke, line 488: :lang<shell> so we don't try to compile it

doc/Language/101-basics.rakudoc
```diff
@@ -461,6 +461,34 @@ say "we have @flavors.sort.join(', ')";
                                   # OUTPUT: «we have peach, vanilla␤»
 =end code
 
+=head1 Calling a raku method from another file
+
+To call a raku subroutine from another file, the convention is to have a minimal
+raku module or class file.
```
coke, line 467: To address @lizmat's concern, maybe we can note this is intended to be a simplistic example and point to how to write a module for a more... realistic? larger? example?

PR #4505, opened 2024-07-30
Update operators.rakudoc

finanalyst:
## The problem
Example in code will cause an error in RakuAST, redeclaration of 'my $a'

## Solution provided
remove second 'my'

### DIFF
--- a/doc/Language/operators.rakudoc
+++ b/doc/Language/operators.rakudoc
@@ -316,7 +316,7 @@ work:
 
     # And an example of a custom operator:
     sub infix:<space-concat> ($a, $b) { $a ~ " " ~ $b };
-    my $a = 'word1';
+    $a = 'word1';
     $a space-concat= 'word2';                 # OUTPUT: «'word1 word2'␤»
 
 One thing the simple and compound assignment operators have in common is that


finanalyst:
@coke bringing this to your attention. The new highlighter @lizmat is developing will require error free examples. 

The problem here is simple to fix and is unlikely to be the only one in the documentation suite

Each code section (between `=begin code` and `=end code` or in an indented section without blanks lines) will need to compile correctly in order to highlight

coke:
Notes:

1. Only partial code is provided in many cases. If the example is meant to compile, a preamble is preprended to the code that is NOT intended to be visible to the author. Until now this preamble has only been used for the 
2. There are several examples where code is shown as a deliberate bad example where it is not intended to compile.
3. Unless there's a tag on the code block, we compile it during 'make xtest' and this snippet should fail as is if it doesn't compile. (items from #2 are excluded from this compilation)
4. This code compiles successfully and passes xt/examples_compilation.rakutest before your change. (So I think @lizmat is correct that this is a difference in rakuAST processing, not that the code is invalid.) (It emits "potential difficulties" but runs sucessfully)

coke (approved):
See my comments on the PR

coke:
> @finanalyst If the highlighting results in an execution error, I suggest we skip highlighting for that particular piece of code, altogether for now. And that we collect these cases, as issues rakudo/rakudo marked RakuAST specific?

Could we add this to the error-report page?

PR #4510, opened 2024-08-11
Update ipc.rakudoc Demonstrating how to pipe data into a process using `:in`

finanalyst:
## The problem
`proc` (run/shell) is covered in a few places, but the way arbitrary data is piped into spawned process is very cryptic.

## Solution provided
An example is provided that illustrates how to employ a terminal utility that is useful for web page design.


coke:
Made it pass 'make xtest' in the branch. Looks good otherwise.

PR #4512, opened 2024-08-20
Update List.rakudoc

0rir:
Under "routine keys":

1) Text shows result as a Range.

2) Example shows `say` output as the creating  literal for a Pair.


coke (approved):

coke:
Thanks!

PR #4513, opened 2024-09-04
subtest Pair can use any order

JJ:
## The problem

The original text says that Pairs need to be in a certain order. That's not really true; tested the alternative order and it works anyway.

## Solution provided

Changed example to reflect that too.

### DIFF
--- a/doc/Type/Test.rakudoc
+++ b/doc/Type/Test.rakudoc
@@ -648,7 +648,7 @@ subtest {
 =end code
 
 You can also place the description as the first positional argument, or use a
-L<C<Pair>|/type/Pair> with description as the key and subtest's code as the value. This can be
+L<C<Pair>|/type/Pair> with description and code in any order. This can be
 useful for subtests with large bodies.
 
 =begin code :preamble<use Test;>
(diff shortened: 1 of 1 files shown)


coke:
Thanks for the catch! My one concern is that roast doesn't seem to test this variant of subtest. Part of resolving this PR this should include at least a ticket in roast with the omission. Rakudo seems to have added this in https://github.com/rakudo/rakudo/commit/e239f6ee3ee18e464995ee0c7b0c02dbdaeb9c75, back in 2016, so I think we're safe in assuming it's intentional.

coke (commented):
Can we show both variants of Pair examples?

JJ:
> Thanks for the catch! My one concern is that roast doesn't seem to test this variant of subtest. Part of resolving this PR this should include at least a ticket in roast with the omission. Rakudo seems to have added this in [rakudo/rakudo@e239f6e](https://github.com/rakudo/rakudo/commit/e239f6ee3ee18e464995ee0c7b0c02dbdaeb9c75), back in 2016, so I think we're safe in assuming it's intentional.

Well, `Test` is not raku so testing should be made in rakudo/rakudo, right?. Which begs the question, is `Test` tested? I haven't found any subdirectory that seems to do so in the `t` directory.

JJ:
> Can we show both variants of Pair examples?

Sure.

coke:
Committed a version that shows both. Thanks.

PR #4519, opened 2024-10-02
Fix for issue 4518, bad example

tbrowder:
## The problem

Bad example: 

    isa-ok $resources, Hash;

## Solution provided

Correct error by showing the correct class name:

    isa-ok $resources, Distribution::Resources;


coke:
Thanks!

PR #4521, opened 2024-10-08
Update Any.rakudoc

0rir:
Deleting 'obviously' because, if the sentence was obvious, it could be deleted entirely.


coke:
Thanks!

PR #4528, opened 2024-10-24
NativeCall void function returns Mu type object

0rir:
Doc' that NativeCall's `returns` trait's default is `Mu:U`.

Emphasize `void`-is-only-for-pointers with its own paragraph.

Ref #5293

### DIFF
--- a/doc/Language/nativecall.rakudoc
+++ b/doc/Language/nativecall.rakudoc
@@ -91,8 +91,10 @@ an example.
 Here, we have declared that the function takes two 32-bit integers and returns a
 32-bit integer. You can find the other types that you may pass in the
 L<native types|/language/nativetypes> page. Note that the lack of a C<returns>
-trait is used to indicate C<void> return type. Do I<not> use the C<void> type
-anywhere except in the C<Pointer> parameterization.
+trait may be used to indicate I<C> C<void> return; the actual return is a C<Mu>
+type object.
+
+Do I<not> use the C<void> type anywhere except in a C<Pointer> parameterization.
 
 For strings, there is an additional C<encoded> trait to give some extra hints on
 how to do the marshaling.


raiph:
@0rir

> Doc' that NativeCall's returns trait's default is `Mu:U`.

A probably inconsequential comment by me about a minor comment by you but that confused me.

But when I looked at the doc patch it said `Mu`, which is what I'd expect, is equivalent to `Mu:_`, not `Mu:U`, and is the same as the default for ordinary routines:
```
sub foo {}
say &foo.of =:= Mu:_; # True
say Mu =:= Mu:_; # True
say Mu =:= Mu:_; # True
say Mu =:= Mu:U; # False
```

0rir:
@raiph, thank you.  My phrasing was meant to convey what the return value is, not the constraints it must meet.
 
NativeCall chokes on the `Mu:D` and `Mu:U` type-ishes.  But by the definition of `defined`ness  the return is `:U`.

Should this be on the language/nativetypes page also or instead?

Should this be tagged as an Rakudo-ism?

raiph:
@0rir

> My phrasing was meant to convey what the return value is, not the constraints it must meet.

Yes, but the return _value_ is, both conceptually, from both native and Raku perspectives, and actually, from both native and Raku perspectives, either a native value or `Mu` aka `Mu:_`, _not `Mu:U`_.

> NativeCall chokes on the `Mu:D` and `Mu:U` type-ishes.

Yes. It should, because neither of them makes any sense at all in the context of a native value.

> But by the definition of ~~definedness~~ `.DEFINITE` the return is `:U`.

No it isn't. The return is either a native value or, conceptually, a native void, and the appropriate type constraint for that is `Mu` aka `Mu:_`, not `Mu:U`.

The type object `Mu` implicitly (and the `Mu:_` type object explicitly) abstracts not only from whether a value is a type object or instance, but whether it's even an object in the first place. And in this case we're talking about a native value or the concept of a void. So `Mu` is perfect. `Mu:U` is entirely wrong.

----

That all said, you don't mention `Mu:U` in the actual doc patch, so while the mention of `Mu:U` in your patch's _description_ is clearly a mistake, it's arguably an entirely inconsequential mistake, and now I've got it out of my system I'm not gonna comment on that aspect further. :)

> Should this be on the language/nativetypes page also or instead?

Your actual patch seems to me to be about right as it stands. But that's a mild feeling.

> Should this be tagged as an Rakudo-ism?

Oh no. I definitely don't think so.

To me this aspect is elegant Raku design. @Larry carefully designed the type system so that `Mu` (and its alias `Mu:_`) 100% correctly, cleanly and simply cover both object and native values, and, in the latter case, both actual values and void. Nice!

0rir:
There is a good case for the doc being more rigorous in its usage of  `defined` versus `DEFINITE`;  I am not sure that would be a net positive.

It is a flaw in the implementation or design that an `is native` sub's sig  cannot bear a `-->Mu:U` which is why it is not in the PR.

Smileys are a fine refinement in @TimToady's design.

```
use NativeCall;
sub perror( Str $s) is native { … }
sub raku( Str $s -->Mu:U) { perror( $s) }
raku 'farewell';    # farewell: Success
```

raiph:
@0rir and all other readers.

I apologize if my way of thinking, feeling, and communicating is causing problems for anyone.

As I write this I am feeling scared that, far from connecting with you, @0rir, as a friend, as I intended in the exchange above, what I have said has instead deeply hurt you.

To put it mildly, I'm hoping that's my mind playing tricks on me. But I do feel I must say something. If I have hurt you, I'm deeply sorry, ask for your forgiveness, hope I can find some way to not feel like I currently feel, and hope you can too.

That said, in the hope my mind is indeed just playing tricks on me, and that you are OK, and we can still talk about technical matters as friends, I will say a couple final things before I really must pause writing about technical matters.

First, while NativeCall is of course currently considered Rakudo implementation specific, and I have no view about whether or not Rakudo's support for an `is native` trait should be complemented by supporting any particular return type constraint and/or value that it currently rejects, I do have a view about Raku (as against Rakudo) and consider that a separate matter.

Second, I agree with @0rir that a doc change to try more rigorously distinguish between defined and definite may backfire. My own best shot is probably the following, which is a slight edit of part of an SO answer I wrote a year or so ago:

# Definite vs Defined

The similarity between definiteness and definedness is deliberate. If a given object has not got a `.defined` method (and almost none do) it will inherit `Mu`'s. And that `.defined` method just returns `self.DEFINITE`.

But while the `DEFINITE` "method" (actually a compile time macro) cannot be overridden¹, the `.defined` method can be:

* Users may override `.defined` for their classes/objects. This makes it easy for someone writing a foreign language adaptor to allow Raku interop with another PL which has a different notion of "definedness" than Raku's. (This is pretty much the only use case. It is generally ill-advised for ordinary Raku code to override `.defined`.)

* The built in `Failure` class overrides `.defined`. It returns `False` for both the type object (as normal) and instances. Thus:
```
say "{.DEFINITE.gist}\t{.defined.gist}\t{.gist}"
  for (Mu, Mu.new, Int, 42, Failure, Failure.new)
```
displays:
```
False False   (Mu)
True  True    Mu.new
False False   (Int)
True  True    42
False False   (Failure)
True  False   (HANDLED) Failed
```
The last line, which corresponds to `Failure.new`, has `True` in the first column (its `.DEFINITE` is `True`) but `False` in the second column (its `.defined` is `False`).

Footnotes

¹ Definiteness is fundamental to the static typing foundation of Raku's gradual type system and its notion that an "object" is something that can be either a type object or an object instance. To allow the compiler to rely on this simple either/or boolean at compile time, the behavior of "definiteness", and thus the `.DEFINITE` macro, cannot be overridden by user code -- unlike the `.defined` method.

0rir:
@raiph, be relieved for I am not pained by our discussion.  It is just a matter of disagreeing.
It is usual that I am not as quick to reply as most here; this time I  regret my delay for the extention of you unease.

I agree with the info in your "shot".  Depending on your target audience, I see varied improvements possible.  If you are interested, let me know.

PR #4535, opened 2024-11-23
Small fixes in JS comparison

linus:
## The problem

There are a few typos / small issues in the Raku <-> JS comparison:

- `logDupe` does not take any args, and calling it with one (particularly `2`) may be confusing.
- The dice roll will never sum to 16 (and 16 is also not boxcars) since each die only rolls up to 6

## Solution provided

- Remove redundant, possibly confusing, argument to `logDupe` function
- Change "Boxcars!" condition to `diceRoll === 12` and `$dice-roll == 12`, respectively

### DIFF
--- a/doc/Language/js-nutshell.rakudoc
+++ b/doc/Language/js-nutshell.rakudoc
@@ -122,7 +122,7 @@ function logDupe() {
     console.log(foo);
 }
 
-logDupe(2);       // OUTPUT: 2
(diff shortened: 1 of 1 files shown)


dontlaugh:
Appreciated :+1:

dontlaugh:
deployed to prod https://docs.raku.org/language/js-nutshell

PR #4537, opened 2024-11-27
Replace Pod6 with Rakudoc

librasteve:
## The problem

References to Pod6 in this document are outdated since the new name is Rakudoc

## Solution provided

Replaced all occurences of `Pod6` with `Rakudoc`


coke (approved):

coke:
Thanks!

PR #4542, opened 2024-12-13
Update --man code example in pod.rakudoc

1-1sam:
## The problem
The `--man` code example provided in the Rendering Pod to Text section has some issues:
* It will only render Pod in the file being executed by Raku. This is
problematic if you intend to distribute the program via Zef, as `$*PROGRAM` will
not point to the original script but a wrapper without the Pod instead.
* Performance issues from invoking Raku.

## Solution provided

* Print the output of `pod2text $=pod` using `Pod::To::Text` directly. This way
we do not have to rely on `$*PROGRAM` pointing to our script file. It's also
faster than re-invoking Raku.

### DIFF
--- a/doc/Language/pod.rakudoc
+++ b/doc/Language/pod.rakudoc
@@ -862,8 +862,10 @@ traditional Unix command line "--man" option to your program with a
 multi MAIN subroutine like this:
 
 =begin code
+use Pod::To::Text;
+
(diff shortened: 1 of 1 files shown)


coke:
Thanks for the submission will review shortly

coke (approved):
Looks good. Testing, I see we want to add a `!` to the `$man` parameter

PR #4544, opened 2024-12-22
Update math.rakudoc

doomvox:
Just moving the Arithmetic section up top above Sets so that a newcomer will see that first.


finanalyst:
This new order seems more reasonable for a non-mathematician

jubilatious1:
Agree with this change.

coke (approved):

coke:
Thanks!

PR #4546, opened 2024-12-25
fix reference to BIND-* methods

timo:
The text claimed that BIND-KEY is used to bind to an array element, but it is actually BIND-POS. Add a new sentence that still mentions BIND-KEY and what it is actually used for, since it's closely related.

### DIFF
--- a/doc/Language/containers.rakudoc
+++ b/doc/Language/containers.rakudoc
@@ -239,8 +239,10 @@ lexpad at runtime.
 
 The answer is that binding to array elements is recognized at the syntax
 level and instead of emitting code for a normal binding operation, a special
-method (called L<C<BIND-KEY>|/routine/BIND-KEY>) is called on the
-array. This method handles binding to array elements.
+method (called L<C<BIND-POS>|/routine/BIND-POS>) is called on the
+array. This method handles binding to array elements. There is also an
+equivalent method for binding to hash elements as well (called
+L<C<BIND-KEY>|/routine/BIND-KEY>)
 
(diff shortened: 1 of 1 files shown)


lizmat (approved):

coke:
Thanks!

doc/Language/containers.rakudoc
```diff
@@ -239,8 +239,10 @@ lexpad at runtime.
 
 The answer is that binding to array elements is recognized at the syntax
 level and instead of emitting code for a normal binding operation, a special
-method (called L<C<BIND-KEY>|/routine/BIND-KEY>) is called on the
-array. This method handles binding to array elements.
+method (called L<C<BIND-POS>|/routine/BIND-POS>) is called on the
+array. This method handles binding to array elements. There is also an
```
raiph, line 243: I tripped over this sentence a bit:

> This method handles binding to array elements. There is also an equivalent method for binding to hash elements as well (called L<C<BIND-KEY>|/routine/BIND-KEY>)

I can explain why if you want me to (just ask) but for now I'll simply suggest what would avoid me tripping up):

> (C<BIND-POS>handles binding to array elements. There is also a method for binding to I<hash> elements called L<C<BIND-KEY>|/routine/BIND-KEY>.)

raiph: Sigh. The suggestion I wrote began with "`C<BIND-POS>` handles..." and ended with a proper RakuDoc link but of course GH gobbles such text. I'm tempted to delete my review given that it was just a very small thing but please just guess what I meant or feel free to ignore my comment.

Issue #4540, opened 2024-12-06
bad design link


coke:
https://docs.raku.org/language/regexes#Quoted_lists_are_LTM_matches

links to

https://design.raku.org/S05.html#Longest-token_matching

but that's just a github repo at this point - not a 404 so it doesn't fail.

Ideally, remove the link and describe the strategy internally to the docs.

finanalyst:
@coke I think that the Docs used to contain a lot of links to Sxx documents, and they have gradually been replaced. Is one of the utilities for testing the documents able to pick out any other `/ 'L<' .+? 'S' \d+ .+? '.html' /` occurrences?

coke:
The problem is, the link *works*, it just goes to a github repo front page now instead of the original deep link, so the test we have doesn't complain.

Anything in

```
rak 'https://design.raku.org/' doc
```

needs fixing to point to the new structure.

finanalyst:
@coke 
A list is given at the bottom of the files. 
I looked at the examples above,
`https://design.raku.org/S05.html#Longest-token_matching`
could be matched to 
`https://github.com/Raku/old-design-docs/blob/master/S05-regex.pod#longest-token-matching`

May be the following algorithm would work:
1. collect the full names in the new structure by S numbers into a hash `%S-files`, so 'S05' => `https://github.com/Raku/old-design-docs/blob/master/S05-regex.pod`
2. Run a regex like the following
```
m/ 'L<' .+? '|' \s* 'https://design.raku.org/' ('S'\d+) '.html' (.*?) $ / 
```
3. any '#' part needs to be transformed so the title is lower case and all '_' are replaced with '-'
4. substitute back into the rakudoc files

@coke let me know and I can try to do this. I could manage a single PR with all the changes to all the files, but I'm not sure I could programmatically create one PR for each file changed.

The following files were reported by `rak 'https://design.raku.org/' doc --files-with-matches`. 
```
doc/Language/pod.rakudoc
doc/Language/glossary.rakudoc
doc/Language/haskell-to-p6.rakudoc
doc/Language/regexes.rakudoc
doc/Language/5to6-nutshell.rakudoc
doc/Language/modules.rakudoc
doc/Language/5to6-perlop.rakudoc
doc/Language/performance.rakudoc
doc/Language/mop.rakudoc
doc/Language/syntax.rakudoc
doc/Language/rb-nutshell.rakudoc
doc/Language/list.rakudoc
doc/Language/5to6-perlsyn.rakudoc
[...]

PR #4548, opened 2024-12-29
see issue #4540 remaining files

finanalyst:
[gist of report with link checks](https://gist.github.com/finanalyst/c676034b4f09b0f7621b9d4ee4848955#file-report-changes-with-internal-targets-md) [gist of internal links that do not work](https://gist.github.com/finanalyst/8323598b46113a0c429bdf7cf0b41457#file-failing-internal-links-md)

### DIFF
--- a/doc/Language/5to6-nutshell.rakudoc
+++ b/doc/Language/5to6-nutshell.rakudoc
@@ -1430,7 +1430,7 @@ Removed.
 =item C<-P> C<-u> C<-U> C<-W> C<-X>
 
 Removed. See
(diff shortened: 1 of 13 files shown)


finanalyst:
@coke there are four internal links that do not work here. The links in the pod go to the correct page, but not to a heading. The problem is escaping of non-ascii chars, or a link to a specific line.

These could be addressed individually I think

coke:
Can you tag them in the ticket? Otherwise good to merge

coke:
Can you tag them in the ticket? Otherwise good to merge

coke (approved):

PR #4553, opened 2025-01-18
Update unicode_ascii.rakudoc

finanalyst:
## The problem
Regularise X<> in headings
Placing the term whitespace outside the `X<>` has an effect on the rendering in the title. It seems to be the only example of this form in a heading in all the Raku/docs/doc repo. Tested with `rak '/^\=head\d\s+X\<\| /' raku-docs` (**raku-docs** is a local clone of Raku/docs)

## Solution provided
Placing the term whitespace inside the `X<>`

### DIFF
--- a/doc/Language/unicode_ascii.rakudoc
+++ b/doc/Language/unicode_ascii.rakudoc
@@ -42,7 +42,7 @@ aren't decimal digit characters, so can't be combined.) For example:
   my $var = ⅒ + 2 + Ⅻ; # here ⅒ is No and Rat and Ⅻ is Nl and Int
   say $var;              # OUTPUT: «14.1␤»
 
-=head1 X<|Language,Whitespace> Whitespace characters
+=head1 X<Whitespace characters|Language,Whitespace>
 
 Besides spaces and tabs, you can use any other unicode whitespace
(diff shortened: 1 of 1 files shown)


finanalyst:
@dontlaugh @coke I'm finding these subtle outliers as I'm refactoring the generation of secondaries, search, routines page, types page and the index - they were all jumbled together in an ad hoc fashion.

coke:
Should this be a style we check for and disallow in a test, or will it eventually cause a compilation error?

finanalyst:
Yes I think this form (an empty ` X<>` together with chars outside the
`X<>`) in a header! should be disallowed. The reason is that it doesn't
achieve what is intended. IMHO However both are correct forms and so will
always compile.

dontlaugh:
@finanalyst will you add that style check to this PR? then we can approve this

finanalyst:
@dontlaugh  Sorry I missed your feedback.
Not sure how you want the style check to be added.
I looked through all the tests under `xt/` and did not find anything I could add to easily.

dontlaugh:
@finanalyst That's okay. We don't need to hold back this PR.

dontlaugh (approved):

PR #4557, opened 2025-02-20
Documented deprecated operators, as discussed on #raku

wayland:
## The problem

Deprecated operators appeared with no explanation.  

## Solution provided

Noted that they were deprecated, with a version.


lizmat:
Thanks!

Issue #4467, opened 2024-05-21
Tone of 5-6 articles


coke:
These definitely feel like they were written by someone outside the 6 community and we pulled them in without doing a strong edit pass on them.

TODOs, "not fully docced", "this is not specifically documented".

Better to have the minimum descriptive text or comment out sections that we don't feel offer any info.

jonathanstowe:
Yep, they were originally made by David Adler in 2015/2016 (the original material is at https://github.com/dha/perl5-to-perl6-docs,)

coke:
bd19ab277 (simple pass)

PR #4559, opened 2025-02-21
Split 5to6 section into 3 paragraphs

0rir:
Split the _Language/5to6-perlop/The_arrow_operator_ section into 3 paragraphs.

Changed the `fat arrow` redirect to the more local 5to6-perlop#Comma_operator.  And put it first.

Reduced the text of the central point into the second paragraph.  Making a change so the first sentence need not be corrected by the second.           

The third pargraph briefly states the `->`'s Raku purpose with links.

My "dereferencing is rare in Raku" seems justified in my slight experience and by JJ's tone.  If not so, advise.

### DIFF
--- a/doc/Language/5to6-perlop.rakudoc
+++ b/doc/Language/5to6-perlop.rakudoc
@@ -47,13 +47,13 @@ my $list   = (1,);     # a List in a scalar container
 For the Perl C«=>» arrow, see the 
 L<Comma operator|/language/5to6-perlop#Comma_operator>.
 
-The dereferencing operator, C«->» is replaced by the dot, which is
+The dereferencing operator, C«->», is replaced by the dot, which is
 also used for method calls.  So Perl's C«$arrayref->[7]» becomes
 Raku's C<$arrayref.[7]> and C«$user->name» becomes C<$user.name>.
-Dereferencing is less common in Raku so most dots represent method calls.
+Note that dereferencing is rare in Raku.
 
 Raku uses the C<->> to attach L<Signatures|/language/Signature> to
-to Blocks, see L<Blocks|/language/Block>.
+L<Blocks|/language/Block>.
 
 =head2 Auto-increment and auto-decrement
 
--- a/doc/Language/5to6-perlop.rakudoc
+++ b/doc/Language/5to6-perlop.rakudoc
@@ -44,13 +44,16 @@ my $list   = (1,);     # a List in a scalar container
 
 =head2 The arrow operator
 
-As you typically will not be using references in Raku, the arrow is
-probably less useful as a dereferencing operator. If you do need to
-dereference something, however, the arrow is the dot. It is also the dot
-for method calls. So, Perl's C«$arrayref->[7]» becomes
-C<$arrayref.[7]> in Raku and, similarly, C«$user->name» becomes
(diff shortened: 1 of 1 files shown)


doc/Language/5to6-perlop.rakudoc
```diff
@@ -44,13 +44,16 @@ my $list   = (1,);     # a List in a scalar container
...
-C<$arrayref.[7]> in Raku and, similarly, C«$user->name» becomes
-C<$user.name>. The C«=>» arrow is used for constructing L<C<Pair>|/type/Pair>s, see
-L<Pair term documentation|/language/terms#Pair>.
+For the Perl C«=>» arrow, see the 
+L<Comma operator|/language/5to6-perlop#Comma_operator>.
+
+The dereferencing operator, C«->», is replaced by the dot, which is
+also used for method calls.  So Perl's C«$arrayref->[7]» becomes
+Raku's C<$arrayref.[7]> and C«$user->name» becomes C<$user.name>.
+Note that dereferencing is rare in Raku.
```
lizmat, line 53: Perhaps add: and that `$arrayref[7]` is also valid syntax for positional access (losing the dot).

0rir: I do not understand this well.  A few times, IIRC, I have needed a dot to get an array index to work.  I am fairly sure  that it was in `map` 's  Callable parameter.  So there is something like indirection that comes up rarely.  JJ seems to be making that distinction.

If that distinction is real, your addition would be better placed with 5to6's guidance about Arrays.

I'll try checking the dot-reference-op's sources, with modest hope of success.

lizmat:
`.[42]` is short for `$_[42]`, perhaps that's where you got that notion from?  The dot is needed, as `[42]` is syntax for an array with a single element: `42`.

0rir:
@lizmat, I don't think so; I loved the vanishing topic at first non-sight.   A named var or a Whatever,  or finding a dot instead of a trait's double parens.

Maybe @JJ can say why he mentioned dereferencing.

JJ:
It would help if I had a bit of context. Are you talking about an issue, about what's actually written there which I wrote (found by blame)? I'll try to go for both, but will need some time figuring out, I'm afraid.

JJ:
OK, I have a bit of context here now.  My 2¢ (taking into account that the only thing I did with this piece of documentation is changing Perl(non-breaking-whitespace)6 to raku
1. I don't think people are coming to Raku from the Perl community, so my first impulse for these pages would be just to consider them technical debt and eliminate them.
2. Fixing something that has not been broken (i. e. does not have an issue) is always a difficult proposition, if not for anything else because it's difficult to evaluate what's right or wrong. That applies also to documentation. And there's effectively an issue with these pages: #4467 which mainly talks about the "tone". That might not be in scope, but as long as you're changing some piece of code affected by that issue, you should try and help with that.
3. Talking about dereferencing should maybe receive some context, such as [a link to this FAQ](https://docs.raku.org/language/faq#What's_up_with_array_references_and_automatic_dereferencing%3F_Do_I_need_the_@_sigil%3F) to the tune of "talking about taking references [and de-referencing]) doesn't make much sense. We're talking about arrows, however, so maybe it would be more useful to just say what an arrow means in Raku if the (rare) Perl user bumps into it.

PR #4562, opened 2025-03-03
Update terms.rakudoc

0rir:
Statement missing a verb.


doc/Language/terms.rakudoc
```diff
@@ -1,4 +1,4 @@
-=begin pod :kind("Language") :subkind("Language") :category("reference")
+/=begin pod :kind("Language") :subkind("Language") :category("reference")
```
patrickbkr, line 1: I believe this is unintentional.

patrickbkr (changes requested):

PR #4567, opened 2025-04-09
Fix missing index entries

donaldh:
Zero-width index directives don't seem to work if they appear after a heading, so fix all occurrences by moving them before the heading.

The broken index directives also had the side-effect of causing blank headings in some situations, e.g.

 - https://docs.raku.org/language/variables#Sigils
 - https://docs.raku.org/language/operators#infix_%2E%2E%2E

### DIFF
--- a/doc/Language/modules-core.rakudoc
+++ b/doc/Language/modules-core.rakudoc
@@ -9,8 +9,8 @@ note that the implementation is not part of the Raku specification and thus
 might be subject to change without prior notice. The following is a list of
 them, along with links to their source code.
 
-=head2 C<CompUnit::*> modules and roles
 X<|Modules,CompUnit (Rakudo classes)>
+=head2 C<CompUnit::*> modules and roles
 
 These modules are mostly used by distribution build tools, and are not intended
 to be used (at least until version 6.c) by the final user.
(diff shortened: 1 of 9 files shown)


finanalyst:
@coke @donaldh tl;dr These are solved problems with new renderer. But the consequences of these changes are minimal. I am happy for them to be merged.

## Detail

The problem noted arises because `=head1 text\nmore text` is all part of the contents for `=head1`. 

So, in each case here, the `X<>` are part of the preceeding `=head1` line.

The processing of `=head` contents with embedded markup, such as `X<>` and `C<>` is quite complex, and stretches the original RakuDoc specification. Older renderers do not properly render embedded content.

The new Renderer (see below) handles these corner cases by rendering the contents of a header.

There are two workarounds (for the present):
1. move the `X<>` lines to before the `=head`, and so remove it from the `=head` contents
2. insert a blank line between `=head` and embedded `X<>`, and ditto

There are three consequences either way. 
1. Since the `X<>` is now not a part of a heading, it will be rendered as a paragraph. However, each of the cases in this commit have the form `X<|...>` so there will be no text visible to the reader.
2. These `X<>` embeddings also generate Composite/Secondary files. So it is possible (though I think very unlikely) someone has inserted a link to a composite generated by such a `X<>` markup in the header.
3. The link to the heading itself uses some of the meta data in the `X<|META DATA>`. This means that if the heading is linked to somewhere, it will not hit this section of text. Instead there will be a soft fail to the top of the webpage.

All of the above are sorted in the `Rakuast::RakuDoc::Render` and the new version of the website does not suffer the problem in the issue (I sincerely hope. I used the Sigils example to resolve the issue).

I prefer for there to be fewer Composites. The problems of bad links is very small.

PR #4570, opened 2025-05-02
Modules rewrite

wayland:
## The problem

The module/package documentation was suboptimally organised, as per https://github.com/Raku/doc/issues/3714

## Solution provided

Organised documentation less suboptimally.  

Things to note:
* Put the first subdirectories under "Tutorials", specifically:
  * Using Modules: How to use raku.land and zef to get modules to use in your projects
  * Making Modules: How to code modules in your projects
  * Making distributions: How to wrap your code up in a distribution
* Each of those has a few pages in it.  I'm not sure whether this was a good idea or not.  Ideas I like are:
  * More subdirectories (but upon what grounds?)
  * Series of tutorials, rather than everything in doc/Language being standalone
* Tweaked the Makefile (for ease of use and readability) and one of the tests (because otherwise it was hanging on that test; the outcome of the test should be unaffected though)

I'm expecting feedback on things I need to change.  

Hope this helps.


coke (changes requested):
Is there a curl option to hide the output instead?

coke (changes requested):
RAKULIB should be . (current directory's META6.json), not lib

finanalyst:
This seems to be a good start. I absolutely agree the modules part needed restructuring.
We should also include something about documentation.

From the site building point of view, it does not matter which directory the files are placed, so long as the meta date at the top of the file (in the line with `=begin rakudoc`) matches existing categories. The pages (introduction, etc) that pull in the separate pages use the meta data.

Richard

wayland:
@finanalyst : 
* Not sure what you mean by "We should also include something about documentation" unless you mean that we should have a page on documenting your distribution, in which case, I agree, but...
* Nearly everything here is a copy-paste from somewhere else in the documentation (with the exception of some bits about $?DISTRIBUTION and %?RESOURCES).  For context, my main role on Wikipedia seems to be to add headings to pages that don't have them, and that's functionally what I'm doing here: taking existing information and grouping it better.  
  * My hope is that, with the better grouping, and people being able to find things more easily, we'll also get more contributions, such as the one you suggested about documentation.  I could put in a crosslink to the Rakudoc page, though

finanalyst:
@wayland Ok understood that the main work currently is about regrouping, and that's definitely important.

I do mean a page on documenting a module. I agree more needs writing - its a TODO of mine.

wayland:
Hi!  I've made the following updates:
* Added the --silent option to curl, which should at least reduce the amount of output
* Changed RAKULIB, as requested
* Added the following targets to the Makefile
  * test-mine: Tests with just the appropriate git files
  * xtest-mine: Like test-mine, but with xtest
  * prep: Runs zef install

HTH,

doc/Type/Distribution/Resource.rakudoc
```diff
@@ -118,17 +118,19 @@ converting it to the canonical library name corresponding to the value, in
...
-A C<Distribution::Resource> is designed to be used directly as the resource
+A C<Distribution::Resource> was designed to be used directly as the resource
 it represents, such as a file, for instance
 
 =for code
 my @data = %?RESOURCES<data/swim.csv>.lines.split(",");
 
-However, this is I<not> because it returns an L<C<IO::Path>|/type/IO::Path>, but because it
-shares many method with it: C<Str, gist, raku, absolute, is-absolute,
+However, this was found to be problematic due to the fact that the path can
```
ugexe, line 127: I'm not sure it makes sense to talk about what why things are implemented the way they are in the docs e.g. "was found to be problematic". Someone looking at the docs presumably wants to know how to use something and not necessarily the history or implementation details.

doc/Type/Distribution/Resource.rakudoc
```diff
@@ -118,17 +118,19 @@ converting it to the canonical library name corresponding to the value, in
...
+get compiled into the library at compile time, which means that it can end
+up pointing to the wrong place at runtime.  Thus, many of the methods that
+were shared with L<C<IO::Path>|/type/IO::Path> are now deprecated; this
+includes C<Str, gist, raku, absolute, is-absolute,
```
ugexe, line 131: If `Str` or `gist` is giving a deprecation message then that is a bug

doc/Type/Distribution/Resource.rakudoc
```diff
@@ -118,17 +118,19 @@ converting it to the canonical library name corresponding to the value, in
...
 relative, is-relative, parts, volume, dirname, basename, extension, open,
-resolve, slurp, lines, comb, split, words, copy>; above we use C<.lines>, for
-instance.
+resolve, slurp, lines, comb, split, words, copy>.
```
ugexe, line 133: This list is not accurate. You can see the deprecations explicitly defined in [the source](https://github.com/rakudo/rakudo/blob/b527cefcb29789120f32d5e4003913e53c47476f/src/core.c/Distribution/Resource.rakumod#L50-L104). Notably slurp, lines, and other methods that don't return a string representation of a path or path part.

doc/Type/Distribution/Resource.rakudoc
wayland: I was looking at https://github.com/rakudo/rakudo/blob/b527cefcb29789120f32d5e4003913e53c47476f/src/core.c/Distribution/Resource.rakumod#L50 when compiling this list; was that incorrect?

doc/Type/Distribution/Resource.rakudoc
wayland: OK, I'll be rewriting this section in line with your recommendations.  

Just my notes on the old version of the list as to what's still current and what isn't.  I'm mainly posting it here for feedback on .gist and .raku (given that .Str is deprecated).  

* Deprecated: Str, absolute, is-absolute, resolve
* Not sure: gist, raku         
* Still good: slurp, lines, comb, split, words, copy

doc/Type/Distribution/Resource.rakudoc
ugexe: I meant .raku or .gist aren’t deprecated

doc/Type/Distribution/Resource.rakudoc
wayland: OK, I've rewritten this section so that it:
* Keeps most of the old material (but updates the deprecation list
* Adds a paragraph on what *not* to do (which was what I was trying to achieve), but without listing the deprecated methods

Would very much appreciate your feedback, since I seem to get this wrong on the regular.  

Thanks!

doc/Type/Distribution/Resource.rakudoc
wayland: Oh!  Right.  Well, I've completely left them out this time around because

1. I wasn't sure of their status
2. They're not specific to mirroring IO::Path, they're on just about everything, so I figured they didn't need a mention.  

HTH,

wayland:
Also regularised use of subkind across the new pages (to "Modules").

doc/Type/Distribution/Resource.rakudoc
jubilatious1: `Str` is deprecated?

doc/Language/distributions/introduction.rakudoc
```diff
@@ -0,0 +1,75 @@
+=begin pod :kind("Language") :subkind("Modules") :category("tutorial")
+
+=TITLE Distributions: An Introduction
```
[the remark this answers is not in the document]
wayland: I didn't see any CI tests, but I made sure all the tests passed on my local.

Makefile
```diff
@@ -1,20 +1,34 @@
-.PHONY: test xtest push help
+.PHONY: test xtest push help test-mine xtest-mine
```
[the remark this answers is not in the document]
wayland: Done.  It was nice to have a way to only test the files I'd changed, though (saved time on having to run the full test suite repeatedly before committing).

xt/01-raku-version.rakutest
```diff
@@ -16,7 +16,10 @@ require a recent enough version for RakuAST classes.
 plan 1;
 
 {
-    my $web = run('curl', '-L', '-H', "Accept: application/vnd.github+json", '-H', "X-GitHub-Api-Version: 2022-11-28", "https://api.github.com/repos/rakudo/rakudo/releases?per_page=1", :out, :err);
+    my $web = run(
+        'curl', '--silent', '-L', '-H', "Accept: application/vnd.github+json", '-H', "X-GitHub-Api-Version: 2022-11-28",
+            "https://api.github.com/repos/rakudo/rakudo/releases?per_page=1",
+        :out);
```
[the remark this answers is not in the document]
wayland: Thanks!

wayland:
Hi @coke 

How now?  

:)

doc/Language/distributions/testing.rakudoc
```diff
@@ -0,0 +1,110 @@
...
+because then the compiler will inspect the C<META6.json> file in the root
+directory of the B<distribution>.
+
+If you are just beginning to develop a module (or series of modules), and you
+have not decided what to call or where to place the code (perhaps splitting it
+into several C<*.rakumod> files), but assuming all the module files are under
+the C<lib/> directory, then it is possible to use
+
+=for code :lang<shell>
+prove6 -Ilib
```
coke: I would consider removing this whole section, or perhaps replace with just one sentence explanation that explains why the previous is better.

doc/Language/distributions/uploading.rakudoc
```diff
@@ -0,0 +1,80 @@
...
+=SUBTITLE How to upload your distribution to the ecosystem
+
+If you've written a Raku module and would like to share it with the
+community, we'd be delighted to have it listed in the
+L<Raku modules directory|https://raku.land>. C<:)>
+
+There have been three ways to distribute a module, but the CPAN and P6C ways
+have been deprecated; please use the zef/fez method instead.  However, no
+matter the method, all modules will be indexed on the L<raku.land|https://raku.land> website.
+
```
coke: Let’s not lead with history, but with the current standard; leave the history as the last (small) entry on the page.

doc/Language/distributions/uploading.rakudoc
```diff
@@ -0,0 +1,80 @@
...
+More details on preparing your module can be found on the pages linked from
+L<Distributions: An Introduction>.
+
+=head2 Upload your module to zef ecosystem
+
+If you want to use the I<zef> ecosystem then you need to register your username
+using I<fez>.
+
+=item Install fez if you haven't done so already
```
coke: ```C<fez>```

doc/Language/using-modules/code.rakudoc
```diff
@@ -0,0 +1,450 @@
...
+use MyModule 'foo';
+say foo.new(); # OUTPUT: «MyModule::Class.new␤»
+
+always();      # OK   - is imported
+shy();         # FAIL - «shy used at line 8. Did you mean 'say'?»
+=end code
+
+You can combine C<EXPORT> with type captures for interesting effect. This
+example creates a C<?> postfix which will only work on L<C<Cool>|/type/Cool>s;
+please note that, by using C<$_> as an argument, we don't need to use a
```
coke: Do we need to introduce the topic topic here?

coke (changes requested):
partial review, didn’t go through everything.

doc/Language/distributions/testing.rakudoc
wayland: This is pre-existing (just moving stuff around).

doc/Language/using-modules/code.rakudoc
wayland: This is pre-existing (just moving stuff around).

wayland:
@coke , I'm afraid I can't any more (because I don't remember).  At this point, I'm just running "git diff master", and, if it appears twice, then I didn't write it.

doc/Language/distributions/uploading.rakudoc
wayland: Fixed!

PR #4571, opened 2025-05-03
Added some more examples of the differences in the map function

wayland:
## The problem

The map function didn't work exactly the same, as promised.

## Solution provided

Changed the promise (by adding more explanation of differences).


coke (changes requested):
Looks good - only note, one of these snippets ends in a ;, the other does not.

Can you make them consistent (looks like the doc has an example just above without)

dontlaugh (approved):

Issue #4576, opened 2025-05-20
docs for methods mm-dd-yyyy and dd-mm-yyyy both inappropriately mention ISO8601


arkiuat:
https://docs.raku.org/type/Dateish#method_mm-dd-yyyy repeats the parenthetical link to (the wikipedia article on ) ISO8601 that appears in the doc for method yyyy-mm-dd

Same is true of the doc for Dateish.dd-mm-yyyy. 

These two mentions are misleading because only .yyyy-mm-dd conforms to ISO8601, so the link should only appear in the doc for .yyyy-mm-dd, not the other two.

arkiuat:
Resolved in https://github.com/Raku/doc/pull/4579#issuecomment-3368598908

PR #4579, opened 2025-05-21
Methods 'mm-dd-yyyy' and 'dd-mm-yyyy' inappropriately refer to ISO 8601

tbrowder:
ISO 8601 details YYYY-MM-DD format, but the same link also discusses other Gregorian time formats. To correct the existing description, this PR keeps the same link (where the formats are described), but changes the link description as appropriate for each method's format as shown on that same page,

- Addresses doc issue #4576

### DIFF
--- a/doc/Type/Dateish.rakudoc
+++ b/doc/Type/Dateish.rakudoc
@@ -183,7 +183,7 @@ say Date.today.yyyy-mm-dd("/");          # OUTPUT: «2020/03/14␤»
 =for code :ok-test<dd>
 method mm-dd-yyyy(str $sep = "-" --> Str:D)
 
-Returns the date in C<MM-DD-YYYY> format (L<Gregorian, month-day-year (MDY)|https://en.wikipedia.org/wiki/ISO_8601>).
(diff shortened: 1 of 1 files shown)


arkiuat:
>but the same link also discusses other Gregorian time formats

Where are you seeing this discussion?

The link is to https://en.wikipedia.org/wiki/ISO_8601, and I went looking, but couldn't find any such discussion.  

mm-dd-yyyy and dd-mm-yyyy aren't even mentioned as output formats, much less discussed. All I could find were one or two mentions (not discussions) as *input* formats, but these two methods are both accessors, not constructors.

The wikipedia article only mentioned them as things to be converted to yyyy-mm-dd, so this doesn't provide any support for the idea that it's not misleading to mention ISO8601 in the context of these two (unlike yyyy-mm-dd).

tbrowder:
> Where are you seeing this discussion?

The link has been corrected.

arkiuat (changes requested):
There are two unmerged commits on this pull request, one of which is correct, but this one is incorrect.

Can we please nix this commit (7c01d32), so that the correct commit (99fa8ba) can be merged and we can close [issue #4576](https://github.com/Raku/doc/issues/4576)?

coke:
Applied manually in 07c904c9a

PR #4584, opened 2025-05-29
Add "andthen" and "orelse" to Promise.rakudoc

voldenet:
This pull requests adds documentation for `Promise.andthen` and `Promise.orelse` methods, since they're really useful.

### DIFF
--- a/doc/Type/Promise.rakudoc
+++ b/doc/Type/Promise.rakudoc
@@ -246,8 +246,7 @@ to the C<&code>.
         say "Got $res";
         $res + 10
     });
-    await $after;
-    say $after.result;  # OUTPUT: «Got 42␤52␤»
+    say await $after;    # OUTPUT: «Got 42␤52␤»
 
     # Broken promise skips the code
     my $broken = Promise.broken(X::AdHoc.new(payload => "Error"));
@@ -270,14 +269,12 @@ to the C<&code>.
         my $err = $orig.cause;  # Get exception from original promise
         "Using cached data because of {$err.message}"
     });
-    await $recovered;
-    say $recovered.result;  # OUTPUT: «Using cached data because of DB Error␤»
+    say await $recovered;  # OUTPUT: «Using cached data because of DB Error␤»
 
     # Kept promise skips the code
     my $kept = Promise.kept(42);
     my $unchanged = $kept.orelse(-> $orig { "recovered" });
-    await $unchanged;
-    say $unchanged.result;  # OUTPUT: «42␤»
+    say await $unchanged;  # OUTPUT: «42␤»
 
 =head2 method keep
 


voldenet:
I'm not sure if these methods should be documented in the current form - they could get changed, so they get result/cause instead of the promise, which is not needed for anything.

librasteve:
maybe wise to also add two tests to ROAST - then the behaviour is official?

FCO:
> Another thing that jumped to mind is that the proposed change would presumably make andthen/orelse blocking (synchronous) methods unlike the current non-blocking / asynchronous behavior.

The block passed to andthen is already only called after the promise is kept, so it's always immediate. Right?

voldenet:
>we are only supposed to document roasted behavior

That makes sense, these docs will have to wait until this is specified.

The rest of the comment is about the issue https://github.com/rakudo/rakudo/issues/5892

>.oO ( Did you forget a .new? )

I was trying to not write long blocks of code in the title, but I promise that the description has proper code.

>Another thing that jumped to mind is that the proposed change would presumably make andthen/orelse blocking (synchronous) methods unlike the current non-blocking / asynchronous behavior.

The continuation is always executed after the Promise is resolved, so the behavior of those two methods would not significantly change. The biggest change is that the original Promise would not be accessible anymore, however already documented `then` method covers advanced use cases. My initial need was to have `.then` and `.catch` known from promises in javascript to allow building chain of callables in a simplified way, so upon finding undocumented (but working) methods, I thought it'd be good idea to describe them.

>(Actually, the initial thought was my first thought as I read the title of this issue before reading any of its content. The title -- Add "andthen" ... to Promise.rakudoc -- juxtaposes "andthen" and "Promise", which instantly sets things up for confusion.

I have used the already existing methods, so I didn't think much about how they're named, but I agree - these names could be more unique.

voldenet:
@raiph
To clarify - I didn't mean that `andthen` should block. The `&code` parameter inside it should get different parameters.
See the original issue about it: https://github.com/rakudo/rakudo/issues/5892

raiph:
@voldenet 

Ahhhh. A penny has dropped. So, in this sentence:

> I'm not sure if these methods should be documented in the current form - they could get changed, so they get result/cause

the _first_ "they" refers to "these methods", whereas the _second_ "they" refers to the block (not a method, or at least probably not) that's passed to, er, them (the two methods `andthen` / `orelse`). Right?

(I had tried to carefully draw that distinction in my comments, precisely because I didn't want to presume I knew what you meant, but it's tough to be precise in natural language, especially English!)

I'm now confident I understand what you meant, and how I got confused, and that I can just apologize for the noise, and it's time for me (and I suggest you and @FCO) to hide our comments in this thread about the tangent I started.

But I think I'll wait for confirmation in the form of you (and @FCO) reading this, and responding by hiding some/all of your comments about this tangent as "Off-topic". I plan to mark my comments likewise presuming I see that. Thank you for your patience!

coke:
Can someone please summarize the current state of this PR?

arkiuat:
> Can someone please summarize the current state of this PR?

My first impression on reading it over is that the behavior is underspecified and that changes in the documentation are on hold until tests are added to roast, but I don't see any linked or open issues or PRs on roast related to this. 

It seems that adding such tests is also on hold, pending the outcome of the discussion in issue https://github.com/rakudo/rakudo/issues/5892

PR #4601, opened 2025-06-07
[Custom_operators_and_operator_overloading] Mention circumfix

habere-et-dispertire:
Under [Custom_operators_and_operator_overloading](https://docs.raku.org/language/js-nutshell#Custom_operators_and_operator_overloading)

> Operators can be defined as prefix, infix, or postfix.

Operators can be defined as prefix, infix, postfix or _circumfix_.


coke:
Thanks!

PR #4606, opened 2025-07-02
[ link bitrot ] prove6

habere-et-dispertire:
I think this is the same repo:

https://raku.land/cpan:LEONT/App::Prove6
=>
https://raku.land/zef:leont/App::Prove6


coke:
Thanks, applied.

Issue #4324, opened 2023-07-12
proc::async output error


coke:
https://docs.raku.org/type/Proc/Async

shows output of
> Proc finished. Exit code: 0

But the code above has
>  say ‘Proc finished: exitcode=’, .exitcode, ‘ signal=’, .signal;

PR #4610, opened 2025-07-13
Fix incorrect Proc::Async output #4324

horberlan:
## The problem
Referring to issue #4324, the output of `class Proc::Async` is incorrect in the template sample.  
## Solution Provided  
Adjust the output for [Proc::Async](https://docs.raku.org/type/Proc/Async).


coke:
Thanks!

Issue #4602, opened 2025-06-07
remove reference to long gone opaquepointer


coke:
It's taking up space in the FAQ

PR #4612, opened 2025-07-13
Remove deprecated OpaquePointer reference #4602

horberlan:
## Solution provided
Eliminate the outdated reference to OpaquePointer in issue #4602.


coke:
Thanks!

PR #4614, opened 2025-07-18
Update variables.rakudoc

arkiuat:
Distribution::Resources s/b singular, not plural

## The problem

There's a 404 deadlink [Distribution::Resources](https://docs.raku.org/type/Distribution/Resources) which I believe can be fixed by removing the s on the end of Resources.

## Solution provided

s/Distribution::Resources/Distribution::Resource/ 

in the link, the linktext, and for two other occurrences of this module name


coke:
thanks!

PR #4622, opened 2025-09-26
Use the TV/Netflix analogy on Supplies also on the dedicated page /type/Supply. (Plus one other small proposal)

schultzdavid:
File `/type/Supply`: This topic was new to me when I read these docs. I struggled to understand the explanations on the page `/type/Supply` [(weblink)](https://docs.raku.org/language/concurrency#Supplies) right up until the point when I read the very good TV/Netflix analogy on `/language/concurrency`. Thus, here's a proposal to incorporate that analogy also in the dedicated page for Supplies.

File `/language/control`: Since strings have a `.chars` method rather than a `.char` method, that name looks more fitting to me here, unless I'm missing something.

### DIFF
--- a/doc/Type/Supply.rakudoc
+++ b/doc/Type/Supply.rakudoc
@@ -15,7 +15,13 @@ It is a thread-safe implementation of the
 L<Observer Pattern|https://en.wikipedia.org/wiki/Observer_pattern>,
 and central to supporting reactive programming in Raku.
 
-There are two types of Supplies: C<live> and C<on demand>. When tapping into a
+There are two types of Supplies: C<live> and C<on demand>. A C<live>
+supply is like a TV broadcast: those who tune in don't get previously emitted
+values. An C<on-demand> broadcast is like Netflix: everyone who starts streaming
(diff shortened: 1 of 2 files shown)


doc/Type/Supply.rakudoc
```diff
@@ -15,7 +15,13 @@ It is a thread-safe implementation of the
 L<Observer Pattern|https://en.wikipedia.org/wiki/Observer_pattern>,
 and central to supporting reactive programming in Raku.
 
-There are two types of Supplies: C<live> and C<on demand>. When tapping into a
+There are two types of Supplies: C<live> and C<on demand>. A C<live>
+supply is like a TV broadcast: those who tune in don't get previously emitted
+values. An C<on-demand> broadcast is like Netflix: everyone who starts streaming
```
coke, line 20: I would prefer to not use Netflix specifically - would you be OK with using something generic like “A video streaming service”?

coke (changes requested):

coke:
Thanks, applied!

PR #4623, opened 2025-09-27
Update Str.rakudoc (.lines & .words return Seq, not List)

arkiuat:
Change "list" to L<C<Seq>|/type/Seq> where this is relevant (.lines, .words)

## The problem

Old description of these methods as returning a "list" is now misleading.

## Solution provided

Changed "list" to  L<C<Seq>|/type/Seq> in descriptions of methods .lines, .words


coke:
I'd feel better if we had a pointer to a spec test for this, but approve of updating it to match current rakudo.

coke:
Thanks, applied!

PR #4626, opened 2025-09-28
Update section `method wrap` in Routine.rakudoc: Link to Re-dispatching

schultzdavid:
Since `callsame`, `callwith` etc. are mentioned in the text, link to where they are defined, namely to https://docs.raku.org/language/functions#Re-dispatching


coke:
Thanks, applied!

PR #4627, opened 2025-09-30
Update faq.rakudoc -- misleading link

arkiuat:
text mentions set difference operator, but links to subtraction operator

## The problem

specifically, text discusses 

https://docs.raku.org/language/operators#infix_(-),_infix_%E2%88%96 

but links instead to

https://docs.raku.org/language/operators#infix_-

## Solution provided

changed to the link to the intended doc


coke:
Good catch, thank you

PR #4635, opened 2025-10-04
In section `Inheritance_and_multis` of `typesystem.rakudoc`, clarify wording and code example

schultzdavid:
[This is the section that's affected.](https://docs.raku.org/language/typesystem#Inheritance_and_multis)

I found it difficult to understand what is meant by the original phrase _"does not compete with"_. The replacement with _"there is no competition from"_ should be clearer and easier to parse.

In the code, change class names `A` and `B` to `Parent` and `Child`, and add comment that `int` is a subclass of `Int` (which is important to know in order to understand what goes on).

### DIFF
--- a/doc/Language/typesystem.rakudoc
+++ b/doc/Language/typesystem.rakudoc
@@ -208,20 +208,20 @@ first argument.
 
 =head4 Inheritance and multis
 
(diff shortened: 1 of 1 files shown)


coke:
Thanks!

PR #4638, opened 2025-10-04
In section `trait is nodal` of typesystem.rakudoc, add link to definition of term "hyperoperator"

schultzdavid:
Link to definition of Hyper operators, because this is the first occurence of this term in this file.


coke:
Thanks

PR #4643, opened 2025-10-06
In signatures.rakudoc, remark that "invocant marker" gets explained in the following section

schultzdavid:
This is the first time the novel term "invocant marker" is used in this document, so let readers know that they'll find it explained immediately hereafter.


librasteve:
I have merged this change - happy to do the additional of for `*%`

Maybe we could combine as: 


And they can also contain the invocant marker (explained in the next section) and the hash slurpee `*%` (to match any named arguments):

PR #4644, opened 2025-10-06
move sleep routines from Date to independent-routines

arkiuat:
## The problem

the three sleep routines were documented under Date, which didn't make sense as a location

## Solution provided

moved the doc for these routines from Date to independent-routines


coke (approved):

coke:
Looks good, thanks!

PR #4645, opened 2025-10-07
In signatures.rakudoc, remove mention of nonexistent trait `is optional`, and change `=:=` to `===`

schultzdavid:
[This is the section that's affected.](https://docs.raku.org/language/signatures#Was_an_argument_passed_for_a_parameter?)

Trying to use the trait "is optional" in a function signature throws error "Can't use unknown trait" (in all of v6.c, v6.d and v6.e.PREVIEW).

Perhaps some mention should be retained, though, if that trait was present in older versions.

### DIFF
--- a/doc/Language/signatures.rakudoc
+++ b/doc/Language/signatures.rakudoc
@@ -1002,14 +1002,15 @@ Table showing checks of whether an argument was passed for a given parameter:
 
 A parameter with a default is always I<optional>, so there is no need to mark it with a C<?>.
 
-Then you can use the C<=:=> L<container identity operator|https://docs.raku.org/language/operators#infix_=:=> in the body of the routine to check
-whether this exact default was bound to the parameter.
+Then you can use the C<===> L<value identity operator|https://docs.raku.org/language/operators#infix_===> in the body of the routine to check
+whether this exact default object was bound to the parameter. These examples use names like C<PositionalAt> to reflect that the C<.WHICH> test 
+invoked by C<===> returns an object of type L<C<ObjAt>|/type/ObjAt>.
 
 Example with a positional parameter:
 =begin code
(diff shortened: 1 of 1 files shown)


schultzdavid:
I've added to this PR further changes on the text in the same section.

The change of `=:=` to `===` is necessary in the second code example, because `=:=` would just always produce False there.

From the explanation above the first example, `===` also seems like the better choice there (even though `=:=` also works as intended). 

Namely, it is `===` which calls the `.WHICH` method. I've added this as a remark to the text, and have rearranged the text slightly.

doc/Language/signatures.rakudoc
```diff
@@ -1000,16 +1000,17 @@ Table showing checks of whether an argument was passed for a given parameter:
 
 ¹ A suitable default is an Object that has a distinct identity, as may be checked by the L<C<WHICH>|https://docs.raku.org/type/Mu#method_WHICH> method.
 
-A parameter with a default is always I<optional>, so there is no need to mark it with a C<?> or the C<is optional> trait.
+A parameter with a default is always I<optional>, so there is no need to mark it with a C<?>.
```
librasteve, line 1003: Yikes, can't believe I wrote that! You are correct.

doc/Language/signatures.rakudoc
```diff
@@ -1035,9 +1036,6 @@ c (Int);        #Undefined
 c 42;           #Defined
 =end code
```
librasteve, line 1037: I think that you are correct to use the `===` value identity operator (obviously I was having a bad day when I wrote this!)

As I re-read this section with fresh eyes, I realise that we should be clearer that this test is True if no argument is passed (ie you identity matching the default in the absence of a passed argument).

coke:
Thanks!

PR #4647, opened 2025-10-08
add method .multi-invocant to class Parameter

arkiuat:
## The problem

new method .multi-invocant (used in roast) not yet documented

## Solution provided

add to class Parameter


schultzdavid:
For the record, [S06-signature/multi-invocant.t](https://github.com/rakudo/rakudo/issues/3611) is the (only?) relevant roast file.

PR #4668, opened 2025-10-16
regularize Mu's routine/sub/method headings

arkiuat:
## The problem

inconsistent use of designations "routine", "sub", & "method" in documenting routines of class Mu, as discussed in issue https://github.com/Raku/doc/issues/4560 

## Solution provided

regularized the most obvious inconsistencies, and also merged sub item (from class Any) with Mu's method item


arkiuat:
changed title to limit scope of this PR to Mu so it doesn't stay in draft forever. I'll make a second PR for class Any after I'm done with this one

arkiuat:
after starting on class Any, I found there doesn't seem to be a consistent convention as to whether to list the method form or the sub form first. I chose to list method first, but the three examples in Mu had all listed sub first, so the last commit changes that

PR #4678, opened 2025-10-23
Extensions and fix in glossary.rakudoc. Some fixes and rewording in signatures.rakudoc.

schultzdavid:


jubilatious1 (commented):
Thank you!

PR #4687, opened 2025-10-28
make examples for `is dynamic` look parallel

schultzdavid:
In particular, put both inside a block.

This makes it clear that the difference is not caused, e.g., by a 'level of nesting' relative to where the subroutine is defined.

### DIFF
--- a/doc/Language/variables.rakudoc
+++ b/doc/Language/variables.rakudoc
@@ -1525,8 +1525,8 @@ sub get-file-content {
 
 =head2 Dynamic variables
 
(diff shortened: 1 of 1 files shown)


doc/Type/Variable.rakudoc
schultzdavid: Note that this change is actually necessary for the code to run as described. (Otherwise, the value of the first `$*x` is printed twice, since the second introspect() call will not bother with `$x` but find the `$*x` defined above.)

PR #4688, opened 2025-10-28
Some work on create-cli.rakudoc

schultzdavid:


doc/Language/create-cli.rakudoc
schultzdavid: In line 107 I've also changed the example from `@all.map: ` to `for @all`. This because, this example is already introducing a new concept (slurpy parameters). Given that this is an introductory document, the functional form with `.map: -> $name { ... }` would probably be too much at a time for learners who mainly want to learn about CLI creation.

coke (changes requested):
The change in flag constants is not helping clarify anything, please revert that one.

schultzdavid:
Yes that's true, I've now restored the original values.

Issue #2354, opened 2018-10-04
Improve the "Python from Perl 6" page


JJ:
## Introduction

You might need to have a quick introduction to the language. [This is as short as it gets](https://docs.perl6.org/language/101-basics). If you have any question about Perl 6, it's probably better to either [take it to StackOverflow](https://stackoverflow.com/questions/tagged/perl6) or ask [in the IRC channel](https://webchat.freenode.net/?channels=#perl6), which you can access from the web.

The language used for the documentation is called [Perl6 pod](https://docs.perl6.org/language/pod). You can pretty much follow the conventions that you already see in the document, but that guide above might help you find whatever is not there.

## The problem

We are talking about [this page, "Python to Perl 6 - nutshell"](https://docs.perl6.org/language/py-nutshell), which you can edit directly on the GitHub page by clicking on the pencil right above the butterfly (a.k.a Camelia); saving it will generate a pull request, or via forking, editing locally and doing a pull request in the usual way. 

The page can be improved in many different ways, obviously not limited to the ones below.

- [ ] Revise page for possible Python errors or something that is not Pythonic enough
- [ ] Reorganize the page (there's a level 2 header "Basic syntax" and nothing else)
- [ ] Add pointers to relevant Perl 6 and Python resources, including tutorials.
- [ ] Include references to widely used libraries, like Cro being similar to Django and somesuch.
- [ ] Add examples, for `prompt` for instance. There's only a link to the definition.
- [ ] Expand with another level 2 section with syntax that is not basic.

PR #4698, opened 2025-10-31
Some work on py-nutshell.rakudoc

schultzdavid:
This will partly address issue #2354.

Note to self:

Add a rudimentary section on Raku's type system below the section on variables.
In this new section, include among other things the distinction between definite and indefinite objects.

In the section "Control flow", include (a little bit) of material on `given` and `//` – mostly to show what would be Raku equivalents of the common Python idiom `if x is not None`. Perhaps split the control flow section into the two subsections "Conditionals" (if, given, when...) and "Iteration" (for, while/until, loop...).

### DIFF
--- a/doc/Language/py-nutshell.rakudoc
+++ b/doc/Language/py-nutshell.rakudoc
@@ -125,7 +125,7 @@ initialized or declared and initialized at once.
     $foo = 12;     # initialize
     my $bar = 19;  # both at once
 
-Also, as you may have noticed, variables in Raku usually start with sigils –
+Also, as you may have noticed, variables in Raku usually start with sigils—
 symbols indicating the type of their container. Variables starting with a C<$>
 hold scalars. Variables starting with an C<@> hold arrays, and variables
 starting with a C<%> hold a hash (dict). Sigilless variables, declared with a


doc/Language/py-nutshell.rakudoc
```diff
@@ -123,7 +125,7 @@ initialized or declared and initialized at once.
     $foo = 12;     # initialize
     my $bar = 19;  # both at once
 
-Also, as you may have noticed, variables in Raku usually start with sigils --
+Also, as you may have noticed, variables in Raku usually start with sigils –
```
schultzdavid: here I replaced two dashes (rendered as two dashes) by an em dash

doc/Language/py-nutshell.rakudoc
```diff
@@ -398,20 +400,23 @@ I<Raku>
...
 
 Declaring a function (subroutine) with C<def> in Python is accomplished
 with C<sub> in Raku.
 
 =begin code :lang<python>
 def add(a, b):
     return a + b
+=end code
 
+=begin code
```
coke: Why split this code up?

PR #4712, opened 2025-11-14
In exceptions.rakudoc, remove a `CATCH` keyword. Plus some stylistic improvements.

schultzdavid:
Since the outer `CATCH` statement serves no function here, I think the example will be more transparent without that keyword (i.e. with only `{}` instead of `CATCH {}`).

I guess one could try to cook something up with two nested `CATCH`es that illustrates how the outer `CATCH` block does give back control to the encompassing (main) block. However, as the example is now, neither of the `CATCH` blocks even gets entered in the first place, because no Failure is thrown in the main block. So removing the `CATCH` word before the outer block is an improvement.

### DIFF
--- a/doc/Language/exceptions.rakudoc
+++ b/doc/Language/exceptions.rakudoc
@@ -388,7 +388,7 @@ sub bad-sub {
     say "Returned $return";
     CATCH {
         default {
-            say "Error ", .^name, ': ',.Str;
+            say "Error ", .^name, ': ', .Str;
             $return = '0';
             .resume;
         }


doc/Language/exceptions.rakudoc
```diff
@@ -255,7 +255,7 @@ Please consult the section on
 L<resuming of exceptions|/language/exceptions#Resuming_of_exceptions>
 for more information on this.
 
-A C<try>-block is a normal block and as such treats its last statement
+A C<try> block is a normal block and as such treats its last statement
```
coke: As long as we’re in here: I think “and as such, uses its last statement as its return value.” reads better - if you agree, please add that update.

Issue #4711, opened 2025-11-13
duckmap doco unclear and unwieldy


arkiuat:
## Problem or new feature

In https://docs.raku.org/type/Any#method_duckmap There is an unclear transition which makes the whole entry confusing. First, three examples are given, and the first one is discussed, then the second one, but in the paragraph discussing the second example (which begins "In the second case," and ends "In this case:") there is a confusing transition to a fourth and fifth example. 

It's not entirely clear whether the fourth and fifth examples are supposed to be of some other aspect, or whether they are further illustrations of what is shown in the second and third examples. This makes for a confusing read, and if the latter is intended, this is probably more examples than needed.

Also, the whole passage was written at a time when only the method form of duckmap was documented. Since it is now documented as a routine (with both method and sub invocations), the examples have become somewhat unwieldy.

## Suggestions

The entry should probably be entirely rewritten, choosing salient examples in each of the two (method & sub) invocation usages without needing to duplicate any one example in both usages.

Issue #4560, opened 2025-03-01
heading 'routine' definition 'method'


finanalyst:
Using search for 'isa', select in category 'Heading', takes you to [class Mu](https://docs.raku.org/type/Mu#routine_isa)
Note that the heading is 'routine', but the definition in code section is only for 'method'. Using `isa(Any)` generates *Undeclared routine*.

After thinking how it could be otherwise, this seems reasonable.

But documentation seems wrong: other headings in Mu are mixed 'method' and 'routine', so suggest changing 'routine' in heading to 'routine'.

Same is true for 'does'. 

I've not raised a PR for the changes because
1. Not sure there was some other documentation intention originally
2. Perhaps documentation could be added by giving example of `&isa` with invocant ?
3. How many more of these mismatches exist

I can make changes to isa and does after feedback about above.

coke:
As I understand, if there are only methods or sub, then we use that as the heading as well, and if there are both, we use 'routine'. I think that the signature test (which is mostly disabled at this point) would be the place to verify if we have one or both, and therefore if the heading is correct.

finanalyst:
@coke In the two examples I cited, there are no sub forms documented, and trying to use one generates an error.

I thought there was a way to turn a method to a sub using an explicit invocant (for option 2). I seem to have misremembered something. So ignore option 2.

raiph:
Interesting coincidence. I feel I should write a couple comments.

I was just reading https://github.com/Raku/doc/blob/main/writing-docs/INDEXING.md as part of an attempt to do as @tbrowder prompted [here](https://github.com/Raku/doc/issues/4561#issuecomment-2692711232).

One of several things that threw me was this sentence:

> Other than explicit creation, headers (`=head` elements of Pod) of certain format get an anchor automatically, say `=head routine sin` creates an index item categorized as `Subroutines` automatically.

I thought to myself that "If that's true, it's wrong and should be fixed. If it's false this 'how to contribute to the doc doc' should be fixed. But I don't want to get even more bogged down in complexity as I try to just understand what it is I'm supposed to do to add a doc link about `foo` to the index in the right way when someone searches for `bar` so I should go and do something else and see whether I still have the desire to do this task after a break."

A few seconds later I saw this issue listed and thought "Hmm. That _might_ be about the very same issue, or a related one, given its title." So I clicked. Now I'm here it really does seem to be related.

[...]

PR #4718, opened 2025-11-19
deepmap/duckmap/nodemap, issues #4560 & #4711

arkiuat:
## The problem

duckmap: unclear and confusing transition from discussion of the second example to discussion of the fourth; cluttered examples. Basically addressing docs issue #4711 with minimal changes.

deepmap & duckmap: sub form incorrectly listed as multi

nodemap: docs issue #4560 

## Solution provided

duckmap: added a mention of the third example to smooth the abrupt transition, and selected only one example of either method or sub call as most appropriate for each example. Also nixed a distracting and unhelpful metaphor.

Incidentally updated the nodemap entry, and changed "multi" to "sub" for both deepmap and duckmap, as their sub forms are not multi.

### DIFF
--- a/doc/Type/DateTime.rakudoc
+++ b/doc/Type/DateTime.rakudoc
@@ -57,14 +57,12 @@ multi method new(Instant:D $i,  :$timezone=0, :&formatter)
 multi method new(Numeric:D $posix,  :$timezone=0, :&formatter)
 multi method new(Str:D $format, :$timezone=0, :&formatter)
 
-Creates a new C<DateTime> object. One option for creating a new C<DateTime> object
+Creates a new C<DateTime> object. One option for creating a new DateTime object
 is from the components (year, month, day, hour, ...) separately. Another is to
 pass a L<C<Date>|/type/Date> object for the date component, and specify the time
 component-wise. Yet another is to obtain the time from an
 L<C<Instant>|/type/Instant>, and only supply the time zone and formatter. Or
-instead of an C<Instant> you can supply a L<C<Numeric>|/type/Numeric> as a Unix timestamp
-(but note that this last method provides no way to disambiguate leap seconds,
-unlike using C<Instant>).
+instead of an Instant you can supply a L<C<Numeric>|/type/Numeric> as a UNIX timestamp.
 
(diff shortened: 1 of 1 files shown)


doc/Type/Mu.rakudoc
```diff
@@ -497,8 +497,10 @@ L<react|/language/concurrency#react> block.
 
 =head2 routine take
 
-    sub    take(\item)
     method take()
+    multi  take()
```
raiph: `take()`, as a subroutine call without an argument, "doesn't make sense" (to quote the associated error message displayed by Rakudo).

A few obvious and entirely distinct reasons why you're proposing to document it this way jump to my mind:

1. You, or doccers more generally, have decided to write `multi foo()` to act as a stand in for something else (maybe as a stand in for a `proto foo (|)`)? That seems unlikely, but if that's the case, I'd be grateful if you linked me to some concluding part of that discussion.

2. You are, er, taking the signatures in the Rakudo implementation as the interface to be documented, and that's a mistake. In this case I can guess that it has a `multi take()` variant to grab the dispatch if there's no argument and display the error message I excerpted above. But there can be other reasons why Rakudo has declarations of constructs such as routines, or routine variants, that, at least in general (though there might be exceptions), are not only not part of the Raku language but should not be documented either. It seems unlikely to me that you don't know that, but if you don't, then, because it's you doing this work, I see it as a wonderful opportunity to begin just about the biggest breakthrough I think could ever happen to Raku. That's perhaps me being inappropriately hopeful, but I am eternally optimistic that Raku is currently cruising along the back straight of the final lap in the first race @Larry wanted us to compete in, namely to ship an essentially complete initial version of the PL design and implementation they began to dream of 25 years ago, and I'd be delighted to explain how I see that as being grounded in appropriately dealing with documenting `multi` signatures!

3. Something else? The above 2 reasons jumped to mind, but I do sometimes miss obvious things. Still, I'd love to know what I'm missing!

arkiuat:
And that last bit of noise was me working on what was supposed to be an entirely different PR, which I won't restart until after this one is closed (or until I figure out how to prevent new commits from automatically attaching to my current open PR).

PR #4722, opened 2025-11-20
Mention Lock::Async in language/concurrency

timo:
mention and link Lock::Async in the language/concurrency page. It is preferable to Lock, though other high-level concurrency bits are better.


coke:
thanks!

PR #4725, opened 2025-11-29
remove method conj from Numeric per issue 3782

arkiuat:
## The problem

method `.conj` is currently documented in Complex, Cool, and Numeric, but the last entry doesn't correspond to anything tested in roast, while the first two can at least be interpreted as corresponding with the two places in roast where `.conj` is tested.

## Solution provided

deleted the entry for this method in doc/Type/Numeric.rakudoc (sorry for the typo in the commit msg)


arkiuat:
Marking this as draft until the discussion in issue 3782 concludes.

Might also want to change the example code in `Cool`'s entry for `method conj` to show it used against a `Str`.

coke (approved):

PR #4737, opened 2025-12-19
Added section on "Avoiding Slangs"

wayland:
## The problem

Sometimes when you want to modify the language, there's an easier option than creating a slang.  These should be listed in one place.  

## Solution provided

I've added a new section to address this.


wayland:
@FCO Agreed on both accounts.  I've removed the bit you complained about, and added something about custom attributes.  How does it seem now?

PR #4742, opened 2025-12-29
using Failures: discussion and examples

arkiuat:
per issue 3978
- added a short discussion of using failures
- added examples provided in 3978 by @niner (after cleaning them up a bit)

### DIFF
--- a/doc/Language/exceptions.rakudoc
+++ b/doc/Language/exceptions.rakudoc
@@ -392,9 +392,8 @@ If the resulting L<C<Failure>|/type/Failure> object is returned into sink
 context (that is, not handled), it's the same as if you had called C<die> or
 C<throw>. If it is instead returned into Boolean context or checked for
 definedness, then it is considered "handled" and does not throw. If it is
-assigned to a variable, it can be handled later, and otherwise the throw is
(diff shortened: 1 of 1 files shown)


doc/Language/exceptions.rakudoc
```diff
@@ -380,6 +380,34 @@ C<throw> can be viewed as the method form of C<die>, just that in this
 particular case, the sub and method forms of the routine have different
 names.
 
+=head1 Delaying exceptions as failures
```
raiph: `fail`ing can be described as delaying exceptions. From that perspective the current title reads as "Delaying exceptions as delaying exceptions".

What about just "Delaying exceptions"? Then the reader will be predisposed to understand that `fail` and "failure" are the chosen concrete terms for the abstract notion of "delaying exceptions".

Then again, failures aren't really "delaying exceptions" in the sense that, if producing code produces them, the whole idea is for consuming code to handle them properly so they do _not_ throw -- and as such, they're not delayed exceptions, but instead just regular error handling like it all used to be in the old days.

Perhaps the title could be "Converting exceptions to error values"?

librasteve (approved):
looks good

PR #4743, opened 2025-12-30
infix ~~ does not topicalize per issue 4709

arkiuat:
3 changes to 2 documents to address docs issue [4709](https://github.com/Raku/doc/issues/4709)

1. infix ~~ does not topicalize: removed the confusing assertion that it does
2. also added two Unicode operators that had been missing from language/unicode_ascii, U+2A75 ⩵ & U+2A76 ⩶


raiph:
@arkiuat 

> infix `~~` does not topicalize: removed the confusing assertion that it does

Wow.

That's a huge change.

In 2019 I wrote an (accepted) answer to the SO question [Use of implicit parameter in for-loop causes hang](https://stackoverflow.com/questions/57462127/use-of-implicit-parameter-in-for-loop-causes-hang).

The code in that question still causes a hang in the glot.io Rakudo (2024.01).

So either Rakudo has been changed since then but no one changed the doc, or my analysis of the hang was incorrect, or, well, maybe Lizmat's code analysis was wrong and the doc change needs to be reverted or at least double checked?

Here's the code:
```
my $a = "AABBCCBGG";
say join "\n", do for $a.comb.squish {
  $a ~~ s/^ ($_+) //;
}
```

See the footnotes in my answer for more code that clarifies some details of, and rationale for, `~~` topicalizing.

Other than my thoughts expressed in those footnotes I'm not strongly arguing for or against it, but I am amazed that Liz didn't mention (or remember?) that it did topicalize for a decade -- at least I (and the doc) _thought_ it did!

arkiuat:
> maybe Lizmat's code analysis was wrong and the doc change needs to be reverted

It looks this way, because not only does the code presented as problematic in the SO question still hang, but JJ's illustrative example showing more detail still behaves exactly the same way.

PR #4744, opened 2025-12-30
slightly demystified `temp` & `let`

arkiuat:
## The problem

Current doc says that `temp` & `let` are like declarators, which was confusing, since they don't create variables.

This PR is intended to close docs issue [4318](https://github.com/Raku/doc/issues/4318)

## Solution provided

Relabeled `temp` & `let` neutrally as "keywords"

Also relocated the entry for `constant` to better reflect its current position in the summary table (which was a recent change motivated by the same issue, 4318).

### DIFF
--- a/doc/Language/variables.rakudoc
+++ b/doc/Language/variables.rakudoc
@@ -510,7 +510,7 @@ augment slang Regex {  # derive from $~Regex and then modify $~Regex
 See L<I<slangs>|/language/slangs> for more.
 
 X<|Syntax,supersede>
-=head1 Variable declarators, scope, and shadowing
+=head1 Variable declarators and scope
 
 Most of the time it's enough to create a new variable using the C<my>
 keyword:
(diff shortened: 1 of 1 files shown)


doc/Language/variables.rakudoc
```diff
@@ -533,12 +533,12 @@ beyond what L<Twigils|#Twigils> can do.
     augment       Adds definitions to an existing name
     supersede     Replaces definitions of an existing name
 
-There are also two prefixes that resemble declarators but act on
-predefined variables:
+There are also two keywords that act on predefined variables, allowing
+temporary shadowing of their current values:
```
raiph: As noted in https://github.com/Raku/doc/issues/4318#issuecomment-3700883561 I don't think "shadowing" is the right term.

arkiuat: I agree with you now, and I've removed all references to shadowing, but I am still 

* removing the phrase "quasi declarator" and 
* referring to these as keywords instead of as prefixes. 

And now (again or still) I think this is the last piece needed to close issue 4318.

PR #4745, opened 2025-12-30
clarifying entry for `require` per issue 4273

arkiuat:
## The problem

description of `require` is confusing per issue 4273

## Solution provided

Draft PR. Added some text to address the concerns in the issue, but the result still needs further clarification (and is probably a little redundant now).

### DIFF
--- a/doc/Language/using-modules/code.rakudoc
+++ b/doc/Language/using-modules/code.rakudoc
@@ -70,7 +70,10 @@ to restrict what you import.
 
 =head2 X<C<require>|Syntax,require>
 
-C<require> loads a compunit and imports definite symbols at runtime.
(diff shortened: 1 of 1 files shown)


arkiuat:
I added some wording based on @raiph's [comment](https://github.com/Raku/doc/issues/4273#issuecomment-3697112867) to address @lucs's specific concerns (which I entirely agree with!), but the result looks to me as if it still needs more work: it could be both shorter and a lot clearer. Suggestions welcome!

librasteve (approved):
looks good

librasteve:
@arkiuat - I have reviewed these but did not Merge as the PR is still marked Draft

Issue #4740, opened 2025-12-28
Add description of new test "exits-ok"


tbrowder:
Start with @lizmat's announcement of the new test in a recent Weekly and embellish and modify as necessary to fit the current docs style.

coke:
Closed with 96b772a5d

PR #4748, opened 2026-01-02
Document exits-ok

tbrowder:
## The problem

New Raku test 'exits-ok' is not yet documented.  #4740

## Solution provided

This PR provides documentation for the new test including a practical example of its use.

### DIFF
--- a/doc/Language/testing.rakudoc
+++ b/doc/Language/testing.rakudoc
@@ -229,11 +229,13 @@ subroutine:
 
 That can be tested like this:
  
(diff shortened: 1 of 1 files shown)


coke:
> On Fri, Jan 2, 2026 at 09:34 Will Coleda ***@***.***> wrote: Yes, lizmat said something like to name it for its current name but mark it deprecated when the new name comes out in next version. i’ll hack toward that end… 

My recommendation: document exits-ok as available in 2026.01 and ignore exit-ok entirely.

coke:
Why was this deleted?

PR #4749, opened 2026-01-02
reorg "Regex interpolation" section per 3705

arkiuat:
## The problem

Issue [3705](https://github.com/Raku/doc/issues/3705) pointed out several clarifications that are needed to the "Regex interpolation" section of /language/regexes

## Solution provided

- Added @-variable forms to $-variable forms in the main table
- Added Boolean condition checks to the main table
- Added subsection headings to parallelize discussion of all five entries
- Moved mention of hashes in regexes being reserved to immediately after the table
- Added links to this section (and to adverbs section) in the introductory part of the article
- Changed the ordering of some of the sections in the last half of the article

### DIFF
--- a/doc/Language/regexes.rakudoc
+++ b/doc/Language/regexes.rakudoc
@@ -25,9 +25,9 @@ i.e. a sublanguage or L<I<slang>|/language/slangs>. This page describes this lan
 regexes can be used to search for text patterns in strings in a process called
 I<pattern matching>.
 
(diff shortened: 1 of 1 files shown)


arkiuat:
I'm planning to move "Regex interpolation" to immediately following "Substitution".

I'm also thinking that the "Adverbs" section and the section-with-link-to "Best practices" should appear somewhat earlier in this article as well, and am thinking some more about the overall ordering of the last several sections

PR #4750, opened 2026-01-03
Add description of new Rakudo test 'exits-ok'

tbrowder:
## The problem

New Rakudo test 'exits-ok' is not yet documented.

## Solution provided

This PR is proposed to address Raku doc issue #4748.


tbrowder:
Maybe it's now (hopefully) just failing on the version test

PR #4761, opened 2026-02-02
Fix return type of `match` method

2colours:
It has been wrong ever since it was written down at https://github.com/Raku/doc/commit/a0cb9311e5e1d9c5a6b2cf66ee2d52d9a831b6ac


lizmat:
Thanks!

PR #4765, opened 2026-02-05
Update modules setup for zef | fez

coke:
Rescue old branch


coke:
Found going through old branches, but this is obsolete after the latest changes to this section.

Sorry!

PR #4771, opened 2026-02-07
Test that the glossary is sorted

coke:
… And make the test pass - it was definitely not sorted.


arkiuat (approved):
The re-sort looks good, definitely an improvement. I'm not yet familiar enough with the rakudoc rakuAST node structure to judge the new test script, but I did read through it.

PR #4772, opened 2026-02-07
right-hand side cleanup

### LINKED ISSUE #2116: Right-hand side of *what*?
See https://github.com/perl6/doc/commit/eb8f7f24803517db0539fe77a89fed2696b98966 and the comments on that commit. Mainly

> right-hand side (no second dash) is more natural English. But a problem here is "right-hand side < of what?>". Should be clarified as "use it as a right-hand[-]side operand" or such? (most other example the context is more clear already.

And here's an example from our docs:

> We can therefore use it as a right-hand side.

Right-hand side of what?

coke:
Update entries for LHS/RHS in the glossary, standardize on ‘left-hand side’ and ‘right-hand side’, add a test.

Closes #2116


arkiuat (approved):
Looks good to merge

PR #4776, opened 2026-02-11
check for various mistypings of 'tiebreak'

2colours:
As discussed on IRC.


coke:
Looks good, would be great if the updated test also passed on this branch (the CI will only test the files that changed, which I realize is not great when the file that changed was a test)

PR #4777, opened 2026-02-11
Precedence and associativity vs "evaluation order"

### LINKED ISSUE #4058: Fix explanation of evaluation order
The issue concers this page :
[https://docs.raku.org/language/operators](https://docs.raku.org/language/operators)

Here is pretty much the first sentence
```
The precedence and associativity of Raku operators determine the order of evaluation of operands in expressions.
```

Which is false, as demonstrated by this example : 
```
$ raku -e 'print(1) ** print(2) ** print(3); say ""'
WARNINGS for -e:
Useless use of "**" in expression "print(1) ** print(2) **" in sink context (line 1)
123
```
I made a github issure very recently about it, as I confounded operators precedence and associativity VS order of evaluation of operands. If there are no side effects, then it doesn't change anything and you wouldn't even know that the order is different, as a language user, but if there are side effects, then it becomes apparent that there is a distinction.

[https://stackoverflow.com/questions/71941798/why-does-print-isnt-affected-by-operator-precedence/71942002](https://stackoverflow.com/questions/71941798/why-does-print-isnt-affected-by-operator-precedence/71942002)

2colours:
The intent is to resolve #4058 .

This is my concrete proposal. Please read it carefully, I tend to write gibberish like "equivalent with X".

If it makes sense and is understandable, I would say it has the advantage over the previous version because _it is true_, after all. :P

### DIFF
--- a/doc/Language/operators.rakudoc
+++ b/doc/Language/operators.rakudoc
@@ -79,7 +79,7 @@ operator|#infix_=_(list_assignment)>.
 
 =head1 Operator associativity
 
-Associativity can be seen as a tiebreaker: where two operators with a same
+Associativity can be seen as a tiebreaker: where two operators with the same
 precedence level act on an operand, the associativity of the operators determines
 which operator the given operand belongs to. For instance, in the expression
 C<100 / 2 * 10>, the binary division operator C</> and the binary multiplication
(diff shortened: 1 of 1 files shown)


arkiuat (approved):
On the whole, this looks like a definite improvement to me.

PR #4785, opened 2026-02-21
Add foldcase lookahead example

librasteve:
From a recent thread on IRC/Discord https://discord.com/channels/538407879980482560/633753286209699870/1474676849329180682

## The problem

ignorecase has some issues with Unicode comparisons

## Solution provided

state the issue and suggest a more general solution with example using foldcase


coke:
```
    # Failed test 'X<Ignoremark|Adverbs,:ignoremark;Adverbs,:m> must not contain multiple indexes in one X<>'
    # at t/18-rakuast-validate.rakutest line 61
    # You failed 1 test of 738
```

I just went through a big PR to eliminate all the doubled ``X<>``s - this seems to be a reversion of the change I made on this header in my PR.

coke (changes requested):

FCO:
Could the link be using the irc logs?

Issue #3881, opened 2021-05-05
Make the role of "leap seconds" more obvious in DateTime and Instant


jonathanstowe:
There was a question on Facebook where the poster was surprised that for a range of given date times the `DateTime.Instant.Int` varied from the perl `DateTime->epoch` by an increasing amount.  

Having looked at the documentation pages for both `DateTime` and `Instant` there doesn't appear to be anywhere where it is made explicit that leap seconds are in play at all, apart from the discussion of "Future Leap Seconds" in the `Instant` page.

It would probably be useful to have some introductory text in both `DateTime` and `Instant` that describes the role of the "leap seconds" and possibly also in the methods such as `.posix` which return the number of seconds describe why, due to the leap seconds, they will differ from `.Instance.Int`.

Haven't got any particular suggestions as to what the words should be.

JJ:
Good idea.

arkiuat:
I'm also planning a few short additions to DateTime.rakudoc, but the largest chunk of new text is planned for Instant.rakudoc, to be inserted under `=head1 Leap Seconds` immediately before the current `=head1 Future Leap Seconds` (which will change to `=head2`.  Here is the first draft, which I think is too long:

https://gist.github.com/arkiuat/f4baa44430034ffc512fb4414242e483

Soliciting suggestions for where/what to cut.

arkiuat:
> My opinion would be that it is a fundamental misunderstanding of Instant
> for it to provide an Int method in the first place.  It's supposed to
> opaque outside of implementations.

This makes sense, but I'm not sure we can get that particular cat back into its particular bag. In the REPL, 

`.say for (my Instant $i).^methods.map(*.gist).elems`

prints 71, and the vast majority of these are methods that should never have been implemented by this standard, so it's not just `Int`.

coke:
Someone please open a problem solving issue with that note, thanks.

arkiuat:
@0rir, thank you for this excellent suggestion.  Three of your five points about the link's contents are specifically related to leap seconds, which only relate to UTC and not to any of the other time scales. Four really, because although unpredictability of future variance applies to UT1 as well, I think the interest here is focused on UTC.

So I'd like to add *two* entries to the glossary:

1. *time scale*, as you suggest, mentioning (and possibly linking to wikipedia for) UT1, TAI, maybe TT (fka ET), and POSIX time (which last is not really a timescale, strictly speaking).
[...]

PR #4795, opened 2026-03-08
2nd attempt to close issue 3881

arkiuat:
I've attempted to implement the changes that @librasteve suggested in [problem-solving issue 497](https://github.com/Raku/problem-solving/issues/497) with the modifications that I suggested and he okayed in [#3881 ](https://github.com/Raku/doc/issues/3881).

I've also updated the dates in the examples where they can be updated, and added a brief explanation at the very end as to why the "future leap seconds" examples have to be so very old (for now).

Finally I've tried to reign in some of my immoderate language a bit. I hope that this proves acceptable and that we can close the docs issue that spawned it.

### DIFF
--- a/doc/Type/Instant.rakudoc
+++ b/doc/Type/Instant.rakudoc
@@ -18,10 +18,11 @@ C<Instant> to a L<C<Duration>|/type/Duration> returns another Instant. Subtracti
 will yield a L<C<Duration>|/type/Duration>. Adding two C<Instant>s is explicitly disallowed. All
 other operations with Instants are undefined.
 
-I<NOTE:> POSIX time, a commonly used time standard, does not address leap seconds,
-while C<Instant> objects do, so there's some extra machinery in the conversion
-methods to handle this. See methods C<from-posix>, C<to-posix>, and the discussion
-of leap seconds below.
+I<NOTE:> L<POSIX time|https://en.wikipedia.org/wiki/Unix_time>, a commonly used
+L<time standard|https://en.wikipedia.org/wiki/Time_standard>,
+does not address leap seconds, while C<Instant> objects do, so there's some
+extra machinery in the conversion methods to handle this. See methods C<from-posix>,
+C<to-posix>, and the discussion of leap seconds below.
 
 =head1 Methods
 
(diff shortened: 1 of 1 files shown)


librasteve (approved):
This is great - all looks good to me.

Thanks for persevering with this against all odds.

doc/Type/Instant.rakudoc
```diff
@@ -39,16 +87,12 @@ does not change for two whole seconds. For example:
...
-is recorded by atomic clocks which disregard the Earth's rotation: when Unix
-time was first established it was based on UTC, which does track the rotation,
-and had by that time fallen 10 seconds behind atomic-clock time. The numerical
-value of C<Instant> happens to track atomic-clock time, and as of 2025 there
-have been 27 leap seconds recorded. This adds up to a total of 37 seconds,
-which is why, as of 2025, the following holds:
+Atomic clocks also disregard the Earth's rotation, so between 1958 when they
+were first calibrated and the POSIX epoch of 1970, UTC had fallen 10 seconds
+behind atomic-clock time. The numerical value of C<Instant> happens to track
+atomic-clock time, and as of 2026 there have been 27 leap seconds recorded.
```
raiph: > The numerical value of `Instant` happens to track atomic-clock time

The word "tracks" is... interesting. I may well have misunderstood the context, and thus whether that's a reasonable thing to write, but it sounds... potentially significantly misleading.

What do you intend the word "tracks" to imply, and is it reasonable to think the reader will correctly infer what they need to infer?

I posed a question to google's gemini about a mythical PL's `Instant` datatype, in respect of whether it would "track" atomic time, and a couple other details, without mentioning Raku. This was its answer:

> An "[Instant]" (like Java's `java.time.Instant` or similar in other languages) represents a specific point on a linear, UTC-based timeline. It is typically initialized from the system's Real-Time Clock (RTC), which, despite using crystal oscillators that can drift by seconds daily, is corrected by periodic Network Time Protocol (NTP) updates, generally keeping devices synchronized within milliseconds to seconds of atomic-clock time.
>
> Here is a breakdown of the accuracy and behaviors you described:
>
> * **What is an Instant?** It is an absolute moment, often recorded as seconds and nanoseconds since the Unix Epoch (1970-01-01T00:00:00Z). It ignores time zones and daylight saving time, representing UTC time.
>
> * **System Clock Drift:** You are correct that local hardware clocks (crystal-based) drift. Without synchronization, they can easily drift several seconds per day.
>
> * **NTP Synchronization:** NTP corrects this by communicating with time servers (often stratum 1 or 2, which are connected to atomic clocks) to adjust the local system time. **Unix/Linux:** Often uses `ntpd` or `chronyd`, which can check for updates every 17 minutes by default, but it’s adjustable. **Apple Devices:** Apple frequently uses their own `timed` daemon, which is designed to be aggressive with syncs to maintain high accuracy. **Windows:** The default Windows Time service (W32Time) sync interval is much less frequent, historically defaulting to once a week, though it can be adjusted.
>
> * **Precision and Accuracy:** While an `Instant` may provide nanosecond or microsecond precision in code, the **accuracy** relative to a real-world atomic clock is limited by network latency and the frequency of NTP adjustments.
>
> **Important Nuances:**
>
> * **Leap Seconds** The `Instant` type in many languages (especially Java/Unix time) traditionally excludes leap seconds, meaning it tracks a linear, continuous time rather than exact Coordinated Universal Time (UTC).
>
>* **Clock Jumps:** When NTP corrections are large, the system clock might jump backward or forward. High-accuracy systems use monotonic clocks (e.g., `CLOCK_MONOTONIC`) for measuring durations, not the "Instant" of the day.

To summarize why I'm writing this comment, I want to make sure folk are taking into account that, unless `Instant`s are some special thing that I don't currently think they are, a new one will typically be adrift from atomic clock time from the getgo, and two `Instant`s are likely to be adrift by significantly differing amounts from each other.

One upshot is that a `Duration` is much more approximate than one might naively expect, and, furthermore, may even be negative. This is why someone has created and uploaded a Raku library that provides monotonic clock features for durations that are much more accurate and guaranteed non-negative.

As usual, I may have gotten some of this wrong, or be out of date with what Rakoons have been up to, etc.

Issue #3714, opened 2020-11-30
Module / Package documentation is structured suboptimally


patrickbkr:
Currently we have:

- https://docs.raku.org/language/packages
- https://docs.raku.org/language/module-packages
- https://docs.raku.org/language/modules
- https://docs.raku.org/language/modules-extra

These pages are not interlinked consistently. The topic each page covers is not differentiated clearly from the topics of the other pages.

Latest case of confusion:

> <pie_flavor> In the docs, the Packages page has lots to say about how to interpolate arbitrary variables into FQ names, but not how to turn FQ names into local names. Does Raku have any sort of import/use/using/alias construct, and if so, where should I have been looking for it in the docs?

JJ:
That's not exactly the reference documentation, but tutorials, which are structured... let's say freely, if at all. So what you mention is not limited to that set of pages. Ditto could be said for regexes, for instance. But what you quote is a different thing, and I guess it could be addressed.

wayland:
Hi!  

We have 4 pages in the Language section on modules, one of which is quite large.  I'd like to propose that we do a complete reorganisation of this.  My overall plan is:

* Give modules their own subdirectory within Language
* Break them into more pages, and ensure that the names give a better idea of what they're for

Specifically, I'd like to see the following organisation:
* Modules: an Introduction
  * From: 
  	* Modules >> Basic Structure
	* Module-packages >> (What are modules?, When to use modules, Working with modules)
  * Add: links to the other Modules pages, below
* Using Modules: Finding and Installing:
	* From: Modules >> Looking for and installing modules
	* Add: a bit more guidance on raku.land
* Using Modules: In your code
	* From: Modules >> Loading and Basic Importing
* Using Modules: The Tools (a writeup on zef)
* Making Modules: The Tools
	* From:
		* Modules-extra
		* Modules-core
* Making Modules: The Code
	* From:
		* Module-packages >> (the rest)
		* Packages
* Making Modules: The Configuration and Structure
	* From: Modules >> Preparing the module
* Making Modules: Uploading
	* From: Modules >> Upload your module to CPAN/zef/p6c

We could also include fixes for:
* https://github.com/Raku/doc/issues/4341
* https://github.com/Raku/doc/issues/4260

coke:
Looks like a reasonable plan of attack.

wayland:
Have made a PR at https://github.com/Raku/doc/pull/4570 .  Feel free to tell me to break it into smaller pieces or something.

PR #4796, opened 2026-03-15
Tidy up pages about modules

schultzdavid:
* A number of links have become dead, probably through the great organization overhaul ([PR 4570](https://github.com/Raku/doc/pull/4570) in response to #3714). This should fix most if not all of them.

* While looking at any individual page, it's hard to tell where it fits in the overall context, or where to find what one is looking for if it isn't there. That's why I've copied the overview section from the bottom of [using-modules/introduction](https://docs.raku.org/language/using-modules/introduction#More_information_about_Modules) (with fixed links, of course) to every other page. That's the big block starting with `=head1 Documentation about Modules` added to the bottom of most files.

Furthermore, I have
* added some cross-references where it seems useful
* replicated the section about non-core testing tools from [distributions/tools](https://docs.raku.org/language/distributions/tools#Non-Core_Distributions) to distributions/testing
* modified some page titles or subtitles to better fit the content

There's still a lot to do, but I'd like to get this "structure cleanup" merged first. It's good to go from my side.

------

Some Todos for later:
* Put `use` first, above the more technical `need`, in [using-modules/code](https://docs.raku.org/language/using-modules/code), and add a section on `import`. In the section about `require`, briefly state how to a conditionally load a module into the right scope, along the lines of [this SO question](https://stackoverflow.com/q/79898394) and e.g. raiph's reply.
* Also in [using-modules/code](https://docs.raku.org/language/using-modules/code), at the bottom, keep only the parts about _selective importing_ and _introspection_, while moving the contents about exporting to language/making-modules/code
* Decide what to do with pages like [language/module-packages](https://docs.raku.org/language/module-packages) and [language/modules-extra](https://docs.raku.org/language/modules-extra), which consist merely of one or more links to other module pages. My preference would be to remove them altogether, in favor of making [language/modules](https://docs.raku.org/language/modules) the single, polished table of contents for everything modules. (To keep old incoming links functional, use HTTP redirects wherever possible. Only for obsolete URLs with multiple successor documents should there be—at the website level, not the Rakudoc level, or at least well-distinguished from the actual documentation—pages with links to the several successor pages; these linklist pages should not, however, be referenced from within the Rakudocs and in particular not be listed on docs.raku.org/introduction—as is currently the case for the given two examples.)

Also, the alphabetic ordering of the index of [Tutorials](https://docs.raku.org/introduction) is suboptimal for the modules pages, since it makes the advanced topics _Distributions_ and _Making Modules_ come first and the beginner topics _Using Modules_ at the end, with the beginner introduction page sitting very inconspicuously in the middle (using-modules/introduction, listed there as "Modules"). Also, I'd say that all the pages about modules (whatever their internal arrangement) are definitely worth their own subsection in or beside this Tutorials index.

### DIFF
--- a/doc/Language/distributions/configuration-structure.rakudoc
+++ b/doc/Language/distributions/configuration-structure.rakudoc
@@ -1,6 +1,6 @@
 =begin pod :kind("Language") :subkind("Modules") :category("tutorial")
 
-=TITLE Making modules: the configuration and structure
+=TITLE Distributing modules: the configuration and structure
 
 =SUBTITLE How to structure and configure Raku modules for distribution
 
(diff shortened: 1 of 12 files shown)


wayland:
@schultzdavid Thanks!

PR #4801, opened 2026-04-16
Update community.rakudoc

gfldex:
The new Raku website does not remove trailing slashes.


coke:
Thanks. I've pinged @librasteve to maybe also support that on his end.

PR #4802, opened 2026-04-18
Update control.rakudoc to fix take-rw link target

Z-raku:
Changing take-rw link from  incorrect anchor Mu#routine_take-rw to correct anchor Mu#sub_take-rw.

## The problem

Control flow page contained take-rw link to non-existent anchor Mu#routine_take-rw

## Solution provided

Changed the link to use Mu#sub_take-rw


arkiuat:
approving this for now because I'm not sure when I'll get back to Mu.rakudoc to determine whether the link target really *ought* to be routine instead of sub (and make the appropriate edits if so). In the long run, should probably make a practice of running searches for links that target any of these that have been recently modified for similar reasons (from sub or method to routine, etc).

In the meantime and the short run, a broken link is fixed. Thanks!

PR #4803, opened 2026-04-18
Update CONTRIBUTING.md - added missing header for Building the documentation section

Z-raku:
Updated doc/CONTRIBUTING.md to add missing section header

## The problem

The CONTRIBUTING.md table of contents shows "Building the documentation" section immediately following the Pull request section, but there is no section header in the body (hence toc does not have a valid link).

## Solution provided

Added ## Building the documentation in the appropriate location.


coke:
Thanks, took the opposite approach and removed the dead headers and moved the remaining sections up the hierarchy slightly.

PR #4815, opened 2026-06-18
Update 01-debugging.rakudoc

comborico1611:
Removed unnecessary characters, which are not pedagogical in context.  But rather undermine purpose of fat arrow.


coke:
Thanks!

PR #4819, opened 2026-07-12
Update 03-environment-variables.rakudoc

AntonOks:
Document commit RAKUDO_ASSUMING_DEBUG, see inline comments in https://github.com/rakudo/rakudo/commit/434bb30b0c39b466e5bca2c7a34092722127be8f

### DIFF
--- a/doc/Programs/03-environment-variables.rakudoc
+++ b/doc/Programs/03-environment-variables.rakudoc
@@ -117,6 +117,14 @@ Type: L<C<Bool>|/type/Bool>.
 
 Controls whether C<.setting> files are included in backtraces.
 
+X<|Programs,RAKUDO_ASSUMING_DEBUG>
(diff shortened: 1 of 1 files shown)


lizmat:
Thank you!

But I think you should also mention that this *only* applies to the `.assuming` method.  And I think mentioning the `RAKUDO_ASSUMING_DEBUG` environment variable with the documentation of `.assuming` is also warranted.

coke:
Please also fix the tests. Thanks.

lizmat:
@Coke What should be used instead of "deparsed" in your opinion?  Cause that's what tripping up the tests, it seems.

coke:
Sorry, I was referring to the white space. Deparsed seems perfectly fine. Leave that alone and I'll fix the spell test later.

